PR #677, opened 2025-06-19
Deleted sell_equipment_from_fighter

joeseos:


joeseos:
I've dropped it in supabase as well

PR #1595, opened 2026-04-10
Moved Custom territories creation to Campaign

Djidiouf:


joeseos:
@Djidiouf Do we still need getAllTerritoriesWithCustom in get-campaign-data.ts?

PR #1597, opened 2026-04-10
Fix multiple attackers or defenders not listed, wrong attributions of…

Djidiouf:
… gangs in attacker or defender when there are none, draws not being displayed

### DIFF
--- a/supabase/edge-functions/discord-campaign-bot/index.ts
+++ b/supabase/edge-functions/discord-campaign-bot/index.ts
@@ -44,6 +44,54 @@ Deno.serve(async (req) => {
 
     const winnerGang = battle.winner_id ? gangMap.get(battle.winner_id) : null;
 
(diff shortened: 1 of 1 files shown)


joeseos:
@Djidiouf some input (claude):

Revised summary:
1. Bug fix (positive) — the Array.isArray guard fixes a real crash risk in the existing code for the ~12,700 battles where participants is a JSONB string, not an array.
2. Draw detection — new feature, not present on main. Null winner_id shows "This battle ended in a draw." Reasonable if battles are only logged when complete. Worth confirming that assumption.
3. "Opponent" bucket is dead code — only attacker/defender roles exist today. Fine as future-proofing.
inline: false — 1v1 embeds will be taller than before. Minor UX tradeoff.
4. Field reorder & "Notes" → "Report" — deliberate improvements, no concerns.

I guess it's fine with #2 and if no winner = draw?

PR #1645, opened 2026-05-01
Added fixed size map features for Labels and Landmarks

Djidiouf:


joeseos:
## Code Review — PR #1645: Fixed size map features for Labels and Landmarks

The overall design is solid: cleanup callbacks are correctly tracked and called in all three removal paths (object replaced, object deleted, map unmounted), and the signature-based dirty check avoids unnecessary icon rebuilds. A few issues worth addressing before merge:

---

### 🔴 HIGH — Stored XSS via raw `text` interpolated into `innerHTML` (pre-existing, but this PR touches the exact code path)

**`campaign-map-canvas.tsx` — `makeLabelIcon`:**
```typescript
html: `<div ...>${text}</div>`,
```
`text` comes from `props.text`, user-authored content saved to the database and rendered for all campaign viewers. If a user saves `` as label text, it executes for every viewer. Leaflet's `L.divIcon` sets `innerHTML` directly.

**Fix:** HTML-escape before interpolation:
```typescript
const escapedText = text.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;');
html: `<div ...>${escapedText}</div>`,
```
`white-space:pre` preserves newlines/spaces; only `< > &` need escaping.

---

### 🟡 MEDIUM — `isMapRelativeMarker` and `isMapRelativeLabel` are identical functions

**`utils/campaigns/map-markers.ts` lines 83–89:**
```typescript
export function isMapRelativeMarker(properties) { return properties.fixedSizeRelativeToMap === true; }
export function isMapRelativeLabel(properties)  { return properties.fixedSizeRelativeToMap === true; }
```
Both check the exact same property. A future maintainer will change one expecting it to only affect one object type — and silently break the other. Either merge into a single `isMapRelative(properties)` and call it everywhere, or add a comment explaining they are intentional aliases.

---

### 🟡 MEDIUM — `zoom` + `zoomend` both registered for the same handler

Appears in both canvas (`buildObjectEntry`) and editor (`refreshRelativeLandmarks`, `refreshRelativeLabels`):
```typescript
map.on('zoom', refreshRelativeIcon);
map.on('zoomend', refreshRelativeIcon);
```
`zoom` fires on **every animation frame** during a pinch/scroll; `zoomend` fires once at the end. The same handler runs on every frame *and* again at the end — redundant double-fire with no benefit. With many map-relative objects this causes unnecessary icon rebuilds per frame.

**Fix:** Use only `zoomend` (correct final state, no mid-animation flicker) or only `zoom` (smooth animation). Not both.

---

### 🔵 LOW — `getLabelTextDimensions` uses a character-count heuristic for width

```typescript
width: Math.ceil(longestLine * fontSize * 0.65) + horizontalPadding,
```
`0.65` is a rough average for Latin proportional fonts. It over-estimates for narrow chars (`i l 1`), under-estimates for wide ones (`W M`), and is significantly wrong for CJK text (~1.0× wide). This can clip labels or leave excessive empty space. Consider a slightly larger safety margin (e.g. `0.75`) to reduce clipping risk.

---

### 🔵 LOW — Tooltip offset updated via direct `options` mutation + `setLatLng` as re-render trigger

```typescript
nameLayer.options.offset = getNameOffset();
nameLayer.setLatLng(geo.latlng); // triggers re-render
```
Mutating `options.offset` directly and using `setLatLng` as a flush side-effect is an undocumented Leaflet pattern that could break on Leaflet version upgrades. A short comment explaining the intent would help future readers.

---

**Verdict: Changes Requested** — items 1–3 should be addressed before merge. Items 4–5 are lower priority but worth noting.

---

PR #1646, opened 2026-05-02
Improved the UX for Captured Fighters

Djidiouf:


joeseos:
## PR Review — Improved UX for Captured Fighters

The overall direction is good: replacing the three-button confirmation modal with an inline "Captured by" combobox is a cleaner flow, and the "Rescue Fighter" context-menu action is a useful addition. The optimistic-update rollback for captured state is well structured. A few issues need attention before merging.

---

### Findings

#### 🔴 MEDIUM — `addFighterInjury`: silent status-update failure creates divergent state

**File:** `app/actions/fighter-injury.ts` lines 163–165

```ts
if (statusError) {
  console.error('Error setting fighter status:', statusError);
  // Don't fail the entire operation, just log the error
}
```

If the `fighters.update({ captured: true, ... })` call fails, the `add_fighter_injury` RPC has already committed a "Captured" injury effect to `fighter_effects`, but the fighter's `captured` flag stays `false`. Because `addFighterInjury` still returns `{ success: true }`, the client's optimistic update is kept — showing the fighter as captured — while the DB disagrees. The same silent path existed for `recovery`, but it matters more now that `captured` controls downstream rescue logic and the injury-effect sync in `editFighterStatus`.

**Suggestion:** Return `captured_status` from `addFighterInjury` (similar to `recovery_status`) so the caller knows whether the flag was actually set. At minimum, surface this as a partial-success indicator rather than full success.

---

#### 🔴 MEDIUM — Non-atomic operation in `editFighterStatus` `capture` case

**File:** `app/actions/edit-fighter.ts` lines 666–740

The fighters table update (`captured = willBeCaptured`) commits before the `add_fighter_injury` / `delete` calls. If the RPC or delete fails afterward, the function throws and returns `{ success: false }`, but the fighters row is already updated. This leaves `captured = true` with no matching `fighter_effects` row (or vice versa on release).

This is an existing pattern in the codebase, but the new code makes it more impactful because `editFighterStatus` is now responsible for keeping `fighters.captured` and `fighter_effects` in sync. If a DB-level transaction (or a compensating update in the catch block) isn't feasible, the code should at minimum re-update the fighters row to its original value inside the catch path rather than leaving it dirty.

---

#### 🟡 MEDIUM — "Rescue Fighter" has no confirmation

**File:** `components/gang/gang-page-content.tsx` lines 415–464

`rescueFighter` is a single-click, fire-and-forget action: it optimistically removes the Captured status and injury immediately with no confirmation prompt. Rescuing a fighter removes the injury record and resets `captured_by_gang_id` permanently. A misclick during a busy roster-management session can accidentally release a fighter with no in-UI undo. All other destructive status changes in this file (capture via FighterActions, recovery) are gated behind a modal.

**Suggestion:** Add a simple confirmation dialog (e.g., `"Rescue [Fighter Name] from captivity?"`) before calling `editFighterStatus`.

---

#### 🟡 MEDIUM — Captured injury can be added twice via the injury form

**File:** `components/fighter/fighter-injury-list.tsx` `handleAddInjury`

`addFighterInjury` calls `add_fighter_injury` RPC unconditionally — it does not check whether a "Captured" injury already exists for the fighter. If a fighter was captured via the FighterActions toggle (which adds the injury via `editFighterStatus`), a user can still open the injury modal and submit a second "Captured" injury, creating a duplicate in `fighter_effects`.

The deduplication guard added to `editFighterStatus` (lines 681–689) doesn't cover this path. Add a similar pre-check in `addFighterInjury` when `params.set_captured` is true, or disable/hide the "Captured" injury option in the dropdown when `fighterCaptured === true`.

---

#### 🟡 MEDIUM — `fighterCampaigns` fallback may use structurally mismatched data

**File:** `components/gang/gang-page-content.tsx` line 543

```ts
fighterCampaigns={currentFighter.campaigns ?? gangData.processedData.campaigns}
```

The `campaignIds` memo in `InjuriesList` maps `campaign.campaign_id ?? campaign.id`. Gang-level campaign objects likely carry only `id` (not `campaign_id`), so the fallback will still resolve IDs. However, the shape of `gangData.processedData.campaigns` is `any` here. If the structure diverges from `Array<{ campaign_id?: string; id?: string }>`, the gang-selection dropdown could silently fail to fetch gangs without any visible error.

**Suggestion:** Type `processedData.campaigns` explicitly, or validate the fallback shape before passing it.

---

#### 🟢 LOW — Unnecessary `<>` fragment wrapping a single element

**File:** `components/fighter/fighter-actions.tsx` lines 532–561

The `<Combobox>` is wrapped in a `<>...</>` fragment but is the sole child in that branch. The fragment was likely added to match the pattern in `fighter-injury-list.tsx` where additional content was added alongside the combobox. Here it adds no structural value and can be removed.

---

#### 🟢 LOW — `null ?? undefined` coercion loses semantic intent

**Files:** `components/gang/gang-page-content.tsx` line 579, `components/fighter/fighter-page.tsx` line 1051

```ts
captured_by_gang_id: capturedByGangId !== undefined ? capturedByGangId ?? undefined : f.captured_by_gang_id,
```

When `capturedByGangId` is `null` (fighter captured, no gang specified), `null ?? undefined` evaluates to `undefined`. Both `null` and `undefined` map to "no capturing gang" in the current type (`?: string | null`), so this is functionally correct. It is, however, slightly surprising: a deliberate `null` from the server is quietly coerced to `undefined` in local state. A comment or a direct `=== null ? null : capturedByGangId` form would make the intent explicit.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | `app/actions/fighter-injury.ts` | Silent `captured` flag failure returns success, diverging client/DB state |
| 2 | Medium | `app/actions/edit-fighter.ts` | Non-atomic: fighters row updated before injury RPC; inconsistent on RPC failure |
| 3 | Medium | `components/gang/gang-page-content.tsx` | No confirmation before rescuing a fighter |
| 4 | Medium | `components/fighter/fighter-injury-list.tsx` | No deduplication guard — duplicate Captured injury possible via form |
| 5 | Medium | `components/gang/gang-page-content.tsx` | `fighterCampaigns` fallback shape unverified |
| 6 | Low | `components/fighter/fighter-actions.tsx` | Unnecessary `<>` fragment |
| 7 | Low | gang-page-content / fighter-page | `null ?? undefined` coercion semantics |

---

**Verdict: Changes requested.** Issues 1, 3 and 4 are the most important to address before merging. Issues 2 and 5 are worth a follow-up even if not blocking. The core UX approach — inline gang selection replacing the three-button modal — is a clear improvement and the optimistic rollback for captured state is implemented correctly.

---

joeseos:
@Djidiouf 

Revised priority summary
1. High — Backfill migration needed. The data drift is real and affects a lot of fighters. Without it, the "Captured" injury filter and hasCapturedInjury guard will be inconsistent for existing users.
2. Medium — Toggle race condition. The action: 'capture' toggle pattern is risky for the rescue flow. Consider adding explicit directionality.
3. Medium — N+1 campaign gang fetching. Sequential fetches per campaign ID should be parallelized or batched.
4. Low — useEffect → TanStack Query migration. Consistent with the ongoing audit but not blocking.
5. Low — Type inconsistency on captured_by_gang_id. Annoying but not a runtime bug today.
6. Nit — Unnecessary fragment, dead-code toast guard, implicit undefined return.

PR #1648, opened 2026-05-04
Adjusted fighter sort order in the stash tab

joeseos:


joeseos:
## Code Review — PR #1648: Adjusted fighter sort order in the stash tab

### Overview

This PR adds `positioning` support to the `GangInventory` (stash tab) component so that the fighter dropdown in the item-assignment UI reflects the same drag-ordered fighter sequence used elsewhere in the app. Two files changed: the prop is threaded from `gang-page-content.tsx` down into `stash-tab.tsx`, where it drives a sort in the render path.

---

### Findings

#### ✅ Correctness

- The sort logic (`Object.entries(positioning).find(...)` → `parseInt` → compare) is **identical** to the pattern in `fighter-page.tsx:608–616` and `print-gang.tsx`. Consistency is good.
- `[...fighters].sort(...)` correctly shallow-copies before sorting, avoiding mutation of the component state.
- The `if (!positioning) return 0` guard correctly handles the optional prop — undefined positioning leaves the list unsorted rather than crashing.
- Fighters absent from the positioning map get `Infinity`, putting them at the end, which matches the rest of the codebase.

#### ⚠️ Medium — Performance: O(n × m) comparator

```tsx
// stash-tab.tsx ~797
const indexA = Object.entries(positioning).find(([, id]) => id === a.id)?.[0];
const indexB = Object.entries(positioning).find(([, id]) => id === b.id)?.[0];
```

`Object.entries(positioning).find(...)` is called **twice per comparison**, making this O(n × m) where n = fighters and m = positioning entries. For typical gang sizes (≤ 30) this is negligible, but it's worth noting that the same pattern exists in `fighter-page.tsx` — if it ever becomes a concern, a pre-built `Map<fighterId, positionIndex>` computed once before the sort would fix it. This is a low-severity issue at current data scale, and since it matches existing patterns, no change is required now.

#### ✅ Type safety

`positioning` is typed `Record<number, string>` throughout the codebase. `Object.entries()` coerces keys to strings at runtime, and `parseInt()` handles that correctly. No issues here.

#### ✅ Null safety

`gangData.processedData.positioning` could theoretically be `undefined` if the lazy-initialisation in `app/gang/[id]/page.tsx` didn't run. The component's `if (!positioning) return 0` guard handles this safely.

#### ✅ Scope

Only the "Fighters" `<optgroup>` is sorted. The vehicles section is unaffected — appropriate since `positioning` tracks fighters, not vehicles.

#### ✅ Prop threading

`gang-page-content.tsx` passes `gangData.processedData.positioning` directly, which is the same source used by `gang.tsx` (line 243) and `print-gang.tsx`. Single source of truth, no duplication.

---

### Summary

| Area | Status |
|---|---|
| Correctness | ✅ No bugs |
| Edge cases / null handling | ✅ Guarded correctly |
| Security | ✅ N/A |
| Performance | ⚠️ O(n×m) comparator — acceptable at current scale, consistent with existing code |
| Readability | ✅ Clear and idiomatic |
| Project conventions | ✅ Matches the pattern in `fighter-page.tsx` and `print-gang.tsx` |

---

### Verdict: **Approve** ✅

Small, focused, correct fix. The implementation is consistent with established patterns in the codebase. The only note is the O(n×m) sort comparator, which is a pre-existing pattern and inconsequential at typical data sizes.

---

PR #1649, opened 2026-05-04
Killed status automatically added when relevant injury or rig glitch …

Djidiouf:
…are applied


joeseos:
## Code Review — PR #1649: "Killed status automatically added when relevant injury or rig glitch are applied"

**Verdict: ❌ Changes Requested**

The overall design is solid — the `killed: 'true'` flag in `type_specific_data`, the migration backfill, and the optimistic-UI patterns all fit project conventions well. However, there are two data-corruption bugs in gang rating accounting that affect every application of a fatal injury to an active fighter, which must be fixed before merging.

---

### 🔴 HIGH — Gang rating double-counted when adding a fatal injury

**File:** `app/actions/fighter-injury.ts` (around the rating-update block in `addFighterInjury`)

`addFighterInjury` calls `updateGangRatingSimple(+delta)` for the injury's `credits_increase` **before** determining `killedRatingDelta`, then calls `getFighterTotalCost` to compute how much to subtract when the fighter is killed. Because `BASE_FIGHTER_EFFECTS` is not yet invalidated at that point, the cache returns the **pre-injury cost**, missing the just-added `credits_increase`.

Concrete example — fighter with base cost 100, fatal injury with `credits_increase = 10`:

| Step | Operation | Rating |
|---|---|---|
| Before | — | 200 |
| +10 (injury credits_increase added) | +10 | 210 |
| −100 (stale cache, doesn't include the new injury) | −100 | **110** |
| **Expected** (remove full cost 100+10) | — | **90** |

This leaves the rating **20 too high** — 2× the injury's `credits_increase`. Since all injuries default to `credits_increase = 10`, this affects every application of Memorable Death or Critical Overload to an active fighter.

**Fix:** Skip the standalone `credits_increase` delta update when the fighter is being killed, and fold it into `killedRatingDelta`:
```ts
const delta = (injuryData?.type_specific_data?.credits_increase || 0) as number;
// Only apply the credits_increase delta if we're NOT about to kill the fighter
if (delta && !shouldSetKilled) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta);
}
// ...
if (shouldSetKilled && !fighter.killed) {
  if (countsTowardRating(fighter)) {
    // getFighterTotalCost returns pre-injury cached cost, so add the delta manually
    killedRatingDelta = -(await getFighterTotalCost(params.fighter_id, supabase)) - delta;
  }
}
```

---

### 🔴 HIGH — Gang rating over-restored when deleting a fatal injury (resurrect)

**File:** `app/actions/fighter-injury.ts` (around the rating-update block in `deleteFighterInjury`)

The reverse direction has the same caching race. `deleteFighterInjury` reverses `credits_increase` first, then calls `getFighterTotalCost` to determine how much rating to restore on resurrect. The cache still includes the deleted injury at that point, so the restore amount is too high by that injury's `credits_increase`.

Example — fighter with base cost 80 + fatal injury (`credits_increase = 20`) = stale cached cost 100, rating currently 0 (fighter was killed):

| Step | Operation | Rating |
|---|---|---|
| −20 (reverse credits_increase) | −20, clamped to 0 | 0 |
| +100 (stale cache, still includes deleted injury) | +100 | **100** |
| **Expected** (restore without deleted injury: +80) | — | **80** |

**Fix:** Subtract the deleted injury's `credits_increase` from `resurrectedRatingDelta`:
```ts
resurrectedRatingDelta = await getFighterTotalCost(params.fighter_id, supabase);
// Subtract the injury we just deleted since cache still includes it
resurrectedRatingDelta -= (injury?.type_specific_data?.credits_increase || 0);
```

---

### 🟡 MEDIUM — Equipment-targeting path silently omits `set_killed` in optimistic update

**File:** `components/fighter/fighter-injury-list.tsx` (`onApplyToTarget` callback)

When a fatal injury `applies_to === 'equipment'`, the equipment-selection modal path calls `addInjuryMutation.mutate` without passing `set_killed`. The server correctly detects and applies the killed status via `hasKilledStatusFlag`, so the database is fine. However, the optimistic client update checks `variables.set_killed` — which will be `undefined` — so the fighter won't appear killed in the UI until server reconciliation. This is a latent bug; no current fatal injury targets equipment, but it will break immediately if one is added.

**Fix:** Compute and pass `set_killed` in `onApplyToTarget` the same way as the other branches:
```ts
const typeSpecificData = (selectedInjury as any).type_specific_data || {};
const requiresKilled = hasKilledStatusFlag(typeSpecificData);
addInjuryMutation.mutate({ ..., set_killed: requiresKilled });
```

---

### 🟡 MEDIUM — No loading state on the Resurrect confirm button

**File:** `components/gang/gang-page-content.tsx` (`confirmResurrectFighter`)

The optimistic update fires synchronously and the modal closes before `editFighterStatus` resolves. On failure the rollback restores state correctly, but a double-tap or a slow network response could result in a confusing UX flash. Consider disabling the confirm button while the request is in flight.

---

### 🟡 MEDIUM — Optimistic `shouldClearKilled` may briefly show wrong state in multi-tab scenarios

**File:** `components/fighter/fighter-injury-list.tsx` (`deleteInjuryMutation.onMutate`, lines 295–308)

The client checks the local injury list to decide whether to clear `killed` optimistically. If another session added a second fatal injury that the current tab hasn't received yet, the optimistic update will transiently clear `killed` incorrectly. The `onSuccess` reconciliation with `result.killed_status` recovers correctly, so this is a minor UX flicker rather than a data issue.

---

### 🟢 LOW — `hasKilledStatusFlag` duplicated across three files

**Files:** `app/actions/fighter-injury.ts`, `app/actions/edit-fighter.ts`, `components/fighter/fighter-injury-list.tsx`

Identical helper defined three times. Extract to a shared utility (e.g., `utils/fighter-status.ts`) so a future flag-format change only needs one edit.

---

### 🟢 LOW — Consistent typo: `'fighter_resurected'` (missing one `r`)

**Files:** `app/actions/fighter-injury.ts`, `app/actions/edit-fighter.ts`, `app/actions/logs/fighter-logs.ts`

Self-consistent so no runtime error, but will be confusing in log queries. Should be `'fighter_resurrected'`.

---

### 🟢 LOW — Dead variable `willBeActive = false`

**File:** `app/actions/fighter-injury.ts`

`const willBeActive = false` is always `false`; `if (wasActive && !willBeActive)` simplifies to `if (wasActive)`. Remove the variable.

---

### 🟢 LOW — Migration matches `fighter_effects` rows by `effect_name` string

**File:** `supabase/migrations/20260504163000_mark_fatal_effects_killed.sql`

The backfill UPDATE matches existing `fighter_effects` rows via `effect_name IN ('Memorable Death', 'Critical Overload')`. Fine for a one-time backfill, but worth noting that `effect_name` is denormalized — matching on `fighter_effect_type_id` would be more robust.

---

**Summary:** Fix findings 1 and 2 (the rating accounting bugs) before merging. The rest are medium/low and can be addressed as follow-ups or in the same PR at your discretion.

---

joeseos:
## Follow-up Review — Commit 5e712771

Thanks for the follow-up commit. The structural sequencing of the rating updates is improved, but both HIGH-severity bugs from the previous review are still present due to a Next.js `unstable_cache` staleness issue, and one MEDIUM issue is also still open. Details below.

---

### Prior Findings — Updated Status

#### 🔴 HIGH — Rating double-count in `addFighterInjury` — PARTIALLY FIXED, CORE BUG REMAINS

The sequencing has improved (killed delta is now computed and applied separately from the `credits_increase` delta), but the **cache staleness problem is unchanged**. The execution order is:

1. `supabase.rpc('add_fighter_injury', ...)` — writes the new injury to `fighter_effects`
2. `updateGangRatingSimple(+delta)` for `credits_increase`
3. `killedRatingDelta = -(await getFighterTotalCost(...))` — reads from `unstable_cache`

At step 3, `BASE_FIGHTER_EFFECTS` and `COMPUTED_FIGHTER_TOTAL_COST` have not yet been invalidated (that happens later at the explicit `revalidateTag` calls). In Next.js, `revalidateTag` marks entries for invalidation on the *next* read — it does not synchronously bust the cache mid-request. So `getFighterTotalCost` at step 3 returns the **pre-injury cost**, missing the newly-added injury's `credits_increase`.

Net effect: the `credits_increase` is still counted twice — once via `+delta` at step 2, and again because `getFighterTotalCost` returns a stale value that causes `killedRatingDelta` to be computed on a cost that doesn't include what was just added.

**Fix:** Capture the fighter's total cost *before* the RPC call, so there's no dependency on post-write cache state:
```ts
// Capture cost BEFORE the RPC call modifies fighter_effects
const preInjuryCost = shouldSetKilled ? await getFighterTotalCost(params.fighter_id, supabase) : 0;

await supabase.rpc('add_fighter_injury', ...);

// credits_increase delta - skip when killing (it's already in preInjuryCost)
if (delta && !shouldSetKilled) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta);
}

// Later, when computing killedRatingDelta:
if (shouldSetKilled && !fighter.killed && countsTowardRating(fighter)) {
  killedRatingDelta = -(preInjuryCost + delta); // base cost + the new injury's credits_increase
}
```

---

#### 🔴 HIGH — Rating over-restore in `deleteFighterInjury` — PARTIALLY FIXED, CORE BUG REMAINS

Same root cause in reverse. `getFighterTotalCost` is called after the injury is deleted from `fighter_effects`, but `BASE_FIGHTER_EFFECTS` has not been invalidated yet. Within the same server-action invocation, `unstable_cache` may serve the stale value that still includes the deleted injury's cost, causing the rating to be over-restored by that injury's `credits_increase`.

**Fix:** Capture the fighter's total cost *before* the delete:
```ts
// Capture cost BEFORE deleting so we get the accurate pre-deletion cost
const preDeletionCost = willResurrect ? await getFighterTotalCost(params.fighter_id, supabase) : 0;

// ... perform the delete ...

if (willResurrect) {
  // Subtract the deleted injury's credits_increase since we don't want to restore it
  resurrectedRatingDelta = preDeletionCost - (injury?.type_specific_data?.credits_increase || 0);
}
```

---

#### 🟡 MEDIUM — `onApplyToTarget` missing `set_killed` in optimistic update — NOT FIXED

`components/fighter/fighter-injury-list.tsx`, `onApplyToTarget` callback.

Both branches (`requiresCaptured` and the `else`) still call `addInjuryMutation.mutate(...)` without `set_killed`. The server will correctly set the killed status, but the optimistic update reads `variables.set_killed` (which will be `undefined`) so the fighter card won't show as killed until the server round-trip completes.

**Fix:**
```ts
const typeSpecificData = (selectedInjury as any).type_specific_data || {};
const requiresKilled = hasKilledStatusFlag(typeSpecificData);
// Pass set_killed: requiresKilled in both branches of addInjuryMutation.mutate
```

---

#### 🟡 MEDIUM — No loading state on Resurrect confirm button — NOT FIXED

`components/gang/gang-page-content.tsx`, Resurrect modal. `confirmResurrectFighter` is async but there is no `isLoading` / disabled state on the confirm button. A double-tap during the server round-trip can trigger duplicate requests.

---

### LOW Findings — Still Open

| Finding | Status |
|---|---|
| `hasKilledStatusFlag` still duplicated in 3 files (not extracted to shared util) | Not fixed |
| Typo `'fighter_resurected'` — database key still misspelled in `log-types.ts`, `fighter-logs.ts`, `edit-fighter.ts`, `fighter-injury.ts` | Not fixed |
| Dead variable `willBeActive = false` in `addFighterInjury` | Not fixed |

---

### New Issue from This Commit

#### 🟢 LOW — Comment at rollback path is misleading

`app/actions/fighter-injury.ts`, rollback block. The comment `// if (delta)` refers only to reversing `credits_increase`, not `killedRatingDelta`. The behavior is actually correct (the killed delta hasn't been applied yet at that point), but the comment is confusing for future maintainers. Worth clarifying.

---

### Verdict: ❌ Changes Requested

The two HIGH bugs both boil down to the same root cause: `getFighterTotalCost` is called in the same server-action request where the cache has not yet been invalidated. The cleanest fix for both is to **capture the fighter's total cost before any writes**, removing any dependency on post-write cache state. That one change resolves both HIGH findings simultaneously. The `set_killed` omission in `onApplyToTarget` should also be addressed before merging.

---

joeseos:
## Follow-up Review — Commit 93eb2eded7

@Djidiouf Good progress on this commit — the architectural approach is now correct and several prior issues are resolved. But there are two arithmetic errors that still corrupt the gang rating, both in `app/actions/fighter-injury.ts`.

---

### What's Fixed ✅

| Prior finding | Status |
|---|---|
| `hasKilledStatusFlag` duplicated across 3 files | ✅ Extracted to `utils/fighter-status.ts` |
| Typo `'fighter_resurected'` in log action type | ✅ Corrected to `'fighter_resurrected'` everywhere |
| Dead variable `willBeActive = false` | ✅ Removed |
| `onApplyToTarget` missing `set_killed` in equipment path | ✅ Both branches now pass `set_killed: requiresKilled` |
| `getFighterTotalCost` called after write (cache staleness) | ✅ `preInjuryCost` and `preDeletionCost` now fetched **before** any DB writes |

---

### 🔴 HIGH — `addFighterInjury`: rating still off by `delta` (direction reversed)

The new code correctly skips adding `credits_increase` to the rating when the fighter is being killed (`if (delta && !shouldSetKilled)` guard). But then:

```ts
killedRatingDelta = -(preInjuryCost + delta);
```

Since `delta` was **never added** to the rating (the guard above skipped it), subtracting it here is an overcorrection.

Example — fighter cost C=100, fatal injury `credits_increase`=10, rating R=200:

| Step | Operation | Rating |
|---|---|---|
| delta guard fires | +0 (skipped) | 200 |
| `killedRatingDelta` applied | −(100 + 10) | **90** |
| **Expected** (remove only pre-injury cost) | −100 | **100** |

The original bug left the rating 10 too **high**. This fix makes it 10 too **low**. Same magnitude, opposite sign.

**Fix — one-line change:**
```ts
// delta was never added to the rating, so don't subtract it here
killedRatingDelta = -preInjuryCost;
```

---

### 🔴 HIGH — `deleteFighterInjury`: `updateGangRatingSimple(-delta)` fires spuriously on resurrection

Lines 394–397:
```ts
const delta = -(injury?.type_specific_data?.credits_increase || 0) as number;
if (delta) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta); // subtracts credits_increase
}
```

This runs for **all** injury deletions. For non-fatal injuries it's correct — the `credits_increase` was added when the injury was applied, so this reverses it.

For **fatal injuries** (resurrection path): `credits_increase` was **never added** to the rating (the `!shouldSetKilled` guard in `addFighterInjury` skips it). Reversing it here incorrectly drains the rating on every resurrection.

`resurrectedRatingDelta = preDeletionCost - credits_increase` is correct in isolation (it strips the deleted injury's cost from the pre-deletion total and gives back `C`). But stacked with the spurious `-delta` call, the net resurrection bonus becomes `C - delta` instead of `C`.

**Combined kill + resurrect cycle** (assuming finding #1 is also fixed so kills are correct):

| Step | Operation | Rating |
|---|---|---|
| After kill (with fix #1) | R − C | 100 |
| Resurrect: `-delta` fires wrongly | 100 − 10 | 90 |
| Resurrect: `+resurrectedRatingDelta` (= 110 − 10 = 100) | 90 + 100 | **190** |
| **Expected** | R restored | **200** |

Each kill+resurrect cycle loses `2 × credits_increase` (20 credits at the default value). This is persistent data corruption that compounds with each cycle.

**Fix — guard the call for non-fatal-only:**
```ts
const creditsIncrease = (injury?.type_specific_data?.credits_increase || 0) as number;
const delta = -creditsIncrease;
// Fatal injuries skip adding credits_increase on apply, so don't reverse it on removal
if (delta && !removedKilledStatusEffect) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta);
}
```

`resurrectedRatingDelta = preDeletionCost - creditsIncrease` already returns the correct value (`C`) — no other change needed there.

---

### 🟢 LOW — Extra DB round-trip in `addFighterInjury`

The new pre-write `fighter_effect_types` query is necessary for correctness on the kill path, but it fires on **every** `addFighterInjury` call including non-fatal injuries. Minor optimisation: only fetch it when `!params.set_killed` (i.e., when we need to check the DB flag ourselves). Not a blocker.

---

### Still Open

- **MEDIUM** — No loading state on the Resurrect confirm button (`gang-page-content.tsx`) — double-tap during server round-trip can trigger duplicate requests.

---

### Verdict: ❌ Changes Requested

Two small fixes needed, both in `app/actions/fighter-injury.ts`:

1. `killedRatingDelta = -preInjuryCost` (remove `+ delta`)
2. Guard `updateGangRatingSimple(delta)` with `&& !removedKilledStatusEffect`

Once those are in, the kill and resurrect accounting will be correct and consistent.

---

joeseos:
## Review — Commit 41285db5f3 ✅ Approved

@Djidiouf All the issues raised in the previous reviews are now resolved. Here's the full verification.

---

### HIGH Findings — Both Fixed ✅

#### `addFighterInjury` — rating delta corrected

```ts
// Before (commit 93eb2eded7) — delta was subtracted even though it was never added
killedRatingDelta = -(preInjuryCost + delta);

// After (this commit) — correct: only remove what was actually in the rating
killedRatingDelta = -preInjuryCost;
```

Verified arithmetic (fighter cost C=100, fatal injury `credits_increase`=10, rating R=200):

| Step | Operation | Rating |
|---|---|---|
| `delta` guard (`!shouldSetKilled`) | +0 skipped | 200 |
| `killedRatingDelta` applied | −100 | **100** = R − C ✓ |

#### `deleteFighterInjury` — spurious `updateGangRatingSimple` guarded

```ts
// Before — fired for all injury removals including fatal ones
if (delta) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta);
}

// After — correctly skipped for fatal injuries (credits_increase was never added on kill)
if (delta && !removedKilledStatusEffect) {
  await updateGangRatingSimple(supabase, fighter.gang_id, delta);
}
```

Kill + resurrect cycle now nets zero, verified:

| Step | Operation | Rating |
|---|---|---|
| Kill (fix above) | R − C | 100 |
| Resurrect: `delta` guard fires | +0 skipped | 100 |
| Resurrect: `resurrectedRatingDelta` = `preDeletionCost − creditsIncrease` = C | +100 | **200** = R ✓ |

---

### MEDIUM Finding — Resurrect modal loading state ✅ Fixed

`confirmResurrectFighter` now tracks in-flight state with `isResurrectSubmitting`, the confirm button is disabled via `confirmDisabled={isResurrectSubmitting}`, and an early-return guard prevents concurrent calls. The `try/finally` block ensures the flag is always cleared.

---

### All Prior Low-Severity Findings ✅

| Finding | Status |
|---|---|
| `hasKilledStatusFlag` duplicated | ✅ Extracted to `utils/fighter-status.ts` |
| Typo `fighter_resurected` | ✅ Fixed to `fighter_resurrected` everywhere |
| Dead variable `willBeActive = false` | ✅ Removed |
| `onApplyToTarget` missing `set_killed` | ✅ Fixed |
| Extra DB round-trip on non-kill path | ✅ Now guarded: query only fires when `!params.set_killed` |

---

### One Residual Note (non-blocking)

The optimistic `shouldClearKilled` check in `deleteInjuryMutation.onMutate` still reads the local injury list, which can be stale in a multi-tab scenario. The `onSuccess` reconciliation with `result.killed_status` corrects it, so this only causes a brief UI flicker and not a data issue. Acceptable trade-off for optimistic UI.

---

This is clean and correct. The core accounting logic is now sound, the cache-staleness problem is properly solved by capturing costs before writes, and the UX edge cases are addressed. Good work! 🎉

---

PR #1650, opened 2026-05-07
fix: surface PostgrestError messages in server action catch blocks

joeseos:
PostgrestError is a plain object with a .message property but does not extend Error, causing instanceof Error checks to return false and swallow real database error messages. Now falls back to (error as any)?.message before the generic unknown error string.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -1185,7 +1185,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     console.error('Error in addFighterToGang server action:', error);
     return {
       success: false,
-      error: error instanceof Error ? error.message : 'An unknown error occurred'
+      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
     };
   }
 } 
--- a/app/actions/add-gang-vehicle.ts
+++ b/app/actions/add-gang-vehicle.ts
@@ -219,7 +219,7 @@ export async function addGangVehicle(params: AddGangVehicleParams): Promise<AddG
     console.error('Error in addGangVehicle server action:', error);
     return {
       success: false,
-      error: error instanceof Error ? error.message : 'An unknown error occurred'
+      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
     };
   }
 }
@@ -271,7 +271,7 @@ export async function getGangVehicleTypes(gangId: string) {
     console.error('Error fetching vehicle types:', error);
     return {
       success: false,
-      error: error instanceof Error ? error.message : 'An unknown error occurred'
+      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
     };
   }
 }
--- a/app/actions/assign-vehicle-to-fighter.ts
+++ b/app/actions/assign-vehicle-to-fighter.ts
@@ -208,7 +208,7 @@ export async function assignVehicleToFighter(params: AssignVehicleToFighterParam
     console.error('Error in assignVehicleToFighter server action:', error);
     return { 
       success: false, 
-      error: error instanceof Error ? error.message : 'An unknown error occurred'
+      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
     };
   }
 }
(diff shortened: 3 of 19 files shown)


joeseos:
## PR Review — `fix: surface PostgrestError messages in server action catch blocks`

**Reviewed by:** Claude (senior engineer review)
**Tagging author:** @joeseos

---

### Summary

The premise is **correct**: in `@supabase/supabase-js` v2.x (this project uses `^2.90.1`), `PostgrestError` is a plain object type `{ message: string; details: string; hint: string; code: string }` — it does **not** extend `Error`. So `instanceof Error` returns `false` for these, and the old fallback was silently returning `'An unknown error occurred'` instead of the real database error.

The per-file fix is mechanically sound. However, there are several issues worth addressing before merging.

---

### Findings

#### 🔴 HIGH

None that block correctness of the touched code.

---

#### 🟡 MEDIUM

**1. Fix is incomplete — ~80 instances of the old pattern remain unfixed**

The PR patches 24 catch blocks across 19 files, but a search of `app/actions/` finds approximately 80 more catch blocks with the same unpatched pattern:

```
app/actions/fighter-advancement.ts       (6 instances)
app/actions/campaigns/[id]/campaign-territories.ts  (6 instances)
app/actions/campaigns/[id]/campaign-members.ts      (5 instances)
app/actions/campaigns/[id]/campaign-allegiances.ts  (4 instances)
app/actions/loadouts.ts                  (4 instances)
app/actions/fighter-injury.ts            (3 instances)
app/actions/customise/custom-fighters.ts (3 instances)
app/actions/customise/custom-share.ts    (4 instances)
app/actions/campaigns/[id]/campaign-gangs.ts (3 instances)
app/actions/equipment.ts                 (line 203, unfixed)
... and ~40 more
```

The bug affects every `throw error` that re-throws a `PostgrestError`. Leaving the majority unpatched creates a confusing inconsistency — some actions surface real errors, others silently swallow them.

**Suggestion:** Either fix all instances in this PR, or (preferred) extract a shared utility (see finding #3 below) and apply it everywhere in one pass.

---

**2. PostgREST error messages are surfaced directly to end users**

Confirmed via `components/gang/add-fighter.tsx:392`:
```typescript
toast.error(error instanceof Error ? error.message : 'Failed to add fighter');
```
The action's `result.error` string is re-thrown as a new `Error` and shown in a toast. PostgREST messages can expose internal schema details, e.g.:
```
duplicate key value violates unique constraint "gangs_name_key"
```

This was already the case for JavaScript `Error` objects, but PostgREST errors tend to be more verbose about schema internals. It may be intentional for a tool like this (power-user audience), but it's worth a deliberate decision. At minimum, consider whether `error.details` and `error.hint` (also on `PostgrestError`) should be appended or omitted.

---

**3. `(error as any)` bypasses TypeScript safety**

```typescript
// current fix
error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
```

This pattern is repeated 24 times in the fix (and exists 100+ times total). A typed alternative avoids `any` and centralises the logic:

```typescript
// lib/errors.ts
export function getErrorMessage(error: unknown): string {
  if (error instanceof Error) return error.message;
  if (typeof error === 'object' && error !== null && 'message' in error) {
    return String((error as { message: unknown }).message);
  }
  return 'An unknown error occurred';
}
```

Then every catch block becomes:
```typescript
return { success: false, error: getErrorMessage(error) };
```

This also makes it trivial to later add structured logging of `error.code`/`error.details` in one place.

---

#### 🟢 LOW

**4. Missing trailing newlines are pre-existing**

Several files in the diff show `\ No newline at end of file`. These existed before this PR and aren't introduced by it — no action needed here.

**5. `PostgrestError.details` / `.hint` not surfaced**

PostgREST errors carry `details` and `hint` fields that are often more actionable than `message` alone (e.g., `hint: "You may need to add an explicit cast"`). Not a blocker, but worth noting for future improvement.

---

### Verdict

**🔶 Changes Requested**

The fix is directionally correct and addresses a real bug. The main concern is **incompleteness** — patching 24 of ~100+ affected catch blocks leaves the codebase in an inconsistent state where the same underlying bug is silently present in the majority of actions. Strongly recommend either:

1. Extracting `getErrorMessage()` into `lib/errors.ts` and applying it across all action files in this PR, or
2. Scoping this PR to the highest-traffic actions and opening a follow-up issue to track full coverage.

Happy to see this merged once the coverage gap is addressed.


---

PR #1651, opened 2026-05-07
Revert "fix: surface PostgrestError messages in server action catch blocks"

joeseos:
Reverts joeseos/mundamanager#1650

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -1185,7 +1185,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     console.error('Error in addFighterToGang server action:', error);
     return {
       success: false,
-      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
+      error: error instanceof Error ? error.message : 'An unknown error occurred'
     };
   }
 } 
--- a/app/actions/add-gang-vehicle.ts
+++ b/app/actions/add-gang-vehicle.ts
@@ -219,7 +219,7 @@ export async function addGangVehicle(params: AddGangVehicleParams): Promise<AddG
     console.error('Error in addGangVehicle server action:', error);
     return {
       success: false,
-      error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
+      error: error instanceof Error ? error.message : 'An unknown error occurred'
     };
   }
 }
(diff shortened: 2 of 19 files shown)


joeseos:
## PR Review — Revert "fix: surface PostgrestError messages in server action catch blocks"

**@joeseos** — reviewed by Claude Code.

---

### Key Finding: The original PR #1650 was based on a false premise

I pulled the actual source of `@supabase/postgrest-js@2.90.1` (the version pinned in `package-lock.json`) to verify the claim made in #1650:

> *"PostgrestError is a plain object with a .message property but does not extend Error"*

**This is incorrect.** The actual implementation in v2.90.1 is:

```ts
// package/src/PostgrestError.ts — @supabase/postgrest-js@2.90.1
export default class PostgrestError extends Error {
  details: string
  hint: string
  code: string

  constructor(context: { message: string; details: string; hint: string; code: string }) {
    super(context.message)  // ← message is set on the Error base
    this.name = 'PostgrestError'
    ...
  }
}
```

`PostgrestError extends Error`, so `error instanceof Error` already returns `true` and `error.message` already surfaces the PostgREST error message correctly. The `(error as any)?.message` fallback introduced by #1650 was redundant.

---

### Impact of the Revert

The changed pattern in all 19 files goes from:

```ts
error: error instanceof Error ? error.message : (error as any)?.message || 'An unknown error occurred'
```

back to:

```ts
error: error instanceof Error ? error.message : 'An unknown error occurred'
```

Because `PostgrestError extends Error`, the `instanceof Error` branch fires correctly for all PostgREST errors, including those thrown directly via `throw error` (e.g., `create-gang.ts:80`, `equipment.ts:622`). Error messages are **not lost** after this revert.

---

### Findings

| File | Line | Issue | Severity |
|------|------|-------|----------|
| — | — | No regressions introduced. Revert is semantically equivalent for all runtime error types in this codebase. | — |

**One observation (not a blocker):** Several campaign action files (`campaign-members.ts`, `campaign-territories.ts`, etc.) use `if (error) throw error` to rethrow raw PostgREST errors, and their own catch blocks use the same `instanceof Error ? error.message` pattern. These were not touched by #1650 or this revert, but they work correctly for the same reason — `PostgrestError extends Error`.

The `(error as any)` cast removed here was a TypeScript code smell (bypassing type safety unnecessarily), so removing it is a net positive for code quality.

---

### Verdict: ✅ Approve

The revert is correct. PR #1650 was built on an incorrect understanding of `PostgrestError`'s inheritance in the version of `@supabase/postgrest-js` this project uses. This revert safely restores the original, type-safe error handling with no functional regression.

---

PR #1652, opened 2026-05-07
Fix FK violation when editing fighter to a custom fighter type

joeseos:
## Summary

- When editing a fighter in a custom gang, the fighter type dropdown returns `custom_fighter_types` mixed with standard `fighter_types` from `/api/fighter-types`
- The edit modal was blindly writing the selected ID to `fighter_type_id`, which has a FK constraint to `fighter_types` — causing `fighters_fighter_type_id_fkey` violation when a custom type was chosen
- Fix preserves `is_custom_fighter` through both transformation paths in the modal, then branches in `handleConfirm` to write to `custom_fighter_type_id` vs `fighter_type_id` (nulling the other), mirroring the pattern already in `add-fighter.ts`
- The `updateFighterDetails` server action gains a `custom_fighter_type_id` param to support the write

**Regression introduced by a9f085a** (custom gang types feature, Apr 18 2026) — the API was updated to return custom types in the dropdown but the edit modal was never updated to handle them.

## Test plan

- [ ] Edit a fighter in a custom gang and change its type to a custom fighter type — save should succeed without FK error
- [ ] Edit a fighter in a custom gang and switch from a custom type back to a standard type — save should succeed and `custom_fighter_type_id` should be cleared
- [ ] Edit a fighter without changing its type — no regression, type fields unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_015e9UJUJLVfJiEw6HN37kyF)_

### DIFF
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -53,7 +53,8 @@ export interface UpdateFighterDetailsParams {
   fighter_class?: string;
   fighter_class_id?: string;
   fighter_type?: string;
-  fighter_type_id?: string;
+  fighter_type_id?: string | null;
+  custom_fighter_type_id?: string | null;
   fighter_sub_type?: string | null;
   fighter_sub_type_id?: string | null;
   note?: string;
@@ -1221,6 +1222,7 @@ export async function updateFighterDetails(params: UpdateFighterDetailsParams): 
     if (params.fighter_class_id !== undefined) updateData.fighter_class_id = params.fighter_class_id;
     if (params.fighter_type !== undefined) updateData.fighter_type = params.fighter_type;
     if (params.fighter_type_id !== undefined) updateData.fighter_type_id = params.fighter_type_id;
+    if (params.custom_fighter_type_id !== undefined) updateData.custom_fighter_type_id = params.custom_fighter_type_id;
     if (params.fighter_sub_type !== undefined) updateData.fighter_sub_type = params.fighter_sub_type;
     if (params.fighter_sub_type_id !== undefined) updateData.fighter_sub_type_id = params.fighter_sub_type_id;
     if (params.note !== undefined) updateData.note = params.note;
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -142,6 +142,7 @@ export function EditFighterModal({
     fighter_sub_type?: string;
     fighter_sub_type_id?: string;
     available_legacies?: Array<{id: string; name: string}>;
+    is_custom_fighter?: boolean;
   }>>([]);
   const [isLoadingFighterTypes, setIsLoadingFighterTypes] = useState(false);
   const [fighterTypesError, setFighterTypesError] = useState<string | null>(null);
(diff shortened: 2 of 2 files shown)


joeseos:
## PR Review — Fix FK violation when editing fighter to a custom fighter type

> @joeseos — good diagnosis, and the FK crash is real. But the fix has a critical gap that causes the primary save path to silently discard the custom type instead of writing it.

---

### Findings

#### 🔴 HIGH — `custom_fighter_type_id` is never forwarded through the mutation path

**File:** `components/fighter/edit-fighter/fighter-edit-modal.tsx` — lines 251–284

`handleConfirm` correctly sets `submitData.custom_fighter_type_id`, but that field is then passed to `mutation.mutate(submitData)` where it disappears. The `mutationFn` parameter type doesn't include `custom_fighter_type_id`, and the `updateFighterDetails` call inside doesn't pass it:

```ts
// mutationFn parameter — line 251–266
mutationFn: async (submit: {
  ...
  fighter_type_id?: string;   // ← present
  // custom_fighter_type_id is missing entirely
}) => {
  const result = await updateFighterDetails({
    ...
    fighter_type_id: submit.fighter_type_id,  // ← forwarded
    // custom_fighter_type_id: NOT forwarded   // ← dropped
  });
```

**Runtime effect when saving a fighter with a custom type (primary path, `fighter-page.tsx` line 1238 uses `onEditMutate`):**
- `fighter_type_id = null` is written to DB ✅ (FK crash prevented)
- `custom_fighter_type_id` stays unchanged ❌ (was already `null` for standard→custom switch)
- Fighter ends up with **both type IDs null** — type data silently erased

The server action (`edit-fighter.ts` line 1225) and the FK itself are fine; the field just never arrives.

**Fix:** Add `custom_fighter_type_id?: string | null` to the `mutationFn` parameter type and forward it in the `updateFighterDetails` call:

```ts
mutationFn: async (submit: {
  ...
  fighter_type_id?: string | null;
  custom_fighter_type_id?: string | null;
  ...
}) => {
  const result = await updateFighterDetails({
    ...
    fighter_type_id: submit.fighter_type_id,
    custom_fighter_type_id: submit.custom_fighter_type_id,
    ...
  });
```

---

#### 🟡 MEDIUM — `onSubmit` interface also missing `custom_fighter_type_id`

**File:** `components/fighter/edit-fighter/fighter-edit-modal.tsx` — lines 77–91

The legacy `onSubmit` callback type doesn't include `custom_fighter_type_id`. No current caller uses this path (`fighter-page.tsx` uses mutation-only), but the interface is incomplete and will mislead anyone who wires up `onSubmit` in future.

```ts
onSubmit?: (values: {
  fighter_type_id?: string;
  // custom_fighter_type_id missing
}) => Promise<boolean>;
```

---

#### 🟢 LOW — Optimistic update doesn't reflect custom fighter type

**File:** `components/fighter/edit-fighter/fighter-edit-modal.tsx` — lines 322–324

The `onMutate` optimistic builder only updates the `fighter_type` display field when `submit.fighter_type && submit.fighter_type_id` are both truthy. For a custom type, `submit.fighter_type_id` will be `null` (intentionally), so the optimistic state won't update the displayed type. This is cosmetic but means the UI shows a stale value until `onEditSuccess` fires.

---

### Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **HIGH** | `fighter-edit-modal.tsx` L251–284 | `custom_fighter_type_id` dropped inside `mutationFn` — custom type never written to DB |
| 2 | **MEDIUM** | `fighter-edit-modal.tsx` L77–91 | Legacy `onSubmit` interface missing `custom_fighter_type_id` |
| 3 | **LOW** | `fighter-edit-modal.tsx` L322–324 | Optimistic update skips type display for custom fighters |

**Verdict: Changes requested.** The FK-crash fix in `handleConfirm` is correct, and the server-action change is clean. The core problem is that `handleConfirm` sets `submitData.custom_fighter_type_id` but the mutation body never reads or forwards it — so the DB write that is the point of this PR doesn't happen.

---

PR #1653, opened 2026-05-07
Fix: forward custom_fighter_type_id through mutation and fix optimistic update

joeseos:
Follow-up to #1652, addressing three review findings that were missed in the original fix.

## Summary

- **HIGH** — `custom_fighter_type_id` was missing from the `mutationFn` parameter type (lines 251–266) and not forwarded into the `updateFighterDetails` call. The FK crash from #1652 was prevented, but the custom type was silently dropped — the DB write that was the whole point of the change never happened. Added the field to the param type and forwarded it.
- **MEDIUM** — Legacy `onSubmit` interface was also missing `custom_fighter_type_id`, making it incomplete for future callers.
- **LOW** — Optimistic update condition `submit.fighter_type && submit.fighter_type_id` failed for custom types because `fighter_type_id` is intentionally `null`. Now also checks `custom_fighter_type_id` as a signal so the displayed type updates immediately.

## Test plan

- [ ] Edit a fighter in a custom gang, change its type to a custom fighter type — save should succeed and the custom type should be persisted in the DB (`custom_fighter_type_id` set, `fighter_type_id` null)
- [ ] Edit a fighter in a custom gang, switch from a custom type back to a standard type — `fighter_type_id` set, `custom_fighter_type_id` null
- [ ] Confirm the displayed fighter type updates immediately on save (optimistic update) for both cases

---
_Generated by [Claude Code](https://claude.ai/code/session_015e9UJUJLVfJiEw6HN37kyF)_

### DIFF
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -82,7 +82,8 @@ interface EditFighterModalProps {
     fighter_class?: string;
     fighter_class_id?: string;
     fighter_type?: string;
-    fighter_type_id?: string;
+    fighter_type_id?: string | null;
+    custom_fighter_type_id?: string | null;
     special_rules?: string[];
     stats?: Record<string, number>;
     fighter_sub_type?: string | null;
@@ -257,7 +258,8 @@ export function EditFighterModal({
       fighter_class?: string;
       fighter_class_id?: string;
       fighter_type?: string;
-      fighter_type_id?: string;
+      fighter_type_id?: string | null;
+      custom_fighter_type_id?: string | null;
       special_rules?: string[];
       fighter_sub_type?: string | null;
       fighter_sub_type_id?: string | null;
@@ -276,6 +278,7 @@ export function EditFighterModal({
         fighter_class_id: submit.fighter_class_id,
         fighter_type: submit.fighter_type,
         fighter_type_id: submit.fighter_type_id,
+        custom_fighter_type_id: submit.custom_fighter_type_id,
         fighter_sub_type: submit.fighter_sub_type,
         fighter_sub_type_id: submit.fighter_sub_type_id,
         fighter_gang_legacy_id: submit.fighter_gang_legacy_id,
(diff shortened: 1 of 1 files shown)


joeseos:
## Code Review — PR #1653

@joeseos — follow-up review against the full codebase.

---

### Overview

This PR correctly identifies and fixes three distinct gaps left after #1652:

1. `custom_fighter_type_id` was absent from the `mutationFn` inline type and never forwarded to `updateFighterDetails` — meaning the FK column change was type-safe on the surface but the DB write never happened.
2. The legacy `onSubmit` interface was correspondingly missing the field.
3. The optimistic condition `submit.fighter_type_id` short-circuited to falsy for custom types (where the field is intentionally `null`), so the displayed name never updated immediately.

The root cause analysis and framing are accurate.

---

### Findings

#### MEDIUM — Optimistic object omits top-level `custom_fighter_type_id`

**File:** `components/fighter/edit-fighter/fighter-edit-modal.tsx`, line 325–326

```ts
...(submit.fighter_type && (submit.fighter_type_id || submit.custom_fighter_type_id)
  ? { fighter_type: { fighter_type: submit.fighter_type, fighter_type_id: submit.fighter_type_id ?? null } as any }
  : {}),
```

The condition is now correctly triggered for custom types, and the nested `fighter_type` object is built correctly for display purposes. However, the **top-level `custom_fighter_type_id`** is never spread into the optimistic object.

In `fighter-page.tsx:1291`:
```tsx
isCustomFighter={Boolean((fighterData as any)?.fighter?.custom_fighter_type_id)}
```

This reads from `fighter.custom_fighter_type_id` (the flat top-level field), not from the nested `fighter_type` object. During the optimistic window (modal closes → server responds), `fighter.custom_fighter_type_id` still carries the previous value:

- **standard → custom**: `isCustomFighter` stays `false` temporarily — the fighter card may briefly render without custom-type restrictions/behaviours.
- **custom → standard**: `isCustomFighter` stays `true` temporarily — the inverse.

The window is short and cosmetic (server success overwrites it), but it's a logical gap in the optimistic patch.

**Suggested fix** — add the field alongside the nested `fighter_type` spread in the `onMutate` optimistic object:

```ts
...(submit.fighter_type && (submit.fighter_type_id || submit.custom_fighter_type_id)
  ? {
      fighter_type: { fighter_type: submit.fighter_type, fighter_type_id: submit.fighter_type_id ?? null } as any,
      custom_fighter_type_id: submit.custom_fighter_type_id ?? null,
      fighter_type_id: submit.fighter_type_id ?? null,
    }
  : {}),
```

---

#### LOW — `fighter_type_id` in `FighterProps` still typed `string | undefined` (no null)

**File:** `types/fighter.ts:134`

```ts
fighter_type_id?: string;
```

The modal's interfaces were updated to `string | null`, and `edit-fighter.ts:56` already accepts `string | null`, but the shared `FighterProps` type still doesn't accept `null`. This is a pre-existing inconsistency (not introduced here), but since this PR touches the same field it's worth noting for a follow-up to keep the types aligned.

---

### What's correct

- `custom_fighter_type_id` forwarded to `updateFighterDetails` (line 281) — this is the actual DB fix and it's done correctly.
- Both inline type blocks updated consistently (lines 86, 262).
- `fighter_type_id?: string | null` is now correct in both interfaces — the `null` case was previously untypable.
- Optimistic trigger condition `(submit.fighter_type_id || submit.custom_fighter_type_id)` — correct; the `&&`-only guard was the right thing to break.
- `updateFighterDetails` in `edit-fighter.ts:1225` already handles `custom_fighter_type_id` via `!== undefined` guard, so the new forwarding will work correctly including the `null` case (explicit null clears the column).

---

### Verdict

**Changes requested** — the core data fix is correct and complete; the medium finding above is a real gap in the optimistic patch that should be addressed before merging to avoid transient incorrect `isCustomFighter` state on the page.

---

PR #1654, opened 2026-05-08
Swapped the drop down in the gang stash tab to a combobox

joeseos:


joeseos:
## PR Review — `feature/changed-to-combobox-in-stash`

@joeseos — solid upgrade overall. The status icons are a nice UX addition and the Combobox integration is clean. A few issues to address before merging:

---

### Findings

#### 🟡 MEDIUM — Accessibility regression: broken `<label htmlFor>`

**File:** `components/gang/stash-tab.tsx:771`

```tsx
<label htmlFor="fighter-select" ...>
```

The original `<select id="fighter-select">` is gone, and the Combobox component doesn't accept or forward an `id` prop to its inner `<input>`. The `htmlFor` association is now broken — clicking the label won't focus the input.

**Fix:** Either pass an `id` prop through the Combobox component to its inner `<input>`, or remove `htmlFor="fighter-select"` from the label (less ideal but won't mislead the browser).

---

#### 🟡 MEDIUM — Options array rebuilt on every render (no `useMemo`)

**File:** `components/gang/stash-tab.tsx:778–822`

The entire `options` array is constructed inline in JSX — including a `.sort()`, multiple `.map()` calls, JSX element creation for each fighter, and a `getAllVehicles()` call. This runs on **every render** of the parent component (e.g. each checkbox tick). For gangs with many fighters this is wasteful, and it also defeats the `useMemo` inside `Combobox` for `filteredOptions` since the `options` reference changes every render.

**Fix:** Hoist the options construction into a `useMemo` that depends on `[fighters, positioning, selectedItems, hasSelectedVehicle, hasVehicleExclusiveItem]`.

---

#### 🔵 LOW — Dead code: `getSelectableFighters`

**File:** `components/gang/stash-tab.tsx:203`

```ts
const getSelectableFighters = () => { ... }
```

This function was presumably used by the old `<select>` but is no longer called anywhere. It can be deleted.

---

#### 🔵 LOW — Duplicate condition in Button `disabled` prop

**File:** `components/gang/stash-tab.tsx:850–851`

```tsx
(hasVehicleExclusiveItem && !selectedFighter.startsWith('vehicle-')) ||
(!selectedFighter.startsWith('vehicle-') && hasVehicleExclusiveItem)  // identical to the line above
```

These two conditions are logically identical (just reordered). The second line can be removed.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Medium | `htmlFor="fighter-select"` label no longer linked to input — accessibility regression |
| 2 | Medium | Options array rebuilt every render — should be `useMemo` |
| 3 | Low | `getSelectableFighters` is dead code |
| 4 | Low | Duplicate condition in Button `disabled` |

---

**Verdict: Changes requested** — Issues 1 and 2 are worth fixing before this lands. Issues 3 and 4 are trivial cleanup. The core combobox migration and status icon additions look correct.

---

PR #1655, opened 2026-05-11
Add trading post editing to admin campaign type management

joeseos:
## Summary

- Admins can now set default trading posts on campaign types from the Campaign Management modal, both when creating new types and editing existing ones
- `GET /api/admin/campaign-types` now returns the `trading_posts` field
- `POST` and `PATCH` endpoints accept and persist `trading_posts` (array of UUIDs or null)
- UI fetches trading post types and shows a **Default Trading Posts** section with add/remove chip controls, consistent with the existing territories and triumphs relationship UI

## Test plan

- [ ] Open Campaign Management modal as admin
- [ ] Select an existing campaign type — verify its current trading posts are pre-loaded as chips
- [ ] Add a trading post via the modal — verify it appears as a chip
- [ ] Remove a trading post chip — verify it's removed
- [ ] Save — verify the `trading_posts` field is persisted correctly in the database
- [ ] Create a new campaign type with trading posts selected — verify they are saved on creation
- [ ] Verify the Add Trading Post button is disabled when all trading post types are already selected

---
_Generated by [Claude Code](https://claude.ai/code/session_01F94uCmHHkvZdGWCcW2gVQE)_

### DIFF
--- a/app/api/admin/campaign-types/route.ts
+++ b/app/api/admin/campaign-types/route.ts
@@ -13,7 +13,7 @@ export async function GET() {
 
     const { data: campaignTypes, error } = await supabase
       .from('campaign_types')
-      .select('id, campaign_type_name, image_url')
+      .select('id, campaign_type_name, image_url, trading_posts')
       .order('campaign_type_name', { ascending: true });
 
     if (error) throw error;
@@ -37,7 +37,7 @@ export async function POST(request: Request) {
     }
 
     const body = await request.json();
-    const { campaign_type_name, image_url } = body;
+    const { campaign_type_name, image_url, trading_posts } = body;
 
     if (!campaign_type_name?.trim()) {
       return NextResponse.json(
@@ -65,11 +65,19 @@ export async function POST(request: Request) {
       }
     }
 
+    if (trading_posts !== undefined && trading_posts !== null && !Array.isArray(trading_posts)) {
+      return NextResponse.json(
+        { error: 'trading_posts must be an array or null' },
+        { status: 400 }
+      );
+    }
+
     const { data: campaignType, error } = await supabase
       .from('campaign_types')
       .insert([{
         campaign_type_name: campaign_type_name.trim(),
-        image_url: image_url?.trim() || null
+        image_url: image_url?.trim() || null,
+        trading_posts: trading_posts ?? null
       }])
       .select()
       .single();
@@ -95,7 +103,7 @@ export async function PATCH(request: Request) {
     }
 
     const body = await request.json();
-    const { id, campaign_type_name, image_url } = body;
+    const { id, campaign_type_name, image_url, trading_posts } = body;
 
     if (!id) {
       return NextResponse.json(
@@ -128,6 +136,15 @@ export async function PATCH(request: Request) {
         }
       }
       updateData.image_url = trimmedUrl;
+    }
+    if (trading_posts !== undefined) {
+      if (trading_posts !== null && !Array.isArray(trading_posts)) {
+        return NextResponse.json(
+          { error: 'trading_posts must be an array or null' },
+          { status: 400 }
+        );
+      }
+      updateData.trading_posts = trading_posts;
     }
 
     if (Object.keys(updateData).length === 0) {
--- a/components/admin/admin-campaign-management.tsx
+++ b/components/admin/admin-campaign-management.tsx
@@ -34,6 +34,11 @@ interface CampaignTriumph {
   updated_at?: string | null;
 }
 
+interface TradingPostType {
+  id: string;
+  trading_post_name: string;
+}
+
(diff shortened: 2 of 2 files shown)


joeseos:
## PR Review — Add trading post editing to admin campaign type management

@joeseos — solid feature overall. The implementation follows existing patterns well and the API validation is reasonable for an admin-only endpoint. A few things to address before merge:

---

### Findings

---

#### 1. `isLoadingTradingPostTypes` not included in the `isLoading` composite — **Medium**

**File:** `components/admin/admin-campaign-management.tsx`

```tsx
// line ~112 (PR branch)
const { data: tradingPostTypes = [] } = useQuery<TradingPostType[]>({
  // isLoading is never destructured here
  ...
});

const isLoading = isLoadingCampaignTypes || isLoadingTerritories || isLoadingTriumphs || isSubmitting;
//                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                isLoadingTradingPostTypes is missing
```

While the "Add Trading Post" button happens to be disabled by coincidence while `tradingPostTypes` is empty (the filter returns length 0), this isn't intentional. Other form controls that gate on `isLoading` (e.g. the campaign type dropdown, the name input) won't correctly reflect the loading state.

**Fix:**
```tsx
const {
  data: tradingPostTypes = [],
  isLoading: isLoadingTradingPostTypes,
} = useQuery<TradingPostType[]>({ ... });

const isLoading = isLoadingCampaignTypes || isLoadingTerritories || isLoadingTriumphs || isLoadingTradingPostTypes || isSubmitting;
```

---

#### 2. No server-side existence check for `trading_posts` UUID elements — **Medium**

**File:** `app/api/admin/campaign-types/route.ts` — POST and PATCH handlers

The API validates that `trading_posts` is an array, but never checks that each element actually corresponds to a row in `trading_post_types`. If the column is `uuid[]`, Postgres will reject malformed UUIDs but will silently accept valid-looking UUIDs that don't exist. This means an admin could save stale or fabricated IDs.

For an admin-only endpoint this is a limited risk, but it means the UI could end up showing a raw UUID fallback (`tradingPost?.trading_post_name ?? id`) for entries that don't resolve — which is the fallback already coded in the chip renderer on the frontend.

**Fix suggestion:** Add a lightweight existence check after the array-type validation:
```ts
if (trading_posts?.length) {
  const { data: validIds } = await supabase
    .from('trading_post_types')
    .select('id')
    .in('id', trading_posts);
  if (!validIds || validIds.length !== trading_posts.length) {
    return NextResponse.json(
      { error: 'One or more trading_posts IDs are invalid' },
      { status: 400 }
    );
  }
}
```

---

#### 3. Redundant/overly complex visibility condition — **Low**

**File:** `components/admin/admin-campaign-management.tsx` — Default Trading Posts section

```tsx
{(isCreateModeCampaignType || (selectedCampaignTypeId && !isCreateModeCampaignType)) && (
```

`(selectedCampaignTypeId && !isCreateModeCampaignType)` is never true when `isCreateModeCampaignType` is already true, so the inner negation is redundant. Simplifies to:

```tsx
{(isCreateModeCampaignType || selectedCampaignTypeId) && (
```

---

#### 4. `invalidateAllData` doesn't include trading post types cache — **Low**

**File:** `components/admin/admin-campaign-management.tsx`

```tsx
const invalidateAllData = () => Promise.all([
  queryClient.invalidateQueries({ queryKey: ['admin-campaign-types'] }),
  queryClient.invalidateQueries({ queryKey: ['admin-territories'] }),
  queryClient.invalidateQueries({ queryKey: ['admin-campaign-triumphs'] }),
  // ['admin-trading-post-types'] is missing
]);
```

Trading post types don't change during campaign type operations, so this isn't a correctness bug — but if the same query key is ever reused elsewhere it may lead to stale cache. Worth adding for consistency.

---

### What works well

- API validation structure (`undefined` / `null` / non-array guards) is consistent between POST and PATCH and follows the pattern of the existing `image_url` validation.
- `trading_posts` is stored/sent as `null` rather than `[]` when empty — consistent with the DB column semantics.
- All relevant state resets (`handleCategoryChange`, `handleCreateNewCampaignType`, post-submit cleanup) correctly include `relatedTradingPostIds`.
- The `TradingPostType` interface is correctly scoped locally rather than polluting the shared `types/campaign.ts` (which already has `CampaignType` updated with the `trading_posts` field).
- The "Add Trading Post" button being disabled when all types are already selected is a nice UX touch (matches the test plan item).

---

Items 1 and 2 should be addressed before merge. Item 1 is a straightforward one-liner fix. Item 2 is optional to implement fully but at minimum worth a comment acknowledging the tradeoff. Items 3 and 4 are low-priority cleanup.

---

PR #1656, opened 2026-05-12
Fixed lower bound of sell value

OsGrGreen:


joeseos:
## PR Review — Fixed lower bound of sell value

@OsGrGreen — the intent is right and the roll logic is correctly updated, but the fix is incomplete in two meaningful ways. Details below.

---

### Findings

---

#### 1. CSS class typo on the hint paragraph — **Medium**

**File:** `components/fighter/fighter-equipment-list.tsx`, line 146

```tsx
<p className="text-xs text-mut  ed-foreground mt-1">Minimum 5 credits</p>
//                       ^^^^^^^^
//                   double space splits into two invalid classes
```

`text-mut  ed-foreground` (two spaces) is parsed by Tailwind/the browser as two separate class tokens: `text-mut` and `ed-foreground`. Neither exists, so the paragraph renders in the default text colour instead of muted. The user hint still appears, but unstyled.

**Fix:**
```tsx
<p className="text-xs text-muted-foreground mt-1">Minimum 5 credits</p>
```

---

#### 2. Minimum is not enforced on manual edit or on confirm — **Medium**

**File:** `components/fighter/fighter-equipment-list.tsx`

The `Math.max(5, ...)` guard is only applied inside `handleRoll`. The number input has no such guard, and neither does the confirm callback:

```tsx
// input min is still 0, not 5
<input
  type="number"
  value={manualCost}
  onChange={(e) => setManualCost(Number(e.target.value))}
  min={0}   // ← should be 5
/>

// confirm sends whatever manualCost holds, including values < 5
onConfirm={() => { onConfirm(Number(manualCost) || 0); return true; }}
//                                                  ^^^
//                       0 || 0 = 0, so selling for 0 is still possible
```

A user can roll (getting ≥ 5), then manually edit the field down to 0, and the sale goes through at 0 credits — or skip rolling entirely and confirm at the original cost, which could be below 5 for cheap items.

**Fix — enforce at both the input and confirm layers:**
```tsx
<input
  type="number"
  value={manualCost}
  onChange={(e) => setManualCost(Math.max(5, Number(e.target.value)))}
  className="w-full p-2 border rounded-md"
  min={5}
/>
```
```tsx
onConfirm={() => { onConfirm(Math.max(5, Number(manualCost) || 5)); return true; }}
```

---

#### 3. IIFE in JSX duplicates logic from `handleRoll` — **Low**

**File:** `components/fighter/fighter-equipment-list.tsx`, lines 123–127

```tsx
{(() => {
  const raw = originalCost - lastRoll * 10;
  const final = Math.max(5, raw);
  return `Roll ${lastRoll}: -${lastRoll * 10} → ${final} credits`;
})()}
```

`manualCost` is already set to the same `Math.max(5, ...)` value in `handleRoll`, so this IIFE recomputes what's already in state. It also creates a subtle risk: if these two expressions ever diverge (e.g. someone changes one but not the other), the roll summary text won't match the value in the input.

Can be simplified back to an inline expression, or — better — just read `manualCost` directly since it's already been clamped:

```tsx
{`Roll ${lastRoll}: -${lastRoll * 10} → ${manualCost} credits`}
```

(This only holds after the roll sets `manualCost`; if the user then edits manually the summary would update — which may or may not be desired. If you want the summary to always show the roll result, keep the IIFE but note the duplication risk.)

---

### What works well

- The `Math.max(5, ...)` clamp in `handleRoll` is correctly placed and the toast message reflects the clamped value.
- `Math.max(5, ...)` is consistently applied in both the state update (`setManualCost`) and the roll summary display in this version.

---

Finding #2 is the key issue — the lower bound advertised by the UI label isn't actually enforced at the data layer. Finding #1 is a quick typo fix. Finding #3 is optional cleanup.

---

PR #1661, opened 2026-05-15
Allows to select a gang for the lasting injury Bitter Enmity

Djidiouf:


joeseos:
## PR Review — Bitter Enmity Gang Tracking

Hey @Djidiouf, solid feature overall — the defense-in-depth validation (TS action + SQL RPC) and the optimistic update with colour swatch are nice touches. A few issues to address before merge.

---

### 🔴 HIGH — Wrong label in `injuryAggregationLabel`

**File:** `utils/bitterEnmityDisplay.ts:15`

```ts
return `Berserker (${gangName})`;
```

This returns `"Berserker (GangName)"` as the aggregated injury label on the gang card and print view, but the injury is named **"Bitter Enmity"** everywhere else in the codebase (DB, RPC checks, all other display logic). A user seeing "Berserker (GangName) x1" next to their other injuries is going to be confused — it should read "Bitter Enmity (GangName)".

**Fix:**
```ts
return `Bitter Enmity (${gangName})`;
```

Unless "Berserker" is intentionally meant to surface the name of the *skill* the injury grants — in which case that intent should be clearly documented, and even then it's inconsistent with the fact that the skills list already shows the Berserker skill separately with its own badge.

---

### 🟡 MEDIUM — Redundant DB queries before RPC call

**File:** `app/actions/fighter-injury.ts:~180–215`

When `bitterGangId` is set, the TS server action runs **two extra Supabase queries** (fetching `campaign_gangs` for both gangs, then checking intersection) before calling the RPC. The SQL function already performs the exact same campaign-membership validation atomically inside a transaction.

This means:
- 2 extra round trips on every Bitter Enmity submission.
- A theoretical TOCTOU window: the gang could leave the campaign between the TS check and the SQL execution (the SQL check is the authoritative one anyway).

The TS-side checks are pure redundancy here. They can be removed — the RPC error will still bubble up with a meaningful message if validation fails.

---

### 🟡 MEDIUM — `'Bitter Enmity'` string literal scattered across 5 files (14 occurrences)

The string `'Bitter Enmity'` appears hardcoded in:
- `app/actions/fighter-injury.ts`
- `app/lib/shared/fighter-data.ts`
- `app/lib/shared/gang-data.ts`
- `components/fighter/fighter-injury-list.tsx` (8 times alone)
- `utils/bitterEnmityDisplay.ts`

A rename of the effect in the DB would silently break all display and validation logic. A shared constant (e.g. `BITTER_ENMITY_EFFECT_NAME = 'Bitter Enmity'` in `utils/bitterEnmityDisplay.ts` or a constants file) would be far safer.

---

### 🟡 MEDIUM — Optimistic update shows empty gang name on cache miss

**File:** `components/fighter/fighter-injury-list.tsx:~185`

```ts
const g = campaignGangs.find((row) => row.id === bitterId);
bitterMeta = {
  bitter_enmity_target_gang_id: bitterId,
  bitter_enmity_target_gang_name: g?.name ?? '',   // ← empty string if not found
  bitter_enmity_target_gang_colour: g?.gang_colour ?? null
};
```

If the gang isn't found in the local `campaignGangs` state (e.g. the list was cleared between selection and submission), the optimistic update renders a badge with an empty name. It recovers once the server response arrives, but it's a visible flash of broken UI.

**Fix:** Guard the optimistic spread with `g` being defined, or don't apply the optimistic update for the badge if the gang can't be resolved locally.

---

### 🟡 MEDIUM — Gang picker can be skipped for non-campaign fighters

**File:** `components/fighter/fighter-injury-list.tsx:1050`

```tsx
{selectedInjury?.effect_name === 'Bitter Enmity' && campaignIds.length > 0 && (
```

If `campaignIds` is empty (fighter not in any campaign), no picker is shown and submission proceeds without a `bitter_enmity_target_gang_id`. The SQL RPC allows this (it only validates when the parameter is `NOT NULL`). This may be intentional for skirmish-mode fighters, but it means a Bitter Enmity injury can be recorded with no enemy gang tracked, with no warning to the user.

If this is intentional, a short inline note ("In skirmish play, no gang selection is required") would help clarify the UX.

---

### 🟠 LOW — Unintentional formatting change in skills list

**File:** `components/fighter/fighter-skills-list.tsx:591,598`

```diff
- ({item.injury_name || 'Lasting Injury'})
+ {item.injury_name || 'Lasting Injury'}
- (Advancement)
+ Advancement
```

The parentheses around "Lasting Injury" and "Advancement" were removed. This is unrelated to the Bitter Enmity feature and changes the visual appearance of *all* injury-derived and advancement-derived skills. If intentional, it should be a separate commit/PR — as-is it's an accidental side effect that makes the diff harder to review.

---

### ✅ What's done well

- **SQL RPC validation is correct** — same-gang check, campaign membership JOIN, `INTO STRICT` for the gang lookup, and cascading `DROP FUNCTION` cleanup are all solid.
- **Reset paths are thorough** — `selectedBitterEnmityGangId` is cleared on close, injury-type change, and all three submission branches (captured/not-captured/equipment flow).
- **`addInjuryBlockedByBitterEnmityGang` memo** correctly blocks the submit button in the inline form *and* the modal confirm button.
- **Colour swatch in the combobox** is a nice UX touch.
- **Migration file** is present alongside the updated function definition.

---

The `Berserker` vs `Bitter Enmity` label mismatch (HIGH) is a user-facing correctness bug that needs fixing before merge. The MEDIUM items are worth addressing in this same PR while the context is fresh.


---

joeseos:
## Follow-up Review — Bitter Enmity Gang Tracking (commit 2)

@Djidiouf Nice turnaround — the second commit addresses every HIGH and MEDIUM item from my first pass. Here's the updated assessment.

---

### ✅ All previously raised issues resolved

| Prior issue | Status |
|---|---|
| 🔴 `Berserker` label → should be `Bitter Enmity (GangName)` | ✅ Fixed — `injuryAggregationLabel` now returns `\`${BITTER_ENMITY_EFFECT_NAME} (${gangName})\`` |
| 🟡 Redundant TS-side campaign validation (extra DB queries) | ✅ Removed — `addFighterInjury` now passes `bitterGangId` straight to the RPC |
| 🟡 `'Bitter Enmity'` string literal scattered across 5 files | ✅ `BITTER_ENMITY_EFFECT_NAME` constant added in `bitterEnmityDisplay.ts`, imported and used in all TS files (`fighter-injury-list`, `fighter-skills-list`, `fighter-data`, `gang-data`, `dice`, `lastingInjuryRank`) |
| 🟡 Optimistic update shows empty gang name badge on cache miss | ✅ Fixed — `bitterMeta` is now only set when `g` is found locally (`if (g) { ... }`) |
| 🟡 Non-campaign fighters — no warning when skipping gang selection | ✅ Fixed — "In skirmish play, no gang selection is required." inline message added |

---

### 🟠 LOW — Parentheses still removed from unrelated skill labels

**File:** `components/fighter/fighter-skills-list.tsx:589,597`

```diff
-  ({item.injury_name || 'Lasting Injury'})
+  {item.injury_name || 'Lasting Injury'}

-  (Advancement)
+  Advancement
```

This visual change was in the first commit and remains unchanged. It affects the appearance of **all** injury-derived and advancement-derived skills, not just Bitter Enmity. It's still unrelated to this feature. If it's intentional cleanup, it deserves a mention in the PR description; otherwise it should be reverted to keep the diff focused.

---

### 🟢 TRIVIAL — `fighter_effect_types` query always fires now

**File:** `app/actions/fighter-injury.ts:156–168`

Previously the `fighter_effect_types` query was inside `if (!shouldSetKilled)`. It's now unconditional (hoisted out), meaning one extra DB round-trip when `set_killed: true` is passed. This is negligible in practice and the code is cleaner for it — just worth being aware of.

---

### ✅ Verdict: **Approve**

The feature is correct, the validation is solid (SQL RPC owns the authoritative checks, TS is a clean pass-through), the constant refactor is thorough, and the remaining LOW item is cosmetic. Ship it.


---

PR #1666, opened 2026-05-18
Polished the map features

Djidiouf:


joeseos:
## PR Review — "Polished the map features" (@Djidiouf)

Overall this is a clean, well-scoped set of improvements: the unified `getLandmarkTerritoryNameOffset` helper eliminates a magic-number branch, the territory sort is correct and consistently applied, and the palette slot wrapper is a nice visual polish. A few issues worth addressing before merge.

---

### 🔴 HIGH — XSS in `buildTooltipContent` (`campaign-map-canvas.tsx:98–148`)

`territory.description`, `territory.territory_name`, and `playing_card` are interpolated directly into an HTML string that Leaflet sets via `innerHTML`:

```ts
// line 98 – description is unescaped
const desc = territory.description
  ? `<div ...>${territory.description}</div>`   // ← raw user input
  : '';

// line 102 / 148 – territory_name and card are also raw
return `<div ...>${card}${territory.territory_name}</div>...${desc}`;
```

Leaflet's `bindTooltip` sets content as `innerHTML`, so a territory description or name containing ``&lt;img src=x onerror=...&gt;`` would execute. This was pre-existing in the allocated branch; the PR extends it to the unallocated branch without fixing it.

Contrast with the new `buildTerritoryTooltipHtml` in `campaign-map-gang-summary.tsx` — that function correctly escapes everything via `escapeHtml`. The inconsistency means two tooltip surfaces have opposite security postures.

**Fix:** apply the same `escapeHtml` (or a shared utility) to `territory.territory_name`, `playing_card`, and `territory.description` inside `buildTooltipContent`.

---

### 🟡 MEDIUM — `buildPaletteMarkerHtml` will silently break for future icon types (`map-markers.ts:116–131`)

The switch-case hardcodes every current icon key. When a new entry is added to `MARKER_ICONS`, the `default` branch falls back to:

```ts
default:
  return MARKER_ICONS[iconKey]?.html(colour) ?? '';
```

…which returns the raw, unwrapped HTML — no `paletteSlotHtml` wrapper. The icon will appear in the palette with wrong dimensions rather than erroring, making it easy to miss.

Two options:
1. Add a comment in `MARKER_ICONS` (or the switch) warning developers to keep them in sync.
2. Move the per-icon palette logic into the `MarkerIconDef` itself (a `paletteHtml?: (colour: string) => string` field) so there's a single source of truth and no separate switch to maintain.

---

### 🟡 MEDIUM — Tooltip offset mutation may not visually update on zoom (`campaign-map-canvas.tsx:~605`, `campaign-map-editor-modal.tsx:986–990`)

```ts
const refreshRelativeLandmark = () => {
  marker.setIcon(makeIcon(isSelected ? HIGHLIGHT_COLOUR : colour));
  if (nameLayer) {
    nameLayer.options.offset = getTerritoryNameOffset();  // ← direct mutation
    nameLayer.setLatLng(geo.latlng);                      // relies on this triggering re-position
  }
};
```

Mutating `options.offset` directly and then calling `setLatLng` on the same latlng works in current Leaflet because `_setPosition` reads `this.options.offset` during `_updatePosition`. But this is relying on an internal execution path rather than a public API. If Leaflet ever caches the offset at bind time this will silently stop working. Calling `nameLayer.update()` after mutating the offset would be more explicit and resilient.

---

### 🟢 LOW — `escapeHtml` doesn't escape single quotes (`campaign-map-gang-summary.tsx:26–32`)

```ts
function escapeHtml(str: string): string {
  return str
    .replace(/&/g, '&amp;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;')
    .replace(/"/g, '&quot;');
    // missing: .replace(/'/g, '&#39;')
}
```

Safe for the current usages (text content and double-quoted attributes), but incomplete for any future use in single-quoted attribute contexts. Worth adding `'` → `&#39;` for correctness.

---

### 🟢 LOW — `LANDMARK_TERRITORY_NAME_GAP` sign convention vs. comment (`map-markers.ts:31–32`)

```ts
/** Gap (px) between the bottom of a landmark icon box and the territory name label. */
export const LANDMARK_TERRITORY_NAME_GAP = -6;
```

A negative value here means the label _overlaps_ the icon, not that there is a gap below it. The comment implies a positive gap. Either the comment or the sign convention should be clarified (e.g. "negative = overlap" or rename to `LANDMARK_TERRITORY_NAME_OVERLAP`).

---

### ✅ What's good

- `getLandmarkTerritoryNameOffset` centralises the per-icon offset logic cleanly; no more scattered `isReactIcon ? 6 : 10` magic numbers.
- `getZoomScale` closure in `buildObjectEntry` avoids duplicating the zoom-scale expression across `makeIcon` and `getTerritoryNameOffset` — good DRY.
- `sortedTerritories` memo in the editor and the per-gang sort in the summary produce consistent UX across all territory list surfaces.
- Event listener cleanup is correctly updated to use `refreshRelativeLandmark` everywhere (no stale reference leak).
- `buildTerritoryTooltipHtml` in gang-summary properly escapes all user-controlled content — good pattern, just needs to be applied to the canvas tooltip too.

---

**Verdict: changes requested** — the XSS in `buildTooltipContent` and the palette switch fragility are worth addressing before merge. The rest is low-risk polish.


---

joeseos:
## Follow-up Review — updated push (@Djidiouf)

Good turnaround. All five issues from the first round are resolved:

| # | Issue | Status |
|---|---|---|
| 🔴 | XSS in `buildTooltipContent` | ✅ Fixed — `escapeHtml` applied to `territory_name`, `playing_card`, and `description` |
| 🟡 | `buildPaletteMarkerHtml` switch fragility | ✅ Fixed — `paletteHtml` moved into `MarkerIconDef`; switch-case eliminated |
| 🟡 | Tooltip offset mutation / `setLatLng` risk | ✅ Fixed — both canvas and editor now call `nameLayer.update()` after mutating `options.offset` |
| 🟢 | `escapeHtml` missing `'` | ✅ Fixed — `&#039;` added |
| 🟢 | `LANDMARK_TERRITORY_NAME_GAP` comment ambiguity | ✅ Fixed — "Negative number = overlap" clarification added |

### One minor note

`escapeHtml` now lives in `utils/campaigns/map-markers.ts`, which is otherwise a Leaflet/icon-specific module. It works fine there (both consumers are in the same campaign-map domain), but if this utility ever needs to be called from outside the map subsystem it will feel odd to import from `map-markers`. Not a blocker — just worth keeping in mind if the file grows.

---

**Verdict: ✅ Approve** — the PR is clean, all security and maintainability concerns are addressed, and the palette/offset improvements are solid.


---

PR #1669, opened 2026-05-18
Add dev-only quick login buttons to sign-in page

prvlaar:
- Adds "Dev: User" and "Dev: Admin" buttons to the sign-in form, visible only when NODE_ENV=development (i.e. localhost)
- Buttons fill and submit the form automatically using credentials from .env.local
- No effect on production — the conditional is compiled out by Next.js at build time
- Documents the required env vars (NEXT_PUBLIC_DEV_USER_EMAIL, etc.) in .env.example so other developers know to set their own

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,3 +8,9 @@ DISCORD_CLIENT_SECRET=your-discord-client-secret
 DISCORD_BOT_TOKEN=your-discord-bot-token
 NEXT_PUBLIC_DISCORD_CLIENT_ID=your-discord-client-id
 NEXT_PUBLIC_APP_URL=http://localhost:3000
+
+# Dev-only quick login (localhost only, never shown in production)
+NEXT_PUBLIC_DEV_USER_EMAIL=your-test-user@example.com
+NEXT_PUBLIC_DEV_USER_PASSWORD=your-test-user-password
+NEXT_PUBLIC_DEV_ADMIN_EMAIL=your-admin@example.com
+NEXT_PUBLIC_DEV_ADMIN_PASSWORD=your-admin-password
--- a/app/sign-in/page.tsx
+++ b/app/sign-in/page.tsx
@@ -97,6 +97,20 @@ export default function SignIn() {
     // No return value needed here (void)
   }
 
+  function devLogin(role: 'user' | 'admin') {
+    const email = role === 'admin'
+      ? process.env.NEXT_PUBLIC_DEV_ADMIN_EMAIL!
+      : process.env.NEXT_PUBLIC_DEV_USER_EMAIL!;
+    const password = role === 'admin'
+      ? process.env.NEXT_PUBLIC_DEV_ADMIN_PASSWORD!
+      : process.env.NEXT_PUBLIC_DEV_USER_PASSWORD!;
+    const emailInput = document.getElementById('email') as HTMLInputElement;
+    const passwordInput = document.getElementById('password') as HTMLInputElement;
+    if (emailInput) emailInput.value = email;
+    if (passwordInput) passwordInput.value = password;
+    emailInput?.closest('form')?.requestSubmit();
+  }
+
   return (
     <main className="flex min-h-screen flex-col items-center">
       <div className="container mx-auto max-w-4xl w-full p-4">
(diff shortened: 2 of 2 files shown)


joeseos:
## Code Review — PR #1669: Add dev-only quick login buttons

Hey @prvlaar — thanks for the QoL improvement. The idea is sound but there are a couple of security issues that need addressing before this merges.

---

### Findings

---

#### 🔴 HIGH — Passwords stored in `NEXT_PUBLIC_*` env vars (`.env.example` + `app/sign-in/page.tsx`)

**Files:** `.env.example` lines 13–16, `app/sign-in/page.tsx` lines 100–106

The `NEXT_PUBLIC_` prefix tells Next.js to **inline these values into the client-side JavaScript bundle at build time** — that is the literal purpose of that prefix. This means any developer who sets `NEXT_PUBLIC_DEV_USER_PASSWORD` to a real credential will have that password:

- Visible to anyone who opens DevTools → Sources
- Readable by any browser extension
- Extractable from the compiled JS bundle with `grep`
- Exposed to any XSS payload

While dead-code elimination *should* strip the block in a production build (since `NODE_ENV === 'development'` compiles to `false`), relying on the minifier to scrub passwords from a bundle is not a defence you want to depend on, and it offers zero protection in development where the credentials are still live in the browser.

**Fix:** Route this through a server action or a dedicated API route that reads non-prefixed env vars (`DEV_USER_EMAIL`, `DEV_USER_PASSWORD`) server-side and performs the sign-in there, returning a session. The client never needs to hold the password at all. Example sketch:

```ts
// app/actions/dev-login.ts  (only compiled in dev)
"use server";
export async function devLoginAction(role: 'user' | 'admin') {
  if (process.env.NODE_ENV !== 'development') return { error: 'Forbidden' };
  const email    = role === 'admin' ? process.env.DEV_ADMIN_EMAIL    : process.env.DEV_USER_EMAIL;
  const password = role === 'admin' ? process.env.DEV_ADMIN_PASSWORD : process.env.DEV_USER_PASSWORD;
  // call signInAction or supabase directly
}
```

The button then calls this server action — no credentials ever touch the browser.

---

#### 🔴 HIGH — Credentials exposed in bundle even when buttons are not rendered

**File:** `app/sign-in/page.tsx` lines 100–106

`devLogin()` is defined unconditionally at module scope. In a production build where `NEXT_PUBLIC_DEV_*` vars happen to be set (e.g. a developer who has them in their shell and runs `next build`), webpack's `DefinePlugin` substitutes the literal values before Terser runs dead-code elimination. The optimiser *usually* removes the dead block, but this is not a contractual guarantee and varies between Next.js/webpack versions.

**Fix:** The server-action approach above eliminates this entirely, since non-`NEXT_PUBLIC_` vars are never inlined into the bundle.

---

#### 🟡 MEDIUM — Turnstile CAPTCHA is bypassed

**File:** `app/sign-in/page.tsx` line 113 (`requestSubmit()`)

`devLogin` calls `requestSubmit()` directly without waiting for the Turnstile widget to complete. If the server-side `signInAction` validates the Turnstile token (which is the entire point of having it), the dev login will either fail silently or trigger an error. Even if it doesn't fail today, it creates a codepath that bypasses a security control.

**Fix:** Either have `TurnstileWidget` expose a method/ref to programmatically complete it in test mode, or have the server action skip Turnstile verification when `NODE_ENV === 'development'` (server-side check only, never client-side).

---

#### 🟡 MEDIUM — Silent failure / missing guard on undefined env vars

**File:** `app/sign-in/page.tsx` lines 98–113

If a developer hasn't set the env vars, `process.env.NEXT_PUBLIC_DEV_USER_EMAIL!` is `undefined`. The `!` non-null assertion suppresses TypeScript's warning, so the form submits with `emailInput.value = undefined` → the string `"undefined"` as the email address. No feedback is shown.

Also, if `emailInput` is non-null but `passwordInput` is null, the password guard (`if (passwordInput)`) prevents it from being set, but `emailInput?.closest('form')?.requestSubmit()` still fires the form with a blank password.

**Fix:**
```ts
function devLogin(role: 'user' | 'admin') {
  const email    = role === 'admin' ? process.env.NEXT_PUBLIC_DEV_ADMIN_EMAIL    : process.env.NEXT_PUBLIC_DEV_USER_EMAIL;
  const password = role === 'admin' ? process.env.NEXT_PUBLIC_DEV_ADMIN_PASSWORD : process.env.NEXT_PUBLIC_DEV_USER_PASSWORD;
  if (!email || !password) {
    console.warn('[dev] Missing dev login env vars — check .env.local');
    return;
  }
  // ...
}
```

---

#### 🟢 LOW — `devLogin` function is always compiled into the module

**File:** `app/sign-in/page.tsx` line 100

The `NODE_ENV` guard only wraps the JSX buttons, not the function definition. `devLogin` is dead code in production. Not a bug, but it's tidier (and more clearly intentional) to co-locate the guard or move the helper inside the conditional render.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 High | Passwords in `NEXT_PUBLIC_*` vars — baked into client bundle |
| 2 | 🔴 High | Bundle exposure not reliably eliminated by dead-code optimiser |
| 3 | 🟡 Medium | Turnstile CAPTCHA bypassed by `requestSubmit()` |
| 4 | 🟡 Medium | Silent failure when env vars are unset; inconsistent null guards |
| 5 | 🟢 Low | `devLogin` always compiled, only buttons are guarded |

---

The usability goal here is good and worth having. The fix is to move credential handling to a server action (non-`NEXT_PUBLIC_` vars) so passwords never reach the browser. Happy to discuss the approach if helpful.

---

joeseos:
Closing this. Not needed

PR #1670, opened 2026-05-18
Add Downtime option for Spyrer rig glitch removal

prvlaar:
- Adds a Downtime button to the Clear Rig Glitches modal as an alternative to the kill-cost path: costs 100 credits from the gang stash, leaves kill count unchanged
- Disabled if the gang has fewer than 100 credits; shows live credit preview (X → X-100) in the modal
- Clears the kill count requirement from the existing Clear All button — Spyrers can now use it regardless of kills
- Logs the action to the gang log as "Downtime glitch removal"
- Adds secondary confirm button support to the shared Modal component


joeseos:
## PR Review — #1670 · Add Downtime option for Spyrer rig glitch removal

@prvlaar — solid feature addition overall. The UI flow is clean and the modal extension is minimal and reusable. A few issues need addressing before this merges, one of which is a data-integrity bug.

---

### 🔴 HIGH — Non-atomic operation: glitches deleted before credits are deducted

**File:** `app/actions/fighter-injury.ts`, lines 573–586

The action deletes every glitch in a serial loop, *then* deducts credits. If `updateGangFinancials` fails (DB error, race condition, etc.) all glitches are already gone but the stash is untouched — free glitch removal.

The mutation's `onError` only performs a UI rollback (restoring local component state) — it does not undo the database deletes.

```ts
// Glitches deleted here (lines 573–578)
for (const injuryId of params.glitch_ids) {
  const result = await deleteFighterInjury(...);
  ...
}

// Credits deducted AFTER — too late to roll back if this throws
const financialResult = await updateGangFinancials(supabase, { creditsDelta: -100, ... });
```

**Suggested fix:** Validate that the gang has ≥ 100 credits *before* touching any injury records. A pre-flight credits check won't eliminate all races but removes the most common failure path:

```ts
// Before the deletion loop
const { data: gang } = await supabase
  .from('gangs').select('credits').eq('id', fighter.gang_id).single();
if (!gang || gang.credits < 100) {
  return { success: false, clearedCount: 0, error: 'Insufficient credits' };
}
// then delete glitches, then deduct
```

Ideally this whole operation (delete effects + update credits) lives in a single Postgres transaction/RPC to be truly atomic.

---

### 🟡 MEDIUM — No server-side `is_spyrer` validation

**File:** `app/actions/fighter-injury.ts`, line 552

`clearRigGlitchesDowntime` only checks authentication. It does not verify the target fighter is actually a Spyrer. Any authenticated user can call this action on a non-Spyrer fighter and pay 100 credits to bulk-clear their lasting injuries — bypassing the game rule that this is a Spyrer-only mechanic.

The fighter query already touches the `fighters` table; adding `is_spyrer` to the select and gating on it is trivial:

```ts
.select('id, gang_id, fighter_name, is_spyrer')
...
if (!fighter.is_spyrer) {
  return { success: false, clearedCount: 0, error: 'Fighter is not a Spyrer' };
}
```

---

### 🟡 MEDIUM — No server-side guard for empty `glitch_ids`

**File:** `app/actions/fighter-injury.ts`, line 573

If `glitch_ids` is an empty array the loop runs zero times, then 100 credits are deducted and a log entry is written claiming "0 rig glitches cleared". The UI guards this with `injuries.length === 0` on the button, but there is no server-side check.

```ts
if (params.glitch_ids.length === 0) {
  return { success: false, clearedCount: 0, error: 'No glitches to clear' };
}
```

---

### 🟡 MEDIUM — Optimistic UI rollback is incomplete on error

**File:** `components/fighter/fighter-injury-list.tsx`, lines 508–527

`onMutate` immediately calls `onInjuryUpdate([], ...)` to clear the injuries list in the UI. If the server call fails, `onError` tries to restore `previousInjuries` via `onInjuryUpdate`. However, if server-side glitch deletions partially succeeded before the error (see HIGH issue above), the UI is now showing re-instated injuries that no longer exist in the DB.

This is secondary to fixing the atomicity issue, but worth noting: the rollback only works correctly if the server operation is fully atomic (either all succeed or none do).

---

### 🔵 LOW — Secondary modal button has no internal loading state

**File:** `components/ui/modal.tsx`, lines 129–140

The primary confirm button wraps `onConfirm` in `setIsSubmitting(true/false)` to show a loading state. The secondary button relies entirely on the externally-passed `secondaryConfirmDisabled`. This works here because `clearGlitchesDowntimeMutation.isPending` is threaded through, but the modal component has no self-contained loading feedback for the secondary action. If a future caller omits the prop there will be no visual indication of in-flight state. Consider adding a parallel `isSecondarySubmitting` state inside the modal.

---

### 🔵 LOW — Kill-cost display shows negative kills when `kill_count` is 0

**File:** `components/fighter/fighter-injury-list.tsx`, line 1497

This line is pre-existing, but the PR makes it newly visible: removing the `kill_count < 1` guard from the "Clear all" button means a Spyrer with 0 kills can now open the modal. With `kill_count = 0` and the default `clearAllKillCost = 4`, the modal shows **"New kills: -4"**. The `confirmDisabled` correctly blocks submission, but the display is confusing.

```tsx
// Current
New kills: <strong>{kill_count - clearAllKillCost}</strong>

// Suggested
New kills: <strong>{Math.max(0, kill_count - clearAllKillCost)}</strong>
```

---

### 🔵 LOW — Missing newline at end of file

**File:** `app/actions/fighter-injury.ts` (end of file)

The diff ends with `\ No newline at end of file`. Minor but will trigger linters in most CI setups.

---

## Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 HIGH | `fighter-injury.ts:573` | Glitches deleted before credits deducted — partial failure = free clear |
| 2 | 🟡 MEDIUM | `fighter-injury.ts:552` | No `is_spyrer` server-side validation |
| 3 | 🟡 MEDIUM | `fighter-injury.ts:573` | Empty `glitch_ids` deducts 100 credits for 0 glitches |
| 4 | 🟡 MEDIUM | `fighter-injury-list.tsx:508` | Optimistic rollback doesn't match DB state if partial delete succeeds |
| 5 | 🔵 LOW | `modal.tsx:129` | Secondary button has no internal `isSubmitting` state |
| 6 | 🔵 LOW | `fighter-injury-list.tsx:1497` | Negative kill display when `kill_count` is 0 |
| 7 | 🔵 LOW | `fighter-injury.ts` EOF | Missing newline at end of file |

**Verdict: ⛔ Changes requested**

The HIGH atomicity issue and the two MEDIUM server-side validation gaps need to be addressed before merge. Everything else is straightforward to fix. The overall structure (extending the modal, the mutation pattern, the log-type registration) is clean and follows existing conventions well.

---

joeseos:
## PR Review — Pass 2 · commit `098baccc`

@prvlaar — great turnaround. Every issue from the first pass has been addressed. Here is the follow-up assessment.

---

### ✅ All previous findings resolved

| # | Was | Now |
|---|-----|-----|
| 🔴 HIGH — glitches deleted before credits checked | Pre-flight credits gate added before the deletion loop | ✅ Fixed |
| 🟡 MEDIUM — no `is_spyrer` server-side guard | `is_spyrer` added to fighter query; early-return if false | ✅ Fixed |
| 🟡 MEDIUM — empty `glitch_ids` deducts credits | Length guard returns error before any DB work | ✅ Fixed |
| 🟡 MEDIUM — optimistic rollback mismatch (connected to atomicity) | Risk window drastically reduced by the pre-flight check | ✅ Effectively fixed |
| 🔵 LOW — no `isSecondarySubmitting` in Modal | Own state added; shows "Processing..." text | ✅ Fixed |
| 🔵 LOW — negative kill display when `kill_count` is 0 | `Math.max(0, ...)` applied | ✅ Fixed |
| 🔵 LOW — missing EOF newline | Newline added | ✅ Fixed |

---

### Remaining notes (non-blocking)

**TOCTOU window is narrow but still exists** — `app/actions/fighter-injury.ts`

The pre-flight credits read and the actual `updateGangFinancials` deduction are still two separate DB round-trips, with the deletion loop in between. In a multi-tab scenario (e.g. equipment purchase in Tab A concurrent with Downtime in Tab B), credits could drop below 100 between the pre-flight check and the update. `updateGangFinancials` has its own `credits + delta < 0` guard that would correctly reject the deduction in that case — but the glitches would already be deleted from the DB while the mutation surface returns failure.

In practice this is a very low-probability event given single-user gang ownership. A true fix would require a Postgres stored procedure / RPC to wrap the delete + debit in one transaction. Worth tracking as a future hardening task, but not a blocker here.

**`gangCredits` preview can show stale value** — `components/fighter/fighter-page.tsx:956`

The credit figure shown in the modal comes from the initial page load. If the user spent credits elsewhere in the same session (without a full refresh), the preview will be wrong. Again non-blocking — consistent with how the rest of the page handles financials — but something to revisit if the live-credit pattern becomes a priority.

---

### Verdict: ✅ Approve

The logic is correct for all practical scenarios, security guards are in place server-side, the modal extension is clean and doesn't break any existing callers, and the new log type is properly registered. The remaining notes are future-hardening items, not correctness bugs at realistic load.

---

joeseos:
## PR Review — Pass 3 · @prvlaar

The implementation is structurally sound and fits the codebase's conventions well. Two small cleanups needed before merge.

---

### ✅ What's done right

- **Server action** (`app/actions/fighter-injury.ts:550–617`): uses `getAuthenticatedUser` + RLS for ownership, matching the pattern in the rest of `fighter-injury.ts` (see `deleteFighterInjury` line 354: *"Authorization is enforced by RLS policies"*). Pre-checks `is_spyrer` and `gang.credits >= 100` before any destructive work. Re-uses `deleteFighterInjury` per-glitch rather than duplicating kill/resurrection/rating-recovery logic — same approach as `clearAllGlitchesMutation` client-side. Deduction via `updateGangFinancials({ creditsDelta: -100, applyToRating: false })` happens **after** the deletions, matching the order in `chem-alchemy.ts`: if the loop fails partway, credits are not charged.

- **Client mutation** (`fighter-injury-list.tsx:499–531`): proper `onMutate` snapshot + `onError` rollback, mirrors `clearAllGlitchesMutation`. Credits < 100 disables the button and shows a red hint.

- **Modal extension** (`components/ui/modal.tsx`): `onSecondaryConfirm` / `secondaryConfirmText` / `secondaryConfirmDisabled` added cleanly, with their own `isSecondarySubmitting` state and cross-disable coupling so neither button fires while the other is in flight.

---

### ⛔ Changes requested

#### 1. Missing `userPermissions.canEdit` guard on the Downtime button

**`components/fighter/fighter-injury-list.tsx` ~line 1525**

`secondaryConfirmDisabled` checks credits and pending state but not `canEdit`. Every other destructive control in this file has that guard:

- "Clear all" open button (line 1169): `disabled={… || !userPermissions.canEdit || …}`
- Per-row delete button (line 1181): `disabled={… || !userPermissions.canEdit}`

Because the "Clear all" button (which opens the modal) already gates on `canEdit`, a view-only user can't reach the Downtime button in practice — but defensive consistency matters, and the server is protected by RLS regardless.

```ts
secondaryConfirmDisabled={
  injuries.length === 0 ||
  gangCredits < 100 ||
  !userPermissions.canEdit ||          // ← add this
  clearGlitchesDowntimeMutation.isPending ||
  clearAllGlitchesMutation.isPending
}
```

---

#### 2. Double-logging: Downtime path emits N+1 logs, kill-cost path emits N

**`app/actions/fighter-injury.ts` ~line 600 · `utils/log-types.ts` line 46**

`deleteFighterInjury` already writes a `logFighterRecovery` entry per deleted glitch (line 488). `clearRigGlitchesDowntime` loops that N times, then additionally writes its own `createGangLog('rig_glitches_cleared_downtime', …)` summary. The existing kill-cost path produces only the N recovery logs — no summary.

**Recommended fix — drop the summary, keep parity with the kill-cost path:**

In `fighter-injury.ts`, remove the `createGangLog` call (~line 600) and the now-unused import at the top:
```ts
// remove this import
import { createGangLog } from './logs/gang-logs';
```

In `utils/log-types.ts`, remove the dead label:
```ts
// remove this entry
'rig_glitches_cleared_downtime': 'Downtime glitch removal',
```

This keeps both paths symmetric: N recovery log entries, no summary, no new log type to maintain.

> If you'd prefer to keep the summary log deliberately (e.g. for a "what happened this downtime" view), that's also fine — but then confirm the `'rig_glitches_cleared_downtime'` action type renders correctly wherever gang logs are filtered/displayed, and leave the `log-types.ts` entry in place.

---

### 🔵 Non-blocking notes

**Stale `gangCredits` preview** — `gangCredits` comes from the initial page load snapshot. If credits change elsewhere in the session, the modal hint can drift from reality. The server re-validates via `updateGangFinancials` and `onError` rolls back the UI, so this is **safe** — just a cosmetic wart. The kill-cost path has the same characteristic, so it's consistent.

**Nits:**
- `app/sign-in/page.tsx` diff is trailing-whitespace noise, unrelated to the feature. Harmless — fine to leave.
- The secondary button in `modal.tsx` uses inline `bg-neutral-900` instead of a `<Button variant="…">`, duplicating the primary button's colour. Not a blocker, but a `variant` would be more idiomatic with the rest of the component.

---

### Pattern-fit summary

| Concern | Codebase pattern | PR follows? |
|---|---|---|
| Auth: `getAuthenticatedUser` + RLS | ✅ yes | ✅ |
| Credits via `updateGangFinancials` after main op | ✅ yes | ✅ |
| Multi-step ops without transactional RPC | ✅ yes (loop + separate update) | ✅ |
| Cache invalidation inside `deleteFighterInjury` | ✅ yes | ✅ inherited per call |
| `useMutation` with optimistic update + rollback | ✅ yes | ✅ |
| `userPermissions.canEdit` gate on destructive UI | ✅ yes | ⚠️ missing on secondary button |
| Logging via `logFighter*` only (no extra summary) | ✅ yes | ⚠️ extra summary log added |

---

### Verdict: ⛔ Changes requested (two small items)

1. Add `!userPermissions.canEdit` to `secondaryConfirmDisabled`.
2. Drop the `createGangLog` summary call (and its import + `log-types.ts` entry) to keep logging parity with the kill-cost path — or consciously keep it and confirm the log type renders correctly end-to-end.

No structural rework needed. Once these two are addressed this is ready to ship.

---

joeseos:
@prvlaar — clarification on the logging fix from the previous review pass.

### Correct approach for the summary log

Replace the `createGangLog` call in `clearRigGlitchesDowntime` with `logFighterAction` (already imported in `fighter-injury.ts` via `./logs/fighter-logs`). This is the function used by financial operations like the resurrection path, and it produces a proper `Credits: X → Y | Wealth: A → B` line via `formatFinancialChanges` — which is what you want here. `updateGangFinancials` already returns both `oldValues` and `newValues`, so the data is right there.

**Step 1 — Add the new action type to the union in `app/actions/logs/fighter-logs.ts` line 13:**
```ts
action_type: ... | 'fighter_kills_changed' | 'fighter_kill_count_changed' | ... | 'rig_glitches_cleared_downtime';
```

**Step 2 — Add a case for it in the `switch` block in the same file (~line 61):**
```ts
case 'rig_glitches_cleared_downtime':
  description = `Fighter "${params.fighter_name}" cleared ${params.old_value} rig glitch${Number(params.old_value) !== 1 ? 'es' : ''} via Downtime.${financialChanges}`;
  break;
```

**Step 3 — Replace the `createGangLog` call in `clearRigGlitchesDowntime` (`app/actions/fighter-injury.ts` ~line 600):**
```ts
// before
await createGangLog({
  gang_id: fighter.gang_id,
  fighter_id: params.fighter_id,
  action_type: 'rig_glitches_cleared_downtime',
  description: `Fighter "${fighter.fighter_name}" cleared ${deletedCount} rig glitch...`,
});

// after
await logFighterAction({
  gang_id: fighter.gang_id,
  fighter_id: params.fighter_id,
  fighter_name: fighter.fighter_name,
  action_type: 'rig_glitches_cleared_downtime',
  old_value: deletedCount,
  oldCredits: financialResult.oldValues?.credits,
  oldRating:  financialResult.oldValues?.rating,
  oldWealth:  financialResult.oldValues?.wealth,
  newCredits: financialResult.newValues?.credits,
  newRating:  financialResult.newValues?.rating,
  newWealth:  financialResult.newValues?.wealth,
});
```

**Step 4 — Remove the now-unused `createGangLog` import** from `app/actions/fighter-injury.ts` line 12.

The `'rig_glitches_cleared_downtime'` entry in `utils/log-types.ts` can stay — it's still used as the display label wherever the gang log is rendered.

---

This will produce a single log entry per downtime operation that looks like:

> Fighter "Zyx-9" cleared 2 rig glitches via Downtime.
> Credits: 250 → 150 | Wealth: 800 → 700

Same structure as the resurrection log, same helper, no new logging machinery needed.

---

joeseos:
## PR Review — Pass 4 · @prvlaar

Reviewing commits since Pass 3 (`6cd59fc`, `8fa8d0ac`, `a4e7da76`).

---

### ✅ Pass 3 findings — all resolved

| # | Was | Now |
|---|-----|-----|
| ⛔ Missing `!userPermissions.canEdit` on Downtime button | `secondaryConfirmDisabled` lacked the guard | ✅ Fixed in `6cd59fc` |
| ⛔ Double-logging / wrong log helper | `createGangLog` dropped; `logFighterAction` wired with `old/newValues` from `updateGangFinancials` | ✅ Fixed across `6cd59fc` + `8fa8d0ac` |

---

### 🟡 MEDIUM — Duplicate `case 'rig_glitches_cleared_downtime'` in switch statement

**File:** `app/actions/logs/fighter-logs.ts`

Commit `8fa8d0ac` (joeseos) added the first correct case. Commit `a4e7da76` (@prvlaar) then added a second case for the same label, after `fighter_copied`. The result in the current file:

```ts
case 'rig_glitches_cleared_downtime':            // ← first case (joeseos) — correct, always matches
  description = `Fighter "…" cleared ${params.old_value} rig glitch… via Downtime.${financialChanges}`;
  break;
case 'fighter_copied':
  …
  break;
case 'rig_glitches_cleared_downtime': {          // ← second case (prvlaar) — dead code, never reached
  const count = Number(params.old_value ?? 0);
  description = `Fighter "…" cleared ${count} rig glitch… via Downtime (-100 credits).${financialChanges}`;
  break;
}
```

The second case is unreachable — JS/TS switch statements always match the first label. Beyond being dead code, ESLint's `no-duplicate-case` rule will flag this as an error and likely fail the build.

The first case is also the better one: it avoids the redundant `(-100 credits)` string since `financialChanges` already carries the credit delta via `formatFinancialChanges`.

**Fix:** Delete the entire second `case 'rig_glitches_cleared_downtime'` block (the one that comes after `fighter_copied`).

---

### 🔵 LOW — Trailing newline dropped from two files

Both regressions were introduced in commit `8fa8d0ac`:

- **`app/actions/fighter-injury.ts`** — the new `clearRigGlitchesDowntime` function at end of file lacks a trailing newline. This re-introduces the issue that was explicitly fixed in `098baccc`.
- **`utils/log-types.ts`** — `main` has a trailing newline; `8fa8d0ac` removed it.

One-line fix in each file: add a newline after the closing `}`.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🟡 MEDIUM | `fighter-logs.ts` | Duplicate `case 'rig_glitches_cleared_downtime'` — dead code, ESLint error |
| 2 | 🔵 LOW | `fighter-injury.ts` | Missing trailing newline (regression from `8fa8d0ac`) |
| 3 | 🔵 LOW | `log-types.ts` | Missing trailing newline (regression from `8fa8d0ac`) |

**Verdict: ⛔ Changes requested**

One small fix needed: remove the duplicate switch case in `fighter-logs.ts`. The two newline regressions are trivially fixed in the same commit. The rest of the implementation is clean and correct — this is the last thing standing between this PR and a merge.

---

joeseos:
@prvlaar — applied the three fixes from Pass 4 directly to the branch (commit `f53e7d4`):\n\n- Removed the duplicate `case 'rig_glitches_cleared_downtime'` block (the dead-code one after `fighter_copied`)\n- Restored the missing trailing newline in `fighter-injury.ts`\n- Restored the missing trailing newline in `utils/log-types.ts`\n\nThe branch is clean. **✅ Approved** — ready to merge.\n\n---\n_Generated by [Claude Code](https://claude.ai/code)_

---

prvlaar:
Was anything else needed on this one?

PR #1673, opened 2026-05-19
Added private note for gangs

Djidiouf:


joeseos:
## Code Review — PR #1673: Added private note for gangs

Hey @Djidiouf, thanks for the PR! The feature idea is solid and the implementation is mostly clean, but there is one **critical security gap** that must be fixed before merge, plus a couple of smaller items.

---

### 🔴 HIGH — `note_private` is readable by any authenticated user

**Files:** `supabase/migrations/20260519100000_add_private_note_to_gangs.sql`, `app/api/gangs/[id]/route.ts`

The existing gangs RLS SELECT policy is:
```sql
CREATE POLICY "Allow authenticated users to view gangs"
  ON public.gangs FOR SELECT TO authenticated USING (true);
```
`USING (true)` means every authenticated user can read **all columns** of every gang row. The migration adds `note_private` and `note_private_updated_at` but adds no compensating policy, so the private note is immediately readable by any logged-in user via:

1. **Direct Supabase API** — the policy grants full row access.
2. **`GET /api/gangs/[id]`** — the handler runs `select('*')` with no auth check and returns the full row to anyone authenticated (lines 18–22 of `route.ts`).

The page server component (`app/gang/[id]/page.tsx:128`) correctly gates the data behind `userPermissions.canEdit`, but that only protects the Next.js rendered page. The raw API endpoint leaks the field.

**Required fixes (pick at least one layer):**

*Layer 1 — API route (minimum fix):* Exclude `note_private` from the GET response for any caller that isn't the owner/arbitrator, or simply never return it from this endpoint (it isn't used by any client-side caller of this route — the page uses `getGangBasic` directly server-side):
```ts
// In GET handler, after fetching gangData:
const { note_private, note_private_updated_at, ...publicGangData } = gangData;
return NextResponse.json({ gang: { ...publicGangData, gang_variants: variantDetails, campaigns } });
```

*Layer 2 — Database (defense in depth):* PostgreSQL doesn't support column-level RLS, but you can create a security definer function / view, or—more practically—add an explicit application-level filter in the Supabase query so `note_private` is never fetched in public paths. The `getGangBasic` function already does the right thing (it selects specific columns), but if you later add another `select('*')` path you'll re-expose it.

---

### 🟡 MEDIUM — `handleNotePrivateUpdate` in parent doesn't sync the timestamp

**File:** `components/gang/gang-page-content.tsx:389–398`

```ts
const handleNotePrivateUpdate = useCallback((updatedNotePrivate: string) => {
  setGangData((prev) => ({
    ...prev,
    processedData: {
      ...prev.processedData,
      note_private: updatedNotePrivate
      // note_private_updated_at is NOT updated here
    }
  }));
}, []);
```

The save response sets `notePrivateUpdatedAt` in local state inside `notes-tab.tsx`, which is fine while the component stays mounted. But `initialNotePrivateUpdatedAt` (which comes from the parent's `processedData`) stays stale. If something causes the child to remount (e.g. tab switching, future refactors), the displayed timestamp will revert to the pre-save value.

**Suggested fix:** Thread the updated timestamp through the callback:
```ts
// notes-tab.tsx — pass timestamp up
onNotePrivateUpdate?.(notePrivate, data.note_private_updated_at);

// gang-page-content.tsx — accept it
const handleNotePrivateUpdate = useCallback((updatedNotePrivate: string, updatedAt?: string) => {
  setGangData((prev) => ({
    ...prev,
    processedData: {
      ...prev.processedData,
      note_private: updatedNotePrivate,
      note_private_updated_at: updatedAt ?? prev.processedData.note_private_updated_at,
    }
  }));
}, []);
```

---

### 🟡 MEDIUM — `formatUpdatedAt` has no guard against invalid dates

**File:** `components/gang/notes-tab.tsx:34–41`

```ts
function formatUpdatedAt(isoString: string): string {
  const d = new Date(isoString);
  // No validity check — bad input → "Updated: NaN-NaN-NaN at NaN:NaN"
```

If `note_private_updated_at` ever arrives as an unexpected value (e.g., malformed timestamp from a migration, a manual DB edit, or a future refactor that passes the wrong field), the UI will render garbage. Simple guard:
```ts
function formatUpdatedAt(isoString: string): string {
  const d = new Date(isoString);
  if (isNaN(d.getTime())) return '';
  // ... rest unchanged
}
```

---

### 🟢 LOW — Unrelated toolbar padding change

**File:** `components/ui/rich-text-editor.tsx:514`

```diff
- className={`border-b rounded-t-md p-2 flex flex-wrap ...`}
+ className={`border-b rounded-t-md p-1 flex flex-wrap ...`}
```

This is a global change to the `RichTextEditor` toolbar that affects every editor in the app, not just the private notes editor. If it's intentional and tested, it's fine — but it would be cleaner to split into a separate PR so it doesn't get lost in the review.

---

### ✅ What looks good

- Page server component correctly gates `note_private` behind `userPermissions.canEdit` (line 128 of `page.tsx`) — the intent is right, execution just needs the API layer to match.
- PATCH endpoint authorization is correctly handled by the existing `"Only gang owner, admin, or arbitrator can update"` RLS policy — only owners/admins/arbs can write.
- Migration is clean and minimal.
- `getGangBasic` fetches specific named columns rather than `select('*')`, so it won't silently pick up future sensitive columns.
- UI gating with `{userPermissions?.canEdit && <NoteEditor ... />}` is correct for render-time.
- The `useEffect` sync pattern for `initialNotePrivate` matches the existing pattern for other notes.

---

The RLS / GET endpoint data leak is a blocker — once a user saves a private note it becomes readable by all authenticated users through the API. The other items are secondary but worth addressing before the feature ships.

---

joeseos:
## Code Review — PR #1673 (updated): Added private note for gangs

Hey @Djidiouf, good progress on this update. Two of the three medium findings from the last review are cleanly resolved, but the original critical blocker is still present.

---

### Previous findings status

| # | Finding | Status |
|---|---------|--------|
| 🔴 HIGH | `note_private` readable by any authenticated user via `GET /api/gangs/[id]` | **⚠️ Still outstanding** |
| 🟡 MEDIUM | `handleNotePrivateUpdate` doesn't sync timestamp to parent | **✅ Resolved** — exactly the suggested fix was applied |
| 🟡 MEDIUM | `formatUpdatedAt` has no guard against invalid dates | **✅ Resolved** — `isNaN(d.getTime())` guard added |
| 🟢 LOW | Unrelated toolbar padding change in `rich-text-editor.tsx` | **Not addressed** (acceptable given it's low severity) |

---

### 🔴 HIGH — GET endpoint data leak still present

**File:** `app/api/gangs/[id]/route.ts:18–88`

The new commits added `page.tsx` server-side gating (nice addition):
```ts
note_private: userPermissions.canEdit ? gangBasic.note_private : undefined,
```

But the raw `GET /api/gangs/[id]` endpoint still does `select('*')` and returns the full row to anyone authenticated:
```ts
// line 19-20 — unchanged
.select('*')
// ...
return NextResponse.json({ gang: { ...gangData, gang_variants: variantDetails, campaigns } });
// line 82-88 — gangData still includes note_private
```

Any authenticated user can `fetch('/api/gangs/<id>')` and read `gang.note_private` directly. The page-level gating only protects the Next.js SSR render, not the API.

**Fix** (same suggestion as before — strip the field before returning):
```ts
// app/api/gangs/[id]/route.ts — in the GET handler, before returning
const { note_private, note_private_updated_at, ...publicGangData } = gangData;
return NextResponse.json({
  gang: {
    ...publicGangData,
    gang_variants: variantDetails,
    campaigns,
  }
});
```

The PATCH handler returning the full `updatedGang` row (line 325) is a lesser concern since the caller is the one who just saved — but for consistency you could apply the same strip there too.

---

### New code looks good

- The `description` prop is rendered via JSX text interpolation, not `dangerouslySetInnerHTML` — no XSS risk.
- The empty-state refactor (`isHtmlEffectivelyEmpty` branching) is cleaner than the old ternary.
- `useEffect` sync for `initialNotePrivateUpdatedAt` correctly mirrors the existing pattern.
- The `userPermissions?.canEdit` guard on the `NoteEditor` render is correct and consistently applied.

---

The GET endpoint leak is the only remaining blocker. Everything else in this update is solid — one more fix and this is ready to merge.

---

PR #1683, opened 2026-05-21
Fixed an issue where after removing a custom gang image, the default …

Djidiouf:
…images were not selectable

### DIFF
--- a/components/ui/image-edit-modal.tsx
+++ b/components/ui/image-edit-modal.tsx
@@ -79,11 +79,17 @@ export const ImageEditModal: React.FC<ImageEditModalProps> = ({
 
   if (!isOpen) return null;
 
+  // When selectedDefaultImageIndex is null we visually show index 1 as the fallback (index 0
+  // if only one image exists), so treat null as that index when default images are available.
+  const fallbackDefaultIndex =
+    defaultImageUrls && defaultImageUrls.length > 1 ? 1 : defaultImageUrls && defaultImageUrls.length === 1 ? 0 : null;
+  const effectiveSelectedIndex = selectedDefaultImageIndex ?? fallbackDefaultIndex;
+
   // Check if default image index has changed
-  const defaultImageIndexChanged = 
-    defaultImageUrls && 
-    selectedDefaultImageIndex !== null &&
-    selectedDefaultImageIndex !== currentDefaultImageIndex;
+  const defaultImageIndexChanged =
+    defaultImageUrls &&
+    effectiveSelectedIndex !== null &&
+    effectiveSelectedIndex !== currentDefaultImageIndex;
 
   // Determine if confirm button should be enabled
   const hasImageToSave = image && ((enableCrop && croppedAreaPixels) || !enableCrop);
@@ -111,10 +117,10 @@ export const ImageEditModal: React.FC<ImageEditModalProps> = ({
       await handleRemoveImage();
     } else if (hasImageToSave) {
       await handleSave();
-    } else if (defaultImageIndexChanged && onDefaultImageIndexChange && selectedDefaultImageIndex !== null) {
+    } else if (defaultImageIndexChanged && onDefaultImageIndexChange && effectiveSelectedIndex !== null) {
       setIsSavingDefaultImage(true);
       try {
-        const result = await onDefaultImageIndexChange(selectedDefaultImageIndex);
+        const result = await onDefaultImageIndexChange(effectiveSelectedIndex);
         if (result.success) {
           onClose();
         }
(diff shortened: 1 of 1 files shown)


joeseos:
## PR Review — `image-edit-modal.tsx`

Hey @Djidiouf, solid fix for the broken state after removing a custom gang image. I traced the full flow through `gang.tsx → gang-image-edit-modal.tsx → image-edit-modal.tsx` and the logic is sound. A few findings below.

---

### Root cause — confirmed correct

When a custom image is removed, `currentDefaultImageIndex` is `null`. The old code guarded `defaultImageIndexChanged` with `selectedDefaultImageIndex !== null`, so the confirm button was permanently disabled and no default image could be selected. The `effectiveSelectedIndex` indirection cleanly solves this.

---

### Findings

#### MEDIUM — Pre-selecting index 1 will silently save an unintended value

**File:** `components/ui/image-edit-modal.tsx` — `fallbackDefaultIndex` / `handleConfirm`

When `currentDefaultImageIndex === null` and there are ≥2 images, `fallbackDefaultIndex` is hardcoded to `1`. This immediately sets `defaultImageIndexChanged = true` (`1 !== null`), enabling the confirm button before the user has interacted at all. If the user clicks "Set Image" without navigating, index `1` is written to the DB — not a value they explicitly chose.

The PR title says "allow picking a default image," but the implementation actually *pre-picks* one. For a user who opens the modal intending to upload a new custom image, the path is fine (the `hasImageToSave` branch fires first). But a user who opens the modal and closes it via confirm without touching anything will silently get index 1 saved.

**Suggestion:** Start from index `0` instead of `1` (or keep `null` and only enable the button after the first navigation). Using `0` as the fallback is the least-surprising default for an empty selection.

```ts
// simpler and more predictable
const fallbackDefaultIndex =
  defaultImageUrls && defaultImageUrls.length > 0 ? 0 : null;
```

If index 1 is intentional (e.g. index 0 is reserved/special), that should be documented.

---

#### LOW — Redundant null-guard in `fallbackDefaultIndex` expression

**File:** `components/ui/image-edit-modal.tsx` line ~84

```ts
const fallbackDefaultIndex =
  defaultImageUrls && defaultImageUrls.length > 1 ? 1
  : defaultImageUrls && defaultImageUrls.length === 1 ? 0   // `defaultImageUrls &&` already known truthy here
  : null;
```

The second `defaultImageUrls &&` is unreachable dead code — if `defaultImageUrls.length > 1` is false and we didn't short-circuit on the first `&&`, the value is still truthy. Minor readability issue:

```ts
const fallbackDefaultIndex = !defaultImageUrls
  ? null
  : defaultImageUrls.length > 1 ? 1
  : defaultImageUrls.length === 1 ? 0
  : null;
```

---

#### LOW — `getDisplayDefaultImageEntry` inconsistency with `effectiveSelectedIndex`

**File:** `components/ui/image-edit-modal.tsx` — `getDisplayDefaultImageEntry`

The function still uses `selectedDefaultImageIndex` (not `effectiveSelectedIndex`) for its primary lookup. When `selectedDefaultImageIndex === null`, it falls through to the `fallbackDefaultIndex` path, so the end result is the same — but it means the same "effective index" logic is expressed in two slightly different ways in the same render. Not a bug, just a minor consistency note.

---

### Correctness checks (all pass)

| Scenario | Result |
|---|---|
| `currentDefaultImageIndex = null`, 0 images | Button stays disabled ✓ |
| `currentDefaultImageIndex = null`, 1 image | Shows index 0, button enabled ✓ |
| `currentDefaultImageIndex = null`, N images | Shows index 1, button enabled ✓ |
| `currentDefaultImageIndex = 1`, N images | `effectiveSelectedIndex = 1`, `1 !== 1` → button disabled ✓ |
| `currentDefaultImageIndex = 0` | `0 ?? fallback = 0` (nullish preserves 0), `0 !== 0` → disabled ✓ |
| Navigate prev from null state | Uses `fallbackDefaultIndex` as starting point correctly ✓ |
| Navigate next from null state | Same ✓ |
| `hasImageToSave` takes priority over default index branch | Yes, in `handleConfirm` ✓ |

---

### Verdict: **Approve with suggestions**

The fix correctly unblocks the broken user flow and handles all edge cases. The medium finding is a UX concern (not a regression), and the lows are cosmetic. If index 1 as the default is intentional, a brief comment explaining why would help future readers.

---

joeseos:
## Follow-up Review — commit `0df85fe` — `image-edit-modal.tsx`

Hey @Djidiouf, reviewed the second commit pushed after the previous review. Here's the status of all prior findings and a check on the new changes.

---

### Previous findings

|---|---|---|---|
| 1 | MEDIUM | Pre-selecting index 1 silently saves without user interaction | **Partially addressed** — comment added documenting the intent; behavior deliberately kept |
| 2 | LOW | Redundant `defaultImageUrls &&` null-guard in `fallbackDefaultIndex` | ✅ **Resolved** — restructured to `!defaultImageUrls ? null : ...` |
| 3 | LOW | `getDisplayDefaultImageEntry` used `selectedDefaultImageIndex` instead of `effectiveSelectedIndex` | ✅ **Resolved** — function now consistently uses `effectiveSelectedIndex` |

On finding #1: the added comment (`// When selectedDefaultImageIndex is null we visually show index 1 as the fallback...`) makes the intent clear. Treating this as a deliberate design choice rather than an oversight is reasonable since the previous review verdict was already "approve with suggestions."

---

### New changes (second commit only)

**`handlePreviousImage` / `handleNextImage` — behaviour change, improvement**

```ts
// before
prev === null || prev === 0 ? defaultImageUrls.length - 1 : prev - 1

// after
const current = prev ?? fallbackDefaultIndex ?? 0;
return current === 0 ? defaultImageUrls.length - 1 : current - 1;
```

When `prev` is `null` and `fallbackDefaultIndex = 1` (the typical case with multiple images), the old code would jump to the **last** image on first "previous" press. The new code starts from `1` and correctly goes to `0`. This aligns navigation with what the user sees displayed — no longer a jarring wrap-around. No regression; strictly better.

The terminal `?? 0` in `prev ?? fallbackDefaultIndex ?? 0` is a safe belt-and-suspenders guard for the case where `fallbackDefaultIndex` is `null` when `defaultImageUrls` is an empty array — which can't happen given the `length > 0` guard above, but harmless.

**`getDisplayDefaultImageEntry` — split into two if-statements**

Cleaner. Functionally equivalent to the previous version (the `effectiveSelectedIndex` path now fires where before `selectedDefaultImageIndex` was `null` and fell through — this is correct since we want the display to reflect the effective index). ✅

**No new issues found.** No regressions, no edge-case gaps, no new security or performance concerns.

---

### Verdict: ✅ **Approve**

Both LOWs resolved. The MEDIUM is documented and accepted as a design choice. The second commit is a net improvement with cleaner navigation logic and consistent use of `effectiveSelectedIndex` throughout.

---

PR #1685, opened 2026-05-21
Fixed an issue where after removing a custom gang image, the default …

Djidiouf:
…images were not selectable_ default image 3

### DIFF
--- a/components/ui/image-edit-modal.tsx
+++ b/components/ui/image-edit-modal.tsx
@@ -79,11 +79,14 @@ export const ImageEditModal: React.FC<ImageEditModalProps> = ({
 
   if (!isOpen) return null;
 
-  // When selectedDefaultImageIndex is null we visually show index 1 as the fallback (index 0 for silhouette art
-  // if only one image exists), so treat null as that index when default images are available.
+  // Default image index to display (0 = Silhouette, 1 = Djidiouf, 2 = Carl R Johnston Grey, 3 = Carl R Johnston Colour)
+  const DEFAULT_IMAGE_INDEX = 3;
+
+  // When selectedDefaultImageIndex is null we visually show index DEFAULT_IMAGE_INDEX as the fallback (index 0 for
+  // silhouette art if only one image exists), so treat null as that index when default images are available.
   const fallbackDefaultIndex = !defaultImageUrls
     ? null
-    : defaultImageUrls.length > 1 ? 1
+    : defaultImageUrls.length > DEFAULT_IMAGE_INDEX ? DEFAULT_IMAGE_INDEX
     : defaultImageUrls.length === 1 ? 0
     : null;
   const effectiveSelectedIndex = selectedDefaultImageIndex ?? fallbackDefaultIndex;


joeseos:
## Code Review — PR #1685

> @Djidiouf — thanks for the fix. The bug is real and the intent is correct, but there's one logic gap worth addressing before merge.

---

### Finding 1 — Medium severity | `components/ui/image-edit-modal.tsx:89`

**Logic gap for arrays with 2 or 3 entries**

The new condition is:

```ts
defaultImageUrls.length > DEFAULT_IMAGE_INDEX  // i.e. length > 3  →  index 3
```

This only passes when `defaultImageUrls` has **4 or more** entries. If the array ever has 2 or 3 entries, neither the `> 3` branch nor the `=== 1` branch matches, so `fallbackDefaultIndex` silently becomes `null`. That means no default image is shown and the picker appears empty.

The old code used `> 1` which was safe for any array with ≥ 2 entries. The new code introduces a dead zone for lengths 2–3.

Right now there are exactly 4 images (as the new comment documents), so this never triggers — but it's a latent bug that would surface if a gang type ever had only 2 or 3 default images.

**Suggested fix — clamp to the last valid index instead of a hard equality gate:**

```ts
const fallbackDefaultIndex = !defaultImageUrls || defaultImageUrls.length === 0
  ? null
  : Math.min(DEFAULT_IMAGE_INDEX, defaultImageUrls.length - 1);
```

This always resolves to a valid index for any non-empty array, and still prefers index 3 when 4+ images are present.

---

### Finding 2 — Low severity | `image-edit-modal.tsx:83`

**`DEFAULT_IMAGE_INDEX` is defined inside the render path, after the early-return guard**

```ts
if (!isOpen) return null;

const DEFAULT_IMAGE_INDEX = 3;   // ← defined here
```

A bare literal re-evaluated on every render is harmless, but it reads oddly because it's a pure constant with no dependency on props or state. Hoisting it to module scope (outside the component) would make its nature clear and save a tiny allocation per render.

---

### Finding 3 — Low severity | `image-edit-modal.tsx:82`

**Comment enumerates specific image names — will rot**

```ts
// Default image index to display (0 = Silhouette, 1 = Djidiouf, 2 = Carl R Johnston Grey, 3 = Carl R Johnston Colour)
```

If images are reordered, added, or renamed in the DB, this comment will be wrong without any compiler warning. The image metadata already lives in `DefaultImageEntry` / `credit`; duplicating the ordering here adds maintenance burden. Not a blocker, but worth removing or replacing with a link to where the data is defined.

---

### Summary

| 1 | `image-edit-modal.tsx:89` | `length > 3` gate creates a null fallback for 2–3 image arrays | **Medium** |
| 2 | `image-edit-modal.tsx:83` | Constant belongs at module scope, not inside render | Low |
| 3 | `image-edit-modal.tsx:82` | Hard-coded image names in comment will rot | Low |

---

The fix correctly addresses the reported bug for the current 4-image case. The medium finding (length 2–3 → `null` fallback) is a latent regression path that's cheap to fix now with a `Math.min` clamp. The low findings are optional but would improve robustness.

---

PR #1699, opened 2026-05-22
Removed old attacker_id and defender_id in table campaign_battles

joeseos:


joeseos:
## Code Review — PR #1699

> @joeseos — solid clean-up overall. The direction is right and most of the code changes are a clear improvement. Two issues need addressing before merge.

---

### Finding 1 — High severity | `supabase/migrations/20260522120000_drop_attacker_defender_from_campaign_battles.sql`

**Destructive migration with no backfill — old records lose attacker/defender data permanently**

The migration drops `attacker_id` and `defender_id` immediately:

```sql
ALTER TABLE campaign_battles
  DROP COLUMN IF EXISTS attacker_id,
  DROP COLUMN IF EXISTS defender_id;
```

There is no migration step to first backfill `participants` for rows that were created before the `participants` column was populated. Any battle row that has `attacker_id`/`defender_id` set but NULL `participants` (i.e., battles logged before the battle-sessions feature landed) will silently lose its attacker/defender information — forever.

Checking the migration history confirms there's no prior backfill migration for `campaign_battles.participants`.

All the application-level fallbacks on `battle.attacker_id`/`battle.defender_id` have also been removed in this same PR, so even if the data were still in memory it couldn't be displayed.

**Suggested fix — add a backfill step at the top of the migration before the DROP:**

```sql
-- Backfill participants from attacker_id/defender_id for rows where participants is NULL
UPDATE campaign_battles
SET participants = json_build_array(
  json_build_object('gang_id', attacker_id, 'role', 'attacker'),
  json_build_object('gang_id', defender_id, 'role', 'defender')
)::text
WHERE participants IS NULL
  AND (attacker_id IS NOT NULL OR defender_id IS NOT NULL);

-- Now safe to drop
ALTER TABLE campaign_battles
  DROP COLUMN IF EXISTS attacker_id,
  DROP COLUMN IF EXISTS defender_id;
```

If you've already confirmed through prod data inspection that every `campaign_battles` row has `participants` populated, please leave a comment in the migration to document that — otherwise this is a data loss risk.

---

### Finding 2 — Medium severity | `app/actions/campaigns/[id]/battle-logs.ts:39`

**`JSON.parse` inside `getAttackerDefenderFromParticipants` is unguarded**

```ts
function getAttackerDefenderFromParticipants(participants: BattleParticipant[] | string) {
  const parsed = typeof participants === 'string' ? JSON.parse(participants) : participants;
  ...
}
```

`JSON.parse` throws synchronously on malformed input. This helper is called inside `createBattleLog` and `updateBattleLog` — both server actions. A bad `participants` string from the DB would bubble up as an uncaught exception.

Every other call site that parses `participants` in this codebase uses try/catch (e.g., `get-campaign-data.ts`, `campaign-tab.tsx`). This helper should do the same.

**Suggested fix:**

```ts
function getAttackerDefenderFromParticipants(participants: BattleParticipant[] | string) {
  try {
    const parsed = typeof participants === 'string' ? JSON.parse(participants) : participants;
    const attacker = parsed?.find((p: BattleParticipant) => p.role === 'attacker');
    const defender = parsed?.find((p: BattleParticipant) => p.role === 'defender');
    return { attacker_id: attacker?.gang_id ?? null, defender_id: defender?.gang_id ?? null };
  } catch {
    return { attacker_id: null, defender_id: null };
  }
}
```

---

### Finding 3 — Low severity | `app/lib/campaigns/[id]/get-campaign-data.ts`

**`participants` is JSON-parsed twice per battle**

Once during the `gangIdSet` building loop and again during the `.map()` to build attacker/defender objects. Not a bug, and battle lists are unlikely to be huge, but it's redundant work. A simple fix would be to store `parsed` on the first pass and reuse it.

---

### Finding 4 — Low severity | `app/api/campaigns/battles/route.ts:71`

**Validation now allows single-participant battles**

The old guard required both `attacker_id` and `defender_id`, effectively enforcing two participants. The new guard:

```ts
if (!scenario || !participants || !Array.isArray(participants) || participants.length === 0)
```

...only requires a non-empty array, so a battle with a single participant now passes validation. If that's intentional (e.g. to support "none"-role multi-player battles), fine — but worth a comment confirming the intent.

---

### Summary

| 1 | Migration SQL | No backfill → old rows permanently lose attacker/defender | **High** |
| 2 | `battle-logs.ts:39` | Unguarded `JSON.parse` in shared helper | **Medium** |
| 3 | `get-campaign-data.ts` | Participants parsed twice per battle | Low |
| 4 | `route.ts:71` | Validation no longer requires two participants | Low |

---

The code simplification and removal of the legacy dual-ID pattern is the right call. Finding 1 (migration safety) is the blocker — it needs either a backfill step or explicit confirmation that prod data is clean. Finding 2 is a quick defensive fix. Everything else is fine to merge as-is after those two are addressed.

---

joeseos:
## Code Review — Second Pass

Thanks for the responses. Verified each one against the branch (`743b68b` is still the only commit).

---

### Finding 1 — Migration backfill ✅ Accepted

Confirmed: prior backfill ran, both columns were nullable, no data loss risk. Migration is safe as written.

---

### Finding 2 — Unguarded `JSON.parse` ❌ Not yet pushed

The fix is described but the code on the branch is still:

```ts
function getAttackerDefenderFromParticipants(participants: BattleParticipant[] | string) {
  const parsed = typeof participants === 'string' ? JSON.parse(participants) : participants;
  ...
}
```

No try/catch. Needs to be committed and pushed before this can be closed.

One note when you do push: because `BattleLogParams.participants` is typed `BattleParticipant[]` (not `string`), the string-branch of this helper is actually dead code at both call sites in `createBattleLog` and `updateBattleLog`. The try/catch is still good defensive practice, but you could also simplify the signature to only accept `BattleParticipant[]` at those call sites and skip the string path entirely.

---

### Finding 3 — Double parsing ❌ Not yet pushed

The Map fix is described but the branch still parses `participants` twice per battle in `get-campaign-data.ts` — once in the `gangIdSet` loop and again in the `.map()` that builds `attackerId`/`defenderId`. Needs to be committed and pushed.

---

### Finding 4 — Single-participant validation ❌ Not yet pushed

The route still has `participants.length === 0`, not `< 2`. Needs to be committed and pushed.

---

### New observation — Low severity | `app/api/campaigns/battles/route.ts:111–112`

The API route has its own inline attacker/defender derivation instead of calling `getAttackerDefenderFromParticipants`:

```ts
const attacker_id = participants.find((p: any) => p.role === 'attacker')?.gang_id;
const defender_id = participants.find((p: any) => p.role === 'defender')?.gang_id;
```

Minor duplication — not a blocker, but if you export the helper you could reuse it here to keep the logic in one place.

---

### Summary

| # | Status | Notes |
|---|--------|-------|
| 1 — Migration backfill | ✅ Resolved | Confirmed safe via developer context |
| 2 — Unguarded `JSON.parse` | ❌ Not pushed | Still bare `JSON.parse` in branch |
| 3 — Double parsing | ❌ Not pushed | Still two parse blocks in `get-campaign-data.ts` |
| 4 — Validation guard | ❌ Not pushed | Still `length === 0`, not `< 2` |
| 5 — Inline duplicate in route | Low | Not a blocker |

---

Three fixes are described but none have landed in the branch. Push findings 2–4 and this is good to merge.

---

joeseos:
## Code Review — Third Pass (post-fixes)

Verified commit `b603206` against all open findings.

---

### Finding 1 — Migration backfill ✅ Accepted
Confirmed safe per prior context.

### Finding 2 — Unguarded `JSON.parse` ✅ Fixed
`getAttackerDefenderFromParticipants` now wraps in try/catch and returns `{ attacker_id: null, defender_id: null }` on failure. Consistent with the rest of the codebase.

### Finding 3 — Double parsing ✅ Fixed
`parsedParticipantsMap` is populated in the `gangIdSet` pass and looked up by `battle.id` in the `.map()`. Single parse per battle, clean implementation.

### Finding 4 — Validation guard ✅ Fixed
Route now rejects requests with fewer than 2 participants (`participants.length < 2`). Correct.

---

### Remaining low-severity note (not a blocker)

`app/api/campaigns/battles/route.ts` still has its own inline attacker/defender derivation rather than calling `getAttackerDefenderFromParticipants`. Minor duplication — worth tidying up if you ever touch that route again, but not worth holding the PR for.

---

### Verdict: **Approve** ✅

All required changes are in. The refactor is clean, the dual-ID columns are safely dropped, and the participants-as-source-of-truth approach is consistently applied across the codebase.

---

PR #1700, opened 2026-05-23
Refactor Advancement workflow for skill selection for non-gangers

Djidiouf:


joeseos:
## PR Review — Refactor Advancement workflow for skill selection for non-gangers

@Djidiouf — solid refactor overall. The new ordering (Skill Set → Acquisition Type → Skill/Roll) is a clear UX improvement and the extraction of all the inline IIFE logic into well-named `useMemo`/`useCallback` hooks is a big readability win. A few issues worth addressing before merging.

---

### Findings

#### 🟡 Medium — `skillAccess` race condition produces a false "not accessible" warning

**File:** `components/fighter/fighter-advancement-list.tsx`

`fetchCategories` (tracked by `loading`) and `fetchSkillAccess` run in parallel when the user picks `advancementType = 'skill'`. The skill-set combobox only appears after `loading` flips to `false` (i.e., categories are done). If `skillAccess` hasn't resolved yet at that point — which is plausible on a slower connection — and the user immediately selects a skill set, `skillAccess` is still `[]`. That means `selectedSkillSetAccess` returns `null` for every set, and the amber warning _"This Skill Set is not accessible to this fighter"_ fires for a valid set before disappearing when the access data lands.

**Fix:** Track a separate `skillAccessLoading` flag and suppress the warning while it is `true`.

---

#### 🟡 Medium — `'allowed'` access only maps to `any_random` — no `any_selected`

**File:** `components/fighter/fighter-advancement-list.tsx` — `getAcquisitionTypeIdsForAccess`

```ts
case 'allowed':
  return ['any_random'];
```

`ALL_SKILL_ACQUISITION_TYPE_IDS` has no `any_selected`, confirming this is intentional. But please add a comment here (or in the constant definition) to document that "Allowed" access is intentionally random-only. Without it, future readers will assume this is an oversight and try to "fix" it.

---

#### 🟡 Medium — Random acquisition type shows Skill Combobox alongside Roll button

**File:** `components/fighter/fighter-advancement-list.tsx` (JSX)

When `isRandomAcquisitionType === true`, both the **Roll for skill** button and the skill `Combobox` are rendered simultaneously. A user can bypass the dice roll entirely and directly select a skill. In the Ganger flow, the same pattern exists (select after roll), but there the roll is gated by a server-logged mutation. Here it is purely client-side, so there's no enforcement.

This may be an intentional safety valve for correcting roll mistakes — but if it is, it's worth a comment to say so. If it's unintentional, the combobox should be hidden until a roll has been made (`skillRollResult !== null`).

---

#### 🟡 Medium — No warning for denied/inaccessible sets in `SkillModal`

**File:** `components/fighter/fighter-skills-list.tsx`

The advancement modal (`fighter-advancement-list.tsx`) shows an amber warning when a fighter lacks access to the selected skill set. The skill modal (`fighter-skills-list.tsx`) does not — it just grays out the label but lets the user proceed silently. Given that both components went through the same `select → Combobox` migration in this PR, it's easy to add the same guard for consistency.

---

#### 🔵 Low — Inconsistent group label names between the two components

**Files:** both changed files

`fighter-advancement-list.tsx`:  `'Universal Skills'` / `'Gang-specific Skills'`  
`fighter-skills-list.tsx`:  `'Universal Skill Sets'` / `'Gang-specific Skill Sets'`

The two modals are shown side-by-side in the fighter UI. Inconsistent group labels will confuse users. Pick one convention and apply it to both.

---

#### 🔵 Low — `ALL_SKILL_ACQUISITION_TYPE_IDS` is a manual hardcoded list

**File:** `components/fighter/fighter-advancement-list.tsx`

```ts
const ALL_SKILL_ACQUISITION_TYPE_IDS = [
  'primary_random',
  'primary_selected',
  'secondary_random',
  'secondary_selected',
  'any_random'
] as const;
```

If a new acquisition type is ever added to the DB, this list must be updated manually or the type silently disappears from the "no access / denied" fallback. A comment linking this to the DB table would go a long way.

---

#### 🔵 Low — `fetchAvailableAdvancements` gives no visual feedback while loading

After selecting a skill set, the Acquisition Type combobox won't appear until `fetchAvailableAdvancements` completes (it depends on `availableAdvancements`). There's no spinner or skeleton in this gap — the UI simply appears frozen. The old flow had the same trait, so not a regression, but now that the acquisition type is the *first* interactive step after the set, the silent pause is more noticeable.

---

### What's good

- `effectiveSkillAccess` correctly normalises `denied` → `null` and the JSDoc makes the intent clear.
- Moving skill-types fetching to `/api/skill-types?fighterId=...` correctly handles campaign-shared custom types; the route is properly authenticated.
- `rollSkillFromSet` correctly derives XP/credits from `editableXpCost`/`editableCreditsIncrease` (already set by `applyAcquisitionTypeSelection`), so the submission path is consistent.
- Disabled group-header items in the Combobox (`__group__*`) cannot be selected — `handleSelect` guards on `isDisabled` and the button has `disabled={option.disabled}`.
- Dependency arrays on all `useMemo`/`useCallback` hooks are correct.
- No SQL injection or auth bypass risks identified.

---

The race condition (#1) and the silent random-bypass (#3) are the ones I'd most want resolved before merge. The rest are quality-of-life improvements. Happy to re-review once those are addressed.


---

joeseos:
## Re-review — commit `9e221cb` (Refactor Advancement workflow, second push)

@Djidiouf — great follow-up push. Five of the seven previous findings are now resolved, and the new code (gang-financial snapshots in skill logs, the `verifyAndLogRolledSkillAdvancementRoll` server action, and the Ganger-class guard) is solid. Two prior issues remain open, and the new commit introduces one small dead-parameter to clean up.

---

| # | Finding | Status |
|---|---------|--------|
| 1 | `skillAccess` race condition — false "not accessible" warning on slow connections | ✅ **Resolved** — `skillAccessLoading` flag added to both `fighter-advancement-list.tsx` and `fighter-skills-list.tsx`; warning is suppressed while the fetch is in flight. |
| 2 | `'allowed'` maps to `any_random` only — undocumented | ✅ **Resolved** — comprehensive block comment added above `ALL_SKILL_ACQUISITION_TYPE_IDS` (linking to the SQL function) and an inline comment inside `case 'allowed':` |
| 3 | Random acquisition type shows Skill Combobox alongside Roll button | ⚠️ **Outstanding** — `<Combobox>` is still rendered unconditionally. A user can still bypass the dice roll by directly selecting from the combobox without ever clicking "Roll for skill". |
| 4 | No access warning in `SkillModal` (`fighter-skills-list.tsx`) | ✅ **Resolved** — `selectedSkillSetLacksAccess` and the amber warning are now present, matching the advancement modal. |
| 5 | Inconsistent group label names between the two components | ✅ **Resolved** — both files now use `'Universal Skill Sets'` / `'Gang-specific Skill Sets'`. |
| 6 | `ALL_SKILL_ACQUISITION_TYPE_IDS` manual list with no DB reference | ✅ **Resolved** — full comment added, explicitly naming `get_available_skills.sql` and the `available_acquisition_types` array. |
| 7 | No loading feedback while `fetchAvailableAdvancements` runs | ⚠️ **Outstanding** — no spinner or skeleton added; the UI still silently freezes between skill-set selection and the acquisition-type combobox appearing. |

---

### New findings (commit `9e221cb`)

#### 🔵 Low — `include_gang_rating` is a dead parameter in `SkillAdvancementLogParams`

**Files:** `app/actions/logs/gang-fighter-logs.ts`, `app/actions/fighter-advancement.ts`

The `SkillAdvancementLogParams` interface still carries `include_gang_rating?: boolean`, and the JSDoc on `gang_financials_before` says it "supersedes the legacy 'New gang rating' line built from `include_gang_rating`" — implying the fallback still exists. It does not. The new `logSkillAdvancement` body never reads `include_gang_rating`; the only summary content comes from `gang_financials_before`/`after`. The caller in `fighter-advancement.ts` still passes `include_gang_rating: true` on every call, but it has no effect.

The mismatch between the JSDoc promise and the implementation will confuse the next developer who relies on it to add a rating line for a zero-cost skill.

**Fix:** Remove `include_gang_rating` from `SkillAdvancementLogParams` and the call-site, and update the JSDoc on `gang_financials_before` to drop the "supersedes" language. (`CharacteristicAdvancementLogParams` and `AdvancementDeletionLogParams` should keep their `include_gang_rating` fields — those still work correctly.)

---

#### 🔵 Low — `acquisitionTypeLabel` falls back to raw type-ID string in gang log

**File:** `components/fighter/fighter-advancement-list.tsx` — `rollSkillFromSet`

```ts
const sample = availableAdvancements.find(
  (a) => a.available_acquisition_types && a.available_acquisition_types.length > 0
);
const acquisitionTypeLabel =
  sample?.available_acquisition_types?.find((t) => t.type_id === skillAcquisitionType)?.name ??
  skillAcquisitionType; // ← raw ID e.g. "primary_random"
```

If `sample` is undefined (no advancement in the pool has `available_acquisition_types` populated), the gang log entry ends up with the raw DB identifier (`"primary_random"`) rather than a human-readable label like `"Random Primary"`. In practice this requires a misconfigured DB row, but the fallback is visible in the audit log so it is worth a readable default:

```ts
?? skillAcquisitionType.replace(/_/g, ' ')
// or a static map: { primary_random: 'Random Primary', ... }
```

---

### What's good in this commit

- The `financialsBefore`/`financialsAfter` snapshot pattern is clean — capturing both before calling `updateGangRatingSimple`/`updateGangFinancials` and threading through to the log gives auditors a proper before/after without a second DB query.
- `verifyAndLogRolledSkillAdvancementRoll` correctly rejects Ganger/Exotic Beast fighters, keeping the two logging paths cleanly separated.
- `rollSkillFromSet` correctly propagates `editableXpCost`/`editableCreditsIncrease` (set by `applyAcquisitionTypeSelection`) into the selected advancement, so the submission path remains consistent.
- `logRolledSkillAdvancement` reuses `gangerAdvancementRollDetailLine` for formatting; the extended `dice_data` fields (`pool_size`, `roll_type`, `skill_set_access`) are logged for future analytics without breaking the formatter.
- `'skill_advancement_roll'` correctly added to `LOG_TYPE_LABELS`.
- `fighter-skills-list.tsx`: the `fetchCategoriesAndAccess` effect now properly serialises the two fetches, setting `skillAccessLoading` before the second request and clearing it in `finally`.

---

Finding #3 from the last review (random bypass) is the one I flagged most strongly — it's still open. The `include_gang_rating` dead parameter (#1 new finding) is easy to clean up and removes a misleading contract from the logging API. Everything else is in good shape.

---

joeseos:
## Re-review — commit `c08f6c5` (third push)

@Djidiouf — this commit is a clean, targeted fix for the dead-parameter finding from the last review. One prior finding resolved, two prior findings still open, no new issues introduced.

---

### Previous findings — updated status

| # | Finding | Status |
|---|---------|--------|
| 1 | `skillAccess` race condition | ✅ Resolved (commit `9e221cb`) |
| 2 | `'allowed'` → `any_random` undocumented | ✅ Resolved (commit `9e221cb`) |
| 3 | Random acquisition: Combobox shown unconditionally alongside Roll button | ⚠️ **Outstanding** — `<Combobox>` still renders for all acquisition types; a user on a random path can still pick directly without rolling |
| 4 | No access warning in `SkillModal` | ✅ Resolved (commit `9e221cb`) |
| 5 | Inconsistent group label names | ✅ Resolved (commit `9e221cb`) |
| 6 | `ALL_SKILL_ACQUISITION_TYPE_IDS` manual list undocumented | ✅ Resolved (commit `9e221cb`) |
| 7 | No loading feedback during `fetchAvailableAdvancements` | ⚠️ **Outstanding** (low priority, acknowledged) |
| N1 | `include_gang_rating` dead parameter in `SkillAdvancementLogParams` | ✅ **Resolved in this commit** — field removed from the interface, call-site `include_gang_rating: true` removed, JSDoc on `gang_financials_before` updated to drop the misleading "supersedes" language. Clean fix. |
| N2 | `acquisitionTypeLabel` falls back to raw type-ID string in gang log | ⚠️ **Outstanding** — `?? skillAcquisitionType` raw fallback still present at `fighter-advancement-list.tsx:1117` |

---

### New findings (commit `c08f6c5`)

None. The diff is 7 lines of dead-code removal and a JSDoc update. No new logic, no new issues.

---

Finding #3 (random bypass) is still the most important blocker. Finding N2 (raw label fallback) is low-severity but easy to fix in the same pass. Everything else is in good shape — the cleanup in this commit was exactly right.

---

PR #1704, opened 2026-05-24
Added the possibility to have multiple winners

Djidiouf:


joeseos:
## Code Review — PR #1704 "Added the possibility to have multiple winners"

@Djidiouf Great addition overall. The backward-compatibility strategy (new flags + legacy `winner_id`/`winner_gang_id` fallback, centralised in `utils/battle-winners.ts`) is well-designed, the migration is minimal, and the UI work covers both the session and campaign-battle modals cleanly. A few issues worth addressing before merge:

---

### 🔴 HIGH

**1. `setSessionWinners` — participant updates silently discard errors**
`app/actions/battle-sessions.ts` · lines 301–311

```ts
await Promise.all(
  sessionParticipants.map((p) =>
    supabase
      .from('battle_session_participants')
      .update({ is_winner: ..., claimed_territory: ... })
      .eq('id', p.id)
  )
);
// ← then unconditionally updates battle_sessions.winner_gang_id
```

The Supabase `update()` calls return `{ data, error }` but the results are never checked. If any individual update fails (transient DB error, RLS policy, etc.) the loop still proceeds and `winner_gang_id` gets written on the parent session, leaving the DB in a partially-updated state — some participants flagged, others not.

**Fix:** check each error, or bail early on first failure:
```ts
const results = await Promise.all(
  sessionParticipants.map((p) =>
    supabase.from('battle_session_participants')
      .update({ is_winner: winnerGangSet.has(p.gang_id), claimed_territory: ... })
      .eq('id', p.id)
  )
);
const failed = results.find((r) => r.error);
if (failed) return { success: false, error: 'Failed to update participant flags' };
```

---

### 🟡 MEDIUM

**2. `normaliseParticipants` is not exported — full logic is duplicated in the API route**
`app/actions/campaigns/[id]/battle-logs.ts` · line 67  
`app/api/campaigns/battles/route.ts` · lines 857–893

`normaliseParticipants` is a private function. The API route hand-rolls the same normalization inline with `any` casts throughout. These two paths will inevitably drift. The API route's version also omits the `territoryClaimerFor` guard, which means `claimed_territory: true` can be written to the participants JSONB even when no territory is being claimed (no functional impact on territory ownership, but stale metadata in the JSONB).

**Fix:** export `normaliseParticipants` (and `territoryClaimerFor`) and call them from the route instead of duplicating.

---

**3. `updateBattleLog` does not invalidate outgoing winners' caches**
`app/actions/campaigns/[id]/battle-logs.ts` · lines 578–591

```ts
// Only the NEW winners are invalidated:
for (const winnerId of effectiveWinnerIds) {
  revalidateTag(CACHE_TAGS.COMPOSITE_GANG_CAMPAIGNS(winnerId));
}
```

`existingBattle` is only selected with `id, campaign_territory_id` (line 425), so the old winners are unknown at this point. If a battle is edited to swap out a winner, the removed gang's stats page will serve stale data until some other cache bust happens.

**Fix:** add `winner_id, participants` to the `existingBattle` select, derive old winners via `getWinnerIds`, and invalidate them before updating:
```ts
const { data: existingBattle } = await supabase
  .from('campaign_battles')
  .select('id, campaign_territory_id, winner_id, participants')
  ...
const oldWinnerIds = getWinnerIds(existingBattle);
// ... later, after the update:
for (const oldId of oldWinnerIds) revalidateTag(CACHE_TAGS.COMPOSITE_GANG_CAMPAIGNS(oldId));
```

---

**4. Unreachable guard in `setSessionWinners`**
`app/actions/battle-sessions.ts` · lines 34–37

```ts
const claimers = winners.filter((w) => w.claimed_territory === true);
const claimerGangId = claimers[0]?.gang_id ?? null;
if (claimerGangId && !winners.some((w) => w.gang_id === claimerGangId)) {
  return { success: false, error: 'The territory claimer must be one of the winners' };
}
```

`claimerGangId` is derived directly from `winners`, so `winners.some(...)` is always `true` when `claimerGangId` is non-null. This check can never fire. Remove it to avoid confusing future readers.

---

### 🔵 LOW

**5. `any` casts on `allParticipants` in `completeBattleSession`**
`app/actions/battle-sessions.ts` · lines 127–134

Now that the select includes `is_winner` and `claimed_territory`, the resulting rows should be properly typed rather than cast to `any`. This makes the fallback logic harder to follow and disables TypeScript's safety net on the participant fields.

---

**6. Redundant `gangIdSet.add` in `_getCampaignBattles`**
`app/lib/campaigns/[id]/get-campaign-data.ts` · lines 978–985

```ts
if (p.gang_id) gangIdSet.add(p.gang_id);              // adds the id
if (p.is_winner === true && p.gang_id) gangIdSet.add(p.gang_id); // no-op
```

The comment acknowledges this is intentional, but the second line is pure noise. Delete it.

---

**7. `victoriesByAllegiance` double-counts shared-allegiance co-winners**
`components/campaigns/[id]/campaign-triumphs.tsx` · lines 187–193

In a multi-winner battle where two gangs from the same allegiance both win, the allegiance scores +2 for that one battle. This may be the intended game rule, but worth confirming — it would affect triumph thresholds for allegiance-based rankings.

---

### Summary

| 1 | `app/actions/battle-sessions.ts:301` | 🔴 High | Partial DB update on `Promise.all` error |
| 2 | `app/api/campaigns/battles/route.ts:857` | 🟡 Medium | `normaliseParticipants` duplicated inline, logic diverges |
| 3 | `app/actions/campaigns/[id]/battle-logs.ts:578` | 🟡 Medium | Old winners not cache-invalidated on edit |
| 4 | `app/actions/battle-sessions.ts:34` | 🟡 Medium | Unreachable guard (dead code) |
| 5 | `app/actions/battle-sessions.ts:127` | 🔵 Low | `any` casts on typed participant rows |
| 6 | `app/lib/campaigns/[id]/get-campaign-data.ts:984` | 🔵 Low | Redundant no-op `gangIdSet.add` |
| 7 | `components/campaigns/[id]/campaign-triumphs.tsx:187` | 🔵 Low | Possible double-counting in triumph tallies |

---

**Verdict: Changes requested** — items 1–3 are the blocking ones; 4–7 are clean-up. The architecture and backward-compat strategy are solid.


---

joeseos:
## Code Review — PR #1704 (updated) · @Djidiouf

Good follow-up push. Every blocking and most clean-up issues from the previous review have been addressed. One new medium issue surfaced and one minor cast worth noting.

---

| 1 | 🔴 High | `setSessionWinners` participant updates discarded errors | ✅ Resolved |
| 2 | 🟡 Medium | `normaliseParticipants` duplicated in API route | ✅ Resolved — new `utils/battle-participants.ts` centralises it |
| 3 | 🟡 Medium | Old winners not cache-invalidated on `updateBattleLog` | ✅ Resolved — `oldWinnerIds` derived + busted |
| 4 | 🟡 Medium | Unreachable guard in `setSessionWinners` | ✅ Resolved — dead check removed |
| 5 | 🔵 Low | `any` casts on typed participant rows in `completeBattleSession` | ✅ Resolved |
| 6 | 🔵 Low | Redundant `gangIdSet.add` no-op | ✅ Resolved |
| 7 | 🔵 Low | `victoriesByAllegiance` allegiance double-count | ✅ Resolved — per-battle `Set<string>` prevents inflation |

---

### New Findings

#### 🟡 MEDIUM — `setSessionWinners`: `battle_sessions` update is still fire-and-forget
`app/actions/battle-sessions.ts` · line 321

```ts
await supabase
  .from('battle_sessions')
  .update({ winner_gang_id: legacyWinnerGangId, updated_at: ... })
  .eq('id', sessionId);
// ← result ignored; returns { success: true } regardless
```

The participant-flag updates are now correctly checked (great), but if this subsequent write fails the session's `winner_gang_id` stays stale while participants already carry the new `is_winner` flags. The two sources of truth diverge silently.

**Fix:**
```ts
const { error: sessionUpdateError } = await supabase
  .from('battle_sessions')
  .update({ winner_gang_id: legacyWinnerGangId, updated_at: new Date().toISOString() })
  .eq('id', sessionId);
if (sessionUpdateError) {
  return { success: false, error: `Failed to update session: ${sessionUpdateError.message}` };
}
```

---

#### 🔵 LOW — `existingBattle as any` in `updateBattleLog`
`app/actions/campaigns/[id]/battle-logs.ts` · line 358

```ts
const oldWinnerIds = getWinnerIds(existingBattle as any);
```

The cast is forced by Supabase's `Json` column type for `participants`, and `parseParticipants` inside `getWinnerIds` handles all runtime shapes correctly, so there is no functional problem. Worth a one-line comment explaining why the cast is safe, to prevent a future refactor from removing it thinking it's sloppy:

```ts
// Supabase types participants as Json; parseParticipants handles string|array|null at runtime.
const oldWinnerIds = getWinnerIds(existingBattle as any);
```

---

### Summary

| 1 | `app/actions/battle-sessions.ts:321` | 🟡 Medium | `battle_sessions` update error not checked |
| 2 | `app/actions/campaigns/[id]/battle-logs.ts:358` | 🔵 Low | `as any` cast uncommented |

**Verdict: Changes requested** — the medium item (#1) is the only remaining blocker; the low item is cosmetic. The architecture, the `utils/battle-participants.ts` extraction, the cache-bust fix, and the UI multi-winner slot work are all solid. Nearly there.

---

joeseos:
**FYI / non-blocking** — a few minor code-reuse observations that could be tidied up in a follow-up PR:

1. **Winner-slot UI is duplicated across two modals.** `complete-battle-modal.tsx` and `campaign-battle-log-modal.tsx` both implement the same multi-winner slot logic (`winners[]`, `isDraw`, `claimedByGangId`, the handler trio, `slotsToRender`, the Combobox render loop + territory-claimer sub-section). That's ~100 lines of near-identical state/JSX. A shared `<WinnerSelector>` exporting and importing it would eliminate the duplication. The campaign modal version also uses an IIFE for the render block while the session modal doesn't — a small inconsistency worth aligning.

2. **`effectiveClaimerFallback` in `battle-logs.ts` reinvents `getClaimerGangId`.** The private helper added to find `claimed_territory: true` on a participants array is essentially the same logic already exported by `utils/battle-winners.ts` as `getClaimerGangId`. Could just call the existing util.

3. **Winner-enrichment block is copy-pasted between `createBattleLog` and `updateBattleLog`.** The ~15-line block that builds `winnerNameLookupIds` → `winnerNameMap` → `winnersEnriched` / `claimerEnriched` / `primaryWinner` is identical in both functions. A small private helper inside the module would remove the duplication.

4. **`get-campaign-data.ts` manually re-implements `getWinnerIds`.** The `_getCampaignBattles` function inlines the `flaggedWinnerIds → effectiveWinnerIds` fallback pattern instead of calling `getWinnerIds()` from `utils/battle-winners.ts` — the exact utility added for this purpose, already used in `campaign-battle-logs-list.tsx`.

5. **Attacker/defender extraction is repeated in two places.** When `getAttackerDefenderFromParticipants` was removed, its replacement (two `.find()` + `?.gang_id` lines) was inlined independently in both `createBattleLog` and `updateBattleLog` rather than kept as a small helper.

PR #1705, opened 2026-05-24
Cleanup task past the multiple winners PR

Djidiouf:


joeseos:
## Code Review — PR #1705: Cleanup task past the multiple winners PR

@Djidiouf — reviewed against the full codebase. This is a solid cleanup: duplicated state logic across two modals is consolidated into a shared hook, and several inline extraction patterns are pushed into shared utilities. The logic is behaviourally identical to what was removed. A few observations below.

---

### Findings

#### Medium

**`enrichWinners` is misplaced in `utils/battle-participants.ts`**
`battle-participants.ts` is about normalising and reading participant arrays. `enrichWinners` has no dependency on participants at all — it maps winner/claimer IDs to names from a `gangs` lookup. It belongs in `utils/battle-winners.ts` alongside `getWinnerIds`, `getClaimerGangId`, and `getExplicitClaimerGangId`. As-is, a future reader will look in the wrong file when tracking winner-enrichment logic.

---

#### Low

**`get-campaign-data.ts`: `battle.participants` is parsed twice per battle**

```ts
// parsedParticipantsMap already parsed this for attacker/defender
const parsedParticipants = parsedParticipantsMap.get(battle.id);

// …then getWinnerIds / getClaimerGangId each call parseParticipants(battle.participants) again
const effectiveWinnerIds = getWinnerIds(battle);
const claimerId = getClaimerGangId(battle);
```

Not a correctness issue (both paths handle the same JSON), but at scale (battles limit is 100) this parses each `participants` column three times per row. Passing `parsedParticipants` into a `CampaignBattleLike`-compatible wrapper, or exposing `getWinnerIdsFromParsed`/`getClaimerGangIdFromParsed` overloads, would avoid the redundant work.

---

**`useWinnerSelection` exposes raw `setWinners` / `setIsDraw`**

Exposing the raw React setState functions breaks the hook's encapsulation — external callers can set `isDraw(true)` without resetting `winners`, or set winners that don't pass through the draw-handling logic. This is currently needed so `campaign-battle-log-modal.tsx` can populate winners when loading an existing battle. A narrow alternative would be a single `loadExistingBattle({ winners, isDraw, claimerId })` method on the hook that atomically sets all three, keeping the mutation path explicit and safe.

---

**Handler functions in the hook are recreated every render**

`handleWinnerChange`, `addWinnerSlot`, `removeWinnerSlot`, and `resetWinnerSelection` are plain arrow functions — not wrapped in `useCallback`. They're passed as `onClick`/`onValueChange` props to Combobox and Button elements. If either modal ever wraps children in `React.memo`, these unstable references will cause unnecessary re-renders. Wrapping with `useCallback` (with empty or minimal dependency arrays, since they all use functional setState forms) would future-proof this.

---

### What's well done

- The `getExplicitClaimerGangId` / `getClaimerGangId` split is a clean and well-documented distinction. The doc comment ("does NOT fall back to `winner_id`") makes the intent unambiguous.
- `normaliseParticipants` was left untouched — correct, since the multi-winner logic in the data model is unchanged.
- `resetWinnerSelection()` replacing the three-line manual reset is exactly the right level of abstraction.
- Removing the inline IIFE from the campaign modal's JSX is a meaningful readability win.
- `initialWinnerIds` / `initialClaimerId` / `initialIsDraw` are correctly threaded into `complete-battle-modal.tsx` so existing sessions render with pre-populated winner state.

---

### Verdict: ✅ Approve

No correctness regressions found. The medium item (file placement) is worth a follow-up but doesn't need to block this. The low items are nice-to-haves for future robustness.

---

joeseos:
## Follow-up Review — PR #1705 (sync at 2a51438)

@Djidiouf — reviewed the new commits against the four findings from the previous pass. All four are resolved; the new code is clean.

---

| Finding | Status |
|---|---|
| **Medium** — `enrichWinners` misplaced in `battle-participants.ts` | ✅ **Resolved** — moved to `battle-winners.ts` with a clear doc comment |
| **Low** — `battle.participants` parsed twice per battle in `get-campaign-data.ts` | ✅ **Resolved** — `parsedForWinners` wrapper passes the already-decoded array; `parseParticipants` short-circuits on arrays |
| **Low** — hook exposed raw `setWinners` / `setIsDraw` | ✅ **Resolved** — both setters are now internal; `loadExistingWinners` provides the single atomic entry point |
| **Low** — handler functions recreated every render | ✅ **Resolved** — `handleWinnerChange`, `addWinnerSlot`, `removeWinnerSlot`, `loadExistingWinners`, `removeGangFromWinners`, and `resetWinnerSelection` are all `useCallback`-wrapped |

---

### New Findings

#### Low — `parsedForWinners` is structurally typed rather than cast to `CampaignBattleLike`

`get-campaign-data.ts` — around the `parsedForWinners` block:

```ts
const parsedForWinners = {
  participants: parsedParticipants ?? [],
  winner_id: battle.winner_id,
};
const effectiveWinnerIds = getWinnerIds(parsedForWinners);
const claimerId = getClaimerGangId(parsedForWinners);
```

If `CampaignBattleLike.participants` is typed as `string` (raw DB column), TypeScript will silently accept this because the literal object is inferred structurally — or it may produce a hidden `any` cast. The runtime is correct because `parseParticipants` handles arrays, but a typed overload (`getWinnerIdsFromParsed(array, winner_id)`) would make the intent explicit and avoid any future compiler surprises if the `CampaignBattleLike` type changes. Not a blocker; just worth a follow-up note in the backlog.

---

### What's well done in this round

- `getAttackerDefenderIds` in `battle-participants.ts` is the right abstraction — symmetrical with the other helpers and eliminates the identical four-line pattern that appeared in both `createBattleLog` and `updateBattleLog`.
- `loadExistingWinners` handles all three branches (`draw`, `winnerIds.length > 0`, empty) atomically. Because React 18 batches the state updates inside, there's no risk of the claimer-clearing effect running between the `setWinners` and `setClaimedByGangId` calls.
- `removeGangFromWinners` correctly leaves `isDraw` unchanged when a participant is swapped out — a draw isn't gang-specific, so clearing it would be wrong.
- Promoting `selectedGangs` to a top-level `useMemo` in `campaign-battle-log-modal.tsx` is cleaner than the IIFE it replaced, and the computed value now flows into `maxParticipants` consistently.
- The JSX render block for winner slots was correctly kept inline per modal (rather than extracted into a shared component) — the two modals build gang options differently and the comment in the hook explains why.

---

### Verdict: ✅ Approve

No correctness regressions. All previous findings addressed. The one new observation (type wrapper) is a minor follow-up, not a blocker.

---

joeseos:
## Follow-up Review — PR #1705 (sync at 95c37f5)

@Djidiouf — reviewed the single new commit against the one outstanding finding from the previous pass. It's resolved cleanly.

---

| Finding | Status |
|---|---|
| **Medium** — `enrichWinners` misplaced in `battle-participants.ts` | ✅ Resolved (previous round) |
| **Low** — `battle.participants` parsed twice per battle | ✅ Resolved (previous round) |
| **Low** — hook exposed raw `setWinners` / `setIsDraw` | ✅ Resolved (previous round) |
| **Low** — handler functions recreated every render | ✅ Resolved (previous round) |
| **Low** — `parsedForWinners` structurally typed rather than cast to `CampaignBattleLike` | ✅ **Resolved** — `parsedForWinners` wrapper is gone; replaced with `getWinnerIdsFromParsed` / `getClaimerGangIdFromParsed` that accept `BattleParticipant[]` directly |

---

### New Findings

None. A few things worth noting as confirmation the code is correct:

- `BattleParticipant.gang_id` is typed as `string` (not `string | null`), so `.map((p) => p.gang_id)` correctly returns `string[]` — no hidden type widening.
- The logic in both new overloads is a faithful copy of `getWinnerIds` / `getClaimerGangId`, just with `parseParticipants(...)` replaced by direct use of the already-decoded array. The fallback-to-`winner_id` paths are identical.
- `parsedParticipants ?? []` correctly collapses the `BattleParticipant[] | undefined` from the map lookup before passing to the typed overloads.
- The `parsed` local variable is confined to the loop body, so no shadowing risk.

---

### Verdict: ✅ Approve

All previous findings addressed. No new issues introduced.

---

PR #1713, opened 2026-05-25
fix default select dropdown

Djidiouf:


joeseos:
## PR Review — `fix default select dropdown` (@Djidiouf)

Reviewed against the full `globals.css` and the broader component tree. Single-file CSS change, targeted and readable.

---

### Findings

**[LOW] `color-scheme: light` missing from `:root`**
`color-scheme: dark` is now correctly set inside `.dark {}`, which is the right place. Best practice is to also set `color-scheme: light` on `:root` alongside it. Without it, `:root` inherits `normal`, which means on a system running in OS-level dark mode while the app is in light mode, some native form controls (like `<select>` chrome/OS-rendered parts) may still render with dark OS styling. This is a very edge case today, but worth a follow-up.

```css
/* suggestion */
:root {
  color-scheme: light;   /* explicit */
  ...
}
.dark {
  color-scheme: dark;    /* already added ✓ */
}
```

---

**[LOW] Base `select` rule could mask intentional Tailwind overrides (unlikely, but noted)**
The new bare `select { background-color: hsl(var(--background)); color: hsl(var(--foreground)); }` lands in `@layer base`, so Tailwind utilities and component-layer styles correctly override it via the cascade. No real risk — just noting that any `<select>` with explicit `bg-*`/`text-*` Tailwind classes will properly win over this base rule.

---

**[GOOD] Firefox `option/optgroup` hardcoded color fixed**
`color: #0a0a0a` → `color: hsl(var(--foreground))` is a genuine bug fix. `#0a0a0a` is the light-mode foreground — in dark mode it rendered near-black text on a dark background. Now correctly adapts to both themes. Good catch.

---

**[GOOD] Print media query interaction**
The existing `@media print` block already overrides `--secondary` to `var(--light-secondary) !important`, so `.dark select { background-color: hsl(var(--secondary)) }` resolves to white during printing even in dark mode. No regression there.

---

**[GOOD] Dark mode cascade is correct**
The layering logic is sound:
- `select` → base colors from `--background` / `--foreground`  
- `.dark select` → overrides background to `--secondary` (slightly lighter than pure background, `0 0% 14.9%`)  
- `.dark select option` → consistent with `.dark select`  

The specificity chain is clean and the intent is clear.

---

**[GOOD] `select:disabled` styling**
`opacity: 0.5` + `cursor: not-allowed` in `@layer base` is a reasonable baseline. Not the primary bug but a harmless and useful UX normalization.

---

### Summary

| Area | Status |
|---|---|
| Correctness | ✅ No bugs introduced |
| Edge cases / null handling | ✅ Print mode handled, disabled state covered |
| Security | ✅ N/A (pure CSS) |
| Performance | ✅ No impact |
| Readability | ✅ Clean, logical grouping |
| Project conventions | ✅ Follows existing CSS variable and layer patterns |

---

### Verdict: **Approve** ✅

This is a clean, well-scoped fix. The approach — using CSS variables throughout so both light and dark themes resolve automatically — is the right pattern for this codebase. The one suggestion (`color-scheme: light` on `:root`) is a follow-up enhancement, not a blocker.


---

PR #1728, opened 2026-06-01
Added activity statistics in the admin panel regarding gangs and camp…

Djidiouf:
…aigns

### DIFF
--- a/app/api/admin/stats/route.ts
+++ b/app/api/admin/stats/route.ts
@@ -4,6 +4,8 @@ import { checkAdmin } from "@/utils/auth";
 import { getUserCount } from '@/app/lib/get-stats-user';
 import { getGangCount } from '@/app/lib/get-stats-gang';
 import { getCampaignCount } from '@/app/lib/get-stats-campaign';
+import { getGangActivityStats } from '@/app/lib/get-stats-gang-activity';
+import { getCampaignActivityStats } from '@/app/lib/get-stats-campaign-activity';
 
 export async function GET() {
   const supabase = await createClient();
@@ -16,16 +18,21 @@ export async function GET() {
 
     // Use cached count functions instead of direct database queries
     // These are cached for 24 hours and automatically invalidated on create/delete
-    const [userCount, gangCount, campaignCount] = await Promise.all([
-      getUserCount(),
-      getGangCount(),
-      getCampaignCount()
-    ]);
+    const [userCount, gangCount, campaignCount, gangActivity, campaignActivity] =
+      await Promise.all([
+        getUserCount(),
+        getGangCount(),
+        getCampaignCount(),
+        getGangActivityStats(),
+        getCampaignActivityStats(),
+      ]);
 
     return NextResponse.json({
       userCount,
       gangCount,
-      campaignCount
+      campaignCount,
+      gangActivity,
+      campaignActivity,
     });
   } catch (error) {
     console.error('Error fetching stats:', error);
--- a/app/lib/get-stats-campaign-activity.ts
+++ b/app/lib/get-stats-campaign-activity.ts
@@ -1,0 +1,52 @@
+import { createServiceRoleClient } from '@/utils/supabase/server';
+
+export interface CampaignActivityStats {
+  last2Weeks: number | null;
+  last1Month: number | null;
+  last3Months: number | null;
+  last6Months: number | null;
+}
+
+const PERIODS = {
+  last2Weeks: 14,
+  last1Month: 30,
+  last3Months: 90,
+  last6Months: 180,
+} as const;
+
+async function countCampaignsUpdatedSince(days: number): Promise<number | null> {
+  const supabase = createServiceRoleClient();
+  const cutoff = new Date(Date.now() - days * 24 * 60 * 60 * 1000).toISOString();
+
+  const { count, error } = await supabase
+    .from('campaigns')
+    .select('*', { count: 'exact', head: true })
+    .gte('updated_at', cutoff);
+
+  if (error) {
+    console.error('Error fetching campaign activity count:', error);
+    return null;
+  }
+
+  return count ?? 0;
+}
+
+/**
+ * Get campaign activity counts by updated_at for admin display.
+ *
+ * @returns Counts per period, or null if service role key is not available
+ */
+export async function getCampaignActivityStats(): Promise<CampaignActivityStats | null> {
+  if (!process.env.SUPABASE_SERVICE_ROLE_KEY) {
+    return null;
+  }
+
+  const [last2Weeks, last1Month, last3Months, last6Months] = await Promise.all([
+    countCampaignsUpdatedSince(PERIODS.last2Weeks),
+    countCampaignsUpdatedSince(PERIODS.last1Month),
+    countCampaignsUpdatedSince(PERIODS.last3Months),
+    countCampaignsUpdatedSince(PERIODS.last6Months),
+  ]);
+
+  return { last2Weeks, last1Month, last3Months, last6Months };
+}
--- a/app/lib/get-stats-campaign-activity.ts
+++ b/app/lib/get-stats-campaign-activity.ts
@@ -1,3 +1,5 @@
+import { unstable_cache } from 'next/cache';
+import { CACHE_TAGS } from '@/utils/cache-tags';
 import { createServiceRoleClient } from '@/utils/supabase/server';
 
 export interface CampaignActivityStats {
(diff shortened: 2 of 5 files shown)


joeseos (changes requested):
Hey @Djidiouf — solid idea adding activity statistics to the admin panel. The UI structure looks clean and follows the existing modal patterns well. However there's one critical bug that needs fixing before merge, plus a few other issues worth addressing.

---

## Findings

---

### 🔴 HIGH — Missing `unstable_cache` wrapper (both activity stat files)

**Files:** `app/lib/get-stats-gang-activity.ts`, `app/lib/get-stats-campaign-activity.ts`

Every single existing stats function in this repo (`getUserCount`, `getGangCount`, `getCampaignCount`) is wrapped in `unstable_cache` with a 24-hour TTL. The two new activity functions are **not**. This means every admin stats panel open fires **8 uncached Supabase queries** directly against the database (4 periods × 2 activity types).

As a consequence, the two new cache tags added to `utils/cache-tags.ts` (`GLOBAL_GANG_ACTIVITY`, `GLOBAL_CAMPAIGN_ACTIVITY`) are **never referenced** anywhere — they're dead code right now.

**Fix:** Wrap each function in `unstable_cache`, following the exact same pattern as the existing files:

```ts
// get-stats-gang-activity.ts
export async function getGangActivityStats(): Promise<GangActivityStats | null> {
  const getCachedGangActivityStats = unstable_cache(
    async () => {
      if (!process.env.SUPABASE_SERVICE_ROLE_KEY) return null;
      const supabase = createServiceRoleClient(); // create once, reuse below
      // ... run the 4 queries using this single client instance
    },
    ['global-gang-activity'],
    {
      tags: [CACHE_TAGS.GLOBAL_GANG_ACTIVITY()],
      revalidate: 86400,
    }
  );
  return await getCachedGangActivityStats();
}
```

Same pattern for `getCampaignActivityStats`. This also fixes the secondary issue of `createServiceRoleClient()` being instantiated 4 times per call (once inside each `countGangsUpdatedSince` / `countCampaignsUpdatedSince` invocation) instead of once.

---

### 🟡 MEDIUM — `updated_at` on campaigns may not reflect user activity

**File:** `app/lib/get-stats-campaign-activity.ts`, line 22

```ts
.gte('updated_at', cutoff)
```

Gangs use `last_updated` (a likely explicit user-touch timestamp), but campaigns use `updated_at`. If `updated_at` on the `campaigns` table is auto-updated by Supabase/Postgres on any write — including cascading updates from gang joins/leaves, territory changes, or batch jobs — this metric will over-count "active" campaigns by including ones that were touched by system operations rather than user intent.

Worth confirming whether `updated_at` accurately represents user-initiated activity on campaigns, or if there's a more specific field (like `last_updated` if it exists on campaigns too). If `updated_at` is the correct column, at least add a comment explaining why it's used instead of a dedicated field.

---

### 🟡 MEDIUM — Mobile layout regression in stat cards

**File:** `components/admin/admin-stats-modal.tsx`, line ~97

The top-level summary grid changed from `grid-cols-1 md:grid-cols-3` to `grid-cols-2 md:grid-cols-3`. With exactly **3 cards**, on mobile this creates a 2-column grid where "Total Campaigns" renders alone in a half-width column — visually unbalanced.

Consider keeping `grid-cols-1` (or using `grid-cols-3` with smaller padding) for the summary section on mobile to avoid the orphaned third card.

---

### 🟢 LOW — Triplicated `ActivityStats` interface

`GangActivityStats` (gang activity file) and `CampaignActivityStats` (campaign activity file) are byte-for-byte identical, and `ActivityStats` in the modal component is a third independent copy of the same shape. Consider exporting a single shared `ActivityStats` type (e.g. from a `types/stats.ts` or from one of the lib files) and importing it in the others. Not a blocker but will cause drift over time.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 HIGH | Activity stats bypass `unstable_cache` — 8 uncached DB hits per admin panel open; cache tags are dead code |
| 2 | 🟡 MEDIUM | `campaigns.updated_at` may include system-triggered writes, inflating "active campaign" counts |
| 3 | 🟡 MEDIUM | Mobile layout: 3 cards in `grid-cols-2` leaves an orphaned half-width card |
| 4 | 🟢 LOW | Triplicated `ActivityStats` type definition |

---

## Verdict: **Changes Requested**

The caching issue (#1) is a must-fix — it directly contradicts the project's established caching strategy and will silently hammer the database. The mobile layout (#3) is also worth fixing before merge since this is a UI feature. Items #2 and #4 can be addressed in a follow-up if preferred. Happy to re-review once the cache wrapper is in place!

---

joeseos (approved):
## Code Review — PR #1728: Activity Statistics in Admin Panel

Hey @Djidiouf — clean implementation overall. The feature is admin-only, appropriately cached, and the UI component is well-structured. A few findings below.

---

### Previous findings
None — this is the first review.

---

### New findings

**1. `unstable_cache` closure recreated on every outer function call — Medium**

`files: app/lib/get-stats-gang-activity.ts, app/lib/get-stats-campaign-activity.ts`

Both new files (and the existing `getGangCount` / `getCampaignCount`) declare the cached function _inside_ the outer function body:

```ts
export async function getGangActivityStats() {
  const getCachedGangActivityStats = unstable_cache(  // new closure every call
    async () => { ... },
    ['global-gang-activity'],
    { ... }
  );
  return await getCachedGangActivityStats();
}
```

Next.js still deduplicates via the cache key string so DB calls aren't duplicated, but a new wrapper closure is allocated on every request. The idiomatic fix is to hoist the `unstable_cache` call to module scope:

```ts
const getCachedGangActivityStats = unstable_cache(
  async () => { ... },
  ['global-gang-activity'],
  { tags: [CACHE_TAGS.GLOBAL_GANG_ACTIVITY()], revalidate: 86400 }
);

export async function getGangActivityStats() {
  return await getCachedGangActivityStats();
}
```

Since the existing stat files share this pattern, fixing all three together in a follow-up would be cleaner than only fixing the new ones now.

---

**2. 8 additional DB queries per cache miss — Low**

Each activity stat type fires 4 parallel `COUNT` queries (14 / 30 / 90 / 180 days). On a cache miss the admin endpoint now issues 8 extra round-trips. For a 24h-cached admin-only endpoint this is fine, but a single query with conditional aggregation would be more efficient if the table grows large:

```sql
SELECT
  COUNT(*) FILTER (WHERE updated_at >= NOW() - INTERVAL '14 days')  AS last_2_weeks,
  COUNT(*) FILTER (WHERE updated_at >= NOW() - INTERVAL '30 days')  AS last_1_month,
  COUNT(*) FILTER (WHERE updated_at >= NOW() - INTERVAL '90 days')  AS last_3_months,
  COUNT(*) FILTER (WHERE updated_at >= NOW() - INTERVAL '180 days') AS last_6_months
FROM campaigns;
```

Not required to merge, just a future optimization note.

---

**3. `ActivityStats` interface shape duplicated three times — Low**

`GangActivityStats` (`get-stats-gang-activity.ts`), `CampaignActivityStats` (`get-stats-campaign-activity.ts`), and `ActivityStats` (`admin-stats-modal.tsx`) are all structurally identical. Consider exporting a single shared type (e.g., from a `types/stats.ts`) and reusing it across all three files.

---

**4. Asymmetric timestamp column names — Low (verify)**

Gangs query `.gte('last_updated', cutoff)` while campaigns query `.gte('updated_at', cutoff)`. This is likely intentional (following the DB schema), but worth a quick sanity-check that both column names are correct — a wrong column name in a Supabase `gte` filter will return an error caught by the handler and silently show `N/A` rather than raising loudly.

---

**5. No manual cache invalidation wired up — Low (by design)**

The new `GLOBAL_GANG_ACTIVITY` and `GLOBAL_CAMPAIGN_ACTIVITY` tags are defined and documented but no `revalidateTag()` calls exist for them anywhere in the codebase. For time-window activity counts, relying solely on the 24h TTL is reasonable — just confirming this is intentional and not an oversight.

---

### Summary

| 1 | `get-stats-*-activity.ts` | `unstable_cache` recreated per call | Medium |
| 2 | Same | 8 DB queries per cache miss | Low |
| 3 | All 3 files | Duplicated `ActivityStats` shape | Low |
| 4 | `get-stats-*-activity.ts` | Column name asymmetry — needs verify | Low |
| 5 | `get-stats-*-activity.ts` | No manual cache invalidation | Low (by design) |

---

**Verdict: ✅ Approve**

The implementation is solid, follows existing patterns, and all issues are non-blocking. Item 1 is worth tracking for a cleanup pass alongside the existing stat files. Good work!

---

PR #1729, opened 2026-06-01
Added custom trading posts to custom assets

joeseos:


joeseos (commented):
## Code Review — PR #1729

Hey @joeseos — thorough feature addition overall. The data model, server actions, and UI plumbing are well-structured and consistent with the existing codebase conventions. A few issues need addressing before this can merge — one is a correctness regression in the SQL function, and another is a security gap in the rule-management actions.

---

### Findings

---

#### 🔴 High — SQL regression: custom TP availability rules no longer filter equipment access

**File:** `supabase/functions/get_equipment_detailed_data.sql`

The PR makes two related changes:

**1. `tp_access` CTE** — the custom TP UNION loses its `CROSS JOIN gang_data gd` and the `NOT EXISTS/EXISTS` availability guard. Before this PR, a custom TP item only appeared in `tp_summary` (with `has_access = true`) if it had no availability rules, or at least one rule matching the querying gang. After this PR, every item in any active custom TP unconditionally gets `has_access = true`.

**2. `custom_tp_override` CTE** — the same `NOT EXISTS/EXISTS` is replaced with a `LEFT JOIN`. Items that would previously have been invisible (no matching rule) are now always included; the JOIN only affects which `availability_override` text is shown.

The migration documents the intended behaviour at line 88:

```sql
-- No rows = available to everyone; rows = allowlist
```

That contract is now broken. A trading-post owner who configured "Goliath only" availability rules will find those restrictions are silently ignored — every gang can now see the restricted equipment.

**Fix:** Keep the `NOT EXISTS/EXISTS` guard in the `tp_access` CTE union for custom TPs (it still exists there on `main` and correctly gates `has_access`). The LEFT JOIN approach in `custom_tp_override` is fine for resolving the `availability_override` display text, but the access gate must remain in `tp_access`. These are two separate concerns: access gating vs. availability label — they need to be handled independently.

---

#### 🟡 Medium — `addAvailabilityRule` and `addPricingRule` don't verify ownership of `equipmentItemId`

**File:** `app/actions/customise/custom-trading-posts.ts` (~lines 697–722 and ~835–858)

Both functions accept an `equipmentItemId` and insert a row without checking that the referenced `custom_trading_post_equipment` row belongs to the authenticated user:

```ts
// No ownership check — any authenticated user can call this with any equipmentItemId
const { error } = await supabase
  .from('custom_trading_post_availability')
  .insert({
    user_id: user.id,
    custom_trading_post_equipment_id: equipmentItemId, // unvalidated
    ...
  });
```

The RLS on `custom_trading_post_availability` only enforces `user_id = auth.uid()` — it does not check that the referenced equipment belongs to the same user. Any authenticated user can inject availability/pricing rules onto another user's equipment items.

Impact: the `custom_tp_override` CTE's LEFT JOIN picks up rules regardless of their `user_id`, so injected rules affect displayed availability text. Also, `saveEquipmentRules` only deletes rules where `user_id = user.id`, so injected rules from other users survive the cleanup.

**Fix:** Before inserting, verify ownership through the parent trading post:

```ts
const { data: equipItem } = await supabase
  .from('custom_trading_post_equipment')
  .select('id, custom_trading_posts!inner(user_id)')
  .eq('id', equipmentItemId)
  .eq('custom_trading_posts.user_id', user.id)
  .single();

if (!equipItem) {
  return { success: false, error: 'Equipment not found or not owned by user' };
}
```

Apply the same ownership check in `getPricingRules`, `getAvailabilityRules`, and `saveEquipmentRules`.

---

#### 🟡 Medium — `campaigns.custom_trading_posts` column type is `jsonb` instead of `text[]`

**File:** `supabase/migrations/20260531120000_add_custom_trading_posts.sql`, line 178

```sql
ALTER TABLE public.campaigns
  ADD COLUMN custom_trading_posts jsonb;
```

Everywhere in the application the column is read and written as a plain array of UUID strings (`string[]` in TypeScript). The peer column `trading_posts` is a proper `text[]`. Using `jsonb` bypasses PostgreSQL's array operators and creates a type inconsistency with the SQL function parameter `$11 uuid[]`.

**Fix:** Change the column type to `text[]` or `uuid[]` to match `trading_posts`.

---

#### 🟡 Medium — `supabase/schema/schema.public.sql` out of sync with new function signature

**File:** `supabase/schema/schema.public.sql`, line 1346

The schema snapshot still declares the function return type including `cost_resource_name text`. The SQL function file removes it. The snapshot should be regenerated so it matches what will actually be deployed.

---

#### 🟠 Low — `deleteCustomTradingPost` updates campaigns before the delete completes

**File:** `app/actions/customise/custom-trading-posts.ts` (~lines 268–298)

Campaign `custom_trading_posts` arrays are cleaned up before the `DELETE` is issued. If the delete fails, campaigns have the reference removed while the trading post still exists. Since child rows are cleaned up by `ON DELETE CASCADE`, the safest order is: delete the trading post first, then clean up campaign arrays.

---

#### 🟠 Low — `AddEquipmentModal.onConfirm` fires mutation and returns `true` immediately

**File:** `components/customise/custom-trading-posts.tsx` (~lines 2563–2570)

```tsx
onConfirm={async () => {
  addMutation.mutate();  // fire-and-forget
  return true;           // modal closes before server responds
}}
```

The modal closes before the result is known. Error toasts fire on a closed modal, and the `'Adding...'` loading state is never shown. Use `mutateAsync` and return `false` on failure.

---

### What's correct and well done

- Ownership checks in all CRUD actions (`create`, `update`, `delete`, `addTPEquipment`, `updateTPEquipment`, `removeTPEquipment`) are solid — all filter by `eq('user_id', user.id)`.
- `addTPEquipment` correctly catches unique-constraint error `23505` with a readable message.
- `shareCustomTradingPost` validates TP ownership before inserting shares, and correctly relies on the campaigns table RLS (OWNER/ARBITRATOR-only UPDATE) for campaign-level authorization — consistent with all other share functions.
- `saveEquipmentRules` is a clean delete-all-then-reinsert batch with no partial state between availability and pricing.
- The `custom_trading_post_names` batch-fetch in `gang-data.ts` is a single `IN` query — no N+1.
- `contextKey` in `equipment.tsx` correctly includes custom TP IDs to invalidate cache when the custom TP set changes.
- `cost_resource_name` is not read by any application code (confirmed grep across `app/` and `components/`), so removing it from the function return type is safe.

---

### Verdict: changes needed before merging

The availability-rules regression (items leaking to gangs that should be restricted by the allowlist) and the ownership gap on rule management are the blockers. The `jsonb` type and schema snapshot are quick fixes to bundle in at the same time.

---

PR #1732, opened 2026-06-03
Added support to buy equipment with resources

joeseos:


joeseos (commented):
Hey @joeseos — solid feature concept and the overall architecture is clean. The resource UI badge, modal flow, and `deductGangResource`/`returnGangResource` split are all sensible. There are two issues that need fixing before merge though — one is an exploit, the other a data-loss scenario on sell.

---

## Findings

---

### 🔴 HIGH — Free equipment exploit: zeroing the resource amount bypasses all cost checks

**File:** `components/equipment/purchase-modal.tsx` (resource input + `resourceArgs`) → `app/actions/equipment.ts` (~line 367)

The purchase modal lets the user edit the resource amount in an `<input>`. If the field is cleared, `manualResourceAmount` becomes `""`, and `Number("") === 0`. The `resourceArgs()` function then returns `[name, 0]`, which flows to the server as `cost_resource_amount: 0`.

On the server, `isResourcePurchase` is:
```ts
Boolean(params.cost_resource_name && params.cost_resource_amount != null)
// 0 != null → true → isResourcePurchase = true
```

The credit-deduction path is skipped (`creditsDelta: isResourcePurchase ? 0 : ...`), and `deductGangResource` is called with `amount = 0`. The sufficiency check `resource.quantity < 0` is always false, so the update runs and deducts zero. Result: **the item is acquired for free** — no credits, no resources consumed.

**Fix — two layers:**

1. In the modal validation (`handleConfirm`), reject a zero or empty resource amount when `hasResourceCost` is true:
```ts
const parsedResource = Number(manualResourceAmount);
if (hasResourceCost && (!parsedResource || parsedResource <= 0)) {
  setCreditError('Resource amount must be greater than 0');
  return false;
}
```

2. In `deductGangResource`, add a server-side guard:
```ts
if (amount <= 0) throw new Error('Resource deduction amount must be positive');
```

---

### 🔴 HIGH — TOCTOU race condition in `deductGangResource` / `returnGangResource`

**File:** `utils/campaigns/resources.ts` (~line 120)

Both functions follow a read-then-write pattern:
1. Fetch current `quantity` via `findGangResourceByName`
2. Compute new value in JS (`resource.quantity - amount`)
3. Write new value back

Under concurrent requests (two browser tabs, rapid double-click on confirm), both reads can return the same `quantity`, both sufficiency checks pass, and both writes commit — resulting in the resource being double-deducted (or double-returned).

**Fix:** Push the arithmetic into the database with an atomic conditional update:
```ts
const { data, error } = await supabase
  .from('campaign_gang_resources')
  .update({ quantity: resource.quantity - amount })
  .eq('id', resource.id)
  .gte('quantity', amount)   // ← atomic guard: update only if still sufficient
  .select('id');

if (!data?.length) {
  throw new Error(`Insufficient ${resourceName} or concurrent update conflict`);
}
```

By adding `.gte('quantity', amount)` to the update filter, the write becomes a no-op if another request already decremented the quantity. Checking that at least one row was updated (`data?.length`) catches the race. No separate read needed for the write.

---

### 🟡 MEDIUM — Equipment deleted before resource returned in `sellEquipmentFromFighter`

**File:** `app/actions/sell-equipment.ts` (~line 235)

```ts
// 1. Equipment deleted
const { error: deleteError } = await supabase.from('fighter_equipment').delete()...
if (deleteError) throw new Error(...);

// 2. Resource returned AFTER delete — error swallowed
if (isResourcePurchase) {
  try {
    await returnGangResource(...);
  } catch (resourceError) {
    console.error('Failed to return resource on sell:', resourceError); // swallowed
  }
}
```

If `returnGangResource` throws (gang left campaign, resource row missing, DB error), the equipment is already gone but the resource is never returned. The player loses both the item and the resource.

Compare with `sellEquipmentFromStash` where the resource return correctly happens **before** the delete. The same ordering should be used here — move the `returnGangResource` block to before the `fighter_equipment` delete.

---

### 🟡 MEDIUM — `returnGangResource` looks up the wrong campaign when gang has history

**File:** `utils/campaigns/resources.ts` (~line 150)

```ts
const { data: campaignGang } = await supabase
  .from('campaign_gangs')
  .select('id')
  .eq('gang_id', gangId)
  .limit(1)
  .single();
```

This returns whichever `campaign_gangs` row comes back first for the gang. If the gang has previously been in another campaign, the lookup is ambiguous — the resource may be credited to the wrong campaign.

The root cause is that `cost_resource` only stores `{ name, amount }`, not `campaign_gang_id`, so the return path has no authoritative reference.

**Fix:** Store `campaign_gang_id` in the `cost_resource` JSON at purchase time:
```ts
// equipment.ts, when writing cost_resource:
cost_resource: {
  name: params.cost_resource_name,
  amount: params.cost_resource_amount,
  campaign_gang_id: params.campaign_gang_id,  // add this
}
```

Then `returnGangResource` can accept `campaignGangId` directly (matching the signature of `deductGangResource`) and skip the unreliable lookup entirely.

---

### 🟡 MEDIUM — No server-side validation that `campaign_gang_id` belongs to `gang_id`

**File:** `app/actions/equipment.ts` (~line 367)

The server verifies the gang's credits, but never checks that `params.campaign_gang_id` actually belongs to `params.gang_id`. A user who is a member of two gangs in the same campaign could supply the `campaign_gang_id` of the other gang and deduct resources from it.

**Fix:** Add a server-side assertion before calling `deductGangResource`:
```ts
const { data: cgRow } = await supabase
  .from('campaign_gangs')
  .select('gang_id')
  .eq('id', params.campaign_gang_id)
  .single();

if (cgRow?.gang_id !== params.gang_id) {
  throw new Error('campaign_gang_id does not match the purchasing gang');
}
```

---

### 🟢 LOW — `findGangResourceByName` fetches all resources and JS-filters

**File:** `utils/campaigns/resources.ts` (~line 95)

The function fetches every `campaign_gang_resources` row for the gang, then filters by name in JS using `.find()`. For gangs with many resource types this is wasteful, and the `any` type on `r` loses type safety. A DB-level filter or typed return would be cleaner — minor, but worth a tidy-up.

---

### 🟢 LOW — `campaignGangId` always uses `campaigns[0]`

**Files:** `components/fighter/fighter-page.tsx`, `components/gang/gang-page-content.tsx`

```ts
campaignGangId={(fighterData.fighter.campaigns || []).length > 0
  ? (fighterData.fighter.campaigns || [])[0]?.campaign_gang_id
  : undefined}
```

If a gang is ever in multiple campaigns simultaneously, this silently picks the wrong one. If gangs are constrained to one active campaign at a time, this is fine — worth a short comment confirming that assumption so it doesn't silently break later.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 HIGH | Zeroing the resource amount input skips all cost checks → free equipment |
| 2 | 🔴 HIGH | TOCTOU race in `deductGangResource`/`returnGangResource`: concurrent requests double-deduct |
| 3 | 🟡 MEDIUM | Equipment deleted before resource returned in `sellEquipmentFromFighter` — resource lost on error |
| 4 | 🟡 MEDIUM | `returnGangResource` resolves campaign via unsorted `gangId` lookup — wrong campaign for gangs with history |
| 5 | 🟡 MEDIUM | No server-side check that `campaign_gang_id` belongs to the purchasing gang |
| 6 | 🟢 LOW | `findGangResourceByName` fetches all resources and JS-filters, uses `any` |
| 7 | 🟢 LOW | `campaigns[0]` assumption for `campaignGangId` is silently fragile |

---

## Verdict: Changes Needed

Issues #1 and #2 are security/correctness bugs that should be fixed before merge. Issue #3 is a data-loss scenario on sell. Issue #4 produces wrong behaviour for gangs with campaign history. Recommend fixing #1–#4 together; #5–#7 can be follow-ups.

---

joeseos (commented):
Second pass review — second commit adding the custom trading post resource UI and Reputation support.

## First pass recap: what's fixed ✅

All five issues from the original review have been addressed:
- ✅ Free-equipment exploit: `isResourcePurchase` now guards `> 0`, `deductGangResource` checks `amount <= 0`
- ✅ TOCTOU in `deductGangResource`: atomic `.gte('quantity', amount)` on the write, with row-count check
- ✅ Sell ordering in `sellEquipmentFromFighter`: resource returned before equipment deleted
- ✅ `returnGangResource` now accepts `campaignGangId` (stored in `cost_resource` JSONB at purchase time)
- ✅ Server-side cross-gang validation: `campaign_gang_id` is now confirmed to belong to `gang_id` before deducting

Good work on those — the atomic guard pattern is solid.

---

## New findings (second commit)

---

### 🔴 HIGH — Wrong column name in all reputation functions: every reputation purchase will fail

**File:** `utils/campaigns/resources.ts` — `deductGangReputation` and `returnGangReputation`

Both functions query the `gangs` table using `.eq('gang_id', gangId)`:

```ts
// deductGangReputation
const { data: gang, error: fetchError } = await supabase
  .from('gangs')
  .select('reputation')
  .eq('gang_id', gangId)  // ← wrong column
  .single();
```

```ts
// returnGangReputation
const { data: gang, error: fetchError } = await supabase
  .from('gangs')
  .select('reputation')
  .eq('gang_id', gangId)  // ← wrong column
  .single();
```

The `gangs` table primary key is `id`, not `gang_id` — consistent with every other gang query in the codebase (e.g. the gang fetch earlier in `buyEquipmentForFighter` uses `.eq('id', params.gang_id)`). `campaign_gangs` has a `gang_id` FK column, but the `gangs` table itself does not.

PostgREST will return a `PGRST204` error (`Column 'gang_id' of relation 'gangs' does not exist`), which means:
- `deductGangReputation` throws "Failed to fetch gang: …" on every call → **all Reputation purchases fail**
- `returnGangReputation` hits `if (fetchError || !gang) return false` → **reputation is silently never returned on sell**

Also affected: the `deductGangReputation` update uses `.eq('gang_id', gangId).select('gang_id')` — both wrong.

**Fix:** Replace every `.eq('gang_id', gangId)` with `.eq('id', gangId)` and `.select('gang_id')` with `.select('id')` in both functions:

```ts
// deductGangReputation — corrected
const { data: gang, error: fetchError } = await supabase
  .from('gangs')
  .select('reputation')
  .eq('id', gangId)        // ← correct
  .single();
// ...
const { data, error } = await supabase
  .from('gangs')
  .update({ reputation: current - amount })
  .eq('id', gangId)        // ← correct
  .gte('reputation', amount)
  .select('id');           // ← correct
```

---

### 🟡 MEDIUM — TOCTOU on the resource/reputation *return* path (concurrent sells)

**File:** `utils/campaigns/resources.ts` — `returnGangResource` and `returnGangReputation`

`deductGangResource` and `deductGangReputation` both have the atomic guard (`.gte()` + row count check). The *return* functions do not — they read the current value and write back a computed result:

```ts
// returnGangResource (still read-then-write)
const resource = await findGangResourceByName(...);   // reads quantity
// ...
.update({ quantity: resource.quantity + amount })     // computed in JS

// returnGangReputation (same pattern once column name is fixed)
const { data: gang } = await supabase.from('gangs').select('reputation')...;
.update({ reputation: (gang.reputation ?? 0) + amount })
```

Two concurrent sells of the same resource-cost item will both read the same `quantity`/`reputation`, both add `amount`, and both write the same value — returning only one copy of the resources instead of two.

For the `gangs` table there's no native atomic increment in Supabase without an RPC, but `campaign_gang_resources` updates can at least drop the pre-read by using a supabase RPC or by removing the JS arithmetic. The safest fix is a small DB function:
```sql
CREATE OR REPLACE FUNCTION increment_gang_resource(row_id uuid, delta numeric)
RETURNS void LANGUAGE sql AS $$
  UPDATE campaign_gang_resources SET quantity = quantity + delta WHERE id = row_id;
$$;
```

Or at minimum document that the return path accepts eventual consistency (rare double-sell scenario).

---

### 🟡 MEDIUM — Admin can create a "zombie" resource-cost item with no enforcement

**File:** `components/customise/custom-trading-posts.tsx` — `EditEquipmentModal.handleSave`

```ts
const handleSave = async () => {
  onSaveLocal({
    costOverride: costOverride.trim() ? Number(costOverride) : null,
    costResourceName: costResourceName || null,
    ...
  });
};
```

If an admin picks a `costResourceName` from the dropdown but leaves the amount field empty, `costOverride` saves as `null`. The SQL function then returns `cost_resource_amount = null` for this item (because `CASE WHEN cost_resource_name IS NOT NULL THEN cost_override END` → `null`).

In the purchase modal:
```ts
const hasResourceCost = Boolean(item.cost_resource_name && item.cost_resource_amount != null);
// → Boolean('Exploration' && false) → false
```

The item shows no resource badge, the resource input is hidden, and no `cost_resource_name` is sent to the server — so `isResourcePurchase` is `false` and the item is purchased with credits at its base price instead of the intended resource cost. The item is misconfigured silently.

**Fix:** In `handleSave`, validate before saving:
```ts
if (costResourceName && !costOverride.trim()) {
  // show an error: "Resource amount is required when a resource is selected"
  return;
}
```

---

### 🟡 MEDIUM — Purchase modal still doesn't validate `manualResourceAmount > 0` client-side

**File:** `components/equipment/purchase-modal.tsx`

With the server-side fix, clearing the resource amount field (`manualResourceAmount = ""`) sends `cost_resource_amount: 0`, which now makes `isResourcePurchase = false` on the server. The purchase then falls through to the credit check and charges credits instead of resources — wrong behaviour, no user-facing error.

The `handleConfirm` validation currently handles credits:
```ts
} else if (!hasResourceCost && parsedCost > 0 && parsedCost > gangCredits) {
  setCreditError(`Not enough credits...`);
}
```

But has no equivalent for the resource amount. Add before the `setCreditError(null)` line:
```ts
if (hasResourceCost) {
  const parsedResource = Number(manualResourceAmount);
  if (!parsedResource || parsedResource <= 0) {
    setCreditError('Resource amount must be greater than 0');
    return false;
  }
}
```

---

### 🟢 LOW — `returnGangResource` fallback still ambiguous for legacy purchases

**File:** `utils/campaigns/resources.ts` — `returnGangResource`

Items purchased before this PR don't have `campaign_gang_id` in their stored `cost_resource` JSON, so `returnGangResource` falls back to the `campaign_gangs` lookup via `gangId`. That `.limit(1).single()` path is still ambiguous if a gang has previous campaign history. This is acceptable as a migration trade-off (old items are rare), but worth a comment explaining the fallback limitation.

---

### 🟢 LOW — Migration comment outdated

**File:** `supabase/migrations/20260603120000_add_cost_resource_to_fighter_equipment.sql`

The column comment says `{"name": "...", "amount": ...}` but the actual stored shape now conditionally includes `campaign_gang_id`. Minor but the comment should match:
```sql
IS 'Resource used to pay, e.g. {"name": "Exploration Points", "amount": 3, "campaign_gang_id": "uuid"}. campaign_gang_id omitted for Reputation. Null = credits.';
```

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 HIGH | `deductGangReputation`/`returnGangReputation` use `.eq('gang_id')` — wrong column, all Reputation purchases fail |
| 2 | 🟡 MEDIUM | `returnGangResource`/`returnGangReputation` TOCTOU on return path — concurrent sells under-return |
| 3 | 🟡 MEDIUM | No validation in `EditEquipmentModal.handleSave` — admin can silently create resource-cost items with no amount |
| 4 | 🟡 MEDIUM | Purchase modal missing client-side guard for `manualResourceAmount > 0` — falls back to credit purchase silently |
| 5 | 🟢 LOW | `returnGangResource` fallback ambiguous for legacy items without `campaign_gang_id` in JSONB |
| 6 | 🟢 LOW | Migration column comment doesn't reflect updated JSONB shape |

## Verdict: Changes Needed

Issue #1 is a hard blocker — Reputation as a resource type is completely broken at runtime. Issue #3 is also a must-fix since it can silently misconfigure items. Issues #2 and #4 should be fixed before merge; #5–#6 can be follow-ups.

---

joeseos (commented):
## PR Review — Added support to buy equipment with resources (UUID refactor pass)

@joeseos The UUID refactor is the right call — replacing the `cost_resource_name` text field with FK references to `campaign_type_resources` and `campaign_resources` makes the cost definition type-safe and referentially correct. The DB constraint `chk_cost_resource_exclusive` and the `.gte()` guard on deduct for optimistic concurrency protection are both well done. One correctness bug needs fixing before merge.

---

### 🔴 Medium — Resource returned but result ignored; equipment still deleted on failure

**`app/actions/sell-equipment.ts` ~line 228 (fighter sell) and ~line 430 (stash sell)**

`returnGangResource` and `returnGangReputation` both return `Promise<boolean>` — they **never throw**, they return `false` on failure. The sell code wraps them in `try/catch` (which only catches the `findGangResourceById` DB error path inside `returnGangResource`), but in all cases execution falls through to the equipment delete regardless:

```ts
try {
  await returnGangReputation(supabase, gangId, costResource.amount); // returns false on failure, doesn't throw
} catch (resourceError) {
  console.error('Failed to return resource on sell:', resourceError); // catches only findGangResourceById throws
  // execution continues regardless
}

// Delete happens unconditionally ↓
const { error: deleteError } = await supabase.from('fighter_equipment').delete()...
```

Two failure paths both silently eat the resource:
1. `findGangResourceById` throws (DB error) → caught → delete proceeds → resource lost
2. `returnGangResource` returns `false` (resource row not found, or update DB error) → not caught, just ignored → delete proceeds → resource lost

The code comment says *"Return resource before deleting equipment (so resource isn't lost if return fails)"* — but the implementation doesn't enforce this. The same issue exists in the stash sell path.

**Suggested fix:**
```ts
const returned = costResource.name === REPUTATION_RESOURCE_NAME
  ? await returnGangReputation(supabase, gangId, costResource.amount)
  : await returnGangResource(supabase, gangId, ...);

if (!returned) {
  return { success: false, error: 'Failed to return resource — item not sold. Please try again.' };
}
```
Or align the return functions to throw like their deduct counterparts (see consistency note below), which would make the existing try/catch work as intended.

---

### 🟡 Low — Inconsistent error contract between deduct and return functions

**`utils/campaigns/resources.ts`**

`deductGangResource` and `deductGangReputation` throw on failure.
`returnGangResource` and `returnGangReputation` return `boolean` on failure.

This asymmetry means callers must handle them differently, and is directly what caused the bug above. Making the return functions throw (matching the deduct pattern) would be the smaller change and fix the sell bug as a side effect.

---

### 🟡 Low — `returnGangResource` increment is not concurrency-safe

**`utils/campaigns/resources.ts` ~line 158**

The deduct path correctly uses `.gte('quantity', amount)` to prevent overdraft under concurrent writes. The return path does not use an equivalent guard — it reads quantity then writes `quantity + amount`, so two concurrent sells race and only one increment is stored. In practice a fighter can't sell the same item twice simultaneously, but worth noting if this pattern gets reused.

---

### 🟡 Low — `campaignGangId` uses only the first campaign

**`components/fighter/fighter-page.tsx` ~line 1362 and `components/gang/gang-page-content.tsx` ~line 913**

If a gang is in multiple campaigns, `[0]?.campaign_gang_id` always picks the first one. Consistent with existing prop patterns in the file, so not a regression — just a known limitation for multi-campaign scenarios.

---

### ✅ What's done well

- **UUID FK refactoring**: replacing `cost_resource_name text` with `cost_type_resource_id uuid` / `cost_campaign_resource_id uuid` with `ON DELETE SET NULL` FK references is the right approach.
- **DB exclusivity constraint**: `chk_cost_resource_exclusive` correctly ensures at most one resource type is set per equipment entry.
- **`campaign_gang_id` ownership check in buy action**: verifying `.eq('gang_id', params.gang_id)` prevents cross-gang resource deduction.
- **`.gte()` guard on deduct**: correctly prevents overdraft under concurrent writes.
- **`costResourcePayload` spread into both stash and fighter insert paths**: both paths record `cost_resource` JSONB consistently.
- **`creditsDelta: isResourcePurchase ? 0 : ...`**: correctly skips credit deduction for resource purchases.
- **SQL function updated for both standard and custom equipment branches**: both the `equipment` and `custom_equipment` UNIONs in `get_equipment_detailed_data.sql` are updated consistently.

---

**Verdict: Changes needed before merge** — the sell path bug (resource silently lost when return fails) is the blocker. Aligning the return functions to throw would fix both the bug and the inconsistency in one shot.


---

joeseos (commented):
## Code Review — Buy Equipment with Resources

Solid feature overall — the resource/reputation purchase flow, the SQL changes, and the UI integration are well thought out. But there are a few correctness and security issues that should be addressed before merge.

---

### 🔴 HIGH — Resource deducted before equipment insert (data loss risk)

**`app/actions/equipment.ts` ~line 365 (PR)**

The PR deducts the resource **before** inserting the equipment record. If the insert fails (constraint violation, FK error, etc.), the resource is gone but no equipment was created. The user loses resources with nothing to show for it.

Compare with the existing credit flow: credits are only **checked** before insert, then **deducted** via `updateGangFinancials()` after the insert succeeds. The resource flow should follow the same pattern.

**Fix:** Move the resource deduction to after the equipment insert succeeds (around where `updateGangFinancials` is called), or wrap the insert in a try/catch that returns the resource on failure.

---

### 🔴 HIGH — Resource returned before equipment delete (duplication exploit)

**`app/actions/sell-equipment.ts` ~line 228 and ~line 433 (PR)**

Both `sellEquipmentFromFighter` and `sellEquipmentFromStash` return the resource **before** deleting the equipment. If the delete fails, the user gets resources back AND keeps the equipment.

**Fix:** Move the resource return to after the delete succeeds, or catch delete failures and re-deduct the returned resource.

---

### 🔴 HIGH — Wealth inflation on resource purchases

**`app/actions/equipment.ts` ~line 634 (PR)**

The wealth formula is `wealthDelta = ratingDelta + creditsDelta + stashValueDelta`. For credit purchases, the `+ratingDelta` and `-creditsDelta` cancel out (wealth-neutral). For resource purchases, `creditsDelta = 0` but `ratingDelta` still adds the full equipment cost — so wealth inflates by `ratingCost` per purchase.

Example: Buy 5 items worth 230 credits total using resources → gang wealth is overstated by 230 while those items are held. Round-trip is neutral (sell reverses it), but the holding-period inflation affects campaign standings if wealth is used for matchmaking.

**Fix:** Either skip the rating/stash contribution to wealth for resource purchases, or explicitly document this as intentional game design.

---

### 🔴 HIGH — New API route missing authentication

**`app/api/custom-trading-posts/[id]/shared-campaigns/route.ts` (new file)**

This route has no auth check. Every other API route that queries `custom_shared` calls `getUserIdFromClaims()` first (see `skill-types`, `gang-types`, `fighter-types` routes). The `createClient()` uses the anon key — RLS is the data-access layer, not an authentication layer.

**Fix:** Add the standard auth gate:
```typescript
const userId = await getUserIdFromClaims(supabase);
if (!userId) {
  return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
}
```

---

### 🟡 MEDIUM — `bool_or(cost_reputation)` inconsistent with deterministic aggregation

**`supabase/functions/get_equipment_detailed_data.sql` ~line 234 (PR)**

`cost_type_resource_id` and `cost_campaign_resource_id` are picked deterministically using `array_agg(...ORDER BY sort_order...)[1]` — i.e., the "winning" row by sort order. But `cost_reputation` uses `bool_or()`, which returns `true` if **any** row has reputation, even if the winning row doesn't.

If equipment appears in two trading posts — one charging reputation, one charging a resource — the output could show `cost_reputation = true` AND `cost_type_resource_id = <some id>`, which is contradictory.

**Fix:** Use the same `[1]` aggregation pattern: `(array_agg(ctpe.cost_reputation ORDER BY ...))[1]` to pick the value from the winning row consistently.

---

### 🟡 MEDIUM — Duplicated resource-return blocks (DRY violation)

**`app/actions/sell-equipment.ts` ~line 228 and ~line 433**

The 15-line resource-return block (cast JSONB → branch reputation vs resource → call return function → catch error) is copy-pasted between `sellEquipmentFromFighter` and `sellEquipmentFromStash`.

**Fix:** Extract a `returnCostResource(supabase, gangId, costResource)` helper into `utils/campaigns/resources.ts`.

---

### 🟡 MEDIUM — `onConfirm` callback grew to 10 positional parameters

**`components/equipment/purchase-modal.tsx` (all 7 `onConfirm` call sites)**

The 4 new resource args (`costResourceName`, `costResourceAmount`, `costTypeResourceId`, `costCampaignResourceId`) are constants derived from `item` props — they never change during the modal lifecycle. Yet they're threaded through 7 call sites. The consumer in `equipment.tsx` already has the `item` as `buyModalData`.

**Fix:** Remove the 4 resource params from `onConfirm`. Read them from `buyModalData` in `equipment.tsx` when calling `purchaseEquipment()`.

---

### 🟡 MEDIUM — Repeated campaign data extraction in JSX

**`components/fighter/fighter-page.tsx` ~lines 1350-1395, `components/gang/gang-page-content.tsx` ~lines 901-920**

The expression `(fighterData.fighter.campaigns || []).length > 0 ? (fighterData.fighter.campaigns || [])[0]?.campaign_gang_id : undefined` (and variants) appears 14 times across 3 blocks. The PR adds 6 new instances to an already-duplicated pattern.

**Fix:** Extract campaign data into a local variable above the JSX and spread it onto both ItemModals.

---

### 🟢 LOW — `resourceName` parameter is YAGNI

**`utils/campaigns/resources.ts` — `deductGangResource` and `returnGangResource`**

Both functions accept `resourceName` only for error messages. The actual lookup uses `campaignTypeResourceId`/`campaignResourceId`. This threads a display string through multiple call layers unnecessarily.

**Fix:** Remove `resourceName` from signatures; use a generic message or look up the name from the already-fetched resource row.

---

### 🟢 LOW — No shared type for `cost_resource` JSONB payload

**`app/actions/equipment.ts`, `app/actions/sell-equipment.ts`**

The `cost_resource` JSONB shape `{ name, amount, campaign_gang_id?, ... }` is defined inline as type assertions in two files and constructed inline in a third. A shared `CostResourcePayload` interface in `types/equipment.ts` would improve type safety.

---

### 🟢 LOW — Race condition on return functions (concurrent sells)

**`utils/campaigns/resources.ts` — `returnGangResource` and `returnGangReputation`**

The deduct functions correctly use `.gte('quantity', amount)` as an optimistic concurrency guard. The return functions have no such guard — two concurrent returns could race, with the first return being silently lost. Low risk in practice (two people can't sell the same item), but worth noting for consistency.


---

PR #1733, opened 2026-06-03
Updated UI and UX elements that were standing out - 1st pass

Djidiouf:


joeseos (changes requested):
Hey @Djidiouf — good cleanup overall. The removal of `PendingEquipmentSection`, the two-step create-then-add UX flow, and the override preview in `displayItems` are all solid improvements. A couple of correctness issues need addressing before this lands.

---

## Findings

### 🔴 MEDIUM — Potential null dereference on `inserted` after batch insert
**File:** `app/actions/customise/custom-trading-posts.ts` — `addTPEquipmentBatch`

```ts
const { data: inserted, error: insertError } = await supabase
  .from('custom_trading_post_equipment')
  .insert(rows)
  .select('id');

if (insertError) throw insertError;

return { success: true, data: inserted.map(r => r.id) };  // ← inserted can be null
```

Supabase types `data` as `T[] | null` even when there's no error (e.g. RLS policy silently rejects inserts). Calling `.map()` on `null` will throw an unhandled exception that escapes the `try/catch` because it happens after the error check.

**Fix:**
```ts
return { success: true, data: (inserted ?? []).map(r => r.id) };
```

---

### 🔴 MEDIUM — Index-based mapping between `additionsToSave` and `batchResult.data` is fragile
**File:** `components/customise/custom-trading-posts.tsx` — `editMutation` / save handler

```ts
const pendingRulesSaves = additionsToSave
  .map((equip, i) => ({ tempId: equip.id, realId: batchResult.data[i] }))
  .filter(({ tempId }) => overridesToSave.get(tempId)?.rulesModified)
  .map(async ({ tempId, realId }) => { ... });
```

This assumes Supabase returns inserted rows in the same order as they were submitted. While Supabase typically preserves insertion order, this is not documented as a guarantee. If the order diverges (e.g. due to a trigger, future PG version behaviour, or a partial-insert retry), rules will be silently associated with the **wrong** equipment — data corruption with no error surfaced.

**Fix:** Select a stable identifier back alongside `id` (e.g. `equipment_id` and `custom_equipment_id`) so you can match by identity rather than position:

```ts
// In addTPEquipmentBatch, change:
.select('id, equipment_id, custom_equipment_id')

// Return typed: { id, equipment_id, custom_equipment_id }[]
// Then match in the component:
additionsToSave.map(equip => {
  const realRow = batchResult.data!.find(r =>
    equip.is_custom
      ? r.custom_equipment_id === equip.original_id
      : r.equipment_id === equip.id
  );
  return { tempId: equip.id, realId: realRow?.id };
})
```

---

### 🟡 LOW — `descCharCount` is derived state that can drift from `formData.description`
**File:** `components/customise/custom-trading-posts.tsx`

`descCharCount` is a separate `useState` that must be manually kept in sync with `formData.description` in three different places (`handleEdit`, `handleView`, `resetForm`, and the textarea `onChange`). It's computable directly:

```tsx
// Replace the useState + all the setDescCharCount calls with:
const descCharCount = formData.description?.length ?? 0;
```

This eliminates the possibility of them diverging on any future code path that touches `formData`.

---

### 🟡 LOW — Hardcoded tooltip `id` breaks if `EquipmentItemsSection` is mounted more than once
**File:** `components/customise/custom-trading-posts.tsx` — `EquipmentItemsSection`

```tsx
data-tooltip-id="equipment-items-tooltip"
// ...
<Tooltip id="equipment-items-tooltip" ... />
```

`react-tooltip` matches anchors to `<Tooltip>` by ID. If this component is ever rendered twice simultaneously (e.g. edit modal open while another panel exists), the second `<Tooltip>` will steal the ID and the first instance won't work.

**Fix:** Use a stable unique ID per instance, e.g. via `useId()` (React 18):
```tsx
const tooltipId = useId();
// use `tooltipId` for both data-tooltip-id and <Tooltip id>
```

---

### 🟡 LOW — `DESCRIPTION_MAX_LENGTH` is only enforced client-side
**File:** `app/actions/customise/custom-trading-posts.ts` — `createCustomTradingPost` / `updateCustomTradingPost`

The 600-char limit exists only in the React form. A user hitting the server action directly (e.g. via the browser console or a curl) can submit an unbounded description. If the DB column has no length constraint this is a mild DoS / data quality issue.

**Fix:** Add a guard at the top of each action:
```ts
if (data.description && data.description.length > 600) {
  return { success: false, error: 'Description must be 600 characters or fewer.' };
}
```

---

## Non-issues / positives
- Removing `pendingEquipment` state and `PendingEquipmentSection` is the right call — the two-step flow is cleaner and the amber hint text communicates it well.
- `pendingAdditionTempIds` filter on `equipmentSaves` is correct — avoids double-saving overrides for newly added items.
- The `displayItems` update to reflect pending overrides immediately in the table is a nice UX improvement.
- Column width swap (30%/60%) makes sense given description is now the more content-rich column.

---

The null-dereference on `inserted` (finding 1) and the ordering-assumption for rules association (finding 2) are both data-correctness bugs that should be fixed before merge. The others are lower priority but worth addressing, especially `descCharCount` as derived state.


---

joeseos:

Reviewed the new commit (`bbf8527`) against the previous findings.

---

### Previous findings: status

| 1 | MEDIUM | Null dereference on `inserted` in `addTPEquipmentBatch` | ✅ Resolved — `(inserted ?? []).map(...)` applied |
| 2 | MEDIUM | Index-based mapping between `additionsToSave` and `batchResult.data` | ✅ Resolved — now identity-matched via `.find(r => r.equipment_id === equip.id \|\| r.custom_equipment_id === equip.original_id)` |
| 3 | LOW | `descCharCount` as driftable `useState` | ✅ Resolved — replaced with `const descCharCount = formData.description?.length ?? 0` |
| 4 | LOW | Hardcoded tooltip ID breaks on multiple instances | ✅ Resolved — `useId()` used |
| 5 | LOW | `DESCRIPTION_MAX_LENGTH` only enforced client-side | ✅ Resolved — server-side guard added in both `createCustomTradingPost` and `updateCustomTradingPost`; constant extracted to shared `custom-trading-posts-constants.ts` |

All five findings resolved. The identity-matching fix for finding 2 is particularly clean — passing `equipment_id`/`custom_equipment_id` back from the select and using `.find()` is exactly the right approach.

---

### New findings

**LOW — `<label>` missing `htmlFor` (minor accessibility regression)**
`components/customise/custom-trading-posts.tsx` — `renderForm`

The description label was switched from the `<Label>` component to a native `<label>` element (to support the inline char counter), but no `htmlFor` was added:

```tsx
<label className="flex justify-between items-center text-sm font-medium mb-1">
  <span>Description</span>
  ...
</label>
<Textarea ... />          {/* no matching id */}
```

Clicking "Description" won't focus the textarea. Since `renderForm` is rendered inside modals (single instance at a time), a stable `id` pair is fine here:

```tsx
<label htmlFor="tp-description" ...>
<Textarea id="tp-description" ... />
```

Not a blocker, but worth a quick fix.

---

### Non-issues (checked and confirmed clean)

- Opening the edit/overrides modal for a **pending addition** (temp UUID) calls `getAvailabilityRules` / `getPricingRules` with the temp ID → Supabase returns an empty array (no error). The modal renders correctly as a blank-slate rules editor, and rules are persisted via `saveEquipmentRules(realId, ...)` post-batch-insert. Flow is correct end-to-end.
- `pendingAdditionTempIds` scoping in the save callback is correct — declared before the conditional block, used after. When `additionsToSave` is empty the Set is empty and the filter is a no-op.
- `costOverride != null` (loose null check in `addTPEquipmentBatch`) correctly catches both `null` and `undefined`, so unset overrides are cleanly excluded from the insert row.

---

### Verdict: ✅ Approve

All medium-severity correctness issues are fixed. The single new LOW finding (missing `htmlFor`) is a minor accessibility nit that won't block merge. Nice work addressing all five points.

---

joeseos (approved):
All five previous findings resolved. One new LOW nit (missing `htmlFor` on the description label). Safe to merge — see inline comment for details.

---

joeseos (approved):

Reviewed commit `9ac150a` (the only new change since the previous APPROVE on `bbf8527`).

---

### Previous finding: status

| 6 | LOW | `<label>` missing `htmlFor` — clicking "Description" didn't focus the textarea | ✅ Resolved — `htmlFor="tp-description"` added to the label, `id="tp-description"` added to the `<Textarea>` |

---

### New findings

None. The commit is 3 lines and does exactly what was requested. I verified that `renderForm` is called in mutually exclusive modal branches (`isAddModalOpen`, `editModalData`, `viewModalData`), so only one instance of `id="tp-description"` is ever in the DOM at a time — the static ID pair is safe.

---

### Verdict: ✅ Approve

All six findings across two review rounds are resolved. Good to merge.

---

PR #1734, opened 2026-06-03
Fixed the availability dropdown and minor UI issues

Djidiouf:


joeseos:
## Code Review — PR #1734: Fixed the availability dropdown and minor UI issues

Hey @Djidiouf, thanks for the PR! The core idea is solid — adding `'S'` as a valid availability letter and making `'I'` number-bearing (consistently with how `equipment.tsx` already renders it). Here's what I found:

---

### 🔴 Medium — Backward-compat regression: `parseAvailability('I')` now returns empty letter

**File:** `components/ui/availability-picker.tsx`

The old code had an explicit guard:
```ts
if (availability === 'I') return { letter: 'I', number: 6 };
```

This PR removes it. Now `parseAvailability('I')` hits none of the early-exit checks and fails to match the new regex `^([CREIS])(\d+)$` (because there's no trailing digit), so it falls through to:
```ts
return { letter: '', number: 6 };
```

**Impact:** Any existing equipment row in the database with `availability = 'I'` (standalone, no number — which is what the old `combineAvailability` used to write) will now load with a **blank availability letter** in the admin-edit and custom-equipment edit forms. A user saving the form without noticing would silently overwrite `'I'` with `'C'` (the `|| 'C'` fallback).

Note that `equipment.tsx` already expected `'I9'`-style format (lines 531, 576: `/^[RIS]\d+$/`), so standalone `'I'` was already broken in the equipment list filter — but the round-trip through the edit forms was the last place it was preserved.

**Suggested fix** — either keep the backward-compat guard in `parseAvailability`:
```ts
if (availability === 'I') return { letter: 'I', number: 6 };  // legacy: re-saves as I6
```
or pair this PR with a DB migration:
```sql
UPDATE equipment SET availability = 'I6' WHERE availability = 'I';
-- (repeat for any other tables that store availability)
```

---

### 🟡 Low — No DB migration for the `'I'` format change

**File:** n/a (database)

`combineAvailability('I', 6)` previously returned `'I'`; it now returns `'I6'`. Any existing rows stored as `'I'` won't round-trip correctly unless they're migrated. A migration alongside this PR would close the gap cleanly.

---

### 🟢 What's good

- **`'S'` support is complete and consistent** — all four files (`availability-picker.tsx`, `admin-create-equipment.tsx`, `admin-edit-equipment.tsx`, `custom-equipment.tsx`) were updated together. ✓
- **`parseAvailability` regex** correctly extended from `[CREI]` to `[CREIS]`. ✓
- **Conditional rendering of the number picker** (shown only for R/I/S, hidden for C/E) is a clean improvement over the previous always-rendered-but-disabled approach. ✓
- **`equipment.tsx` was already aligned** — color badges and filters already handled `'S'` and expected `'I{n}'` format, so no changes needed there. ✓
- **UI label renames** (`Pricing Rules → Cost Rules`, `Adjusted Cost → Cost`) are consistent; minor note that `"AL"` as a column header is a fairly opaque abbreviation — `"Avail."` was more self-describing, but this is a judgment call.

---

The `'I'` backward-compat issue is the only blocking item. The simplest fix is adding back the `parseAvailability('I')` guard (one line). If there are definitely no `'I'`-only rows in the database, a DB migration instead would be the cleaner long-term solution.


---

PR #1736, opened 2026-06-04
Added Archetype selection in Add Fighter and updated Edit Fighter to …

Djidiouf:
…have a Combobox for that matter

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -25,6 +25,7 @@ interface AddFighterParams {
   use_base_cost_for_rating?: boolean;
   use_delegation_cost?: boolean;
   fighter_gang_legacy_id?: string;
+  selected_archetype_id?: string;
 }
 
 interface FighterStats {
@@ -445,6 +446,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
       kills: 0,
       special_rules: effectiveFighterData.special_rules,
       fighter_gang_legacy_id: params.fighter_gang_legacy_id || null,
+      selected_archetype_id: params.selected_archetype_id || null,
       user_id: gangData.user_id
     };
 
@@ -473,6 +475,43 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     }
 
     const fighterId = insertedFighter.id;
+
+    // Apply archetype skill-access overrides when an archetype is selected
+    if (params.selected_archetype_id) {
+      try {
+        const { data: archetypeData } = await supabase
+          .from('skill_access_archetypes')
+          .select('skill_access')
+          .eq('id', params.selected_archetype_id)
+          .single();
+
+        if (archetypeData?.skill_access && Array.isArray(archetypeData.skill_access)) {
+          const overrideRows = (archetypeData.skill_access as Array<{
+            skill_type_id: string;
+            access_level: 'primary' | 'secondary';
+          }>)
+            .filter(sa => sa.access_level === 'primary' || sa.access_level === 'secondary')
+            .map(sa => ({
+              fighter_id: fighterId,
+              skill_type_id: sa.skill_type_id,
+              access_level: sa.access_level,
+              user_id: gangData.user_id,
+            }));
+
+          if (overrideRows.length > 0) {
+            const { error: overrideError } = await supabase
+              .from('fighter_skill_access_override')
+              .insert(overrideRows);
+
+            if (overrideError) {
+              console.error('Failed to insert archetype skill-access overrides:', overrideError);
+            }
+          }
+        }
+      } catch (archetypeError) {
+        console.error('Error applying archetype skill-access overrides:', archetypeError);
+      }
+    }
 
     // Get default skills and equipment from fighter_defaults table
     let fighterDefaultsData: any[] = [];
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -6,6 +6,10 @@ import { invalidateFighterAddition, invalidateUserGangsList } from '@/utils/cach
 import { createExoticBeastsForEquipment } from '@/utils/exotic-beasts';
 import { updateGangFinancials } from '@/utils/gang-rating-and-wealth';
 import { logFighterAction } from '@/app/actions/logs/fighter-logs';
+import {
+  isArchetypeEligible,
+  mapArchetypeSkillAccessToOverrides,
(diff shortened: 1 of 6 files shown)


joeseos (changes requested):
## PR Review — Archetype selection in Add Fighter + Combobox in Edit Fighter

@Djidiouf Good direction — extracting `isArchetypeEligible` into a shared utility and migrating to `Combobox` are both solid improvements. A few issues need addressing before merge.

---

### 🔴 Medium — Silent failure when archetype overrides fail on add

**`app/actions/add-fighter.ts` ~line 507**

```ts
if (overrideError) {
  console.error('Failed to insert archetype skill-access overrides:', overrideError);
  // ← fighter already committed, user gets no feedback
}
```

The fighter is already inserted at this point. If the override insert fails, the user silently ends up with a fighter that has the wrong skill access — no toast, no error returned. The edit path correctly handles this with `toast.error(...)`. The add path should do the same.

**Suggested fix:** Return an appropriate error or, at minimum, surface a user-facing message (same pattern as the edit modal's "Fighter updated but skill access save failed..." toast).

---

### 🔴 Medium — No server-side archetype eligibility validation

**`app/actions/add-fighter.ts` ~line 480**

The eligibility check (`isArchetypeEligible`) lives only in the UI. The server action accepts any `selected_archetype_id` for any fighter in any gang. A crafted request could apply Underhive Outcasts archetypes to, e.g., a Goliath gang Leader.

**Suggested fix:** Add a server-side guard before the override insert — verify the gang's `gang_type_id` matches `UNDERHIVE_OUTCASTS_GANG_TYPE_ID` and the fighter class is eligible. You already have `gangData` and `fighterTypeData` in scope at that point.

---

### 🟡 Medium — `fighter_class_id` missing from `FighterType` type

**`components/gang/add-fighter.tsx` ~lines 270 & 273**

```ts
queryKey: ['skill-archetypes', (currentFighterType as any)?.fighter_class_id],
const classId = (currentFighterType as any)?.fighter_class_id;
```

`fighter_class_id` is manually added to the transform at line 487 but not added to the `FighterType` interface in `types/fighter-type.ts`. The `as any` casts suppress a real type error. Since `fighter_class_id` is already in `types/fighter.ts` (line 177) and on other interfaces, it should just be added to `FighterType` too.

---

### 🟡 Low — Inconsistent `skill_access` filtering between add and edit paths

**`app/actions/add-fighter.ts` ~line 492 vs `components/fighter/edit-fighter/fighter-edit-modal.tsx` ~line 364**

Add path filters to only `primary`/`secondary`:
```ts
.filter(sa => sa.access_level === 'primary' || sa.access_level === 'secondary')
```

Edit path maps **all** access levels including `'allowed'`:
```ts
archetype.skill_access.map(sa => ({
  access_level: sa.access_level as 'primary' | 'secondary' | 'allowed'
}))
```

If archetypes can include `allowed` entries, fighters will end up with different overrides depending on whether the archetype was chosen at creation vs. editing. One path should be brought in line with the other.

---

### 🟢 Low — Accessibility regression: `label` lost its `htmlFor`

**`components/fighter/edit-fighter/fighter-edit-modal.tsx` ~line 1197**

```diff
- <label htmlFor="archetype" className="block text-sm font-medium mb-1">
+ <label className="block text-sm font-medium mb-1">
```

The `htmlFor` association was dropped when switching from `<select id="archetype">` to `<Combobox>`. If `Combobox` exposes an `id` prop, pass a matching value to restore the label→input association for screen readers.

---

### ✅ What's done well

- Extracting the duplicated eligibility constants + logic into `utils/archetypeEligibility.ts` is the right call — clean deduplication.
- Archetype reset on type/sub-type change and form reset is thorough.
- `staleTime: 10 * 60 * 1000` on the archetype query is appropriate.
- The `enabled: showModal && canUseArchetypes` guard prevents unnecessary fetches.

---

**Verdict: 🔴 Changes requested** — the silent failure on add and the missing server-side guard are the blockers; the type and consistency fixes are straightforward.


---

joeseos (approved):

Reviewed commit `bdab09d` against the five previous findings.

---

### Previous findings: status

| 1 | MEDIUM | Silent failure when archetype overrides fail on add | ✅ Resolved — `archetypeSkillAccessWarning` set in all three failure branches (fetch error, insert error, catch); returned as `warning` on the success response; surfaced via `toast.error` in `onSuccess` |
| 2 | MEDIUM | No server-side archetype eligibility validation | ✅ Resolved — `isArchetypeEligible({ gangTypeId: gangData.gang_type_id, fighterClass: effectiveFighterData.fighter_class })` runs before fighter insertion; returns `{ success: false, error: '...' }` when not eligible, and only then sets `archetypeIdToPersist` |
| 3 | MEDIUM | `fighter_class_id` missing from `FighterType` type | ✅ Resolved — `fighter_class_id?: string` added to `types/fighter-type.ts`; both `as any` casts in `add-fighter.tsx` removed |
| 4 | LOW | Inconsistent `skill_access` filtering between add and edit paths | ✅ Resolved — `mapArchetypeSkillAccessToOverrides` extracted to `utils/archetypeEligibility.ts` and used by both paths; both now consistently save `primary`, `secondary`, and `allowed` entries |
| 5 | LOW | `label` missing `htmlFor` after Combobox migration | ✅ Resolved — `id?: string` prop threaded through `Combobox` to the underlying `<input>`; `htmlFor="archetype"` wired in `fighter-edit-modal.tsx` and `htmlFor="add-fighter-archetype"` in `add-fighter.tsx` |

---

### New findings

None. A few things I checked and confirmed clean:

- **`onSuccess` ordering**: fighter reconciliation and exotic-beast additions fully complete before the `if (result.warning)` early return — state is never left partial.
- **`mapArchetypeSkillAccessToOverrides` behaviour change on the add path**: the old add path filtered to `primary`/`secondary` only; the old edit path was `.map()` with no filter at all (effectively all levels). The new shared function filters to the explicit set `{primary, secondary, allowed}` — more restrictive than the edit path was, but equivalent for any archetype data that only uses those three levels, which is the expected shape.
- **`archetypeIdToPersist` null propagation**: when no archetype is selected, `archetypeIdToPersist` stays `null` and is correctly written as `selected_archetype_id: null` in the insert row. The downstream `if (archetypeIdToPersist)` block is skipped cleanly.

---

### Verdict: ✅ Approve

All five findings resolved. The server-side guard and the shared `mapArchetypeSkillAccessToOverrides` utility are both clean implementations. Safe to merge.

---

PR #1737, opened 2026-06-04
Fixed gang origin not being present in print page

Djidiouf:


joeseos (approved):
## PR Review — Fixed gang origin not being present in print page

@Djidiouf Clean fix. Reviewed against `app/lib/shared/gang-data.ts`, the `GangBasic` type, `getGangTypeConfig`, and the live-view `gang.tsx` for parity.

---

### ✅ What's done well

- **Correct data source**: `gang_origin?.origin_name` comes from `getGangBasic`, which already joins `gang_origin` and `gang_origin_categories` in its Supabase query — no extra DB calls introduced.
- **`getGangTypeConfig` is synchronous**: it just reads from the already-fetched `gangBasic` object (`gangBasic.gang_types ?? gangBasic.custom_gang_types ?? null`), so placing it outside `Promise.all` is fine.
- **Null safety throughout**: `gangBasic.gang_origin?.origin_name || ""`, `gangTypeConfig?.gang_origin_categories?.category_name || ""`, and `!!gangTypeConfig?.gang_origin_category_id` all handle undefined/null correctly.
- **Double guard on render** (`gang_type_has_origin && gang_origin_name`) mirrors exactly what the live `gang.tsx` view already does — good parity.
- **Three render locations updated**: header, roster badge section, and fighter-card badge section are all consistent.

---

### 🟢 Low — `gang_origin_category_name` could render as an orphaned colon

**`components/gang/print-gang.tsx` ~lines 1062 & 1499**

```tsx
{gang_origin_category_name}: <Badge variant="secondary">{gang_origin_name}</Badge>
```

If `gangTypeConfig` exists (so `gang_type_has_origin` is `true`) but its `gang_origin_categories` join returns `null`, then `gang_origin_category_name` would be `""` and the output would be `: Precinct 51`. In practice this can't happen if the DB has referential integrity on `gang_origin_category_id`, but a defensive guard `{gang_origin_category_name || "Origin"}:` would be belt-and-braces. Not blocking — the same risk exists in the live view and hasn't been an issue.

---

**Verdict: ✅ Approve** — focused, correct fix with no regressions. The approach is a faithful port of the live-view pattern into the print page.


---

PR #1738, opened 2026-06-05
Fixed legendary names selection type not showing up in Advancement flow

Djidiouf:


joeseos (commented):
## Code Review — PR #1738 `fix/legendary_name_advancement`

**Reviewer:** Claude Code  
**Developer:** @Djidiouf

---

### Summary

The bug is real and the root cause analysis is correct. When a fighter selects a Legendary Names skill set, `get_available_skills` returns type IDs `'selected'`/`'random'` instead of the standard `primary_selected`/`primary_random`/etc. The old code's filter (`allTypes.filter(t => allowedIds.has(t.type_id))`) produced an empty array because `getAcquisitionTypeIdsForAccess` never included those bare IDs, and `isRandomAcquisitionType` also silently failed for `'random'` since it only checked `.endsWith('_random')`. Both are fixed correctly.

---

### Findings

#### 🟡 Medium — Detection heuristic is implicit and fragile

**File:** `components/fighter/fighter-advancement-list.tsx`, lines 262–264, 294–295

The legendary detection relies on `'selected'` / `'random'` being exclusive to the legendary branch in the SQL. This is true today, but these are very generic strings. If any future skill type ever returns a bare `'selected'` or `'random'` type ID (custom types, a new game mechanic, etc.), `getAllowedAcquisitionTypeIds` would silently misfire — treating a normal skill set as legendary and stripping out all standard acquisition types.

A more robust approach would be to propagate `legendary_name: boolean` from `get_available_skills` through the `SkillData` / `AvailableAdvancement` types and use it as the discriminant directly, removing the implicit string matching entirely.

That said, the SQL is the single source of truth and the comment at lines 258–260 documents the dependency well. If refactoring the API shape feels out of scope for this fix, an explicit guard comment on `LEGENDARY_SKILL_ACQUISITION_TYPE_IDS` spelling out "these must not overlap with any type_id in ALL_SKILL_ACQUISITION_TYPE_IDS and must remain in sync with the legendary_name = TRUE branch in get_available_skills.sql" would make the invariant obvious to future maintainers.

---

#### 🟡 Medium — False "no access" warning may still appear for legendary names (out of scope, worth noting)

**File:** `components/fighter/fighter-advancement-list.tsx`, lines 1233–1238

`selectedSkillSetLacksAccess` fires whenever `selectedSkillSetAccess === null`, which happens when the selected skill set has no row in `fighter_type_skill_access`. If legendary name skill types are not stored in that table (they have their own cost structure, separate from primary/secondary access), fighters will see the amber "This Skill Set is not accessible" warning even after this fix makes the acquisition types appear correctly.

This is pre-existing behavior and not introduced by this PR — just worth a follow-up issue if it's observed in practice. A targeted fix would be to suppress the warning when the skill set is detected as legendary (using the same heuristic already introduced here).

---

#### 🟢 Low — JSDoc block style is inconsistent with surrounding code

**File:** `components/fighter/fighter-advancement-list.tsx`, lines 285–288

The new `getAllowedAcquisitionTypeIds` function uses a `/** */` JSDoc block, while all surrounding explanatory comments (e.g. lines 241–249, 274–276, 279–281) use `//` line comments. No functional impact, but it's a minor consistency break.

---

### What's correct

- The `LEGENDARY_SKILL_ACQUISITION_TYPE_IDS` constant and `isLegendaryAcquisitionTypeId` helper are clean extractions that keep the detection logic in one place.
- `getAllowedAcquisitionTypeIds` cleanly wraps `getAcquisitionTypeIdsForAccess` without modifying it — no regression risk to the existing skill flow.
- The `isRandomAcquisitionType` fix (`=== 'random' || .endsWith('_random')`) is the minimal correct change. The two-condition form reads clearly.
- The `acquisitionTypeComboboxOptions` memo dependencies are unchanged and correct.
- No security issues. No performance concerns (arrays of ≤5 items).

---

### Verdict

**✅ Approve** — the fix is correct, minimal, and well-reasoned. The medium finding on heuristic fragility is worth a follow-up but is not a blocker; the invariant holds today and is documented. The "no access" warning concern is pre-existing and out of scope for this PR.

---

PR #1739, opened 2026-06-06
Removed legacy columns

joeseos:


joeseos (commented):
## Code Review — PR #1739 `refac/removed-legacy-resources`

**Reviewer:** Claude Code
**Developer:** @joeseos

---

### Summary

This PR does two things: (1) removes six hardcoded resource columns (`meat`, `scavenging_rolls`, `exploration_points`, `power`, `sustenance`, `salvage`) from the `gangs` table and their corresponding `has_*` flags from `campaigns`/`campaign_types`, and (2) adds CRUD support for `campaign_type_resources` in the admin campaign management UI, completing the migration to the normalized resource model. The legacy column removal is thorough — 21 files touched across interfaces, queries, API routes, components, and SQL functions.

The migration to `DROP COLUMN IF EXISTS` is clean and safe. The bulk of the changes (removing interface fields, props, state, query columns) are mechanical and correct. The new feature (admin resource management) works but has a couple of issues worth addressing.

---

### Findings

#### 🔴 High — Resource insert error silently swallowed in POST handler

**File:** `app/api/admin/campaign-types/route.ts`, lines ~99-106 (new code)

```ts
if (Array.isArray(resources) && resources.length > 0 && campaignType) {
  const newNames: string[] = resources.map((r: string) => r.trim()).filter(Boolean);
  if (newNames.length > 0) {
    await supabase
      .from('campaign_type_resources')
      .insert(newNames.map(resource_name => ({ campaign_type_id: campaignType.id, resource_name })));
  }
}
```

The insert result is not checked for errors. If this fails (e.g., duplicate resource name violating the unique constraint `(campaign_type_id, resource_name)`, or a DB error), the campaign type is created successfully but its resources are silently lost. The response returns the campaign type as if everything succeeded.

**Fix:** Capture the error and throw:
```ts
const { error: resourceError } = await supabase
  .from('campaign_type_resources')
  .insert(...);
if (resourceError) throw resourceError;
```

---

#### 🟡 Medium — PATCH response contract changed (breaking)

**File:** `app/api/admin/campaign-types/route.ts`, PATCH handler

Old behavior returned the updated campaign type object: `NextResponse.json(campaignType)`.
New behavior returns: `NextResponse.json({ success: true })`.

The admin component at `admin-campaign-management.tsx:258-259` does:
```ts
const savedCampaignType = await response.json();
const campaignTypeId = savedCampaignType.id || selectedCampaignTypeId;
```

This currently works because the UPDATE path has `selectedCampaignTypeId` set as a fallback. But the response shape is now inconsistent with POST (which still returns the full object), and any future consumer relying on the returned object from PATCH would break.

**Suggestion:** Return the updated object consistently, or at minimum return `{ success: true, id }` so the contract is self-contained.

---

#### 🟡 Medium — No duplicate resource name validation in UI

**File:** `components/admin/admin-campaign-management.tsx`, lines ~840-860

Both the Enter key handler and the Add button handler allow adding resources with the same name. The DB has a unique constraint on `(campaign_type_id, resource_name)` which will catch it server-side, but combined with the missing error check in the POST handler (finding #1), duplicates would be silently dropped.

**Fix:** Check before adding:
```ts
if (campaignTypeResources.some(r => r.resource_name.toLowerCase() === newResourceName.trim().toLowerCase())) {
  toast.error('Resource already exists');
  return;
}
```

---

#### 🟢 Low — Enter key on resource input should prevent default

**File:** `components/admin/admin-campaign-management.tsx`, lines ~843-848

The `onKeyDown` handler adds a resource on Enter but doesn't call `e.preventDefault()`. If this input is ever nested in a form (or a component that intercepts Enter), it could trigger unintended behavior.

---

#### 🟢 Low — `supabase/schema/schema.public.sql` still references dropped columns

**Files:** `supabase/schema/schema.public.sql` (lines 3511-3516, 3636-3641, 4275-4281, 4297-4305, 5321-5332)

The generated schema dump still defines `has_meat`, `has_exploration_points`, etc. on `campaign_types`/`campaigns` and `meat`/`scavenging_rolls`/`exploration_points` on `gangs`. This is expected since it's a `pg_dump` output that will be refreshed after the migration runs — just noting for awareness.

---

### What's correct

- Legacy field removal is comprehensive: interfaces (`GangBasic`, `GangCampaign`, `GangProps`, `Member`, `Gang`, `GangUpdates`, `Campaign`, `GangEditModalProps`, `CampaignPageContentProps`), Supabase queries, API routes, React state, SQL function signatures and return columns — all cleaned up consistently.
- The `get_gang_details.sql` function correctly drops the columns from both the `RETURNS TABLE` signature and the two SELECT sites (JSON campaign object and final select).
- `copy-gang.ts` correctly removes the legacy columns. Campaign resources (`campaign_gang_resources`) are tied to `campaign_gang` entries, not the gang itself, so they're correctly excluded from gang duplication.
- The migration uses `DROP COLUMN IF EXISTS` — idempotent and safe.
- The PATCH sync logic (diff existing vs. new, delete removed, insert added) is correct and handles the name-based reconciliation properly.
- `useEffect` dependency arrays updated correctly after removing state variables.
- No security issues — admin-only endpoints remain admin-only, RLS on `campaign_type_resources` enforces admin CRUD.

---

### Verdict

**Changes requested** — The unchecked resource insert error in the POST handler (finding #1) is a real bug that would silently lose data. The other findings are lower priority but worth addressing. The overall approach and scope of the cleanup is sound.

---

joeseos (commented):
## Follow-up Review — PR #1739 (commit `40c165f`)

@joeseos — All four issues from the initial review are resolved:
- POST handler now checks resource insert errors ✅
- PATCH returns the full updated object (with resources) ✅  
- Duplicate resource names validated client-side ✅
- Enter key calls `preventDefault()` ✅

---

### Remaining bloat / cleanup opportunities

Now that the legacy columns are gone, a few stale artifacts remain in files **not touched by this PR**:

#### 1. Dead commented-out function — `app/lib/shared/gang-data.ts:309-340`

A 30-line commented-out `getGangResources` function that references the dropped columns (`meat`, `scavenging_rolls`, `exploration_points`, `power`, `sustenance`, `salvage`) and the deprecated `BASE_GANG_RESOURCES` cache tag. Safe to delete entirely.

#### 2. Stale cache tag comments — `utils/cache-tags.ts`

- **Line 22**: `BASE_GANG_BASIC` comment still lists `meat, scavenging_rolls, etc.` — these are no longer in the query.
- **Line 24**: Commented-out `BASE_GANG_RESOURCES` cache tag (already deprecated, now doubly dead).
- **Line 298**: Comment references `"Resources (reputation, meat, etc.)"` — `meat` is no longer a gang column.

#### 3. Minor DRY — `components/admin/admin-campaign-management.tsx`

The "add resource" logic (duplicate check → append → clear input) is copy-pasted between the `onKeyDown` Enter handler and the `onClick` button handler (~8 identical lines each). Extracting to a small helper would reduce duplication:

```ts
const addResource = () => {
  const trimmed = newResourceName.trim();
  if (!trimmed) return;
  if (campaignTypeResources.some(r => r.resource_name.toLowerCase() === trimmed.toLowerCase())) {
    toast.error('Resource already exists');
    return;
  }
  setCampaignTypeResources([...campaignTypeResources, { id: `new-${Date.now()}`, resource_name: trimmed }]);
  setNewResourceName('');
};
```

---

### Verdict

**✅ Approve** — The core changes are correct and the review issues are resolved. The three cleanup items above are low-severity and can be addressed in this PR or a follow-up at your discretion.

---

PR #1740, opened 2026-06-06
Added the possibility to edit Cost and Availability rules

Djidiouf:


joeseos (changes requested):
## Code Review — PR #1740: Edit Cost and Availability Rules for Custom TP

@Djidiouf thanks for the PR! The core refactor — replacing two boolean open/close flags with a `'add' | number | null` discriminated union — is a clean pattern and the state initialisation for edit mode is thorough. One real correctness bug to fix before merging, plus a few lower-priority observations.

---

### Finding 1 — **MEDIUM** — Stale `fighterTypeId` survives custom gang-type switch

**File:** `components/customise/custom-trading-posts.tsx` ~line 1784

In `AddPricingRuleModal`, the `onGangTypeChange` handler clears `fighterTypeId` only for *non-custom* gang types:

```tsx
if (!id) {
  setFighterTypeId('');
} else if (fighterTypeId) {
  if (!isCustom) {          // ← custom branch is missing
    if (!fighterTypes.some(ft => ft.gang_type_id === id && ft.id === fighterTypeId)) {
      setFighterTypeId('');
    }
  }
  // when isCustom === true: nothing — stale ID survives
}
```

When the user switches between two *custom* gang types while a `fighterTypeId` is already set, the old ID is not cleared immediately. The `useEffect` that eventually clears it guards behind `isCustomFighterTypesLoading`:

```tsx
if (isCustomGangType ? isCustomFighterTypesLoading : isFighterTypesLoading) return;
```

So during the fetch window for the new custom gang type's fighter list, `isValid` can still pass (because `fighterTypeId` is truthy) and `handleSave` can persist a `fighter_type_id` that belongs to the *old* gang type.

**Fix:** add the `isCustom` branch — since custom fighter types require a new fetch, always clear when switching to a custom gang type:

```tsx
if (!id) {
  setFighterTypeId('');
} else if (fighterTypeId) {
  if (isCustom) {
    setFighterTypeId('');   // must wait for new fetch; clear eagerly
  } else if (!fighterTypes.some(ft => ft.gang_type_id === id && ft.id === fighterTypeId)) {
    setFighterTypeId('');
  }
}
```

---

### Finding 2 — **LOW** — `selectedFighterTypeName` computed as IIFE on every render

**File:** ~line 1711

```tsx
const selectedFighterTypeName = (() => { ... })();
```

Every other derived value in this component is wrapped in `React.useMemo`. This IIFE re-runs `buildMultiProfileKeys` and three `.find()` calls on every render. Not a correctness issue, but inconsistent and unnecessary.

**Fix:** wrap it in `React.useMemo(() => { ... }, [scopedFighterTypes, fighterTypeId, fighterTypes, customFighterTypes, isCustomGangType, gangTypeId])`.

---

### Finding 3 — **LOW** — `useEffect` may clear a valid `fighterTypeId` on initial edit render

**File:** ~line 1726

When the modal opens in edit mode (`initialRule` is set), `fighterTypeId` is seeded from `initialRule.fighter_type_id`. The `useEffect` runs once `isFighterTypesLoading` flips to `false` and then checks:

```tsx
if (!scopedFighterTypes.some(ft => ft.id === fighterTypeId)) {
  setFighterTypeId('');
}
```

`scopedFighterTypes` is `fighterTypes.filter(ft => ft.gang_type_id === gangTypeId)`. If a fighter type in the API response happens to have a missing/null `gang_type_id`, it won't appear in `scopedFighterTypes` and the valid pre-populated ID will be silently wiped. This is data-dependent, but worth being aware of if users report their fighter type disappearing when re-opening a rule.

---

### Finding 4 — **LOW** — `availRules` vs `localAvailRules` in edit indexing

**File:** ~line 1327

```tsx
initialRule={typeof availRuleModal === 'number' ? availRules[availRuleModal] : undefined}
```

`availRules` = `localAvailRules ?? fetchedAvailRules`. The index comes from mapping over the same `availRules` array, so this is correct. The only theoretical edge case is a react-query background refetch landing between the click and the re-render and shifting indices, but that's not realistically triggerable here. No action needed — noting it for awareness.

---

### Non-issues / positives

- The `'add' | number | null` union is a clean alternative to two separate boolean + index states — good call.
- Preserving the original `id` (`initialRule?.id ?? \`local_${Date.now()}\``) so local edits don't generate spurious new IDs is correct.
- The fighter type dropdown scoping (only show types relevant to the selected gang) and the `disabled={!gangTypeId}` guard are good UX improvements.
- `handleUpdateAvailRule` / `handleUpdatePricingRule` correctly fallback to `fetchedAvailRules` via `prev ?? fetchedAvailRules` when `localAvailRules` is still null.
- Label copy changes (`"Resource Cost"` → `"Resource Type"`, `"Cost Override"` → `"General Cost Override"`) are clear improvements.

---

Fix the `isCustom` branch gap in `onGangTypeChange` (Finding 1) — it's the only path that can silently persist a cross-gang `fighter_type_id`. The rest are low-priority observations. Once that's addressed this is good to merge.

---

joeseos:
## Follow-up Review — PR #1740 (commit `97e77b7`)

@Djidiouf the follow-up commit addresses all the findings from the previous review. Here's the status of each:

---

### Previous findings

| 1 | MEDIUM | Stale `fighterTypeId` survives custom gang-type switch | ✅ **Resolved** |
| 2 | LOW | `selectedFighterTypeName` recomputed as IIFE on every render | ✅ **Resolved** |
| 3 | LOW | `useEffect` may clear valid `fighterTypeId` on initial edit render | ✅ **Resolved** (see note) |
| 4 | LOW | `availRules` vs `localAvailRules` index stability | ➡️ **No action needed** (acknowledged) |

---

**Finding 1** — The `onGangTypeChange` handler now correctly branches on `isCustom` and eagerly calls `setFighterTypeId('')` when switching between custom gang types. Fix matches the suggested approach exactly.

**Finding 2** — `selectedFighterTypeName` is now wrapped in `React.useMemo` with the full dependency array. The memo also adds a multi-source fallback lookup (`scopedFighterTypes → fighterTypes → customFighterTypes`) which is a solid defensive improvement.

**Finding 3** — Resolved more elegantly than suggested: `scopedFighterTypes` now includes the filter clause `|| (fighterTypeId !== '' && ft.id === fighterTypeId)`, meaning the currently-selected ID is always present in the scoped list during the loading window. This makes the useEffect guard a non-issue for valid pre-populated IDs — a cleaner fix than the alternatives.

---

### New changes reviewed (no new issues found)

- `handleUpdateAvailRule` / `handleUpdatePricingRule` — correctly copy the array, replace at index, and reset modal state. No mutation bugs.
- `onSaved` callback closure — `availRuleModal` / `pricingRuleModal` value is read inside the callback and correctly routes to update vs. add. Correct.
- `disabled={!gangTypeId}` on the fighter type Combobox — good UX guard, consistent with the loading-state guard already in place.
- Module-level extraction of `PricingRuleFighterType`, `buildMultiProfileKeys`, `formatPricingRuleFighterTypeLabel` — correct placement; these don't close over component state.
- Label copy (`"Resource Type"`, `"General Cost Override"`, `"General Availability Override"`) — clean improvements, no issues.

---

### Verdict: ✅ Approve

All four previous findings resolved. No new issues introduced. This is good to merge.

---

joeseos (approved):
All four findings from the previous review have been resolved. Finding 1 (the `isCustom` branch gap) is cleanly fixed; Finding 3 is addressed elegantly via the `scopedFighterTypes` filter inclusion. No new issues in the follow-up commit. Good to merge.

---

PR #1742, opened 2026-06-06
Added a dropdown to Add Special Rules in the Edit Fighter modal that …

Djidiouf:
…list default rules the fighter normally has if they were previously removed, offer a custom options, works similarly to scenario selection

### DIFF
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -16,8 +16,7 @@ import { FighterPromotionModal } from './fighter-promotion-modal';
 import { FighterCharacteristicTable } from './fighter-characteristic-table';
 import { CharacterStatsModal } from './character-stats-modal';
 
-const normalizeSpecialRule = (rule: string) =>
-  typeof rule === 'string' ? rule.replace(/^"|"$/g, '') : rule;
+const normalizeSpecialRule = (rule: string) => rule.replace(/^"|"$/g, '');
 
 
 interface FighterTypesData {
@@ -229,12 +228,12 @@ export function EditFighterModal({
     const selectedFighterType = fighterTypes.find(ft => ft.id === selectedFighterTypeId);
     if (!selectedFighterType) return null;
 
-    if (selectedSubTypeId !== undefined && selectedSubTypeId !== null && selectedSubTypeId !== '') {
+    if (selectedSubTypeId !== '') {
       const foundSubType = availableSubTypes.find(st => st.value === selectedSubTypeId);
       if (foundSubType?.fighterTypeId) {
         return fighterTypes.find(ft => ft.id === foundSubType.fighterTypeId) ?? selectedFighterType;
       }
-    } else if (selectedSubTypeId === '' && availableSubTypes.length > 0) {
+    } else if (availableSubTypes.length > 0) {
       const defaultOption = availableSubTypes.find(st => st.value === '');
       if (defaultOption?.fighterTypeId) {
         return fighterTypes.find(ft => ft.id === defaultOption.fighterTypeId) ?? selectedFighterType;
@@ -275,7 +274,11 @@ export function EditFighterModal({
         disabled: true,
       });
       availableDefaultSpecialRules.forEach(rule => {
-        options.push({ value: rule, label: rule });
+        options.push({
+          value: rule,
+          label: <span className="ml-3">{rule}</span>,
+          displayValue: rule,
+        });
       });
     }
 
@@ -451,9 +454,7 @@ export function EditFighterModal({
           fighter_type: type.fighter_type,
           fighter_class: type.fighter_class,
           fighter_class_id: type.fighter_class_id,
-          special_rules: (type.special_rules || [])
-            .map((r: string) => typeof r === 'string' ? r.replace(/^"|"$/g, '') : r)
-            .filter(Boolean),
+          special_rules: (type.special_rules || []).map(normalizeSpecialRule).filter(Boolean),
           gang_type_id: type.gang_type_id,
           total_cost: type.total_cost,
           typeClassKey: type.typeClassKey,
@@ -665,6 +666,8 @@ export function EditFighterModal({
   // Update the handleFighterTypeChange function
   const handleFighterTypeChange = (fighterTypeId: string) => {
     setSelectedFighterTypeId(fighterTypeId);
+    setSelectedSpecialRuleOption('');
+    setCustomSpecialRule('');
 
     // Set flag to indicate user has explicitly selected a fighter type
     setHasExplicitlySelectedType(true);
@@ -762,6 +765,8 @@ export function EditFighterModal({
   // Add handler for sub-type change
   const handleSubTypeChange = (subTypeId: string) => {
     setSelectedSubTypeId(subTypeId);
+    setSelectedSpecialRuleOption('');
+    setCustomSpecialRule('');
   };
 
   // Add handler for gang legacy change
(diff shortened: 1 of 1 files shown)


joeseos (changes requested):
## PR Review — #1742: Offer default fighter-type rules in "Add Special Rule" combobox

@Djidiouf thanks for the contribution. The feature direction is solid — surfacing the fighter type's default rules in the combobox is much better UX than a raw text field. A few issues need addressing before merge.

---

### 🔴 Medium — DRY: `normalizeSpecialRule` not consolidated with existing inline copy

The PR extracts the normalization logic into a module-level function, which is good, but the identical inline version in the `preFetchedFighterTypes` `useEffect` was not updated:

```tsx
// fighter-edit-modal.tsx — existing useEffect (not changed by this PR)
special_rules: (type.special_rules || [])
  .map((r: string) => typeof r === 'string' ? r.replace(/^\"|\"$/g, '') : r)
  .filter(Boolean),
```

This means the same normalization logic now lives in two places and can drift. That inline `.map` should call `normalizeSpecialRule` instead.

**Fix:**
```tsx
special_rules: (type.special_rules || []).map(normalizeSpecialRule).filter(Boolean),
```

---

### 🔴 Medium — DRY: Two near-identical `<Combobox>` instances for one control

Both branches of the `selectedSpecialRuleOption === 'custom'` conditional render an identical `<Combobox>` with the same `options`, `onValueChange`, `placeholder`, `allowCustom`, and `dropdownPlacement`. Only `value` differs (`"custom"` vs `selectedSpecialRuleOption`). This should be one `<Combobox>` with the custom text input conditionally shown below it:

```tsx
<div className="mb-2">
  <div className={selectedSpecialRuleOption !== 'custom' ? "flex space-x-2" : ""}>
    <div className={selectedSpecialRuleOption !== 'custom' ? "grow" : ""}>
      <Combobox
        options={specialRuleComboboxOptions}
        value={selectedSpecialRuleOption}
        onValueChange={handleSpecialRuleOptionChange}
        placeholder="Add a Special Rule"
        allowCustom={true}
        dropdownPlacement="down"
      />
    </div>
    {selectedSpecialRuleOption !== 'custom' && (
      <Button onClick={handleAddSpecialRule} type="button">Add</Button>
    )}
  </div>
  {selectedSpecialRuleOption === 'custom' && (
    <div className="flex space-x-2 mt-2">
      <Input
        type="text"
        value={customSpecialRule}
        onChange={(e) => setCustomSpecialRule(e.target.value)}
        placeholder="Enter custom Special Rule"
        className="grow"
        onKeyDown={(e) => { if (e.key === 'Enter') { e.preventDefault(); handleAddSpecialRule(); } }}
      />
      <Button onClick={handleAddSpecialRule} type="button">Add</Button>
    </div>
  )}
</div>
```

---

### 🔴 Medium — Stale `selectedSpecialRuleOption` when fighter type changes

If a user selects a default rule (e.g. "Iron Will") from the combobox, then changes the fighter type, `selectedSpecialRuleOption` remains `'Iron Will'` even though it may no longer be a valid default for the new type. If they then click "Add", that stale rule gets appended regardless of relevance.

`handleFighterTypeChange` (and `handleSubTypeChange`) should reset the special rule selection:

```tsx
const handleFighterTypeChange = (fighterTypeId: string) => {
  setSelectedFighterTypeId(fighterTypeId);
  setSelectedSpecialRuleOption('');   // add this
  setCustomSpecialRule('');            // add this
  // ... rest of existing logic
};
```

Same applies to `handleSubTypeChange`.

---

### 🟡 Low — `normalizeSpecialRule` type signature is misleading

```tsx
const normalizeSpecialRule = (rule: string) =>
  typeof rule === 'string' ? rule.replace(/^"|"$/g, '') : rule;
```

The parameter is typed as `string` but the `typeof` guard only makes sense if non-string values are possible at runtime. TypeScript will never warn a caller passing the wrong type because the signature says `string`. Either:
- Remove the `typeof` guard (it's always true given the type), or
- Change the parameter type to `unknown` and assert the return type as `string` so callers get accurate checking

Given this is purely defensive against API data quality issues, removing the guard is the simpler fix if the `.filter(Boolean)` downstream already handles empty/falsy values.

---

### 🟡 Low — Redundant null checks on a `string` state variable

```tsx
if (selectedSubTypeId !== undefined && selectedSubTypeId !== null && selectedSubTypeId !== '') {
```

`selectedSubTypeId` is `useState<string>`, so TypeScript guarantees it's always a `string`. The `!== undefined` and `!== null` checks add noise. The condition can just be `if (selectedSubTypeId !== '')`.

---

### ✅ What's working well

- The sub-type-aware `effectiveFighterType` memo correctly resolves through sub-type variants before looking up default rules.
- `availableDefaultSpecialRules` correctly excludes rules already on the fighter, and reacts to `formValues.special_rules` so just-added rules immediately leave the dropdown.
- The disabled header option (`__default_special_rules_header__`) is correctly guarded in the Combobox's `handleSelect` — the sentinel value can never be emitted as a real selection.
- Duplicate detection uses `normalizeSpecialRule` for comparison, handling the quoted-string quirk from the API.
- The `allowCustom` flow (typing a value → falling into the `else` branch → pre-filling the custom input) is clever and works correctly with the Combobox's existing `handleCustomValue` path.

---

The two DRY violations (unconsolidated `normalizeSpecialRule` usage, duplicated Combobox JSX) and the stale-state bug on fighter type change are the required fixes. The type annotation and null-check issues can be cleaned up alongside.


---

joeseos:
## Follow-up Review — #1742 (commit `31d785d`)

@Djidiouf thanks for the quick turnaround. All five findings from the previous review have been addressed.

---

### Previous findings status

| # | Finding | Status |
|---|---------|--------|
| 1 | 🔴 DRY: inline `normalizeSpecialRule` copy in `preFetchedFighterTypes` useEffect | ✅ Resolved — inline `.map((r) => typeof r === 'string' ? r.replace(…) : r)` replaced with `.map(normalizeSpecialRule)` |
| 2 | 🔴 DRY: two near-identical `<Combobox>` instances | ✅ Resolved — single `<Combobox>` with the custom `<Input>` conditionally rendered below |
| 3 | 🔴 Stale `selectedSpecialRuleOption` on fighter type / sub-type change | ✅ Resolved — both `handleFighterTypeChange` and `handleSubTypeChange` now reset `selectedSpecialRuleOption` and `customSpecialRule` |
| 4 | 🟡 Misleading `typeof` guard in `normalizeSpecialRule` | ✅ Resolved — guard removed; function is now a clean `(rule: string) => rule.replace(/^"\|"$/g, '')` |
| 5 | 🟡 Redundant `!== undefined && !== null` checks on `string` state | ✅ Resolved — simplified to `if (selectedSubTypeId !== '')` |

---

### New findings

None. The new commit is a clean, targeted fixup with no regressions or new issues observed.

A couple of minor observations for awareness only (no action required):

- When `selectedSpecialRuleOption` is `''` (initial state), the "Add" button is visible alongside the Combobox but clicking it is a no-op (guarded by `if (!ruleToAdd) return`). Acceptable UX.
- The `handleSpecialRuleOptionChange` `else` branch (for typed-in custom values) falls through to `setCustomSpecialRule(value)` — this correctly pre-fills the custom input but silently accepts any string that isn't in `availableDefaultSpecialRules`, including edge-case Combobox internals. In practice this is fine since the sentinel `__default_special_rules_header__` is guarded at the Combobox level.

---

**Verdict: Approve** ✅

The implementation is clean, the DRY violations are gone, and the stale-state bug is fixed. Good to merge.

---

joeseos (approved):
All previous findings resolved. No new issues. Approving.

---

PR #1744, opened 2026-06-07
Optimize battle completion: defer logging, batch queries, suppress refetch

joeseos:
## Summary
Refactored the `completeBattleSession` action to improve performance and user experience by deferring non-critical work, batching database queries, and suppressing unnecessary refetches during completion.

### Server-side optimizations (`app/actions/battle-sessions.ts`)
- **Deferred work with `after()`**: Moved all logging, notifications, and cache invalidation to run after the response is sent using Next.js `after()` API, ensuring the user sees the completion result immediately
- **Parallel queries**: Consolidated multiple sequential Supabase queries into a single `Promise.all()` call for session, participants, profile, and territory data
- **Inline campaign battle log**: Replaced `createBattleLog()` server action call with direct Supabase insert, eliminating function call overhead
- **Batch gang logs**: Changed from N sequential `logBattleResult()` calls to a single batch insert of all gang logs
- **Territory claim in same flow**: Moved territory update into the campaign battle creation block to reduce round trips
- **Removed unnecessary revalidation**: Deleted `revalidateTag()` call from `fetchBattleSession()` that was invalidating cache on every read

### Client-side improvements (`hooks/use-battle-session-realtime.ts` & `components/battle-session/`)
- **Refetch suppression**: Added `suppressRefetch()` callback to prevent unnecessary query invalidations for 2 seconds after completion
- **Suppress state tracking**: Implemented `suppressedRef` and `suppressTimerRef` to track and enforce suppression window
- **Updated hook return**: Changed `useBattleSessionRealtime()` to return object with `broadcast` and `suppressRefetch` instead of just broadcast
- **Integrated suppression**: Connected `suppressRefetch` callback to `CompleteBattleModal` completion handler

## Implementation Details
- All deferred work is wrapped in error handling to prevent silent failures
- Gang logs now include user_id for proper attribution
- Campaign name resolution is cached within the deferred block to avoid redundant queries
- Territory claim logging is conditional on successful territory update
- Refetch suppression uses a 2-second window to cover typical network latency

https://claude.ai/code/session_01DpZfBBVjrSDP7mdHfA4T4t

### DIFF
--- a/app/actions/battle-sessions.ts
+++ b/app/actions/battle-sessions.ts
@@ -1205,7 +1205,7 @@ export async function completeBattleSession(
 
     // Derive the effective winner set
     const flaggedWinnerIds: string[] =
-      (allParticipants ?? [])
+      allParticipants
         .filter((p) => p.is_winner === true)
         .map((p) => p.gang_id);
     const effectiveWinnerIds: string[] =
@@ -1216,7 +1216,7 @@ export async function completeBattleSession(
           : [];
     const winnerGangSet = new Set(effectiveWinnerIds);
     const claimerGangId: string | null =
-      (allParticipants ?? []).find((p) => p.claimed_territory === true)?.gang_id
+      allParticipants.find((p) => p.claimed_territory === true)?.gang_id
       ?? session.winner_gang_id
       ?? null;
     const legacyWinnerId: string | null =
@@ -1224,7 +1224,7 @@ export async function completeBattleSession(
 
     // --- Campaign battle log (inline insert, no createBattleLog call) ---
     let campaign_battle_id: string | undefined;
-    if (session.campaign_id && allParticipants) {
+    if (session.campaign_id) {
       const battleParticipants = allParticipants.map((p) => ({
         role: p.role,
         gang_id: p.gang_id,
@@ -1264,7 +1264,8 @@ export async function completeBattleSession(
           .eq('campaign_id', session.campaign_id);
 
         if (claimError) {
-          await supabase.from('campaign_battles').delete().eq('id', campaign_battle_id);
+          const { error: deleteError } = await supabase.from('campaign_battles').delete().eq('id', campaign_battle_id);
+          if (deleteError) console.error('Failed to rollback campaign battle after territory claim error:', deleteError);
           return { success: false, error: `Failed to claim territory: ${claimError.message}` };
         }
       }
@@ -1285,7 +1286,8 @@ export async function completeBattleSession(
 
     if (error || !updated || updated.length === 0) {
       if (campaign_battle_id) {
-        await supabase.from('campaign_battles').delete().eq('id', campaign_battle_id);
+        const { error: deleteError } = await supabase.from('campaign_battles').delete().eq('id', campaign_battle_id);
+        if (deleteError) console.error('Failed to rollback campaign battle after session update error:', deleteError);
       }
       return { success: false, error: error?.message || 'Session was already completed' };
     }
@@ -1296,8 +1298,6 @@ export async function completeBattleSession(
     // === Everything below runs AFTER the response is sent ===
     after(async () => {
       try {
-        if (!allParticipants) return;
-
         // Fetch gang names + owner IDs for logging (one query)
         const gangIds = allParticipants.map((p) => p.gang_id);
         const { data: gangs } = await supabase
@@ -1308,9 +1308,7 @@ export async function completeBattleSession(
           gangs?.map((g) => [g.id, g]) || []
         );
 
-        // Batch gang logs — one insert instead of N sequential calls
-        const campaignName = session.campaign_id ? undefined : 'Standalone Battle';
-        let resolvedCampaignName = campaignName;
+        let resolvedCampaignName = 'Standalone Battle';
         if (session.campaign_id) {
           const { data: campaign } = await supabase
             .from('campaigns')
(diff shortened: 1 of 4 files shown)


joeseos (commented):
## PR Review — #1744: Optimize battle completion

@joeseos the performance direction here is solid — parallelising the initial fetches and deferring logging/notifications to `after()` is the right shape. Two correctness bugs need fixing before merge, plus a few medium-severity regressions worth addressing.

---

### 🔴 High — Territory update errors silently swallowed (data corruption risk)

`app/actions/battle-sessions.ts`, campaign territory claim block:

```ts
// Claim territory in same flow
if (options?.campaign_territory_id && claimerGangId) {
  await supabase
    .from('campaign_territories')
    .update({ gang_id: claimerGangId })
    .eq('id', options.campaign_territory_id)
    .eq('campaign_id', session.campaign_id);
  // ← result is discarded — errors are invisible
}
```

The old `createBattleLog` destructured `{ error }` and threw on failure:
```ts
if (error) throw new Error(`Failed to claim territory: ${error.message}`);
```

In the new code, if the `campaign_territories` update fails (RLS, network blip, race condition), the function silently continues. The `campaign_battles` record is committed with `campaign_territory_id` set, but the territory's `gang_id` is never updated — the battle log permanently claims a territory that the gang doesn't actually own.

**Fix:** capture and check the error:
```ts
const { error: territoryError } = await supabase
  .from('campaign_territories')
  .update({ gang_id: claimerGangId })
  .eq('id', options.campaign_territory_id)
  .eq('campaign_id', session.campaign_id);

if (territoryError) {
  console.error('Failed to claim territory:', territoryError);
  return { success: false, error: `Failed to claim territory: ${territoryError.message}` };
}
```

---

### 🔴 High — `onSuppressRefetch()` fires before form validation

`components/battle-session/complete-battle-modal.tsx`:

```ts
const handleConfirm = async () => {
  onSuppressRefetch?.();          // ← suppression starts HERE

  if (!hasAnyWinnerSelected) {
    toast.error('Please select a winner or draw');
    return false;                  // ← early exit without completing
  }
  // ... more validation guards ...
```

The 2-second suppression window opens the moment the user clicks "Complete Battle", even when validation subsequently fails. A common failure path:

1. User clicks "Complete" with no winner → toast error, suppression window starts
2. User picks a winner and clicks again within 2 s
3. Battle completes server-side, but the realtime refetch is still suppressed → UI doesn't update

**Fix:** move the call to immediately before `completeBattleSession()`, after all validation has passed:
```ts
  // ... all validation ...
  onSuppressRefetch?.();           // ← only suppress once we're committed
  const completeResult = await completeBattleSession(...);
```

---

### 🟡 Medium — `allParticipants` query error silently ignored in parallel fetch

`app/actions/battle-sessions.ts`:

```ts
const [
  { data: session },
  { data: allParticipants },   // ← error field discarded
  ...
] = await Promise.all([...]);
```

If the `battle_session_participants` query errors, `allParticipants` is `null`. The `if (session.campaign_id && allParticipants)` guard then silently skips the entire campaign battle log creation — the session is marked completed but no `campaign_battles` record is written. The old sequential code would have surfaced this through the outer `catch`.

**Fix:**
```ts
const [
  { data: session },
  { data: allParticipants, error: participantsError },
  ...
] = await Promise.all([...]);

if (participantsError) return { success: false, error: 'Failed to fetch participants' };
```

---

### 🟡 Medium — `fetchBattleSession` revalidation removal changes semantics

```diff
-  revalidateTag(CACHE_TAGS.BASE_BATTLE_SESSION(sessionId));
   const supabase = await createClient();
   return getBattleSessionCached(sessionId, supabase);
```

If `getBattleSessionCached` uses `unstable_cache` or fetch-level caching, the original `revalidateTag` call was the escape hatch guaranteeing fresh data. `completeBattleSession` now calls `revalidateTag(CACHE_TAGS.BASE_BATTLE_SESSION(sessionId))` synchronously (correct), so the completion flow is safe. The concern is other call sites of `fetchBattleSession` that may rely on the "always-bust-on-read" behaviour. Worth auditing all callers or adding an explanatory comment here.

---

### 🟡 Medium — Territory claim log loses `is_custom` distinction

The `after()` block always writes `"claimed territory"`. The old `logTerritoryClaimed` distinguished custom territories:
```ts
const territoryType = params.is_custom ? 'custom territory' : 'territory';
```
Minor behavioural regression in log descriptions. The `territory_id` field on the DB record (`!territory.territory_id → is_custom`) would be needed to restore it.

---

### 🟡 Medium — Inline `campaign_battles` insert forks from `normaliseParticipants`

`createBattleLog` used the shared `normaliseParticipants` utility. The new inline path recomputes the same values independently (correctly, for now). The maintenance risk is that future fixes to `normaliseParticipants` won't automatically apply here. A comment cross-referencing the utility would help.

---

### ✅ What's working well

- `Promise.all()` for the initial fetch is a clear win: session, participants, profile, and territory in one round trip instead of four sequential ones.
- `after()` is used correctly — the Supabase client is instantiated before `after()` so the auth token is already captured in the closure; deferred work is wrapped in a top-level `try/catch`.
- Batch `gang_logs` insert (N rows → 1 insert) is the right fix for the old N+1 logging loop.
- Campaign and standalone sessions now log through one unified code path instead of two divergent ones.
- `suppressRefetch` architecture is clean: `useRef` for the flag avoids stale closure issues, `useCallback` gives stable identity, and the `useEffect` cleanup correctly clears `suppressTimerRef`.

---

The territory error-swallowing (#1) can produce a `campaign_battles` record claiming a territory that was never actually updated — silent data corruption. The suppression-before-validation issue (#2) can leave the UI stuck after a valid completion. Both are straightforward fixes. The medium items are worth addressing in this PR given the scope of the refactor.


---

joeseos (commented):
## Second-Pass Review — #1744 (commit `04bafa1d`)

@joeseos second commit addresses all four items from the first review well. Noting what's fixed, then the remaining issues — mostly dead code introduced by the fix.

---

### ✅ Fixed from first review

| # | Issue | Status |
|---|-------|--------|
| 1 | Territory update error silently swallowed | Fixed — `claimError` checked, orphaned battle deleted on failure |
| 2 | `allParticipants` query error ignored | Fixed — `participantsError` destructured and returns early |
| 3 | `onSuppressRefetch()` fired before validation | Fixed — moved to after `setSubmitting(true)`, past all guards |
| 4 | Territory claim log missing `is_custom` | Fixed — `territory_id` added to query, `territoryIsCustom` computed and used |

---

### 🟡 Medium — Rollback delete swallows its own error

`app/actions/battle-sessions.ts`:

```ts
if (claimError) {
  await supabase.from('campaign_battles').delete().eq('id', campaign_battle_id);
  // ← delete result is discarded
  return { success: false, error: `Failed to claim territory: ${claimError.message}` };
}
```

If the compensating delete fails, the orphaned `campaign_battles` record stays silently. The caller sees the correct error, but the DB is now inconsistent. At minimum log the rollback failure:

```ts
if (claimError) {
  const { error: rollbackError } = await supabase
    .from('campaign_battles').delete().eq('id', battle.id);
  if (rollbackError) console.error('Failed to roll back campaign_battles:', rollbackError);
  return { success: false, error: `Failed to claim territory: ${claimError.message}` };
}
```

(Using `battle.id` directly is also safer than `campaign_battle_id` which TypeScript types as `string | undefined` even though it can't be here.)

---

### 🟡 Medium — `fetchBattleSession` revalidation removal still unaddressed

Carried from first review — if any call site uses `fetchBattleSession` expecting to always get fresh data, this silently serves stale results. Worth auditing callers or adding a comment.

---

### 🔵 Low — Dead code left behind by the `participantsError` early return (bloat)

Now that `allParticipants` is guaranteed non-null past the new early return, several defensive fallbacks become dead code. They're not wrong, but they're noise that makes the invariant less obvious:

**1. Inside `after()` — unreachable guard:**
```ts
after(async () => {
  try {
    if (!allParticipants) return;  // ← allParticipants cannot be null here
```

**2. Campaign block — redundant null check:**
```ts
if (session.campaign_id && allParticipants) {  // ← && allParticipants is dead
```

**3. Three `?? []` fallbacks on a non-nullable value:**
```ts
const flaggedWinnerIds = (allParticipants ?? []).filter(...)
// ...
(allParticipants ?? []).find(...)
// ...
(allParticipants ?? []).find(...)
```

All three `?? []` can be removed. Dropping the dead checks makes the narrowed type visible to the reader (and TypeScript) rather than implying the value could still be null.

---

### 🔵 Low — `campaignName` is a pointless intermediate variable (bloat)

```ts
const campaignName = session.campaign_id ? undefined : 'Standalone Battle';
let resolvedCampaignName = campaignName;
if (session.campaign_id) {
  resolvedCampaignName = campaign?.campaign_name || 'Campaign';
}
```

`campaignName` is created and immediately copied to `resolvedCampaignName` before being discarded. It also gives `resolvedCampaignName` an inferred type of `string | undefined`, meaning the template literal `(Campaign: ${resolvedCampaignName})` would render `"Campaign: undefined"` if the branch logic changed. Simpler and typed as `string`:

```ts
let resolvedCampaignName: string;
if (session.campaign_id) {
  const { data: campaign } = await supabase.from('campaigns')...
  resolvedCampaignName = campaign?.campaign_name || 'Campaign';
} else {
  resolvedCampaignName = 'Standalone Battle';
}
```

---

### 🔵 Low — Two `gang_logs` inserts where one would do (minor DRY)

The `after()` block sends the battle logs as a batch, then sends the territory claim as a second separate insert:

```ts
if (logs.length > 0) {
  await supabase.from('gang_logs').insert(logs);   // insert 1
}

if (options?.campaign_territory_id && claimerGangId && claimed_territory) {
  await supabase.from('gang_logs').insert({ ... }); // insert 2
}
```

The territory entry is the same shape as the battle entries. Push it onto `logs` before the single insert to keep the "batch everything" pattern consistent and save a round-trip:

```ts
if (options?.campaign_territory_id && claimerGangId && claimed_territory && gangMap.get(claimerGangId)) {
  const claimer = gangMap.get(claimerGangId)!;
  logs.push({
    gang_id: claimerGangId,
    user_id: claimer.user_id,
    action_type: 'territory_claimed',
    description: `Gang "${claimer.name}" claimed ${territoryType} "${claimed_territory}" in campaign "${resolvedCampaignName}"`,
    created_at: new Date().toISOString(),
  });
}

if (logs.length > 0) {
  await supabase.from('gang_logs').insert(logs);
}
```

---

### Summary

The two high-severity bugs from the first pass are fixed. What remains is one medium concern (rollback error swallowed), one carry-over (fetchBattleSession), and a cluster of low-severity dead code and a redundant variable that the `participantsError` fix left behind. Clean those up and this is good to merge.


---

PR #1745, opened 2026-06-07
Added a dropdown to Add Special Rules in the Edit Vehicle modal that …

Djidiouf:
…list default rules the vehicle normally has if they were previously removed, offer a custom options, works similarly to scenario selection

### DIFF
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -1308,8 +1308,8 @@ export function EditFighterModal({
                 Special Rules
               </label>
               <div className="mb-2">
-                <div className={selectedSpecialRuleOption !== 'custom' ? 'flex space-x-2' : ''}>
-                  <div className={selectedSpecialRuleOption !== 'custom' ? 'grow' : ''}>
+                <div className="flex space-x-2">
+                  <div className="grow min-w-0">
                     <Combobox
                       options={specialRuleComboboxOptions}
                       value={selectedSpecialRuleOption}
--- a/components/gang/gang-page-content.tsx
+++ b/components/gang/gang-page-content.tsx
@@ -910,6 +910,7 @@ export default function GangPageContent({
           vehicles={gangData.processedData.vehicles || []}
           fighters={gangData.processedData.fighters || []}
           gangId={gangId}
+          gangTypeId={gangData.processedData.gang_type_id}
           onVehicleUpdate={handleVehicleUpdate}
           onFighterUpdate={handleFighterUpdate}
           userPermissions={userPermissions}
--- a/components/gang/vehicle-edit.tsx
+++ b/components/gang/vehicle-edit.tsx
@@ -9,9 +9,16 @@ import { Combobox } from "@/components/ui/combobox";
 import { LuPlus } from "react-icons/lu";
 import { LuMinus } from "react-icons/lu";
 import { HiX } from "react-icons/hi";
+import { toast } from 'sonner';
 import { VehicleProps, VehicleEffect } from '@/types/vehicle';
 
 const normalizeSpecialRule = (rule: string) => rule.replace(/^"|"$/g, '');
+const DEFAULT_SPECIAL_RULES_HEADER_VALUE = '__combobox_header__:default_special_rules';
+
+type VehicleTypeOption = {
+  id: string;
+  special_rules: string[] | null;
+};
 
 type CombinedVehicleProps = VehicleProps & {
   assigned_to?: string;
@@ -19,6 +26,7 @@ type CombinedVehicleProps = VehicleProps & {
 
 interface VehicleEditProps {
   vehicle: CombinedVehicleProps | null;
+  gangTypeId?: string | null;
   onClose: () => void;
   onSave: (vehicleId: string, vehicleName: string, specialRules: string[], statAdjustments?: Record<string, number>) => Promise<boolean>;
   isLoading?: boolean;
@@ -346,6 +354,7 @@ function VehicleCharacteristicModal({
 
 export default function VehicleEdit({
   vehicle,
+  gangTypeId,
   onClose,
   onSave,
   isLoading = false
@@ -357,14 +366,17 @@ export default function VehicleEdit({
   const [showStatsModal, setShowStatsModal] = useState(false);
   const [pendingStatAdjustments, setPendingStatAdjustments] = useState<Record<string, number>>({});
 
-  const { data: vehicleTypes = [] } = useQuery<Array<{ id: string; special_rules: string[] }>>({
-    queryKey: ['vehicle-types', vehicle?.gang_id],
+  const gangId = vehicle?.gang_id;
+
+  const { data: vehicleTypes = [] } = useQuery<VehicleTypeOption[]>({
+    queryKey: ['vehicle-types', gangTypeId ?? gangId],
     queryFn: async () => {
-      const response = await fetch(`/api/gangs/${vehicle!.gang_id}/vehicles`);
+      if (!gangId) throw new Error('No gang_id');
+      const response = await fetch(`/api/gangs/${gangId}/vehicles`);
       if (!response.ok) throw new Error('Failed to fetch vehicle types');
       return response.json();
     },
-    enabled: !!vehicle?.gang_id,
+    enabled: !!gangId,
     staleTime: 10 * 60 * 1000,
   });
 
@@ -397,7 +409,7 @@ export default function VehicleEdit({
 
     if (availableDefaultSpecialRules.length > 0) {
       options.push({
-        value: '__default_special_rules_header__',
+        value: DEFAULT_SPECIAL_RULES_HEADER_VALUE,
         label: <span className="font-bold">Default Special Rules</span>,
         displayValue: 'Default Special Rules',
         disabled: true,
@@ -447,6 +459,7 @@ export default function VehicleEdit({
 
     const normalisedCurrent = vehicleSpecialRules.map(normalizeSpecialRule);
     if (normalisedCurrent.includes(normalizeSpecialRule(ruleToAdd))) {
+      toast.error('This rule is already applied');
       setSelectedSpecialRuleOption('');
       setCustomSpecialRule('');
       return;
--- a/components/gang/vehicle-edit.tsx
+++ b/components/gang/vehicle-edit.tsx
@@ -1,13 +1,17 @@
 'use client';
 
 import { useState, useEffect, useMemo } from 'react';
+import { useQuery } from '@tanstack/react-query';
 import Modal from "@/components/ui/modal";
 import { Input } from "@/components/ui/input";
 import { Button } from "@/components/ui/button";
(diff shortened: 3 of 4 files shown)


joeseos:
## Code Review — PR #1745

@Djidiouf Thanks for this feature! The core idea is clean and the implementation is mostly solid. A few items to address before merge.

---

### Findings

---

#### 🔴 HIGH — API endpoint fetches the wrong resource shape (type mismatch risk)

**`components/gang/vehicle-edit.tsx`, new `useQuery` block**

```ts
const { data: vehicleTypes = [] } = useQuery<Array<{ id: string; special_rules: string[] }>>({
  queryFn: async () => {
    const response = await fetch(`/api/gangs/${vehicle!.gang_id}/vehicles`);
```

The `/api/gangs/[id]/vehicles` GET handler returns full rows from the `vehicle_types` table (confirmed in `app/api/gangs/[id]/vehicles/route.ts`). The DB column `special_rules` is not guaranteed to be `string[]` — it could be `string[] | null` in practice.

The `|| []` guard in `availableDefaultSpecialRules` handles the null case fine, but the TypeScript type on the query declares `special_rules: string[]` (non-nullable), masking this. If the DB ever returns `null` for `special_rules` and the guard is missed somewhere downstream, you'd get a runtime error.

**Fix:** Type the query more accurately:
```ts
useQuery<Array<{ id: string; special_rules: string[] | null }>>
```
The existing `vehicleType?.special_rules || []` already handles it correctly — this is just making the type honest.

---

#### 🟡 MEDIUM — `vehicle!` non-null assertion in async queryFn closure

**`components/gang/vehicle-edit.tsx`, `queryFn`**

```ts
queryFn: async () => {
  const response = await fetch(`/api/gangs/${vehicle!.gang_id}/vehicles`);
```

The `enabled: !!vehicle?.gang_id` guard prevents the query from running when `vehicle` is null. However, the queryFn captures `vehicle` from the outer closure at the time the query is defined — not at the time the queryFn is eventually called by React Query. If the component re-renders between definition and execution (e.g., modal closes while a refetch is in flight), `vehicle` could be `null` in the closure and the `!` assertion would throw.

**Fix:** Capture the value inside the queryFn:
```ts
queryFn: async () => {
  const gangId = vehicle?.gang_id;
  if (!gangId) throw new Error('No gang_id');
  const response = await fetch(`/api/gangs/${gangId}/vehicles`);
```

---

#### 🟡 MEDIUM — Query cache key is too broad

**`components/gang/vehicle-edit.tsx`**

```ts
queryKey: ['vehicle-types', vehicle?.gang_id],
```

This fetches all vehicle types compatible with the gang on every distinct gang. But what you actually need is the vehicle types for a given `gang_type_id`, not `gang_id` — two different gangs of the same gang type will have identical vehicle type lists but will generate separate cache entries and separate network requests.

The API itself does a two-step: fetch `gang_type_id` from `gangs`, then fetch `vehicle_types` filtered by that `gang_type_id`. Caching by `gang_id` is overly granular.

This is a minor inefficiency, but if users switch between multiple gangs of the same type, you pay extra fetch costs unnecessarily.

---

#### 🟡 MEDIUM — Silent duplicate-rule rejection gives no feedback

**`components/gang/vehicle-edit.tsx`, `handleAddSpecialRule`**

```ts
if (normalisedCurrent.includes(normalizeSpecialRule(ruleToAdd))) {
  setSelectedSpecialRuleOption('');
  setCustomSpecialRule('');
  return;
}
```

When a duplicate is detected, the selection is silently cleared with no message to the user. They have no idea why their Add did nothing. This is pre-existing behaviour for the custom input case, but it's more surprising when selecting from the dropdown — the user explicitly picked a "default rule" and clicked Add, and nothing happened.

**Fix:** A simple toast or inline warning (`"This rule is already applied"`) would significantly improve discoverability.

---

#### 🟡 MEDIUM — Layout shifts when toggling custom mode

**`components/gang/vehicle-edit.tsx`, JSX section ~line 535**

```tsx
<div className={selectedSpecialRuleOption !== 'custom' ? 'flex space-x-2' : ''}>
  <div className={selectedSpecialRuleOption !== 'custom' ? 'grow' : ''}>
```

Toggling between custom and non-custom modes causes a noticeable layout shift: the Combobox resizes (loses `grow`) and a new row appears below. This can feel jarring in a modal.

**Suggestion:** Keep the Combobox at full width always (`grow` or `w-full`) and conditionally render only the secondary text input below it — the "Add" button stays in the same position next to the input regardless of mode. This would make the mode switch feel smoother.

---

#### 🟢 LOW — `__default_special_rules_header__` is a fragile sentinel

```ts
value: '__default_special_rules_header__',
```

If any vehicle type ever has a special rule with this exact string (unlikely but not impossible), it would be silently treated as a disabled header. Consider using a `Symbol` or a type-discriminated structure instead.

---

#### 🟢 LOW — Duplicate detection is case-sensitive

```ts
const normalisedCurrent = vehicleSpecialRules.map(normalizeSpecialRule);
if (normalisedCurrent.includes(normalizeSpecialRule(ruleToAdd))) {
```

`normalizeSpecialRule` only strips surrounding quotes — it does not normalise case. Adding `"Rugged"` when `"rugged"` already exists won't be caught. Pre-existing behaviour, but worth noting as a follow-up.

---

### Summary

| 1 | `vehicle-edit.tsx` | `special_rules` typed as non-nullable in query, DB can return null | High |
| 2 | `vehicle-edit.tsx` | `vehicle!` non-null assertion in async queryFn closure | Medium |
| 3 | `vehicle-edit.tsx` | Cache key scoped to `gang_id` rather than `gang_type_id` | Medium |
| 4 | `vehicle-edit.tsx` | Silent failure on duplicate-rule detection | Medium |
| 5 | `vehicle-edit.tsx` | Layout shift when switching custom/non-custom mode | Medium |
| 6 | `vehicle-edit.tsx` | `__default_special_rules_header__` sentinel fragility | Low |
| 7 | `vehicle-edit.tsx` | Case-sensitive duplicate detection | Low |

---

The feature logic itself is correct and the edge cases (empty defaults, all rules already added, normalisation of stored rules) are handled well. Issues 1 and 2 carry real runtime risk and should be fixed before merge. Issues 3–5 are worth addressing for robustness and UX quality.

---

joeseos (approved):
## Code Review — PR #1745 (follow-up)

@Djidiouf Great turnaround — the substantive issues from the first review have all been addressed. Here's the full status rundown.

---

### Previous Findings

| 1 | High | `special_rules` typed as non-nullable in query | ✅ **Resolved** — `VehicleTypeOption` type now correctly declares `special_rules: string[] \| null` |
| 2 | Medium | `vehicle!` non-null assertion in async queryFn closure | ✅ **Resolved** — `gangId` extracted outside the fn; explicit guard `if (!gangId) throw` added |
| 3 | Medium | Cache key scoped to `gang_id` rather than `gang_type_id` | ✅ **Resolved** — `gangTypeId` is now threaded all the way from `gang-page-content` → `vehicles-tab` → `vehicle-edit`, and `queryKey: ['vehicle-types', gangTypeId ?? gangId]` makes two gangs of the same type share the cache entry |
| 4 | Medium | Silent failure on duplicate-rule detection | ✅ **Resolved** — `toast.error('This rule is already applied')` fires correctly |
| 5 | Medium | Layout shift when toggling custom/non-custom mode | ⚠️ **Partially resolved** — the custom input now correctly drops to a second row rather than replacing the combobox, which is cleaner. The combobox itself still changes width when the Add button appears/disappears (it gains/loses `grow`), so a small shift is still visible. Acceptable for now; a follow-up pass could simply always apply `grow` to the combobox wrapper. |
| 6 | Low | `__default_special_rules_header__` sentinel fragility | ✅ **Resolved** — named constant `DEFAULT_SPECIAL_RULES_HEADER_VALUE = '__combobox_header__:default_special_rules'` is far more collision-resistant and eliminates the magic string |
| 7 | Low | Case-sensitive duplicate detection | ⏭️ **Outstanding** — pre-existing behaviour, acceptable as a separate follow-up |

---

### New Findings

#### 🟢 LOW — Add button is visible and clickable when nothing is selected

**`components/gang/vehicle-edit.tsx`, JSX section**

When `selectedSpecialRuleOption` is `''` (the initial/reset state), the condition `selectedSpecialRuleOption !== 'custom'` is `true`, so the Add button renders. Clicking it hits the `if (!ruleToAdd) return` guard and silently does nothing. No crash, but the button appearing active when nothing is chosen is slightly misleading.

**Suggestion (optional):** Disable the button when the selection is empty:
```tsx
<Button
  onClick={handleAddSpecialRule}
  type="button"
  disabled={!selectedSpecialRuleOption}
>
  Add
</Button>
```

---

### Verdict: ✅ Approve

All high and medium issues are resolved. The feature logic, null safety, caching, and user feedback are all correct. The two remaining items (partial layout shift, empty-selection button) are polish-level and can be addressed as a follow-up without blocking merge.

---

joeseos:
## Code Review — PR #1745 (follow-up, commit `ae1797e`)

@Djidiouf Quick turnaround again — the new commit is a clean, targeted fix. Full status below.

---

### Previous Findings

| 1 | High | `special_rules` typed as non-nullable in query | ✅ **Resolved** (prior commit) |
| 2 | Medium | `vehicle!` non-null assertion in async queryFn closure | ✅ **Resolved** (prior commit) |
| 3 | Medium | Cache key scoped to `gang_id` rather than `gang_type_id` | ✅ **Resolved** (prior commit) |
| 4 | Medium | Silent failure on duplicate-rule detection | ✅ **Resolved** (prior commit) |
| 5 | Medium | Layout shift when toggling custom/non-custom mode | ✅ **Fully resolved** — this commit removes the conditional class assignments in both `vehicle-edit.tsx` and `fighter-edit-modal.tsx`. Both now apply `flex space-x-2` and `grow min-w-0` unconditionally, eliminating the combobox width shift entirely. |
| 6 | Low | `__default_special_rules_header__` sentinel fragility | ✅ **Resolved** (prior commit) |
| 7 | Low | Case-sensitive duplicate detection | ⏭️ **Outstanding** — pre-existing, accepted as separate follow-up |
| NEW | Low | Add button visible/clickable when selection is empty | ⏭️ **Outstanding** — optional polish, not a blocker |

---

### New Findings

The new commit (`ae1797e`) is 8 lines total: 2 additions + 2 deletions in each of the two files. No new issues.

The changes are correct:
- `min-w-0` is the right companion to `grow` inside a flex container — it prevents flex children from overflowing when content is long, which is important for the Combobox here.
- Applying `flex space-x-2` unconditionally gives the layout stable structure regardless of whether custom mode is active.
- The approach is now consistent between `vehicle-edit.tsx` and `fighter-edit-modal.tsx`.

**Minor process note** (no action needed): this vehicle-focused PR retroactively touches `fighter-edit-modal.tsx` to fix the same layout pattern. The change is correct and the cleanup is welcome, but it does extend the blast radius of this PR beyond its stated scope. Worth noting for future branching decisions — not a reason to hold up merge.

---

### Verdict: ✅ Approve

All high and medium findings are fully resolved. The new commit cleanly closes the last remaining medium issue (finding #5). No regressions. Ready to merge.

---

PR #1746, opened 2026-06-08
Added the promotion to champion flow for specialists

Djidiouf:


joeseos:
## Code Review — PR #1746: Promotion to Champion flow for Specialists

Hey @Djidiouf — solid feature overall. The refactoring (index → ID, `buildSkillSetComboboxOptions` extracted, `normalizeSpecialRules` extracted) is all clean. One **high-severity** correctness bug and two medium-severity issues need addressing before merge.

---

### 🔴 HIGH — Partial mutation: promoted on server, skill never recorded

**File:** `components/fighter/fighter-advancement-list.tsx` — `applyChampionPromotionMutation`

`updateFighterDetails` and `addSkillAdvancement` are called sequentially inside the same `mutationFn`. If the first call succeeds and the second fails:

- The server has promoted the fighter (class/type changed permanently).
- `onError` rolls back the optimistic skill and XP updates in the UI.
- But `onFighterDetailsUpdate` is only called in `onSuccess` — so the UI doesn't reflect the promotion either.
- On next page load: the fighter is a Champion with no skill gained and no XP spent.

The `onError` rollback for the second operation cannot undo the first server-side write.

**Fix options (pick one):**
1. **Preferred**: Create a single server action / API route that performs both operations in one DB transaction.
2. If kept client-side: on error, call `updateFighterDetails` again with the *original* fighter class/type to attempt a revert, then surface a clear error.

```ts
// onError — add fighter details rollback
onError: (e, vars, context) => {
  // attempt to revert the fighter class change
  void updateFighterDetails({ fighter_id: fighterId, /* original values */ });
  if (context?.previousSkills && onSkillUpdate) onSkillUpdate(context.previousSkills);
  if (vars && onXpCreditsUpdate) onXpCreditsUpdate(vars.xp_cost, -vars.credits_increase);
  toast.error(e?.message || 'Failed to apply promotion and skill');
}
```

---

### 🟡 MEDIUM — Champion promotion validates against current skill access, not the target Champion type

**File:** `components/fighter/fighter-advancement-list.tsx` — skill access `useEffect` + `executeChampionPromotionPurchase`

The champion promotion path fetches skill access without `previewFighterTypeId`:

```ts
// This is correct for ganger → specialist:
`/api/fighters/skill-access?fighterId=...&previewFighterTypeId=${gangerPendingPromotion.fighter_type_id}`

// But the champion path uses:
`/api/fighters/skill-access?fighterId=${fighterId}`  // ← no preview type!
```

So "Primary" is validated against the *Specialist's* access, not the *Champion's*. A set that's Primary for a Specialist may be Secondary (or absent) for the Champion type they're promoting into. The user could pick a set that passes the UI guard but is semantically wrong for the Champion.

**Fix:** Add a separate `useEffect` for champion promotion skill access (mirroring the ganger specialist pattern), fetching with `previewFighterTypeId=championPendingPromotion.fighter_type_id`. Pass `highlightPrimaryOnly: true` and validate against that response.

---

### 🟡 MEDIUM — XP cost resets on every skill-set change during champion promotion

**File:** `components/fighter/fighter-advancement-list.tsx` — around the `promotion_to_champion` XP useEffect

```ts
useEffect(() => {
  if (advancementType === 'promotion_to_champion') {
    setEditableXpCost(CHAMPION_PROMOTION_XP_COST);
    setEditableCreditsIncrease(CHAMPION_PROMOTION_CREDITS_INCREASE);
  }
}, [advancementType, selectedCategory]); // ← selectedCategory causes repeated resets
```

Every time the user changes the skill set dropdown, the XP cost is reset to 12. If a GM edits the cost before selecting a skill set, the change is lost. The dependency should just be `[advancementType]` since the intent is to set the cost once when switching to this type.

```ts
}, [advancementType]);
```

---

### 🔵 LOW — `normalizeSpecialRules` has a misleading type signature

**File:** `components/fighter/edit-fighter/fighter-promotion-modal.tsx` (lines 23–28)

```ts
function normalizeSpecialRules(rules: string[]): string[] {
  return rules
    .map((r) => (typeof r === 'string' ? normalizeSpecialRule(r) : r))  // ← if r is string[], this is always true
    .filter(Boolean);
}
```

The `typeof r === 'string'` guard inside a `string[]`-typed function is a code smell that signals dirty data at the source. TypeScript will never flag `typeof r === 'string'` as always-true here because it trusts the type. But the guard only exists because non-string values DO arrive. The type should honestly be `(string | unknown)[]` or the data layer should be sanitised upstream. As-is this is a lint/correctness hazard waiting to bite.

---

### 🔵 LOW — Non-primary skill sets are selectable, producing confusing UX

**Files:** Both `fighter-advancement-list.tsx` sections (ganger specialist + champion)

Non-primary sets are greyed/italic in the combobox but remain fully selectable. After selecting one, an amber warning appears and the Buy button is disabled. Users have no upfront affordance that they're choosing an ineligible option until after selection.

Consider marking non-primary options as `disabled: true` in `buildSkillSetComboboxOptions` when `highlightPrimaryOnly` is true, so the combobox itself prevents the selection.

---

### ✅ What's good

- **`buildSkillSetComboboxOptions`** extracted cleanly — good DRY improvement shared between ganger and champion flows.
- **`gangerSelectedSkillSetId` replacing `gangerSelectedSetIndex`** — more stable (no off-by-one risk, survives list reordering).
- **`normalizeSpecialRules`** extracted from three inline duplications.
- **Skill categories + skill access fetched independently** for the specialist flow — correct separation of concerns.
- **`isSkillLikeAdvancementType`** derived boolean keeps skill/champion branches consistent without repeated string literals.
- `cancelToken` pattern in all new `useEffect` fetches — properly prevents stale-state bugs.
- **`characteristicComboboxOptions`** split out from `skillSetComboboxOptions` — cleaner and avoids unnecessary recomputation.

---

**Verdict: 🔴 Changes requested**

The non-atomic mutation (#1) needs a concrete fix before this ships — a partially-promoted fighter with missing data will be very hard to debug in production. #2 (wrong skill access for validation) is also worth fixing in the same PR since it affects correctness of the core feature. Happy to discuss approach on any of these.


---

joeseos:
## Follow-up Review — PR #1746 (Promotion to Champion flow)

Hey @Djidiouf — no new commits have landed since the previous review, so this is a re-verification pass confirming the status of each prior finding against the current diff. All five are still outstanding.

---

### Previous findings status

#### 🔴 HIGH — Partial mutation: promoted on server, skill never recorded — **OUTSTANDING**

Confirmed in `applyChampionPromotionMutation` (around line ~1006 in `fighter-advancement-list.tsx`):

```ts
// mutationFn — two sequential server calls with no rollback between them
const u = await updateFighterDetails({ ... });    // ← succeeds
if (!u.success) throw new Error(...);
const s = await addSkillAdvancement({ ... });      // ← fails here
if (!s.success) throw new Error(...);
```

`onError` only rolls back the optimistic skill/XP state in the client. It does not attempt to revert the `updateFighterDetails` call. `onFighterDetailsUpdate` is only called in `onSuccess`. Net result on failure of the second call: fighter is a Champion on the server, but the parent component still thinks they're a Specialist — until the page reloads.

**Still needs a fix before merge** (either a single server-side transaction, or an explicit revert call in `onError`).

---

#### 🟡 MEDIUM — Champion promotion validates against Specialist's skill access, not Champion's — **OUTSTANDING**

The `skillAccess` useEffect fetches with no `previewFighterTypeId`:

```ts
// ~line 1888 — same effect used for both 'skill' and 'promotion_to_champion'
const res = await fetch(
  `/api/fighters/skill-access?fighterId=${encodeURIComponent(fighterId)}`
  //                                     ↑ no previewFighterTypeId passed
);
```

This Specialist access feeds `skillSetComboboxOptions` (so the "Primary" / "Secondary" labels in the combobox are for the Specialist's access, not the Champion's), and it's also the access that drives `selectedSkillSetAccess !== 'primary'` in `buyAdvancementDisabled` and in `executeChampionPromotionPurchase`. The entire validation chain — visual annotations AND the buy guard — is based on the wrong fighter type.

**Needs a separate `useEffect` for champion skill access, fetching with `previewFighterTypeId: championPendingPromotion.fighter_type_id`.**

---

#### 🟡 MEDIUM — XP cost resets on every skill-set change — **OUTSTANDING**

```ts
useEffect(() => {
  if (advancementType === 'promotion_to_champion') {
    setEditableXpCost(CHAMPION_PROMOTION_XP_COST);
    setEditableCreditsIncrease(CHAMPION_PROMOTION_CREDITS_INCREASE);
  }
}, [advancementType, selectedCategory]);  // ← selectedCategory causes repeated resets
```

Any skill-set selection resets XP to 12, clobbering any GM override. The dependency should be `[advancementType]` only.

---

#### 🔵 LOW — `normalizeSpecialRules` has a misleading type signature — **OUTSTANDING**

```ts
function normalizeSpecialRules(rules: string[]): string[] {
  return rules
    .map((r) => (typeof r === 'string' ? normalizeSpecialRule(r) : r))  // guard always true per the declared type
    .filter(Boolean);
}
```

Present in both changed files. The `typeof r === 'string'` guard signals dirty upstream data that the type doesn't admit. Either change the signature to `(string | unknown)[]` or sanitise at the data layer.

---

#### 🔵 LOW — Non-primary skill sets remain selectable — **OUTSTANDING**

`buildSkillSetComboboxOptions` renders non-primary options as grey/italic when `highlightPrimaryOnly` is true, but does not set `disabled: true` on them. Users can select a non-eligible set, see the amber warning, and find the Buy button blocked — all after the fact. Consider adding `disabled: true` to non-primary entries when `highlightPrimaryOnly` is true.

---

### New findings

#### 🟡 MEDIUM — `buyAdvancementDisabled` Buy-button guard compounds the wrong-access bug

```ts
advancementType === 'promotion_to_champion'
  ? !championPendingPromotion ||
    !selectedCategory ||
    selectedSkillSetAccess !== 'primary' ||   // ← derived from Specialist's skillAccess
```

`selectedSkillSetAccess` is computed from `skillAccess` (the Specialist's access, per the finding above). A skill set that is Primary for the Specialist but not for the Champion will incorrectly enable the Buy button. This is a direct consequence of the MEDIUM bug but is worth calling out as its own code site since it also needs to be updated when the correct access fetch is added.

---

### CI

Only the Vercel Preview check is running; it's green. No TypeScript or lint CI is configured on this PR.

---

**Verdict: 🔴 Changes requested**

The two medium findings (#2 wrong-access chain and #1 non-atomic mutation) are correctness bugs in the core of the new feature. Happy to discuss implementation approach for the transaction fix — a single server action wrapping both DB writes would be cleanest.

---

joeseos (approved):
## Follow-up Review — PR #1746 (Promotion to Champion — updated)

Hey @Djidiouf — great turnaround. All previous findings are addressed. Details below, then three small new LOW items.

---

### Previous Findings Status

| # | Finding | Status |
|---|---------|--------|
| 🔴 HIGH | Partial mutation — promoted on server, skill never recorded | ✅ **Resolved** |
| 🟡 MEDIUM | Champion validates against Specialist's skill access, not Champion's | ✅ **Resolved** |
| 🟡 MEDIUM | XP cost resets on every skill-set change | ✅ **Resolved** |
| 🟡 MEDIUM (follow-up) | `buyAdvancementDisabled` compounding wrong-access bug | ✅ **Resolved** |
| 🔵 LOW | `normalizeSpecialRules` misleading type signature | ✅ **Resolved** |
| 🔵 LOW | Non-primary skill sets selectable, confusing UX | ✅ **Resolved** |

**How each was fixed:**

- **Atomicity** — `applyPromotionWithSkillAdvancement` is a proper server action that snapshots pre-state, promotes, and if the skill write fails performs a best-effort revert with an explicit "check Edit Fighter" error message if the revert also fails. Clean option-2 implementation.
- **Champion skill access** — `championPreviewSkillAccess` is now fetched with `previewFighterTypeId=championPendingPromotion.fighter_type_id` in a dedicated `useEffect`. `selectedSkillSetAccess`, `skillSetComboboxOptions`, `buyAdvancementDisabled`, and `executeChampionPromotionPurchase` all source from this correct access set.
- **XP cost reset** — `useEffect` dep array trimmed to `[advancementType]` only. ✅
- **`normalizeSpecialRules`** — signature is now `(string | unknown)[]` and non-string values are handled with `String(r)`. Honest and correct. ✅
- **Non-primary disabled** — `buildSkillSetComboboxOptions` now spreads `{ disabled: true }` when `highlightPrimaryOnly && effectiveLevel !== 'primary'`. ✅

---

### New Findings

#### 🔵 LOW — `applyPromotionWithSkillAdvancement`: exception during revert loses the partial-state warning

**File:** `app/actions/fighter-advancement.ts` — lines 604–630

The `!revertResult.success` branch (line 614) correctly returns the "could not be reverted automatically" message. But if the revert `updateFighterDetails` call **throws** an exception instead of returning `{ success: false }`, control falls to the outer `catch` (line 625), which returns `error.message ?? 'Unknown error occurred'` — with no mention that the fighter might already be promoted.

This is an edge case (server actions normally return rather than throw), but if it ever fires the support trail goes cold.

Suggested hardening:

```ts
if (!skillResult.success) {
  let revertResult;
  try {
    revertResult = await updateFighterDetails({ ... });
  } catch {
    revertResult = { success: false };
  }
  if (!revertResult.success) {
    return {
      success: false,
      error: `${skillResult.error}. Promotion could not be reverted automatically — please check this fighter's class and type in Edit Fighter.`
    };
  }
  return { success: false, error: skillResult.error || 'Failed to add skill' };
}
```

---

#### 🔵 LOW — Silent empty state when a skill set has no available skills

**File:** `components/fighter/fighter-advancement-list.tsx` — around lines 2272 and 2413

Both the ganger-specialist and champion flows gate the skill picker on `gangerSkillsInSet.length > 0` / `availableAdvancements.length > 0`. When a set is selected but returns zero skills (e.g., fighter already owns every skill in that set), the picker section disappears silently — no feedback. There's a loading pulse for the fetching state but no empty state.

Adding a single line covers both:

```tsx
{gangerPendingPromotion && gangerSelectedSkillSetId && !gangerSkillsInSetLoading && gangerSkillsInSet.length === 0 && (
  <p className="text-sm text-muted-foreground">No skills available in this set.</p>
)}
```

---

#### 🔵 LOW — `Exotic Beast Specialist` offered 'Promotion to Champion' — verify intent

**File:** `components/fighter/fighter-advancement-list.tsx` — `advancementTypeComboboxOptions` (~line 1500)
**File:** `components/fighter/edit-fighter/fighter-promotion-modal.tsx` — `PROMOTION_MAP`

`advancementTypeComboboxOptions` adds `promotion_to_champion` for `fighterClass === 'Exotic Beast Specialist'`, and `PROMOTION_MAP` maps `'Exotic Beast Specialist' → 'Champion'`. If exotic beasts promote to a separate "Exotic Beast Champion" class (rather than a standard Champion class), the target class derivation would be wrong. Worth a quick sanity-check against the Necromunda rules, since this path is now live.

---

### What's Good

- `buildSkillSetComboboxOptions` extracted cleanly and reused across both flows.
- `gangerSelectedSkillSetId` replacing the old index is a meaningful stability improvement.
- Loading pulse for skill fetching is a nice UX touch.
- `cancelToken` pattern in all new `useEffect` hooks — no stale-state races.
- The server action comment block accurately documents the compensation logic.

---

**Verdict: ✅ Approve**

All HIGH and MEDIUM issues are resolved. The three new LOWs are non-blocking — the feature's correctness and core UX are solid.

---

PR #1752, opened 2026-06-11
Fixed log modal footer on mobile not being displayed when the screen …

Djidiouf:
…height was too small

### DIFF
--- a/components/fighter/fighter-logs.tsx
+++ b/components/fighter/fighter-logs.tsx
@@ -134,7 +134,7 @@ export default function FighterLogs({ gangId, fighterId, fighterName, vehicleId,
       onClose={onClose}
       width="4xl"
     >
-      <div className="max-h-[70vh] min-h-[400px] overflow-y-auto">
+      <div className="max-h-[min(70vh,calc(100svh-12rem))] overflow-y-auto">
         <table className="w-full table-auto">
           <thead className="sticky top-0 bg-card z-10 shadow-xs">
             <tr className="bg-muted">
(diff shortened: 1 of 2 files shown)


joeseos (approved):
## PR Review — `fix/log_modal_footer_on_mobile`

Hey @Djidiouf, clean fix — the root cause diagnosis is correct and the approach is sound. Here's the full breakdown:

---

### What this PR does

1. **Height constraint** — replaces `max-h-[70vh] min-h-[400px]` with `max-h-[min(70vh,calc(100svh-12rem))]` on the scrollable table div in both log components.
2. **Pagination footer** — makes the footer responsive: stacks vertically on mobile, reverts to a horizontal row at `sm` breakpoint.

---

### Findings

**`fighter-logs.tsx:137` / `gang-logs.tsx:135` — Height constraint formula**
Severity: **low**

The `12rem` offset is a magic number accounting for modal header + content padding + pagination footer. It's reasonable for the current modal structure, but it's load-bearing: if the modal header ever gains more helper-text lines or the pagination footer changes height, pagination could start getting clipped again. Not blocking, but worth a comment or a named CSS custom property if this pattern spreads to more modals.

**`min-h-[400px]` removal**
Severity: **low**

Removing the minimum height fixes the mobile overflow correctly, but it's a visual trade-off on desktop: with only 1–2 log entries the table will collapse to just those rows instead of holding a fixed minimum height. This is probably acceptable (logs are almost always non-trivial), but worth knowing about.

**`100svh` unit — correct call**
`svh` (small viewport height) is the right unit here — it excludes the collapsible browser chrome on mobile, which is exactly the scenario that was causing the bug. Browser support is fine for the target audience (Chrome 108+, Firefox 101+, Safari 15.4+).

**Nested scroll containers**
Severity: **low** (pre-existing, not introduced here)

The `Modal` wraps its content in `overflow-y-auto flex-1`, and the table div inside also has `overflow-y-auto`. The PR's `max-h` formula is designed to keep the inner scroll active and prevent the outer modal from needing to scroll — which is the right strategy. Just flagging it as something to keep in mind if the modal structure changes.

**Code duplication — `fighter-logs.tsx` / `gang-logs.tsx`**
Severity: **low** (pre-existing)

Both files now have identical pagination UI and height logic. Not introduced by this PR, but a future extraction of a shared `<LogModal>` or `<PaginationBar>` component would eliminate this class of bug for free.

---

### Verdict: ✅ Approve

The fix is correct and targets the actual cause. The responsive pagination stacking is clean and consistent. No regressions expected on desktop — the only change in desktop behaviour is that very sparse log lists no longer hold a 400 px floor, which is a minor cosmetic trade-off well worth the mobile fix.

---

PR #1753, opened 2026-06-11
Allow campaign owners/arbitrators to edit battle sessions

joeseos:
## Summary

Campaign **OWNERs/ARBITRATORs** (and site admins) can now edit battle sessions belonging to their campaign — the same model as the gang page, where arbitrators have full edit rights over gangs in their campaign.

The RLS policies on `battle_sessions` / `battle_session_participants` / `battle_session_fighters` already permitted this (`private.is_admin() OR private.is_arb(campaign_id)`); this PR wires it through the TypeScript server actions and the UI. **No DB migration.**

### What arbitrators can now do
- **Session management** (previously creator-only): cancel, change scenario, phase transitions, add/remove players, set winners, complete battle, advance/revert rounds.
- **Participant data** (previously the player only): crew selection, loadouts, XP/injuries/conditions, ready toggle, roles, credits/reputation/resource outcomes.

Sessions without a `campaign_id` grant arbitrators nothing, and a session *creator* gains no new rights over other players' data.

### Owner attribution
Writes made on a player's behalf carry the **gang owner's** `user_id`, following the existing arbitrator-equipment-purchase pattern. The one gap found was `updateGang`'s resource-change logs, which were attributed to the caller — now attributed to the gang owner (also fixes arbitrator edits from the gang page).

### Changes
- `app/actions/battle-sessions.ts` — `canManageSession`/`isSessionArbitrator` helpers; creator-only checks accept arbitrators; `verifySessionParticipant` falls back to arbitrator rights
- `app/lib/user-permissions.ts` — new `PermissionService.isCampaignArbitrator()` shared by actions and page
- `app/actions/update-gang.ts` — log attribution fix
- `app/gang/[id]/battle-session/[sessionId]/page.tsx` — computes `isArbitrator` server-side
- `components/battle-session/active-session.tsx` / `participant-card.tsx` — arbitrator gating; `ParticipantCard` prop renamed `isOwner` → `canManage`; cancel-redirect fallback for arbitrators without a gang in the session
- `components/battle-session/create-battle-modal.tsx` — `gangId`/`gangName` optional so arbitrators can add players; opponent list excludes gangs already in the session

### Review & verification
- Pre-push code review ran 7 analysis angles; confirmed cleanups were applied (shared permission helper, prop rename). Notably verified: the `gang_logs` INSERT policy only constrains `gang_id` (arbitrators qualify via campaign membership), so owner-attributed log rows pass RLS.
- `npx tsc --noEmit` passes.

https://claude.ai/code/session_01Af9JaAhWR98Hp2VXzjeMmH

---
_Generated by [Claude Code](https://claude.ai/code/session_01Af9JaAhWR98Hp2VXzjeMmH)_

### DIFF
--- a/app/actions/battle-sessions.ts
+++ b/app/actions/battle-sessions.ts
@@ -15,6 +15,7 @@ import type {
   SessionRecord,
 } from '@/types/battle-session';
 import { getBattleSessionCached } from '@/app/lib/battle-sessions/get-battle-session-data';
+import { PermissionService } from '@/app/lib/user-permissions';
 
 // =============================================================================
 // Data Fetching
@@ -29,21 +30,61 @@ export async function fetchBattleSession(sessionId: string): Promise<BattleSessi
 // Authorization Helpers
 // =============================================================================
 
-async function verifySessionCreator(
+// Site admins and campaign OWNERs/ARBITRATORs can act on any session in their
+// campaign. Mirrors the RLS policies on the battle session tables
+// (private.is_admin() OR private.is_arb(campaign_id)) and the gang page
+// permission model.
+async function isSessionArbitrator(
+  userId: string,
+  campaignId: string | null
+): Promise<boolean> {
+  const permissionService = new PermissionService();
+  const profile = await permissionService.getUserProfile(userId);
+  if (profile?.user_role === 'admin') return true;
+  if (!campaignId) return false;
+  const role = await permissionService.getCampaignRole(userId, campaignId);
+  return role === 'OWNER' || role === 'ARBITRATOR';
+}
+
+async function canManageSession(
+  userId: string,
+  session: { created_by: string; campaign_id: string | null }
+): Promise<boolean> {
+  if (session.created_by === userId) return true;
+  return isSessionArbitrator(userId, session.campaign_id);
+}
+
+async function verifySessionManager(
   supabase: Awaited<ReturnType<typeof createClient>>,
   sessionId: string,
   userId: string
 ): Promise<{ authorized: boolean; error?: string }> {
   const { data: session } = await supabase
     .from('battle_sessions')
-    .select('created_by')
+    .select('created_by, campaign_id')
     .eq('id', sessionId)
     .single();
 
   if (!session) return { authorized: false, error: 'Session not found' };
-  if (session.created_by !== userId)
-    return { authorized: false, error: 'Only the session creator can perform this action' };
+  if (!(await canManageSession(userId, session)))
+    return { authorized: false, error: 'Only the session creator or an arbitrator can perform this action' };
   return { authorized: true };
+}
+
+// Arbitrators (not the session creator) may act on another player's
+// participant slot — same rights they have over that player's gang page.
+async function isArbitratorForSession(
+  supabase: Awaited<ReturnType<typeof createClient>>,
+  sessionId: string,
+  userId: string
+): Promise<boolean> {
+  const { data: session } = await supabase
+    .from('battle_sessions')
+    .select('campaign_id')
+    .eq('id', sessionId)
+    .single();
+  if (!session) return false;
+  return isSessionArbitrator(userId, session.campaign_id);
 }
 
 async function verifySessionParticipant(
@@ -55,13 +96,14 @@ async function verifySessionParticipant(
   if (participantId) {
     const { data: participant } = await supabase
       .from('battle_session_participants')
-      .select('id')
+      .select('id, user_id')
       .eq('id', participantId)
       .eq('battle_session_id', sessionId)
-      .eq('user_id', userId)
       .maybeSingle();
 
     if (!participant)
+      return { authorized: false, error: 'You are not a participant in this session' };
+    if (participant.user_id !== userId && !(await isArbitratorForSession(supabase, sessionId, userId)))
       return { authorized: false, error: 'You are not a participant in this session' };
     return { authorized: true, participantId: participant.id };
   }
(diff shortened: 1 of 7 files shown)


joeseos (commented):
## PR Review — `feature/added-arb-rights-to-battle-session`

Hey @joeseos — good feature, clean structure, the authorization model is consistent with the stated design. Two real issues worth addressing before merge and a few lower-priority notes.

---

### Findings

---

#### 🔴 MEDIUM — `verifySessionParticipant` arbitrator fallback returns no `participantId`

**File:** `app/actions/battle-sessions.ts` (new lines ~111–116)

When called **without** a `participantId` and the caller has no participant slot in the session, the arbitrator path short-circuits to:

```ts
if (await isArbitratorForSession(supabase, sessionId, userId))
  return { authorized: true };   // participantId is undefined
```

Any action that calls `verifySessionParticipant` without a `participantId` and then writes using `auth.participantId` will silently get `undefined` and either no-op or hit an unexpected DB path. The return type already marks `participantId?: string`, but there's no enforcement that callers check it.

In the actions present today this is unlikely to bite (most arbitrator edits supply an explicit `participantId`), but it's a footgun in a shared auth helper. Safer options:
- Fail loudly: return `{ authorized: false, error: 'Arbitrators must specify a participantId' }` in this path, since a session-level action has no reason to call `verifySessionParticipant` without an ID.
- Or split the helper into `verifySessionParticipant(supabase, sessionId, userId, participantId: string)` (always required) and a separate `verifyIsInSession` for the no-ID path.

---

#### 🔴 MEDIUM — `isCampaignArbitrator` is not cached

**File:** `app/lib/user-permissions.ts` (new lines ~122–134)

`isCampaignArbitrator` calls `getUserProfile` + `getCampaignRole` — two raw DB round trips — and is not wrapped in `unstable_cache` the way `getGangPermissions` is.

In a single action like `updateParticipantRole` the call graph is:

1. Fetch session (`created_by, campaign_id, status`) — 1 query
2. Fetch participant — 1 query
3. `canManageSession` → `isSessionArbitrator` → `isCampaignArbitrator` → **2 more queries**

Any action that calls both `verifySessionManager` AND `verifySessionParticipant` on the same request will hit `isCampaignArbitrator` (and therefore `getUserProfile` + `getCampaignRole`) **twice** — 4 extra queries on top of the real work. The fix mirrors `getGangPermissions`: wrap in `unstable_cache` keyed on `(userId, campaignId)` with the campaign-membership cache tag. Alternatively, compute it once at the top of each action and thread it down.

---

#### 🟡 LOW — `isArbitratorForSession` fetches `campaign_id` redundantly

**File:** `app/actions/battle-sessions.ts`

`isArbitratorForSession` always issues a SELECT to get `campaign_id` from `battle_sessions`. Most callers invoke it from actions that already fetched the full session row, resulting in two fetches of the same row in the same request. `canManageSession` avoids this correctly by accepting a session object. `isArbitratorForSession` could do the same — accept `campaignId: string | null` directly — so callers with the row already in hand can pass it through without the extra round trip.

---

#### 🟡 LOW — `toggleParticipantReady` second SELECT has no `battle_session_id` scope

**File:** `app/actions/battle-sessions.ts` (diff ~line 491)

Removing `.eq('user_id', user.id)` is correct — `verifySessionParticipant` at the top of the function already handles authorization including the arbitrator path. ✅

But the second SELECT that fetches the `ready` state only filters by `id`:

```ts
.from('battle_session_participants')
.select('id, ready')
.eq('id', participantId)   // no .eq('battle_session_id', sessionId)
.single()
```

A `participantId` from a different session would be found and its `ready` state toggled. RLS and the top-of-function check limit the practical exposure, but adding `.eq('battle_session_id', sessionId)` here is a one-liner and makes the scoping explicit.

---

#### 🟡 LOW — `update-gang.ts` log attribution uses `gang.user_id` without null guard

**File:** `app/actions/update-gang.ts`

The fix (attributing logs to `gang.user_id` instead of `user.id`) is correct in intent. `gang` is non-null by the guard above it, but `gang.user_id` itself isn't checked before being passed to `logGangResourceChanges`. If the schema allows a nullable owner (or ever does in future), the log INSERT would get a null `user_id`. Worth a quick `if (gang.user_id)` guard around the log call, or a schema-level confirmation that `gangs.user_id` is NOT NULL.

---

### Things that look good ✅

- `verifySessionParticipant` **with** `participantId` — correctly replaces `.eq('user_id', user.id)` with the explicit arbitrator check. The authorization model is sound.
- `ParticipantCard` prop rename `isOwner` → `canManage` + `isArbitrator` cleanly separates "can manage session" (creator or arb) from "can edit participant data" (arb only). The right abstraction.
- `existingGangIds` filter in `create-battle-modal.tsx` correctly deduplicates both campaign and user-search gang lists against already-added participants.
- Cancel redirect fallback for arbitrators without a gang (`/campaigns/:id` → `/`) is handled gracefully.
- `canManageSession` short-circuits on `created_by === userId` before touching the DB — the expensive path only runs for non-creators.
- `tsc --noEmit` passes; no dangling `user` references after the attribution fix in `update-gang.ts`.

---

The two MEDIUM issues are worth addressing before merge:
1. The no-`participantId` arbitrator fallback in `verifySessionParticipant` is a silent footgun in a shared helper.
2. `isCampaignArbitrator` needs caching — the uncached pattern compounds on every participant interaction and will be noticeable at load.

The LOW items can be a quick follow-up, but the `battle_session_id` scope in `toggleParticipantReady` is a one-liner worth doing now.

---

PR #1755, opened 2026-06-11
Removed fighter_equipment_tradingpost from equipment RPC

joeseos:
First step to drop the old table fighter_equipment_tradingpost

### DIFF
--- a/components/equipment/equipment.tsx
+++ b/components/equipment/equipment.tsx
@@ -272,10 +272,8 @@ const ItemModal: React.FC<ItemModalProps> = ({
         // In Trading Post mode with fighter type, we want both trading post AND fighter's list items
         if (resolvedTypeId && !isVehicleEquipment && !isCustomFighter) {
           // Pass both filters - SQL will use OR logic to return items in EITHER trading post OR fighter's list
-          // fighters_tradingpost_only ensures only fighter-specific trading post items are shown (not gang-level)
           requestBody.equipment_tradingpost = true;
           requestBody.fighter_type_equipment = true;
-          requestBody.fighters_tradingpost_only = true;
         } else {
           // For vehicle/custom/gang-level, use standard trading post filter
           requestBody.equipment_tradingpost = true;
--- a/supabase/functions/get_equipment_detailed_data.sql
+++ b/supabase/functions/get_equipment_detailed_data.sql
@@ -5,6 +5,7 @@ DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boole
 DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,boolean);
 DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,boolean,uuid[]);
 DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,boolean,uuid[],uuid[]);
+DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,uuid[],uuid[]);
 
 CREATE OR REPLACE FUNCTION get_equipment_detailed_data(
     gang_type_id uuid DEFAULT NULL,          -- $1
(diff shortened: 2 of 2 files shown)


joeseos (commented):
## PR Review — `refac/removed-fighter_equipment_tradingpost-from-rpc`

Reviewed with the stated context in mind: this is step 1 of N — stop reading from `fighter_equipment_tradingpost` in the RPC, so the table can be dropped later. The diff is clean and the mechanics are correct. Two things need to be pinned down before deploying to production.

---

### Findings

---

#### 🔴 HIGH — Functional regression risk: items only in `fighter_equipment_tradingpost` will vanish

The old `tp_access` CTE had a third UNION branch that fed equipment IDs from `fighter_equipment_tradingpost` into trading-post access:

```sql
-- Fighter-specific trading post access (no TP name — these aren't real TPs)
SELECT ff.equipment_id, NULL::text
FROM filtered_fet ff ...
```

That branch is gone. Any equipment item whose **only** path to `tp.has_access = true` was through `fighter_equipment_tradingpost` — i.e., it has no row in `trading_post_equipment`, no campaign TP, and no custom TP — will now return `has_access = false` and be filtered out of trading-post queries.

The admin route at `app/api/admin/fighter-types/route.ts:192` still reads and writes `fighter_equipment_tradingpost`, confirming live data exists in that table.

**Before deploying this PR:** confirm (or document) that every equipment ID stored in `fighter_equipment_tradingpost` is also reachable via `trading_post_equipment` or a campaign/custom TP. A quick check:

```sql
SELECT DISTINCT equip_id::uuid
FROM fighter_equipment_tradingpost fet
CROSS JOIN jsonb_array_elements_text(fet.equipment_tradingpost) AS equip_id
WHERE equip_id::uuid NOT IN (SELECT equipment_id FROM trading_post_equipment)
  AND equip_id::uuid NOT IN (SELECT equipment_id FROM custom_trading_post_equipment WHERE equipment_id IS NOT NULL);
```

If that returns rows, those items need to be migrated first or they silently disappear from fighters' shopping lists.

---

#### 🔴 MEDIUM — Trading-post scope widening for fighters: intentional?

The old `$9 = true` (`fighters_tradingpost_only`) path narrowed the TP check to `fighter_tp_set` — items specifically listed in `fighter_equipment_tradingpost`. Regular gang TP items were **excluded** from a fighter's trading-post view under that flag.

After this PR, when `fighters-tradingpost` mode fires for a named fighter type, the TP check becomes `COALESCE(tp.has_access, false)`, which includes the gang's full trading post (`trading_post_equipment` matching the gang type), all campaign-authorised TPs, and custom TPs. That is a **superset** of what was shown before.

Concretely: a fighter whose gang type has a trading post full of general equipment will now see all of it under "fighters-tradingpost" mode, not just items that were explicitly assigned to that fighter type.

If this widening is intentional (i.e., the fighter-specific TP table was a historical workaround and fighters were always supposed to see the full gang TP), great — just confirm and document it in the PR. If it's a side-effect, the scope needs to be narrowed in a follow-up before the old data is migrated out.

---

#### 🟡 LOW — `schema.public.sql` still has the old 11-parameter signature

`supabase/schema/schema.public.sql:1343` still declares the old `(uuid, text, uuid, boolean, boolean, uuid, uuid, uuid, boolean, uuid[], uuid[])` function. This file appears to be a schema snapshot rather than a migration source, so it may auto-regenerate. But if it's manually maintained or used as a reference for deployments, it will go stale after this migration runs. Worth regenerating it or noting it needs a follow-up update.

---

#### 🟡 LOW — Dead `if`-branch in `equipment.tsx` after cleanup

After removing `fighters_tradingpost_only`, the `if (resolvedTypeId && !isVehicleEquipment && !isCustomFighter)` branch inside `fighters-tradingpost` mode now differs from the `else` branch only in that it also sends `fighter_type_equipment: true`:

```ts
// if branch:  equipment_tradingpost + fighter_type_equipment
// else branch: equipment_tradingpost only
```

This difference is real and intentional (fighter-type shows TP + fighter list, vehicle/custom/gang shows TP only). But the comment that used to justify it (`fighters_tradingpost_only ensures...`) has been removed. The remaining comment (`Pass both filters - SQL will use OR logic...`) still explains it correctly. No bug, but worth a quick read to confirm the branching logic is still desired in its current form.

---

### Things that look good ✅

- The DROP list correctly adds the new 10-parameter signature (`uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,uuid[],uuid[]`) for idempotent re-runs, and retains the old 11-param signature for rollback safety.
- Parameter renumbering ($10→$9, $11→$10) is consistent across all 6 uses in the function body.
- `gang_tp` CTE, `custom_tp_override`, `custom_trading_post` sub-select in the UNION ALL, and the GRANT/REVOKE block are all updated to the new parameter numbers. No missed renumber.
- The `AND NOT ($4 IS NULL AND $5 IS NOT NULL AND $5 = true AND $9 IS NOT NULL AND $9 = true)` filter on the custom-equipment branch is correctly removed — it was specifically guarding against the old `fighters_tradingpost_only` path; without that param it was always a no-op.
- Only one TypeScript call site exists (`equipment.tsx`) and both changes there (`requestBody` cleanup + comment removal) are correct.
- The admin routes (`app/api/admin/fighter-types/route.ts`) that write to `fighter_equipment_tradingpost` are intentionally left alone — correct for step 1.

---

The SQL mechanics are correct and the refactor is a clean removal. The blocking concern is the data-gap risk (HIGH): if any `fighter_equipment_tradingpost` entries aren't covered by `trading_post_equipment`, those items disappear silently in production the moment this migration runs. Verify with the query above (or add it as a pre-deploy check) before merging. The scope-widening question (MEDIUM) should also be explicitly confirmed as intentional.

---

PR #1756, opened 2026-06-11
Removed the table fighter_equipment_tradingpost

joeseos:


joeseos (commented):
## PR Review — `refac/removed-fighter-equipment-tradingpost-table`

This is the step-2 companion to #1755. It directly addresses the HIGH finding from that review (data gap risk), cleans up the admin UI, and drops the table. The mechanics are mostly correct — a few things to pin down.

---

### Findings

---

#### 🔴 HIGH — Merge ordering: this PR must land after #1755

The migration drops `fighter_equipment_tradingpost`. If this PR is merged **before** #1755, the current production RPC still contains `filtered_fet` and `fighter_tp_set` CTEs that `FROM fighter_equipment_tradingpost`. Dropping the table first breaks every equipment query for every fighter in the app until #1755 is also deployed.

The migration has no guard against this (a runtime check isn't really possible), so the ordering needs to be enforced at the merge level. Worth adding a note in the PR description or a migration comment: *"Requires #1755 (RPC update) to be deployed first."*

---

#### 🔴 MEDIUM — Semantic destination: `fighter_type_equipment`, not `trading_post_equipment`

The migration moves rows into `fighter_type_equipment` — the fighter's personal house list — not `trading_post_equipment`.

**Old behaviour:**
- `fighters-list` mode queried `fighter_type_equipment` only. Items from `fighter_equipment_tradingpost` did **not** appear here.
- `fighters-tradingpost` mode used OR logic: fighter's list **or** `fighter_equipment_tradingpost`. Both sources appeared.

**New behaviour after this migration:**
- `fighters-list` mode now includes the migrated items (they're in `fighter_type_equipment`). They **weren't visible here before**.
- `fighters-tradingpost` mode still shows them (via the fighter's list arm of the OR).

In Necromunda terms: items that were "buy from the trading post" (TP-only, not on the house list) now also appear as permanent house-list equipment. If these items were intentionally TP-only — things a fighter *can* buy but that aren't on their standing list — this changes how they're displayed and priced everywhere equipment is shown.

If the intent is "these items should now be treated as house-list equipment" that's a valid product decision, but it's a meaningful semantic shift that's worth an explicit confirmation in the PR description.

---

#### 🟡 LOW — Migration atomicity: no explicit transaction

```sql
INSERT INTO fighter_type_equipment ...;   -- step 1
DROP TABLE fighter_equipment_tradingpost; -- step 2
```

Supabase's migration runner wraps each file in a transaction, so in practice this is atomic. But if for any reason the migration runs outside a transaction wrapper and `DROP TABLE` fails after the `INSERT` succeeds, the data is now in both tables. Re-running is safe due to the `NOT EXISTS` guard, but the table won't be dropped without manual intervention. Low risk, but an explicit `BEGIN; ... COMMIT;` would make the intent self-documenting.

---

#### 🟡 LOW — `supabase/schema/schema.public.sql` goes stale

The schema dump still declares `fighter_equipment_tradingpost` with its table definition, indexes, FK, and RLS policies. After the migration runs, that file is out of sync. Should be regenerated as a follow-up (or as part of this PR if schema dumps are kept current).

---

#### 🟡 LOW — Migration doesn't validate UUID cast

```sql
SELECT fet.fighter_type_id, (elem)::uuid
```

If any element in the `equipment_tradingpost` JSONB array is not a valid UUID string, the cast throws and the entire migration fails. Given the data was already cast the same way in the old RPC (`equip_id::uuid`), any invalid values would have surfaced already — so this is unlikely to matter. Still worth a quick `SELECT` against the table before deploying to production to confirm no bad data:

```sql
SELECT count(*) FROM fighter_equipment_tradingpost fet
CROSS JOIN jsonb_array_elements_text(fet.equipment_tradingpost) AS elem
WHERE (elem ~ '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$') = false;
```

---

### Things that look good ✅

- **Addresses the #1755 HIGH finding.** The data gap risk is closed: every equipment ID in `fighter_equipment_tradingpost` is migrated to `fighter_type_equipment` before the table is dropped, so nothing silently vanishes.
- **`NOT EXISTS` duplicate guard** in the migration is correct and makes re-runs safe.
- **NULL JSONB safety.** `jsonb_array_elements_text(NULL)` returns zero rows, so fighter types with no trading post data are handled cleanly.
- **Migrated rows use NULL for `gang_type_id` / `gang_origin_id`.** The RPC's JOIN conditions treat NULL as "applies to all," so this correctly broadens rather than restricts access — consistent with the original intent of `fighter_equipment_tradingpost` (which had no gang-type scoping).
- **Admin UI removal is complete:** GET (both fetch paths), PUT, and POST in `route.ts` are all cleaned up. `admin-fighter-trading-post-access.tsx` deleted, its import removed from `admin-edit-fighter-type.tsx`.
- **No orphaned imports in `admin-create-fighter-type.tsx`.** `Checkbox` is still used at five sites in that file (equipment list selection). `EquipmentWithId` is still used by the equipment `useQuery` and the list rendering at line 915. The removed trading post state variables (`equipmentByCategory`, `expandedCategory`, etc.) are the only things deleted — confirmed clean.
- **`fetchEquipmentByCategory` removal** is safe: it was a manual-fetch duplicate of the `useQuery` at line 161. The query already handles equipment loading for the remaining UI.

---

The merge ordering (HIGH) is the blocker — document and enforce it. The semantic destination question (MEDIUM) needs a confirmation that moving these items onto the fighter's house list is the intended product behaviour, not just a convenient migration target. Everything else is low-risk cleanup noise.


---

PR #1760, opened 2026-06-12
Added logs to the campaign page and refactored the log component

joeseos:


joeseos (commented):
## PR Review — `feature/campaign-logs`

Good refactor — collapsing three duplicated log components into one generic modal is the right call, and the campaign logs feature is a clean addition on top. Two medium issues worth fixing before merge.

---

### Findings

---

#### 🔴 MEDIUM — Both `gangId` and `fetchUrl` are optional with no guard

**File:** `components/log-modal.tsx`

```ts
const url = fetchUrl || `/api/gangs/${gangId}/logs`;
```

If a caller provides neither, `gangId` is `undefined` and the constructed URL becomes `/api/gangs/undefined/logs`. TypeScript accepts this silently because both props are typed as optional. Current callers are all correct, but the interface is a footgun for future use.

Fix: enforce that at least one is provided via a discriminated union:

```ts
interface GangLogsProps {
  isOpen: boolean;
  onClose: () => void;
  title?: string;
  emptyMessage?: string;
}
type LogModalProps =
  | (GangLogsProps & { fetchUrl: string; gangId?: never })
  | (GangLogsProps & { gangId: string; fetchUrl?: never });
```

Or at minimum add a runtime guard:

```ts
const url = fetchUrl ?? (gangId ? `/api/gangs/${gangId}/logs` : null);
if (!url) return null; // or throw
```

---

#### 🔴 MEDIUM — `useQuery` cache causes stale data flash on modal reopen

**File:** `components/log-modal.tsx`

The old implementation called `setLogs([])` when the modal closed, which prevented stale data from showing when it was reopened. The new `useQuery` approach with `enabled: isOpen` disables the query on close but keeps the response in the React Query cache (default `gcTime`: 5 minutes). When the modal reopens within that window, the old data renders immediately before the refetch completes.

Fix: set `gcTime: 0` so the cache entry is discarded as soon as the query becomes inactive:

```ts
const { data: logs = [], isLoading } = useQuery<GangLog[]>({
  queryKey: ['logs', url],
  queryFn: async () => { ... },
  enabled: isOpen,
  gcTime: 0,
});
```

This matches the old clear-on-close behaviour without needing a manual `useEffect`.

---

#### 🟡 LOW — Fighter log `helper` text loses vehicle context

**File:** `components/log-modal.tsx`

The old `FighterLogs` component rendered a context-aware helper: *"Track all changes made to this fighter and their vehicle"* when `vehicleId` was set. The new generic modal hardcodes *"Track all changes and activity"* — less informative for the fighter/crew case. The `helper` text could be added as an optional prop, or the fighter-specific context could be passed via a `helper` prop on `LogModal`.

---

#### 🟡 LOW — Dead `actions` variable and unnecessary `ListAction` import

**File:** `components/log-modal.tsx`

```ts
import { ListColumn, ListAction } from '@/components/ui/list';
...
const actions: ListAction[] = [];
```

`actions` is defined but never passed to anything. `ListAction` is only imported to type an empty array that goes nowhere. Both can be removed.

---

#### 🟡 LOW — Stale JSDoc in `log-types.ts`

```ts
/**
 * Shared action type → display label map for gang and fighter log UIs.
 * Single source of truth — imported by both gang-logs.tsx and fighter-logs.tsx.
 */
```

Both `gang-logs.tsx` and `fighter-logs.tsx` are deleted by this PR. The comment should be updated to reflect the new consumers.

---

#### 🟡 LOW — Campaign logs API doesn't filter `campaign_gangs` by status

**File:** `app/api/campaigns/[campaignId]/logs/route.ts`

```ts
const { data: campaignGangs } = await supabase
  .from('campaign_gangs')
  .select('gang_id')
  .eq('campaign_id', params.campaignId);
```

This fetches gang IDs for all rows in `campaign_gangs`, including any with pending or removed status. Logs for gangs that were rejected or have left the campaign could surface in the results. Adding `.eq('status', 'ACCEPTED')` would tighten this up — the existing gang logs API has the same pattern so it's not a regression, but worth fixing here since this is new code.

---

### Things that look good ✅

- **React Query migration is correct.** `QueryClientProvider` is app-wide in `app/layout.tsx`, so `useQuery` works in all three call sites without additional setup.
- **`queryKey: ['logs', url]`** uses the full URL (including query params), so gang, fighter, and campaign log caches are distinct. No cross-contamination.
- **Campaign logs API authorization** is solid: membership check first, admin fallback second, `CAMPAIGN_ACTION_TYPES` filter prevents leaking gang-internal events to campaign members.
- **`CAMPAIGN_ACTION_TYPES`** covers the right set of events and the `as const` + spread `[...CAMPAIGN_ACTION_TYPES]` pattern for the `.in()` call is correct.
- **Import cleanup**: `List` correctly removed from the `@/components/ui/list` import since the table is rendered inline, not via the `List` component.
- **Pagination duplication eliminated.** Deleting `fighter-logs.tsx` (255 lines) and generalising `gang-logs.tsx` removes the last copy of the duplicated pagination+table block that was flagged in the #1752 review.
- **`gang.tsx` import path** `'../log-modal'` is the correct relative path from `components/gang/` to `components/log-modal.tsx`. ✓
- **`enabled: isOpen`** correctly prevents the fetch firing when the modal is closed (no background polling).

---

Fix the two medium issues before merge: the missing guard on the `gangId`/`fetchUrl` props, and the stale cache flash. Both are small changes. The low items are tidy-ups that can be bundled into the same commit.


---

joeseos:
## Code Review — PR #1760 (round 2)

@joeseos the "Fixes" commit (`477da89`) addresses the two medium blockers and the valid low items from round 1. The acknowledged design decisions (stale cache, helper text, campaign_gangs status filter) are noted and not re-raised. Three new low-severity findings below.

---

| # | Finding | Status |
|---|---------|--------|
| 1 | 🔴 `gangId`/`fetchUrl` both optional → `/api/gangs/undefined/logs` | ✅ Resolved — `LogModalProps.fetchUrl: string` is now required; `gangId` prop dropped entirely |
| 2 | 🔴 `useQuery` cache → stale data flash on reopen | ⚠️ Acknowledged — cached data + background refetch is intentional UX |
| 3 | 🟡 Fighter helper text loses vehicle context | ⚠️ Acknowledged — generic text works across all three contexts |
| 4 | 🟡 Dead `ListAction` import + empty `actions` array | ✅ Resolved — both removed |
| 5 | 🟡 JSDoc references deleted files | ✅ Resolved — updated to "Shared action type → display label map for log UIs." |
| 6 | 🟡 `campaign_gangs` not filtered by status | ⚠️ Acknowledged — consistent with existing APIs, deferred |

---

### New findings

#### 🔵 LOW — `gangId` passed unguarded into URL template in `fighter-details-card.tsx`

```tsx
// New
fetchUrl={`/api/gangs/${gangId}/logs?fighterId=${id}...`}

// Old — had an explicit guard
gangId={gangId || ''}
```

The old `FighterLogs` call used `gangId || ''`, strongly implying `gangId` can be `undefined` at the call site. Without the guard the URL becomes `/api/gangs/undefined/logs?...` — the exact pattern flagged in round 1. In practice the modal only opens on user interaction so a missing `gangId` is unlikely, but the defensive guard was there for a reason. One-line fix:

```tsx
fetchUrl={`/api/gangs/${gangId ?? ''}/logs?fighterId=${id}...`}
```

---

#### 🔵 LOW — Exported function name `GangLogs` is stale after the rename

**`components/log-modal.tsx`**

```ts
export default function GangLogs({ ... }: LogModalProps) {
```

The file is `log-modal.tsx`, the props interface is `LogModalProps`, and the component is now used for gangs, fighters, **and** campaigns — but the exported function is still called `GangLogs`. Default exports don't enforce the name at import sites (and the three callers already use three different aliases: `GangLogs`, `GangLogs`, `LogModal`), so this is cosmetic-only. Renaming to `LogModal` would complete the refactor:

```ts
export default function LogModal({ ... }: LogModalProps) {
```

---

#### 🔵 LOW — Client-side `sortLogs` is redundant

**`components/log-modal.tsx`**

```ts
const sortLogs = (a: GangLog, b: GangLog) => {
  return new Date(b.created_at).getTime() - new Date(a.created_at).getTime();
};
const sortedLogs = [...logs].sort(sortLogs);
```

All three API endpoints (`/api/gangs/[gangId]/logs`, `/api/campaigns/[campaignId]/logs`) already return rows sorted by `created_at` descending. The client-side sort is a no-op on correctly ordered data and adds an unnecessary `[...logs]` copy on every render. Can be simplified to:

```ts
const sortedLogs = logs; // API guarantees descending order
```

---

### Summary

| 1 | `fighter-details-card.tsx` | 🔵 Low | `gangId` unguarded in URL template — regression of the `\|\| ''` guard from the old code |
| 2 | `log-modal.tsx` | 🔵 Low | Exported function still named `GangLogs`; should be `LogModal` to match file and props |
| 3 | `log-modal.tsx` | 🔵 Low | Redundant client-side sort — all APIs already return `created_at DESC` |

**Verdict: 🟢 Approve** — both medium blockers from round 1 are gone. The three new findings are all cosmetic/low; they can be bundled into a follow-up or fixed here at your discretion. This is ready to merge.

---

PR #1762, opened 2026-06-12
Added support for custom ammo

joeseos:


joeseos (commented):
## PR #1762 Review — `feature/added-support-for-custom-ammo`

Good fix for a real data-integrity bug. The core `weapon_group_id` / `custom_equipment_id` distinction is now correct throughout. Two issues worth addressing, one note on safe removal.

---

### Medium

**`storedTargetWeaponId` initialized once from props, not synced — `custom-weapon-profiles.tsx`**

```ts
const [storedTargetWeaponId, setStoredTargetWeaponId] = useState(
  profiles[0]?.weapon_group_id
);
```

`useState` only reads the initializer on the first mount. If the parent re-renders with a refreshed `profiles` array (e.g. after a save round-trips back through the server), `storedTargetWeaponId` remains stale and the target-weapon dropdown will show the old selection. A `useEffect` sync fixes it:

```ts
useEffect(() => {
  setStoredTargetWeaponId(profiles[0]?.weapon_group_id ?? undefined);
}, [profiles]);
```

---

**Unconditional equipment fetch fires for non-ammunition items — `custom-equipment.tsx`**

`availableWeapons` is fetched via `/api/equipment` (all equipment) on every render of `CustomEquipment`, regardless of whether the item being edited is ammunition. The `showTargetWeapon` flag is already computed — gate the fetch behind it:

```ts
useEffect(() => {
  if (!showTargetWeapon) return;
  // fetch availableWeapons
}, [showTargetWeapon]);
```

Without this, every non-ammo equipment edit triggers an unnecessary full-catalogue network request.

---

### Notes (no action needed)

**DELETE fix is correct and the fallback removal is safe.**
The old DELETE used `.eq('weapon_group_id', equipmentId)` which, for ammo profiles (where `weapon_group_id` = target weapon ≠ ammo's own ID), would have silently deleted the wrong profiles. The fix to `.eq('custom_equipment_id', equipmentId)` is right.

Removing the two-step fallback from `getCustomWeaponProfiles` is safe: ammo profiles created with the old buggy INSERT (where both columns were set to `equipmentId`) are still found by `custom_equipment_id`, so no data migration is needed.

**INSERT fallback `weapon_group_id: profile.weapon_group_id || equipmentId`** correctly covers both cases — ammo profiles carry an explicit target, non-ammo profiles self-reference.

---

Overall this is a solid correctness fix. The stale-state and unconditional-fetch issues are the only things worth a follow-up before shipping.

---

joeseos:
## Code Review — PR #1762 (round 2)

@joeseos the latest commit (`c7bb6fc`) addresses both medium findings from round 1. One new trivial finding below.

---

| # | Finding | Status |
|---|---------|--------|
| 1 | 🔴 `storedTargetWeaponId` initialized once from props, not synced on re-render | ✅ Resolved — `useEffect` sync added (`[profiles]` dependency), exactly as suggested |
| 2 | 🔴 Unconditional equipment fetch fires for non-ammunition items | ✅ Resolved — `needsAvailableWeapons` gate correctly scopes the `useQuery` to ammo-category modals only |

---

### New finding

#### 🔵 LOW — `useMemo` imported but not used — `custom-equipment.tsx`

```ts
import React, { useState, useMemo } from 'react';
```

`useMemo` appears in the import but is never called in the component body — `needsAvailableWeapons` is a plain expression. Dead import:

```ts
import React, { useState } from 'react';
```

---

### Things that look good ✅

- **`needsAvailableWeapons` logic** correctly handles both create (`createModalOpen && category === 'ammunition'`) and edit (`editModalData !== null && category === 'ammunition'`) paths without false positives.
- **`queryKey: ['availableWeapons']`** is stable enough here — `staleTime` defaults to 0 so React Query refetches on every `enabled: true` transition, meaning a newly-created weapon will appear the next time the ammo modal opens.
- **Copy path fix** (`weapon_group_id: profile.weapon_group_id === copyModalData.id ? createdEquipment.id : profile.weapon_group_id`) correctly preserves the target-weapon reference for post-fix ammo profiles and passably handles the legacy self-referential case.
- **Cross-ammo lookups** in `gang-data.ts` and `fighter-data.ts` are symmetric and internally consistent: standard weapon → custom ammo uses `customAmmoByParentWeapon.get(equipment_id)` with a `custom_equipment_id` ownership filter; custom weapon → standard ammo uses the same pattern in reverse. The standard-weapon + custom-ammo case (the main use case) is the one that actually benefits from the `weapon_group_id` fix and will resolve correctly post-save.
- **`types/equipment.ts`** updated to include `weapon_group_id` on `CustomWeaponProfileData`. ✅

---

### Summary

| 1 | `custom-equipment.tsx` | 🔵 Low | `useMemo` imported but never used |

**Verdict: 🟢 Approve** — both medium blockers resolved, new finding is a one-word import cleanup. Ready to merge.

---

PR #1767, opened 2026-06-13
Add custom collections (bundles of custom items)

joeseos:
## What

Introduces **collections** — an arbitrator's bundle of custom items (gang types, fighter types, equipment, skills, trading posts). A collection supports two complementary actions:

- **Share (apply) a collection to a campaign** *(primary)* — expands the collection's items into per-item `custom_shared` rows (tagged with `custom_collection_id`), cascading gang types → fighters → skills (mirroring `shareCustomGangType`) and syncing `campaigns.custom_trading_posts`. Because rows are written at the item grain, **every existing read path lights up with no DB-function changes** (`get_available_skills`, trading-post visibility). Share targets are limited to campaigns the user arbitrates (the share modal's `userCampaigns`).
- **Copy a collection into your account** *(secondary)* — an atomic deep-clone via the `copy_custom_collection()` plpgsql RPC. Computes the full transitive dependency closure so the copy is self-contained, remaps all internal custom FKs, and keeps global FKs as-is.

## Data model

- **One new table** `custom_collections` with membership stored as a jsonb `items` array of `{type, id}` entries — mirroring the existing `campaigns.custom_trading_posts` jsonb convention (no join table). Indexed on `user_id`.
- **One new column** `custom_shared.custom_collection_id` (`ON DELETE SET NULL`) as a provenance tag so a collection's shares can be cleanly re-shared/unshared without disturbing individually-shared rows.
- `copy_custom_collection(p_collection_id uuid)` — `SECURITY INVOKER` plpgsql function (open-SELECT reads the source, owner-INSERT writes the clones; same access pattern as the existing `copyGang` flow). Implemented with array variables + jsonb id-maps (no temp tables) so it compiles under `check_function_bodies` and runs atomically in one transaction.
- RLS on `custom_collections` follows the verbatim owner-or-admin write / open-read pattern used by the other `custom_*` tables (verified against the live DB).

### Deployment note
The migration (`supabase/migrations/20260613130000_add_custom_collections.sql`) contains **only** the table, index, RLS, grants, and the `custom_shared.custom_collection_id` column. The `copy_custom_collection()` RPC is maintained in **`supabase/functions/copy_custom_collection.sql`** and must be applied via the Supabase dashboard, per this repo's convention (functions are not synced through migrations — see `supabase/README.md`).

## App layer

- `app/actions/customise/custom-collections.ts` — CRUD, `addCollectionItem`/`removeCollectionItem` (jsonb read-modify-write), and `copyCollection` (RPC wrapper).
- `shareCollection(collectionId, campaignIds)` added to `app/actions/customise/custom-share.ts` (cascade + batch dedup + trading-post jsonb sync, mirroring the existing share actions).
- `app/lib/customise/custom-collections.ts` — `getUserCustomCollections` (resolves item display names, skips dangling ids).
- `components/customise/custom-collections.tsx` — TanStack-Query UI with optimistic create/update/delete and live item add/remove (matching `custom-trading-posts.tsx`); item picker sourced from the custom-asset arrays the tab already holds.
- `ShareCustomCollectionModal` in `components/customise/custom-shared.tsx` — and the 5 share modals there were consolidated into one generic `ShareToCampaignsModal` (net code reduction).
- Wiring: Custom Assets tab (`home-tabs.tsx` / `page.tsx`) — collections render like every other custom asset (owner shares to campaigns there); and the user profile page (`app/user/[id]/page.tsx` + `/api/users/[id]`) as the "Copy to my account" surface for other users.

## Validation

- Migration table + RPC compile-tested against the database inside a rolled-back transaction (nothing persisted).
- `tsc --noEmit`: 0 errors. `next build`: compiled, linted, type-checked, 72/72 pages generated.

## Review fixes applied
- Renamed migration to avoid a timestamp collision with `20260613120000_add_description_to_custom_assets.sql` on main.
- `shareCollection` no longer strips a trading post from `campaigns.custom_trading_posts` when another share still links it.
- Normalized the editor's local `items` to the DB shape (`{type, id}`), keeping the display `name` only on `resolvedItems`.
- Rejected (validated as incorrect): the `delegation_cost` clone suggestion (that column exists only in the TS interface, not the `custom_fighter_types` table), and the name-length guard (no sibling custom-asset action validates name length).

## Notes

- The checked-in `supabase/schema/schema.public.sql` dump is **not** regenerated in this PR — the migration is the source of truth; the dump can be regenerated after merge.
- Functional end-to-end testing of the clone/share flows under a real authenticated session is best done on a Supabase preview branch (the RPC requires `auth.uid()`).

https://claude.ai/code/session_018iyjiPZJGRCfTtF5Hcjdk5


joeseos (commented):
## PR Review — #1767 · `feature/custom-packs`

@joeseos this is a substantial feature with a thoughtful design — the `ShareToCampaignsModal` refactor is a genuine win (removed ~300 lines of duplicated code), the closure-loop approach in `copy_custom_pack` is correct, and the optimistic UI patterns mirror the existing codebase well. A few issues need attention before merge, the most critical being a data loss bug in the SQL clone function.

---

### 🔴 HIGH — `copy_custom_pack` silently drops `delegation_cost` for fighter types

**File:** `supabase/functions/copy_custom_pack.sql` — `INSERT INTO custom_fighter_types`

`CustomFighterType` has a `delegation_cost` column (confirmed in `types/fighter.ts`). The clone INSERT lists every other fighter type column but omits `delegation_cost`, so it silently resets to `NULL` for every cloned fighter. For any pack that includes fighters with a delegation cost, the copy is data-corrupt.

**Fix:** Add `delegation_cost` to the INSERT column list and the SELECT:

```sql
INSERT INTO public.custom_fighter_types (id, created_at, user_id, fighter_type, gang_type, cost, ...,
                                          custom_gang_type_id, description, delegation_cost)
SELECT ..., cft.delegation_cost
FROM public.custom_fighter_types cft WHERE cft.id = ANY(v_ft);
```

---

### 🔴 HIGH — Migration timestamp collision with already-merged `20260613120000_add_description_to_custom_assets.sql`

**File:** `supabase/migrations/20260613120000_add_custom_packs.sql`

This PR shares the timestamp `20260613120000` with `20260613120000_add_description_to_custom_assets.sql`, which is already present on `main` (the file exists at the current main HEAD `8ae21f5b`). Supabase tracks migrations by timestamp; two files at the same timestamp will conflict on apply.

**Fix:** Use a later timestamp, e.g. `20260613130000`.

---

### 🟡 MEDIUM — `sharePack` TP cleanup can incorrectly strip a trading post from `campaigns.custom_trading_posts`

**File:** `app/actions/customise/custom-share.ts` — `sharePack`, the TP sync block

The cleanup identifies campaigns to strip a trading post from by looking only at _pack-tagged_ `custom_shared` rows:

```ts
const { data: oldTpShares } = await supabase
  .from('custom_shared')
  .select('campaign_id')
  .eq('custom_pack_id', packId)      // only pack-tagged rows
  .not('custom_trading_post_id', 'is', null);
```

But `campaigns.custom_trading_posts` is a single jsonb array shared by ALL share sources. If campaign Y has both a pack-tagged share of TP X _and_ an individual share of TP X, and the user re-shares the pack without including Y, the code:

1. Deletes the pack-tagged `custom_shared` row for TP X → campaign Y ✅
2. Does NOT re-insert one (Y is not in `campaignIds`) ✅
3. **Removes TP X from `campaigns.custom_trading_posts` for Y** ❌ — even though the individual share still exists and Y still needs TP X

**Fix:** Before removing a TP id from the jsonb array, verify no other `custom_shared` row still links that TP to that campaign:

```ts
const { data: remainingShares } = await supabase
  .from('custom_shared')
  .select('id')
  .eq('campaign_id', c.id)
  .in('custom_trading_post_id', tpArray)
  .is('custom_pack_id', null);  // non-pack shares
// only remove from jsonb if remainingShares is empty
```

---

### 🟡 MEDIUM — GIN index on `items` described but not created

**File:** `supabase/migrations/20260613120000_add_custom_packs.sql`

The PR description says "GIN index on `items`." The migration only creates `idx_custom_packs_user_id`. Either add the index or correct the description. If intended:

```sql
CREATE INDEX idx_custom_packs_items ON public.custom_packs USING GIN (items);
```

---

### 🟡 MEDIUM — Packs editor inaccessible to users with no campaigns on the home tab

**File:** `components/home-tabs.tsx` line ~105

```tsx
{userCampaigns.length > 0 && (
  <CustomisePacks ... />
)}
```

A user without any campaigns cannot create or manage packs from the home page at all — even though the copy action is available to everyone on the user profile page. Packs are useful to prepare ahead of joining a campaign. Consider rendering the list unconditionally and only gating the Share button.

---

### 🟠 LOW — `applyItemsUpdate` assigns `ResolvedPackItem[]` to `items: PackItem[]`

**File:** `components/customise/custom-packs.tsx` — `applyItemsUpdate`

```ts
setPacks(prev => prev.map(p => (p.id === packId ? { ...p, resolvedItems: items, items } : p)));
```

`items` here is `ResolvedPackItem[]` (with `name`), but `items` on `CustomPackWithItems` should be `PackItem[]` (no `name`). TypeScript doesn't flag this because `ResolvedPackItem` structurally extends `PackItem`, but it means local state silently diverges from the DB shape. Not a runtime bug today, but a future mutation could serialize the extra field.

**Fix:**
```ts
items: items.map(({ type, id }) => ({ type, id }))
```

---

### 🟠 LOW — No pack name length validation

`createCustomPack` and `updateCustomPack` have no max-length guard on `name`, unlike every other custom asset action. Add a consistent check (e.g. `name.length > 100 → error`) to match the pattern.

---

### ✅ What's done well

- **`ShareToCampaignsModal` refactor** — clean elimination of ~300 lines of duplicated modal boilerplate across 4 components; correct `idPrefix` threading avoids duplicate DOM ids
- **`copy_custom_pack` closure loop** — correct fixed-point termination; the array-variable + jsonb id-map approach is the right call for `check_function_bodies` compatibility and avoids temp-table caching pitfalls
- **Optimistic UI** — create/update/delete/add-item/remove-item mutations all roll back on error correctly and share a consistent shape
- **Dedup in `sharePack`** — the `alreadyShared` set correctly avoids inserting duplicate `custom_shared` rows when campaigns already had individual shares
- **`getUserCustomPacks` batch name resolution** — one query per type across all packs, not N+1
- **`ON DELETE SET NULL` on `custom_pack_id`** — correct choice so deleting a pack never silently revokes active shares
- **`SECURITY INVOKER` on the RPC** — right call; the function uses the caller's RLS identity for both reads and writes

---

joeseos:
## PR Re-review — #1767 · `feature/custom-packs` → `feature/custom-collections` (head `b70d2b4`)

@joeseos great turnaround — all previous HIGH and MEDIUM findings are closed out. One new LOW item below, but nothing blocking.

---

| 1 | 🔴 HIGH | `copy_custom_pack` silently drops `delegation_cost` for fighter types | ✅ **Resolved (finding was incorrect)** — `delegation_cost` only exists on `fighter_types` (added by `20260304100000_add_delegation_cost_to_fighter_types.sql`), not on `custom_fighter_types`. Confirmed in the schema snapshot and migrations. Developer's rebuttal is correct. |
| 2 | 🔴 HIGH | Migration timestamp collision with `20260613120000` | ✅ **Resolved** — migration renamed to `20260613130000_add_custom_collections.sql` |
| 3 | 🟡 MEDIUM | `shareCollection` TP cleanup strips TPs still linked by another share | ✅ **Resolved** — `stillLinked` guard correctly queries remaining `custom_shared` rows after the collection-tagged deletes, then only strips from the jsonb array when no other row still links the TP to that campaign |
| 4 | 🟡 MEDIUM | GIN index on `items` described but not created | ✅ **Resolved** — index was intentionally dropped as YAGNI; PR description updated accordingly |
| 5 | 🟡 MEDIUM | Collections editor inaccessible to users with no campaigns | ✅ **Resolved** — campaign gate removed; Share action gates itself to arbitrated campaigns at the modal level |
| 6 | 🟠 LOW | `applyItemsUpdate` assigns `ResolvedPackItem[]` to `items: PackItem[]` | ✅ **Resolved** — `items` is now stripped to `{ type, id }` via `.map(({ type, id }) => ({ type, id }))` |
| 7 | 🟠 LOW | No collection name length validation | ⚪ **Acknowledged** — consistent with the pattern across all other custom-asset create/update actions; not flagging further |

---

### New finding

**🟠 LOW — `copyMutation` missing `onSettled` cache invalidation**

**File:** `components/customise/custom-collections.tsx` — `copyMutation`

Every other mutation in this component (create, update, delete, add-item, remove-item) calls `queryClient.invalidateQueries({ queryKey: ['customCollections'] })` in `onSettled`. The `copyMutation` has no `onSettled` handler, so after a successful copy the newly-cloned collection doesn't appear in the user's own list until a page reload.

```ts
const copyMutation = useMutation({
  mutationFn: ...,
  onSuccess: () => { ... setCopyModalData(null); },
  onError: ...,
  // missing:
  onSettled: () => { queryClient.invalidateQueries({ queryKey: ['customCollections'] }); },
});
```

Not blocking, but inconsistent with sibling mutations.

---

### ✅ Additional observations on new code

- **`ShareToCampaignsModal` refactor** — clean; the `idPrefix` threading correctly avoids duplicate DOM ids across all five share modals; the `column` / `queryKind` / `applyVerb` parameterisation is unambiguous
- **Rename commit** — pure find-replace (+389/-389), no logic changes; nomenclature is consistent throughout (files, types, RPC, migration, README)
- **`getUserCustomCollections` name resolution** — correct one-query-per-type pattern; dangling item ids (deleted underlying assets) are silently skipped, matching the stated contract
- **`copy_custom_collection` SQL** — closure loop is correct; topological insert order (skill types → skills → equipment → gang types → fighter types → ...) respects FK dependencies; trading post availability/pricing FK remapping via `v_map_tpe` is correct
- **Profile page wiring** — `readOnly={currentUserId !== profile.id}` correctly gates edit/share/delete to the owner; no candidate arrays needed in readOnly mode (item picker is hidden)
- **Migration** — RLS policies follow verbatim owner-or-admin write / open-read pattern; `ON DELETE SET NULL` on `custom_collection_id` is the right choice

---

**Verdict: Approve** — all previous HIGH/MEDIUM findings resolved, one trivial LOW item (cache invalidation on copy) is the only remaining note.

---

PR #1770, opened 2026-06-13
Equipment attachment tooltip updates

prvlaar:
## Summary

- Weapon attachment tooltips on the fighter page were showing base (unmodified) weapon profiles even when attachments (e.g. Tier 1 upgrade) had modifiers applied
- Range values in tooltips were missing the inch symbol (`24` instead of `24"`)
- Removing an attachment didn't revert the parent weapon's profiles in client state

## Changes

- **`fighter-data.ts`**: Fixed targeting effects query to mirror `gang-data.ts` — previously used `.in('target_equipment_id', ...)` which missed effects where `target_equipment_id` is null (rig-glitch style). Now fetches all effects with `fighter_equipment_id` set for the fighter and resolves the target as `target_equipment_id || fighter_equipment_id`. Also populates `base_weapon_profiles` alongside modified profiles to enable client-side revert.
- **`equipment.ts`**: Returns `attachment_effect` with modifiers in the server response so the client can update the target weapon's tooltip immediately after purchase without a page refresh.
- **`fighter-equipment-list.tsx`**: Applies `attachment_effect` modifiers to the target weapon optimistically after purchase; reverts profiles to `base_weapon_profiles` when an attachment is removed.
- **`equipment-tooltip.tsx`**: Appends `"` to numeric range values to match the fighter card display.
- **`package.json`**: Adds `dev:fresh` script for a clean dev server restart that clears the data cache.

## Test plan

- [ ] Fighter with an attachment (e.g. Yeld missile gauntlet + Tier 1) — tooltip shows modified profiles (AP, range etc.)
- [ ] Range values display with inch symbol (`24"` not `24`)
- [ ] Purchasing a new attachment updates the target weapon's tooltip immediately without a page refresh
- [ ] Removing an attachment reverts the target weapon's tooltip to base profiles immediately
- [ ] Page refresh after any of the above shows correct data

### DIFF
--- a/app/actions/equipment.ts
+++ b/app/actions/equipment.ts
@@ -965,6 +965,22 @@ export async function buyEquipmentForFighter(params: BuyEquipmentParams): Promis
             try {
               invalidateFighterDataWithFinancials(params.fighter_id, params.gang_id);
             } catch {}
+
+            // Include the applied effect's modifiers in the response so the client can
+            // update the target weapon's profiles immediately without a page refresh
+            if (result.effect_data) {
+              responseData.attachment_effect = {
+                id: result.effect_id,
+                effect_name: result.effect_data.effect_name,
+                type_specific_data: result.effect_data.type_specific_data || {},
+                fighter_effect_modifiers: (result.effect_data.fighter_effect_type_modifiers || []).map((m: any) => ({
+                  stat_name: m.stat_name,
+                  numeric_value: m.default_numeric_value,
+                  operation: m.operation || 'add'
+                })),
+                target_equipment_id: chosenTargetId
+              };
+            }
           }
         } catch (e) {
           console.error('Failed to attach equipment upgrade during purchase:', e);
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -462,12 +462,13 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
           }
 
           // If we have profiles, apply equipment-targeted effects from fighter_effects targeting this equipment
+          let baseWeaponProfiles: any[] | null = null;
           if (weaponProfiles.length > 0) {
             // Lookup targeting effects from Map (O(1) instead of query)
             const targetingEffects = targetingEffectsMap.get(fighterEquipmentId) || [];
 
             if (targetingEffects.length > 0) {
-              // Use shared utility function to apply weapon modifiers
+              baseWeaponProfiles = weaponProfiles;
               weaponProfiles = applyWeaponModifiers(weaponProfiles, targetingEffects);
             }
           }
@@ -490,6 +491,7 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
             is_master_crafted: item.is_master_crafted || false,
             is_editable: item.is_editable || false,
             weapon_profiles: weaponProfiles,
+            base_weapon_profiles: baseWeaponProfiles,
             target_equipment_id: targetEquipmentId,
             effect_names: effectNames.length > 0 ? effectNames : undefined,
             loadout_ids: loadoutIds.length > 0 ? loadoutIds : undefined,
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -332,7 +332,11 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
               .order('profile_name')
           : Promise.resolve({ data: [] }),
 
-        // Batch fetch targeting effects (effects that target equipment)
+        // Batch fetch targeting effects (effects that modify weapon profiles).
+        // Two cases mirror the gang-data.ts approach:
+        // 1. Equipment-to-equipment upgrades: fighter_equipment_id = source attachment,
+        //    target_equipment_id = target weapon
+        // 2. Rig glitches: fighter_equipment_id = target weapon, target_equipment_id = NULL
         fighterEquipmentIds.length > 0
           ? supabase
               .from('fighter_effects')
@@ -342,11 +346,13 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
                 effect_name,
                 type_specific_data,
                 sort_order,
+                fighter_equipment_id,
                 target_equipment_id,
                 fighter_effect_type:fighter_effect_type_id ( sort_order ),
                 fighter_effect_modifiers ( stat_name, numeric_value, operation )
               `)
-              .in('target_equipment_id', fighterEquipmentIds)
+              .eq('fighter_id', fighterId)
+              .not('fighter_equipment_id', 'is', null)
           : Promise.resolve({ data: [] }),
 
         // Batch fetch loadout assignments (which loadouts each equipment belongs to)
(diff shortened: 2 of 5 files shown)


joeseos (changes requested):
## PR Review — Equipment attachment tooltip updates

@prvlaar nice set of fixes. The core approach is sound and the gang-data.ts mirroring makes sense. A few things to flag:

---

### 🟡 Medium — Multi-attachment optimistic update applies second effect to base, not to current state

**`components/fighter/fighter-equipment-list.tsx` ~line 189**

```ts
const baseProfiles = eq.base_weapon_profiles || eq.weapon_profiles || [];
const modifiedProfiles = applyWeaponModifiers(baseProfiles, [attachmentEffect]);
return { ...eq, base_weapon_profiles: baseProfiles, weapon_profiles: modifiedProfiles };
```

When a second attachment targeting the same weapon is purchased, `eq.base_weapon_profiles` is already set from the first purchase. So `baseProfiles` is the **original unmodified** profiles, and the second effect is applied to those — overwriting the first attachment's modification in the displayed tooltip. The cumulative result (both effects) won't show until a page refresh.

**Fix:** apply to `eq.weapon_profiles` (current state) rather than `baseProfiles`, while keeping `base_weapon_profiles` anchored to the original:

```ts
const baseProfiles = eq.base_weapon_profiles || eq.weapon_profiles || [];
const currentProfiles = eq.weapon_profiles || [];
const modifiedProfiles = applyWeaponModifiers(currentProfiles, [attachmentEffect]);
return { ...eq, base_weapon_profiles: baseProfiles, weapon_profiles: modifiedProfiles };
```

This correctly stacks both effects for the tooltip while keeping the revert anchor correct.

---

### 🟡 Medium — Partial revert when one of two attachments is removed leaves profiles stale

**`components/fighter/fighter-equipment-list.tsx` ~line 35**

```ts
const stillTargeted = filteredEquipment.some(r => r.target_equipment_id === eq.fighter_equipment_id);
if (!stillTargeted && eq.base_weapon_profiles) { ... }
```

The `stillTargeted` guard is correct — it avoids a full revert when another attachment remains. But after removing the first of two attachments, the weapon's `weapon_profiles` still reflects both modifiers. Only a page refresh corrects this.

This is an inherent limitation of the current approach (no stored per-attachment effect state on the client). Acceptable since the server re-syncs on the next load, but worth noting in the PR description as a known gap. It would be fixable by re-applying remaining effects from base on removal, but that requires tracking effect payloads on the client.

---

### 🟢 Low — `dev:fresh` uses `npx` inconsistently

**`package.json`**

```json
"dev:fresh": "npx next clean && npx next dev --turbopack"
```

Other scripts invoke `next` directly (via the local binary). `npx` here is redundant and inconsistent. Should be:
```json
"dev:fresh": "next clean && next dev --turbopack"
```

---

### ✅ Things verified correct

- **`fighter-data.ts` query refactor**: The switch from `.in('target_equipment_id', fighterEquipmentIds)` to `.eq('fighter_id', fighterId).not('fighter_equipment_id', 'is', null)` correctly mirrors gang-data.ts and handles both rig-glitch and equipment-to-equipment cases. The `target_equipment_id || fighter_equipment_id` fallback in the map-building loop is right.

- **`baseWeaponProfiles` assignment**: `applyWeaponModifiers` returns a new array with `{ ...profile }` copies, so `baseWeaponProfiles` correctly holds the pre-modifier reference without mutation risk.

- **`formatRange`**: Regex `/\d$/` correctly guards against double-appending `"` on values already containing it (e.g., a DB-stored `"24\""`). Handles `-`, `N/A`, `E` and other non-numeric strings gracefully.

- **`effect_data` → `fighter_effect_modifiers` mapping**: `insertEffectWithModifiers` returns `fighter_effect_type_modifiers` with `default_numeric_value`; the mapper in `equipment.ts` correctly renames it to `numeric_value` to match the `EffectModifier` interface. Good.

- **Rollback on error**: All three removal paths (delete, sell, stash) correctly snapshot state before the optimistic update and rollback on error. The `revertWeaponProfilesAfterRemoval` function is idempotent and safe.

---

### Verdict: **Changes requested** (one real bug, one acceptable limitation)

The multi-attachment stacking bug is the main blocker — worth fixing since the one-line correction is straightforward. Everything else is correct or minor.

---

joeseos:
@prvlaar Not sure adding another fetch to the get fighter data is the way to go here. I'll have a look.

joeseos (approved):
## Follow-up Review — Equipment attachment tooltip updates

@prvlaar reviewed the new commit (`99e1219`) against the previous `CHANGES_REQUESTED`.

---

### Previous findings

| Finding | Status |
|---|---|
| 🟡 Multi-attachment stacking: second effect applied to base instead of current state | ✅ **Resolved** |
| 🟡 Partial revert with two attachments leaves profiles stale (known limitation) | ⚪ **Outstanding — accepted as-is** |
| 🟢 `dev:fresh` uses redundant `npx` prefix | ✅ **Resolved** |

---

### Stacking fix — verified correct

```ts
// fighter-equipment-list.tsx (new)
const baseProfiles = eq.base_weapon_profiles || eq.weapon_profiles || [];
const currentProfiles = eq.weapon_profiles || [];
const modifiedProfiles = applyWeaponModifiers(currentProfiles, [attachmentEffect]);
return { ...eq, base_weapon_profiles: baseProfiles, weapon_profiles: modifiedProfiles };
```

`currentProfiles` now tracks the running modified state, so a second attachment stacks on top of the first. `baseProfiles` correctly anchors to the original via the `|| eq.weapon_profiles` fallback (which kicks in on first purchase, before `base_weapon_profiles` exists). The revert path is also consistent: after `revertWeaponProfilesAfterRemoval` sets `base_weapon_profiles: null`, the next purchase falls back to `eq.weapon_profiles` as the base — which is now the reverted base. All transitions are sound.

### `dev:fresh` — verified correct

```json
"dev:fresh": "next clean && next dev --turbopack"
```

`npx` removed, consistent with the other scripts.

### No new issues found

Scanned the full diff. The `attachment_effect` shape (`fighter_effect_modifiers` with `numeric_value`) matches what `applyWeaponModifiers` already expects from server-fetched effects. The broadened query in `fighter-data.ts` (`.eq('fighter_id').not('fighter_equipment_id', 'is', null)` vs the old `.in('target_equipment_id', ...)`) may return slightly more rows for fighters with many effects, but extra rows are simply unmapped and the corrected rig-glitch coverage justifies it.

---

### Verdict: **Approve** ✅

All actionable findings addressed. The one outstanding item (stale profiles after partial removal of two attachments) was accepted as a known limitation in the previous review and still is — it self-corrects on next page load.

---

prvlaar:
> @prvlaar Not sure adding another fetch to the get fighter data is the way to go here. I'll have a look.

Thanks! On local for some reason the server RPC isn't running. So it's also not doing optimistic updates. 
It's flagging that it's not working, but will work on Live. Interested to see if that's the case

joeseos:
> > @prvlaar Not sure adding another fetch to the get fighter data is the way to go here. I'll have a look.
> 
> Thanks! On local for some reason the server RPC isn't running. So it's also not doing optimistic updates. It's flagging that it's not working, but will work on Live. Interested to see if that's the case

There's no RPC to fetch the fighter data though. It's in the fighter-data.ts.

joeseos:
Closing this PR as solution is in [PR1784](https://github.com/joeseos/mundamanager/pull/1784)

PR #1780, opened 2026-06-15
Defer pwa install button rendering to avoid hydration mismatch

Djidiouf:


joeseos (approved):
## PR Review — `fix/pwa_hydration_error` @Djidiouf

**Verdict: ✅ Approve (with minor suggestions)**

---

### What the PR does

Moves all `window`/UA detection out of `useState` lazy initializers into a `useEffect`, introducing an `isReady` flag that gates rendering until client-side detection is complete. This is the canonical Next.js fix for hydration mismatches in client-only components.

---

### Findings

#### 🟡 Medium — `detectPlatform()` has no SSR guard

**File:** `components/pwa-install-button.tsx`, lines 13–28

```ts
function detectPlatform() {
  const ua = window.navigator.userAgent.toLowerCase(); // will throw in SSR context
  ...
}
```

The extracted function directly accesses `window` without any `typeof window !== 'undefined'` guard. It is currently **safe** because it's only ever called from inside `useEffect`, which only runs on the client. But since it lives at module scope with no guard, any future refactor that calls it outside a `useEffect` (e.g. a `useMemo`, another `useState` initializer, or a utility import) will throw a `ReferenceError` during SSR.

The original code had explicit `if (typeof window === 'undefined') return false;` guards for this reason. The refactor removed them without preserving that safety boundary.

**Suggested fix — add a guard to the function:**
```ts
function detectPlatform() {
  if (typeof window === 'undefined') {
    return { ios: false, android: false, windowsDesktop: false, installed: false };
  }
  const ua = window.navigator.userAgent.toLowerCase();
  // ...
}
```

---

#### 🟢 Low — `console.log` in production install handler

**File:** `components/pwa-install-button.tsx`, lines 135–138 (pre-existing, not introduced by this PR)

```ts
console.log('User accepted the install prompt');
console.log('User dismissed the install prompt');
```

These ship to production. Minor, but worth cleaning up before or after this lands.

---

### What's correct ✅

- **Fix is sound:** Moving all `window` access to `useEffect` + returning `null` until `isReady` is the right hydration-safe pattern for Next.js `'use client'` components.
- **React 18 batching:** All the `setState` calls in the first `useEffect` are batched into a single re-render, so by the time the second `useEffect` fires (triggered by `isReady` becoming `true`), `isIOS` and `isInstalled` already have their correct values. No race condition.
- **Type safety improvement:** Replacing `(window.navigator as any).standalone` with `(window.navigator as Navigator & { standalone?: boolean }).standalone` is a good cleanup.
- **iOS installable logic preserved:** The `if (ios && !installed) setIsInstallable(true)` path in the first `useEffect` correctly reproduces the old lazy initializer behavior.
- **Render gate ordering:** The `if (!isReady) return null` check at line 142 precedes all the platform-specific render guards, so `isWindowsDesktop`, `isIOS`, etc. are always fully resolved before they're used in render.

---

The medium finding is a safety concern worth addressing (a one-liner guard), but it doesn't block this fix from landing since the current call site is safe.

---

joeseos (approved):
## Follow-up Review — `996a453` @Djidiouf

**Verdict: ✅ Approve — all previous findings addressed, no new issues**

---

### Previous findings status

| Finding | Severity | Status |
|---|---|---|
| `detectPlatform()` missing SSR guard — bare `window` access without `typeof window !== 'undefined'` | 🟡 Medium | ✅ **Resolved** |
| `console.log` in production install handler | 🟢 Low | ✅ **Resolved** |

---

### What changed in this commit

**SSR guard added to `detectPlatform()` (lines 13–16):**
```ts
if (typeof window === 'undefined') {
  return { ios: false, android: false, windowsDesktop: false, installed: false };
}
```
This restores the same safety boundary the original lazy initializers had. The function is now safe to call from any context, not just `useEffect`.

**`console.log` statements removed along with the dead `outcome` variable:**
```ts
// Before
const { outcome } = await deferredPrompt.userChoice;
if (outcome === 'accepted') { console.log(...) } else { console.log(...) }

// After
await deferredPrompt.userChoice;
```
Clean — no production logging noise, and the unused destructured variable is gone too.

---

### No new issues introduced

The two-commit diff is minimal and surgical. Nothing else changed. The complete fix is now correct, clean, and SSR-safe. Ready to merge.

---

PR #1781, opened 2026-06-15
Updated node version

Djidiouf:


joeseos (commented):
## PR Review — "Updated node version" (@Djidiouf)

**Scope:** 3 files changed, 3 additions, 3 deletions — `.nvmrc`, `package.json`, `package-lock.json`.

The intent is clear and the direction is correct: bumping `@types/node` to satisfy eslint's type requirements and aligning the local dev Node.js version.

---

### Findings

#### 🟡 Medium — `package-lock.json` may not have been properly regenerated

**File:** `package-lock.json`, line 58

The diff shows only one changed line in `package-lock.json` — the version specifier in `packages[""].devDependencies` (which mirrors `package.json`). In npm's lockfile v3 format, a real `npm install` that upgrades `@types/node` across a version boundary would also produce a changed entry in the `packages["node_modules/@types/node"]` section (the resolved, pinned version).

The absence of that second change suggests the lockfile was **manually edited** rather than regenerated via `npm install`. If the locked resolved version of `@types/node` was below `20.19.0` (e.g. `20.12.0`, which satisfies the old `^20.10.6`), then:
- `npm ci` in a fresh environment will fail (lockfile inconsistency)
- TypeScript may silently use mismatched types until someone reruns `npm install`

**Suggested fix:** Run `npm install` locally after editing `package.json`, then commit the full diff from `package-lock.json`. If the resolved version was already `>=20.19.0` in the lockfile, this is a non-issue — but please confirm.

---

#### 🟢 Low — `engines` field not updated (intentional?)

**File:** `package.json`, engines field (unchanged)

`package.json` still declares `"node": ">=20.9.0"`. This is fine if the intent is to keep the **minimum** version for consumers/deployers at 20.9.0 and use 20.20.2 only for local dev (via `.nvmrc`). Just worth a conscious call — if the app now genuinely requires 20.20.2 to build/run, the `engines` field should reflect that.

---

#### ℹ️ Informational — CI workflows don't use Node.js

Neither `.github/workflows/` file references Node.js, so the `.nvmrc` change only affects local dev environments and Vercel (which passed its check). No CI concern here.

---

### Verdict

**Changes requested (on the lockfile point).** The code change itself is correct and minimal. The lockfile concern is the only blocker — if you can confirm that `npm install` was run and the resolved `@types/node` version was already `>=20.19.0` in the lockfile (meaning no second hunk was needed), I'm happy to approve immediately.

---

joeseos:
## PR Re-Review — "Updated node version" (@Djidiouf)

Reviewing the new commit `bd17f8f` pushed since the last review.

---

| Finding | Severity | Status |
|---|---|---|
| `package-lock.json` may have been manually edited (missing resolved-version update for `@types/node`) | 🟡 Medium | ✅ **Resolved** |
| `engines` field in `package.json` not updated to match `.nvmrc` | 🟢 Low | ✅ **Resolved** |

Both blockers have been addressed:
- `package.json` `engines.node` bumped from `>=20.9.0` → `>=20.20.2`
- The lockfile was properly regenerated via a real `npm install` run — confirmed by the 1281-line churn across optional platform packages (`@emnapi/*`, `@img/sharp-*`, `lightningcss-*` variants), which only happens with an actual install, not a manual edit.

---

### New Findings

#### 🔵 Informational — Lockfile regenerated on Linux x64; macOS darwin sharp binaries stripped

The `npm install` was clearly run on a Linux x64 environment. As a result, the macOS (`@img/sharp-darwin-arm64`, `@img/sharp-darwin-x64`) and some arm64 optional native binaries were removed from the lockfile, while Linux x64 entries remain.

**Impact:**
- Vercel (Linux) builds — unaffected ✅
- `npm ci` on a macOS dev machine — optional sharp deps will be missing from the lockfile; `npm ci` may warn or fail on those optionals
- `npm install` on macOS — will still resolve correctly

This is a common trade-off when regenerating lockfiles on Linux CI/CD. Since the deploy target is Vercel/Linux and Sharp's Darwin binaries are optional, this is **not a blocker**. Just worth knowing if any team members see `npm ci` warnings locally.

No new bugs, security issues, or regressions introduced by this commit.

---

### Verdict

**✅ Approved** — Both previous findings are resolved. The change is correct, minimal, and the lockfile is now in a valid regenerated state. Good to merge.

---

PR #1784, opened 2026-06-15
Fixed so that effects shows in fighter equipment list tooltip

joeseos:


joeseos (commented):
@joeseos — the fix is logically sound. Verdict: **approve** with a few low-severity notes worth cleaning up.

---

## Review: PR #1784 — Effects in fighter equipment list tooltip

**Scope:** 3 files, 49 additions / 31 deletions. Core change: expand the effects query in `fighter-data.ts` to include self-effects alongside targeting effects, add cache invalidation when effects change, and update the optimistic update to show stat modifications immediately.

---

### ✅ Correctness — no bugs found

**`fighter-data.ts` — query expansion**

The original `.in('target_equipment_id', fighterEquipmentIds)` only fetched targeting effects. The new `.or(...)` query correctly adds self-effects with `target_equipment_id IS NULL`. Condition 2 explicitly requires `target_equipment_id IS NULL` which avoids double-counting targeting effects. ✓

The `weaponEffectsMap` key logic (`effect.target_equipment_id || effect.fighter_equipment_id`) is correct, and filtering to `targetingOnly` for the names tooltip is the right call — prevents duplicating names already shown via `renderEffectRows`. ✓

**`equipment.ts` — cache invalidation**

Adding `revalidateTag(CACHE_TAGS.BASE_FIGHTER_EQUIPMENT(...), { expire: 0 })` to both `applySelfUpgradesToEquipment` and `deleteEquipmentEffect` is the right fix. Previously adding/removing an equipment effect wouldn't bust the equipment list cache, so the tooltip wouldn't refresh. ✓

**`fighter-equipment-list.tsx` — optimistic update**

Moving `newEffects` construction before `equipment.map()` and populating `fighter_effect_modifiers` with actual modifier data is what allows `applyWeaponModifiers` to update weapon profiles optimistically. The original was passing empty `[]` modifiers — that's why the tooltip showed stale stats. Fix is correct. ✓

---

### ⚠️ Issues

**[MEDIUM] Stale variable name `targetingEffectsData` — `fighter-data.ts:242`**

```typescript
const [..., targetingEffectsData, ...] = await Promise.all([...])
```

The variable is still named `targetingEffectsData` but now holds both targeting effects and self-effects. The comment above the query was updated, but the variable name wasn't. A future reader will be confused when they see `targetingEffectsData.data` feeding `weaponEffectsMap` which covers both types.

```typescript
// Suggest renaming to:
const [..., equipmentEffectsData, ...] = await Promise.all([...])
// and: (equipmentEffectsData.data || []).forEach(...)
```

---

**[LOW] Three separate `Date.now()` calls where IDs must match — `fighter-equipment-list.tsx`**

```typescript
const newEffects = selectedEffects.map((effect, index) => ({
  id: `temp-${Date.now()}-${index}`,
  fighter_effect_modifiers: (effect.modifiers || []).map(m => ({
    id: `temp-mod-${Date.now()}-${index}`,
    fighter_effect_id: `temp-${Date.now()}-${index}`,  // ← should equal parent id
```

`fighter_effect_id` on modifiers is supposed to reference the parent effect's `id`. They'll match in practice (synchronous tick), but the intent isn't clear. Simple fix:

```typescript
const newEffects = selectedEffects.map((effect, index) => {
  const tempId = `temp-${Date.now()}-${index}`;
  return {
    id: tempId,
    fighter_effect_modifiers: (effect.modifiers || []).map((m, mi) => ({
      id: `temp-mod-${tempId}-${mi}`,
      fighter_effect_id: tempId,
```

---

**[LOW] Type cast `as Parameters<typeof applyWeaponModifiers>[1]` — `fighter-equipment-list.tsx`**

```typescript
updatedProfiles = applyWeaponModifiers(updatedProfiles, newEffects as Parameters<typeof applyWeaponModifiers>[1]);
```

The mismatch is `FighterEffect.type_specific_data: TypeSpecificData | string` vs `Effect.type_specific_data?: TraitModificationData`. Correct at runtime since `applyWeaponModifiers` only uses `type_specific_data` for structured trait objects. But the cast silences the checker on a real type divergence — worth aligning the types rather than accumulating casts.

---

**[LOW] `.or()` uses string interpolation — `fighter-data.ts`**

```typescript
.or(`target_equipment_id.in.(${fighterEquipmentIds.join(',')}),and(fighter_equipment_id.in.(${fighterEquipmentIds.join(',')}),target_equipment_id.is.null)`)
```

The previous `.in('target_equipment_id', fighterEquipmentIds)` used the parameterized client API. No security risk here (DB-sourced UUIDs), and Supabase's `.or()` doesn't expose a clean parameterized form for nested `and()+in()`, so this is pragmatically acceptable — just noting the pattern inconsistency.

---

### Verdict: **Approve**

Bug is correctly fixed. Cache invalidation is in the right places. Optimistic update now reflects actual stat changes. The `targetingEffectsData` rename is the one item I'd want addressed before merge (maintenance hazard). Everything else is low-severity cleanup.

---

PR #1785, opened 2026-06-15
Fixed the hydration error due to DnD attributes on home page

Djidiouf:


joeseos (approved):
## PR Review — `fix/hydration_error_with_dnd_on_home_page`

@Djidiouf Good fix — the root cause diagnosis is correct and the approach used is the canonical React 18 solution for this exact problem.

---

### What changed and why it's right

The old initializer:
```tsx
const [isMounted] = useState(() => typeof window !== 'undefined');
```
is broken in Next.js App Router because the `useState` initializer re-runs on the client during hydration and evaluates to `true` (window exists), while the server produced `false`. React sees a mismatch in the rendered tree (DnD markup vs. plain markup) → hydration error.

The replacement:
```tsx
const isMounted = useSyncExternalStore(() => () => {}, () => true, () => false);
```
uses `getServerSnapshot` (`() => false`) both on the server and during the client's hydration pass, so the client matches what the server produced. After hydration completes React switches to `getSnapshot` (`() => true`) and re-renders with `DndContext` active. This is precisely the pattern documented in the React 18 `useSyncExternalStore` docs for "browser-only" rendering.

---

### Findings

**[LOW] `components/home/campaigns-tab.tsx` & `components/home/gangs-tab.tsx` — unstable `subscribe` reference**

`useSyncExternalStore` expects a stable `subscribe` reference. The current inline arrow `() => () => {}` creates a new function object on every render, causing React to re-subscribe on each render. It's harmless here because the subscription does nothing and the store never emits changes, but it's technically incorrect per the API contract and React may warn about it in a future version.

Suggested fix — extract to a module-level constant in each file (or in a shared hook):
```tsx
const noopSubscribe = () => () => {};
// ...
const isMounted = useSyncExternalStore(noopSubscribe, () => true, () => false);
```

**[LOW] DRY — identical pattern in two files**

Both files contain the exact same `useSyncExternalStore` idiom (same as before with `useState`). A small shared hook would centralise it:
```tsx
// hooks/use-is-mounted.ts
import { useSyncExternalStore } from 'react';
const noopSubscribe = () => () => {};
export function useIsMounted() {
  return useSyncExternalStore(noopSubscribe, () => true, () => false);
}
```
This isn't a blocker for this PR since the duplication pre-existed, but it's a good follow-up.

---

### Verdict: ✅ Approve

The fix is correct, minimal, and directly addresses the hydration error. The two low-severity notes above are improvements worth doing in a follow-up but don't block this from landing.

---

joeseos:
## Follow-up Review — `fix/hydration_error_with_dnd_on_home_page` (commit `38576d6`)

@Djidiouf Both findings from the previous review have been addressed. Nothing new to flag.

---

### Previous findings status

| Finding | Status |
|---|---|
| [LOW] Unstable `subscribe` reference (`() => () => {}` inline) | ✅ **Resolved** — `noopSubscribe`, `getClientSnapshot`, and `getServerSnapshot` are all extracted to module-level constants in `hooks/use-is-mounted.ts` |
| [LOW] DRY — identical pattern duplicated across files | ✅ **Resolved** — shared `useIsMounted()` hook created and adopted in all three components |

---

### New commit quality

The new commit goes further than the previous review asked for:

- `components/gang/draggable-fighters.tsx` was also carrying the same inline `useSyncExternalStore` pattern (it wasn't explicitly called out before), and it's been cleaned up too — good catch.
- The hook implementation in `hooks/use-is-mounted.ts` is clean: stable references, correct snapshot semantics, and the JSDoc accurately describes the before/after mount behaviour.
- No new issues, regressions, or edge cases introduced.

---

### Verdict: ✅ Approve

The fix is correct and the follow-up refactor tightens it up nicely. Ready to merge.

---

joeseos (approved):
Follow-up review on the new commit — both previous findings resolved, no new issues. ✅

---

PR #1788, opened 2026-06-16
Fixed issues regarding Gang origin not being updated in Gang page thr…

Djidiouf:
…ough optimistic update and Skill access not being updated as RPC was lacking taking it into account


joeseos (changes requested):
Hey @Djidiouf — good work isolating both the optimistic-update bug and the missing origin skill access. The shape of the solution is solid. One **high-severity** logic divergence between the SQL function and the TypeScript route needs to be fixed before this lands, plus a few smaller points below.

---

## High

### 1. `access_level = 'denied'` behaves differently in SQL vs TypeScript

**File:** `app/api/fighters/skill-access/route.ts` line 197–203 / `app/lib/shared/skill-access.ts` line 26 / `supabase/functions/get_available_skills.sql`

**Problem:**

In the SQL function, origin access is the last fallback in `COALESCE`:

```sql
COALESCE(
    sao.access_level,           -- fighter override
    ftsa.access_level,          -- fighter type default  ← if 'denied', stops here
    CASE WHEN skill_type_id = v_origin_skill_type_id THEN 'primary' ELSE ... END
)
```

SQL `COALESCE` returns the **first non-NULL value**, so an explicit `ftsa.access_level = 'denied'` is returned as-is and the origin `CASE` is never reached. The origin cannot override a fighter-type denial in the RPC.

In the TypeScript route, `pickStrongerSkillAccessLevel` treats `'denied'` as equivalent to `null`:

```typescript
if (!a || a === 'denied') return b && b !== 'denied' ? b : null;
```

So if `existing.access_level = 'denied'` (the fighter-type default) and there is no individual override (`override_access_level = null`), the origin **upgrades** `access_level` to `'primary'`. The effective level shown to the UI is `null ?? 'primary' = 'primary'`.

**Impact:** The gang page shows the origin skill set as Primary-accessible; when the fighter actually tries to acquire a skill from it, `get_available_skills` (the RPC) silently excludes those skills because it still sees `'denied'`. The result is a misleading UI — a visible, clickable skill set that returns no skills.

**Fix — decide one canonical rule and align both:**

*Option A – origin cannot override an explicit fighter-type denial (match SQL):*
Change `pickStrongerSkillAccessLevel` to treat `'denied'` as stronger than any positive level:
```typescript
// 'denied' is an explicit restriction; origin cannot override it
if (a === 'denied' || b === 'denied') return 'denied';
if (!a) return b ?? null;
if (!b) return a;
return ACCESS_RANK[a] >= ACCESS_RANK[b] ? a : b;
```

*Option B – origin CAN override a fighter-type denial (fix SQL to match TypeScript):*
Replace the COALESCE logic in the SQL with an explicit ranking expression that treats 'denied' as overridable. This is more complex and changes game-rule semantics.

Option A is the safer default and consistent with how `sao.access_level = 'denied'` (individual override) already takes priority over origin.

---

## Medium

### 2. Two extra sequential DB queries on every request

**File:** `app/api/fighters/skill-access/route.ts` lines 175–203

Previously the gang query ran only inside `if (skillTypeId)`. Now it runs unconditionally, and a second `skill_types` query can follow it. The two are sequential (second depends on first). On a fighter with no origin and no `skillTypeId` this is wasted latency.

**Suggestion:** Guard the entire block the same way — only resolve origin access when you actually need it to fill the response. If `gangOriginName` is only needed to extend `formattedSkillAccess`, it can run after line 172 but only when the origin info is non-trivially needed. Alternatively, combine both into a single query:

```typescript
const { data: candidateSkillType } = await supabase
  .from('gang_origins')
  .select('id, origin_name, skill_types!inner(id, name)')
  .eq('id', fighter_gang_origin_id)  // join gang first
  ...
```

Not a blocking concern if the queries are fast in practice, but worth noting.

---

## Low

### 3. `findSkillTypeMatchingOriginName` is exported but never imported anywhere

**File:** `app/lib/shared/skill-access.ts` lines 36–47

This function exists only in the shared module but is not used by the route (which uses `.ilike()` + `normaliseSkillSetName` inline instead) and is not imported anywhere else. Dead export. Remove it or use it — the route's inline logic could be replaced with this function to reduce duplication.

### 4. No error handling on the `gangData` fetch

**File:** `app/api/fighters/skill-access/route.ts` line 175

```typescript
const { data: gangData } = await supabase.from('gangs')...single();
```

The `error` is silently discarded. If the query fails, `gangOriginName` is `undefined`, origin access is skipped, and the endpoint returns 200 with incomplete data. Consistent with some existing patterns in the route, but worth adding at least a `console.error` or a fallback so failures are observable.

### 5. Custom skills with only `custom_skill_type_id` are excluded from origin access (SQL)

**File:** `supabase/functions/get_available_skills.sql` / migration

In `visible_custom_skills`, the origin CASE only matches `cs.skill_type_id`:

```sql
WHEN cs.skill_type_id = v_origin_skill_type_id THEN 'primary'
```

Custom skills backed by a `custom_skill_type_id` (no `skill_type_id`) will never match. This is probably intentional (origin grants only affect standard skill sets), but confirm it is by design.

---

## What's good

- **`gang.tsx` fix is correct.** Using `gangOriginName` state (updated by the optimistic-update mutation) instead of the `gang_origin_name` prop (static initial value) properly reflects the origin change without a page reload.
- **`applyGangOriginSkillAccess` correctly preserves `override_access_level`** when upgrading an existing entry — individual overrides still take priority.
- **The migration matches the SQL source file** — no drift.
- **`normaliseSkillSetName` in the double-check** is actually load-bearing: `.ilike()` doesn't escape LIKE wildcards (`%`, `_`), so an origin name containing those characters would match unintended rows. The secondary string equality check catches those false positives. Worth adding an inline comment to explain why both checks are present.

---

**Verdict: changes requested.** The `access_level = 'denied'` divergence between the RPC and the TypeScript route is the blocker — it produces a silent mismatch where the UI claims access the database won't grant. Fix that (preferably Option A), clear up the dead export, and this should be ready to merge.

---

joeseos:
## PR Review — @Djidiouf

### Previous findings
No prior review comments on this PR — this is the first review.

---

### Summary of changes

This PR tackles two distinct bugs:
1. **Gang origin display** — the gang page was rendering the stale prop `gang_origin_name` instead of the reactive state `gangOriginName`, so optimistic updates didn't appear until a full refresh.
2. **Skill access gap** — the `skill-access` API route and the `get_available_skills` SQL function both ignored the gang's Origin when computing which Skill Sets a fighter can access.

The approach is sound: a new shared utility (`app/lib/shared/skill-access.ts`) handles the TypeScript-side merging, and the SQL function is extended with a `v_origin_skill_type_id` lookup. Overall this is clean work.

---

### New findings

#### 🟠 Medium — `ilike` without wildcard escaping, `route.ts:200`

```ts
.ilike('name', gangOriginName)
```

`gangOriginName` comes from `gang_origins.origin_name`. PostgreSQL's `ILIKE` operator treats `%`, `_`, and `\` as metacharacters. If an origin name contains those characters, the query may return no row (false negative), silently skipping origin skill access for that fighter entirely. `resolveSkillTypeForOriginName` correctly prevents false positives from wildcard expansion, but it cannot recover from a false negative where no row is returned in the first place.

**Fix:** escape metacharacters before the `.ilike()` call:
```ts
const escapedOriginName = gangOriginName.replace(/[%_\\]/g, '\\$&');
const { data: candidateSkillType, error: skillTypeError } = await supabase
  .from('skill_types')
  .select('id, name')
  .ilike('name', escapedOriginName)
  .maybeSingle();
```
Or, since case-insensitive exact-match is what you need, consider `.filter('name', 'ilike', escapedOriginName)` with the escape, or an RPC that uses `lower(name) = lower($1)`.

---

#### 🟡 Low — Redundant `DROP FUNCTION` before `CREATE OR REPLACE` in migration

`supabase/migrations/20260616120000_gang_origin_skill_set_access.sql`:

```sql
DROP FUNCTION IF EXISTS public.get_available_skills(uuid);
CREATE OR REPLACE FUNCTION public.get_available_skills(...)
```

The function signature is unchanged (`uuid → jsonb`), so `CREATE OR REPLACE` alone handles the in-place update safely. The `DROP IF EXISTS` before it is unnecessary and introduces risk: if any other database object depends on this function, `DROP` without `CASCADE` will raise an error and abort the migration. `CREATE OR REPLACE` never has that problem.

**Fix:** Remove the `DROP` line; rely solely on `CREATE OR REPLACE FUNCTION`.

---

#### 🟡 Low — `override_access_level` silently wins over origin-granted `access_level` (intentional but undocumented)

In `applyGangOriginSkillAccess`, only `access_level` is upgraded; `override_access_level` is left untouched. This means a per-fighter override of `'secondary'` will be shown as the effective level even when origin grants `'primary'`. This is consistent with the SQL (where `sao.access_level` is first in `COALESCE`), so the TypeScript and SQL agree. But the function docblock doesn't mention this — a reader could assume origin upgrades the effective displayed level.

**Fix:** Add a single line to the JSDoc of `applyGangOriginSkillAccess` noting that `override_access_level` takes precedence (mirrors the SQL COALESCE order), so origin only affects the default `access_level`.

---

### What's working well

- `resolveSkillTypeForOriginName` using a post-ilike equality check is a good defensive pattern that correctly nullifies false positive wildcard matches.
- `pickStrongerSkillAccessLevel` correctly treats `'denied'` as a floor — neither input can demote it.
- Moving origin lookup to a single JOIN on fighter fetch (instead of a separate gang query) is a clean latency improvement.
- The SQL handles `v_gang_origin_id IS NULL` correctly (WHERE predicate on NULL never matches, leaving `v_origin_skill_type_id` as NULL).
- `CASE WHEN cs.skill_type_id = v_origin_skill_type_id` (not `cs.custom_skill_type_id`) correctly scopes origin grants to standard skill types only.
- The `gang.tsx` fix is correct: `gangOriginName` (state, updated optimistically at line 183/523) vs. `gang_origin_name` (prop, stale after save) — the displayed name now tracks the optimistic state as intended.

---

### Final verdict

**Changes requested** — the ilike wildcard issue is a real silent-failure edge case, and the unnecessary `DROP FUNCTION` is a deployment hazard. Both are small fixes. Everything else looks solid.


---

joeseos:

Reviewing commits `c8fa160e` and `5dd55ed6` against the previous CHANGES_REQUESTED review.

---

|---|---|---|---|
| 1 | **HIGH** | `pickStrongerSkillAccessLevel` treated `'denied'` like `null`, diverging from SQL COALESCE | ✅ **Resolved** |
| 2 | **Medium** | Two extra sequential DB queries unconditionally added per request | ✅ **Resolved** |
| 3 | **Low** | `findSkillTypeMatchingOriginName` exported but never used | ✅ **Resolved** |
| 4 | **Low** | `gangData` fetch error silently discarded | ✅ **Resolved** |
| 5 | **Low** | Custom skills with `custom_skill_type_id` excluded from origin access — intentional? | ✅ **Resolved** |
| 6 | **Medium** | `ilike` without LIKE-metacharacter escaping | ✅ **Resolved** |
| 7 | **Low** | Redundant `DROP FUNCTION IF EXISTS` before `CREATE OR REPLACE` | ✅ **Resolved** |
| 8 | **Low** | `override_access_level` precedence over origin grant undocumented | ✅ **Resolved** |

**Every previous finding has been addressed.** Brief notes on the resolutions:

- **#1** — `pickStrongerSkillAccessLevel` now returns `'denied'` when either input is `'denied'`, correctly matching the SQL `COALESCE` behaviour. Option A from the suggested fix was applied.
- **#2** — The gang origin is now fetched via a join in the single fighter query (`gangs:gang_id(gang_origin_id, gang_origin:gang_origin_id(origin_name))`), eliminating the separate `gangs` round-trip entirely.
- **#3** — `findSkillTypeMatchingOriginName` was renamed to `resolveSkillTypeForOriginName` with a narrower signature (single candidate, not an array) and is now imported and used by the route. No dead export remains.
- **#4** — The separate gang fetch is gone; `fighterError` is now logged before returning 404, and `skillTypeError` is logged before skipping origin access.
- **#5** — SQL comment added explicitly: *"Origin grants apply to standard skill_types only, not custom_skill_type_id"*.
- **#6** — `escapeForIlikePattern` added to the shared module and applied at the call site. The secondary `normaliseSkillSetName` equality check in `resolveSkillTypeForOriginName` is retained as a sanity guard.
- **#7** — `DROP FUNCTION IF EXISTS` removed from both the SQL source file and the migration; `CREATE OR REPLACE` alone now handles the in-place update.
- **#8** — JSDoc on `applyGangOriginSkillAccess` now explicitly documents that only `access_level` is upgraded and that `override_access_level` takes precedence, mirroring the SQL COALESCE order.

---

### New issues in the updated commits

None of substance. A few observations for completeness:

- **`fighter.gangs as { ... } | null` (route.ts ~line 196)** — the manual `as` cast is the practical approach given Supabase's generated types for nested joins; the optional-chaining on every access (`gangData?.gang_origin_id`, `gangData?.gang_origin?.origin_name`) means a runtime shape mismatch would just silently skip origin access rather than crash, which is the correct safe-default behaviour.

- **`resolveSkillTypeForOriginName` JSDoc** — After the escaping fix, the note about guarding against wildcard matches is slightly less critical than before (escaped ILIKE shouldn't produce false positives from metacharacters), but keeping the guard and the note is fine defensive coding.

- **Guard changed from `if (gangOriginName)` to `if (gangOriginId && gangOriginName)`** — the extra `gangOriginId` check is consistent with how the SQL function uses `v_gang_origin_id` and slightly reduces unnecessary `skill_types` lookups for fighters whose gangs have no origin. Good change.

---

### What's working well

- Collapsing the fighter + gang + gang_origin into a single query is a clean latency win with no readability cost.
- `escapeForIlikePattern` is correctly scoped to the shared module and documented with a concise one-liner.
- The alignment between SQL `COALESCE` semantics and `pickStrongerSkillAccessLevel` is now exact — `'denied'` blocks at both layers consistently.
- Migration file is clean: no `DROP`, pure `CREATE OR REPLACE`, idempotent and dependency-safe.

---

### Final verdict

**Approved.** All blockers and lower-severity issues from the prior round are resolved. The new commits introduce no new bugs or regressions. Ready to merge.

---

joeseos (approved):
All previous findings resolved. No new issues. Approving.

---

joeseos:
Hey @Djidiouf, a few notes on this PR:

**The `gang.tsx` fix is clean** — switching from the stale prop to the optimistic state variable is the right call.

**The skill-access changes work but have some DRY/YAGNI concerns:**

1. **Business logic duplication** — The "origin name matches a skill type → grant primary access" rule is now implemented independently in both the `get_available_skills` SQL RPC function *and* the TypeScript `/api/fighters/skill-access` route. If the rule ever changes (e.g. origin grants `secondary` instead of `primary`, or the matching logic changes), both need updating and could silently drift.

2. **`app/lib/shared/skill-access.ts` isn't actually shared** — It's only imported by the one API route. The `shared/` directory name is misleading and the extraction into five separate exported functions/constants (`pickStrongerSkillAccessLevel`, `normaliseSkillSetName`, `escapeForIlikePattern`, `resolveSkillTypeForOriginName`, `applyGangOriginSkillAccess`, `GANG_ORIGIN_SKILL_ACCESS_LEVEL`) is over-abstracted for a single call site. The ranking map and "pick stronger" generalisation anticipate a generality that isn't needed — the only comparison that actually happens is "null vs primary" (and "denied stays denied").

3. **Belt-and-suspenders on the ILIKE lookup** — The API route does `.ilike('name', escapeForIlikePattern(...))` then double-checks with `resolveSkillTypeForOriginName` (a normalised equality check). The SQL side just does `lower(trim(st.name)) = lower(trim(go.origin_name))` — cleaner and more direct. The TS side could follow the same pattern.

**What's good:**
- Folding the gang origin join into the initial fighter query (removing the separate `gangs` fetch) is a nice optimisation.
- The SQL COALESCE chain is logically sound — origin fills in only when override and fighter-type access are both null, and `denied` is respected.
- Scoping origin grants to standard `skill_type_id` only (not `custom_skill_type_id`) is correct edge-case handling.

None of these are blockers, but worth considering — especially point 1 around the duplicated business logic.

---

PR #1790, opened 2026-06-16
Updated the Select crew modal in battle mode to display more informat…

Djidiouf:
…ion for more informed decision

### DIFF
--- a/components/battle-session/crew-selection-modal.tsx
+++ b/components/battle-session/crew-selection-modal.tsx
@@ -51,7 +51,7 @@ function getFighterRowClass(f: GangFighterOption, beast: boolean, disabled: bool
     return `${base} cursor-pointer bg-muted hover:bg-accent/40 dark:hover:bg-accent/25`;
   }
   if (beast) {
-    return `${base} ml-6 cursor-default border border-dashed border-border bg-muted/30 dark:bg-neutral-800/50`;
+    return `${base} cursor-default border border-dashed border-border bg-muted/30 dark:bg-neutral-800/50`;
   }
   return `${base} cursor-default border border-dashed border-neutral-300/90 dark:border-neutral-600 bg-neutral-200/80 dark:bg-neutral-900/70`;
 }
@@ -354,9 +354,9 @@ export default function CrewSelectionModal({
       ) : (
         <div className="space-y-2">
           <div className="p-2 bg-muted rounded-md flex flex-col sm:flex-row sm:items-center gap-2 sm:gap-3">
-            <div className="flex items-center justify-between sm:justify-start sm:gap-4">
-              <div className="flex items-center gap-2">
-                <span className="text-sm text-muted-foreground">Pick</span>
+            <div className="flex flex-col gap-2 sm:w-full sm:flex-row sm:items-center sm:justify-between sm:gap-8">
+              <div className="flex items-center gap-1">
+                <span className="inline-block w-14 text-sm text-muted-foreground">Choose</span>
                 <Button variant="outline" size="icon" className="h-8 w-8" onClick={() => handlePickChange(pickCount - 1)}>
                   <LuMinus className="h-4 w-4" />
                 </Button>
@@ -365,8 +365,8 @@ export default function CrewSelectionModal({
                   <LuPlus className="h-4 w-4" />
                 </Button>
               </div>
-              <div className="flex items-center gap-2">
-                <span className="text-sm text-muted-foreground">Random</span>
+              <div className="flex items-center gap-1">
+                <span className="inline-block w-14 text-sm text-muted-foreground">Random</span>
                 <Button variant="outline" size="icon" className="h-8 w-8" onClick={() => handleRandomChange(randomCount - 1)}>
                   <LuMinus className="h-4 w-4" />
                 </Button>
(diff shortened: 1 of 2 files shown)


joeseos (changes requested):
## Code Review — PR #1790 · Update Select Crew UI in Battle Mode

Hey @Djidiouf — solid improvement overall. The refactor is clean and the new info density in both the crew modal and fighter table is a good UX win. A few things to verify before merge:

---

### Medium

**1. Beast rows get double indentation in the crew modal (`crew-selection-modal.tsx`)**

`getFighterRowClass` returns `ml-6` for beasts, AND the new `displayName` logic prepends `— ` to the name:
```ts
const displayName = beast && f.owner_id ? `— ${f.fighter_name}` : f.fighter_name;
```
So beasts now render with *both* `ml-6` margin *and* a `— ` dash prefix, which feels like double-dipping. The participant card table (which this PR also touches) only uses the `— ` prefix — no `ml-6`. The two views are now visually inconsistent.

Suggestion: drop `ml-6` from the beast branch in `getFighterRowClass` if you're relying on the `— ` prefix, or remove the dash from `displayName` and keep only the margin. Pick one signal.

---

**2. `loadoutName` silently removed from the second row in `FighterRow` (`participant-card.tsx`)**

Old second-row format: `Ganger - My Loadout (Exotic Beast Specialist)`  
New second-row format: `Ganger (Exotic Beast Specialist)` — loadout name moved to first row as `[My Loadout]`

```diff
- [fighterType, loadoutName].filter(Boolean).join(' - '),
+ fighterType,
```

This is a visible behaviour change for any gang using named loadouts (profiles). The loadout is now in the first row alongside the fighter name, so it's not *lost*, but users who rely on the second row to skim loadout names will notice. Confirm this is intentional — if so, a note in the PR description would help future reviewers.

---

**3. New "Value" column is always rendered, not just in battle mode (`participant-card.tsx`)**

```tsx
<th className="p-1 md:p-2 text-right font-medium whitespace-nowrap">Value</th>
```

This column shows for every state (pre-battle, battle, post-battle). Previously cost was inline in the Fighter cell only. For the non-battle / read-only view this adds a column that never changes and takes screen real estate on mobile. The `colSpan` update is correct (`3`/`2`), but it's worth deciding if this column should be conditional the same way the Actions/Status column is.

---

### Low

**4. `label` field on `GangFighter` type is assumed but not shown**

```tsx
// participant-card.tsx ~925
label: gf.label,
// fighter-row JSX
{gangFighter?.label && <span ...>{gangFighter.label}</span>}
```

The diff doesn't show changes to `@/app/lib/shared/gang-data`'s `GangFighter` type. If `label` and `fighter_type` are already there (likely, given `fighter_type` was already read in the old `FighterRow`), no problem. Just confirm TypeScript is clean — `tsc --noEmit` should catch it if not.

**5. Status icons have no accessible label**

Icons like `<IoSkull />`, `<MdChair />`, `<GiCrossedChains />` are now rendered in two places (crew modal + fighter row) with no `title` or `aria-label`:
```tsx
{f.killed && <IoSkull className="text-gray-300" />}
```
Pre-existing issue, but now more widespread. A `title="Killed"` / `title="In Recovery"` on each icon would make these readable on hover and for screen readers.

**6. Minor DRY opportunity**

The `fighterDetails` formatting is nearly identical in both files:

`crew-selection-modal.tsx`:
```ts
function formatFighterDetails(f: GangFighterOption): string {
  return [f.fighter_type, f.fighter_class ? `(${f.fighter_class})` : ''].filter(Boolean).join(' ');
}
```
`participant-card.tsx` (inside `FighterRow`):
```ts
const fighterDetails = [fighterType, fighterClass ? `(${fighterClass})` : ''].filter(Boolean).join(' ');
```

Not a blocker — they live in different files and have different signatures — but if this pattern grows, a shared util would be worth it.

---

### What's Good

- Extracting `getFighterRowClass` into a pure function is a clean improvement over the inline ternary chain.
- Extracting `disabled` into a local const (`const disabled = isDisabled(f)`) before using it in both the class helper and the JSX avoids calling `isDisabled` twice — good catch.
- The `colSpan` update for the empty state (`3`/`2`) is correctly handled.
- Beast `displayName` with `— ` prefix now matches the participant-card table's naming convention for associated exotic beasts — nice consistency.
- Moving status icons (skull, chair, chains, etc.) into the first info row is a clear UX win.

---

### Verdict

**Changes requested** — primarily item #1 (beast double-indentation is a likely visual bug) and a confirmation on #2 (loadout-name placement change). Items #3–6 are low-risk but worth a quick look.

---

joeseos:
## Follow-up Review — PR #1790 (commit `64b21fa`)

Hey @Djidiouf — quick follow-up on the updated commit. Two of the previous findings are resolved, the rest are still open.

---

### Previous Findings Status

| # | Finding | Status |
|---|---------|--------|
| 1 | Beast double-indentation (`ml-6` + `— ` dash prefix) | ✅ **Resolved** — `ml-6` removed from `getFighterRowClass`, dash prefix is the single visual signal now |
| 2 | `loadoutName` removed from second row in `FighterRow` | ⚠️ **Outstanding** — still not confirmed as intentional |
| 3 | "Value" column always rendered (not conditional on battle mode) | ⚠️ **Outstanding** — not addressed |
| 4 | TypeScript safety of `label` / `fighter_type` on `GangFighter` type | ⚠️ **Outstanding** — no type file changes in diff |
| 5 | Status icons missing `title` / `aria-label` | ✅ **Resolved** — all icons in both files now have `title` and `aria-label` |
| 6 | DRY opportunity for `formatFighterDetails` | ⚠️ **Outstanding** — flagged as non-blocker, still duplicated |

---

### New Changes (commit `64b21fa`) — No New Issues

The Roll button relocation (moved inside the Random control group) and the "Pick → Choose" label rename are clean UX changes. The `sm:ml-6` / `ml-auto` on the Roll button works correctly on both mobile (right-aligns it within the Random row) and desktop (spaced by `sm:gap-8` + `sm:justify-between`). No regressions spotted.

---

### Still Blocking

**Finding #2 — confirm loadout name intent**
The loadout name moved from the second row (`Ganger - My Loadout`) to the first row (`Ganger [My Loadout]`). This is not a bug but it is a visible behavioural change for gangs with named loadouts. A one-liner in the PR description confirming this is intentional would close it out cleanly.

**Finding #3 — Value column visibility**
```tsx
// participant-card.tsx
<th className="p-1 md:p-2 text-right font-medium whitespace-nowrap">Value</th>
```
This is unconditional, so it shows in the read-only / pre-battle view too. The Actions/Status column has the guard `{(canInteract || battleActive) && ...}`. Worth deciding if Value should have a similar guard, or if showing cost everywhere is the intent — either is valid but it should be a deliberate choice.

---

### Verdict

**Changes requested** — same call as before. #1 and #5 are cleanly fixed; the remaining two medium items (#2, #3) just need a quick confirmation or a small code change. Once those are addressed this is good to merge.

---

PR #1797, opened 2026-06-18
Refactor card view mode to be responsive and work on all screens with…

Djidiouf:
…out having to modify the zoom of the browser

### DIFF
--- a/components/gang/ViewModeDropdown.tsx
+++ b/components/gang/ViewModeDropdown.tsx
@@ -1,10 +1,30 @@
 import React from 'react';
 
-type ViewMode = 'normal' | 'small' | 'medium' | 'large';
+export const GANG_PAGE_VIEW_MODES = ['normal', '2-card', '3-card', '4-card'] as const;
+export type GangPageViewMode = (typeof GANG_PAGE_VIEW_MODES)[number];
+
+/** Gang page grid modes plus print-only mode for /gang/[id]/print */
+export type GangViewMode = GangPageViewMode | 'print';
+
+export const PRINT_GANG_VIEW_MODE = 'print' as const;
+
+export function isGangViewMode(value: string | null): value is GangPageViewMode {
+  return value !== null && (GANG_PAGE_VIEW_MODES as readonly string[]).includes(value);
+}
+
+export function isCompactGangViewMode(
+  viewMode: GangViewMode
+): viewMode is Exclude<GangPageViewMode, 'normal'> {
+  return viewMode !== 'normal' && viewMode !== 'print';
+}
+
+export function isFullSizeGangViewMode(viewMode?: GangViewMode): boolean {
+  return viewMode === undefined || viewMode === 'normal' || viewMode === 'print';
+}
 
 interface ViewModeDropdownProps {
-  viewMode: ViewMode;
-  setViewMode: (mode: ViewMode) => void;
+  viewMode: GangPageViewMode;
+  setViewMode: (mode: GangPageViewMode) => void;
   className?: string;
 }
 
@@ -12,13 +32,13 @@ export const ViewModeDropdown: React.FC<ViewModeDropdownProps> = ({ viewMode, se
   <div className={`${className} print:hidden`}>
     <select
       value={viewMode}
-      onChange={(e) => setViewMode(e.target.value as ViewMode)}
+      onChange={(e) => setViewMode(e.target.value as GangPageViewMode)}
       className="w-full h-10 p-1 border rounded-md border-border focus:outline-hidden focus:ring-2 focus:ring-black text-sm"
     >
       <option value="normal">Page View</option>
-      <option value="small">Small Cards</option>
-      <option value="medium">Medium Cards</option>
-      <option value="large">Large Cards</option>
+      <option value="2-card">2-card View</option>
+      <option value="3-card">3-card View</option>
+      <option value="4-card">4-card View</option>
     </select>
   </div>
 );
--- a/components/gang/ViewModeDropdown.tsx
+++ b/components/gang/ViewModeDropdown.tsx
@@ -8,7 +8,7 @@ export type GangViewMode = GangPageViewMode | 'print';
 
 export const PRINT_GANG_VIEW_MODE = 'print' as const;
 
-export function isGangViewMode(value: string | null): value is GangPageViewMode {
+export function isGangPageViewMode(value: string | null): value is GangPageViewMode {
   return value !== null && (GANG_PAGE_VIEW_MODES as readonly string[]).includes(value);
 }
 
--- a/components/gang/draggable-fighters.tsx
+++ b/components/gang/draggable-fighters.tsx
@@ -5,6 +5,7 @@ import { useDndSensorsConfig } from '@/hooks/use-dnd-sensors';
 import { useIsMounted } from '@/hooks/use-is-mounted';
 import { MyFighters } from './my-fighters';
 import { FighterProps } from '@/types/fighter';
+import { GangPageViewMode } from './ViewModeDropdown';
 import { UserPermissions } from '@/types/user-permissions';
 
 interface DraggableFightersProps {
@@ -12,7 +13,7 @@ interface DraggableFightersProps {
   onPositionsUpdate?: (positions: Record<number, string>) => void;
   onFightersReorder?: (newFighters: FighterProps[]) => void;
   initialPositions: Record<number, string>;
-  viewMode?: 'normal' | 'small' | 'medium' | 'large';
+  viewMode?: GangPageViewMode;
   userPermissions?: UserPermissions;
 }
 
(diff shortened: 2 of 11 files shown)


joeseos (changes requested):
## Code Review — PR #1797 · @Djidiouf

Good overall direction. Replacing fixed-pixel card sizes with a viewport-relative CSS-transform scale is a solid approach for responsive card grids, and the type system clean-up (centralising `GangViewMode` / `GangPageViewMode` in one file) is a clear improvement. A few issues need addressing before merge.

---

### 🔴 Medium — `next.config.js`: personal LAN IP committed to production config

```js
allowedDevOrigins: ['192.168.1.108'],   // line 5
```

`allowedDevOrigins` is a Next.js development-only key. Committing your LAN IP here means it ships to production on every deployment. It's unnecessary there and leaks network topology info. Either move it to a `next.config.local.js` (gitignored) or remove it from this PR entirely.

---

### 🔴 Medium — `hooks/use-viewport-width.ts`: SSR/client hydration mismatch

```ts
// line 33-35
const [width, setWidth] = useState(() =>
  typeof window !== 'undefined' ? window.innerWidth : DEFAULT_WIDTH
);
```

During SSR the server renders with `DEFAULT_WIDTH` (800). On client hydration the `useState` lazy initialiser runs again with the real `window.innerWidth` (e.g. 1440). React will detect this as a hydration mismatch and may log errors in strict mode. The safe pattern is:

```ts
const [width, setWidth] = useState(DEFAULT_WIDTH);   // consistent on server & client

useEffect(() => {
  setWidth(window.innerWidth);                        // update after hydration
  subscribeToViewportResize();
  const listener = () => setWidth(viewportWidth);
  listeners.add(listener);
  return () => { listeners.delete(listener); };
}, []);
```

---

### 🔴 Medium — `hooks/use-viewport-width.ts`: module-level event listeners are never cleaned up

```ts
// lines 15–29
let isSubscribed = false;

function subscribeToViewportResize() {
  if (isSubscribed || ...) return;
  isSubscribed = true;
  window.addEventListener('resize', handleResize);
  window.addEventListener('orientationchange', publishWidth);
}
```

`isSubscribed` is only ever set to `true` — there is no path that removes the event listeners. For a browser SPA this is mostly fine in production, but it breaks unit/integration tests that re-mount the module in the same process (e.g. Jest with jsdom), and it would prevent resubscription if the module is ever re-evaluated. Add an `unsubscribe` export or a cleanup path to make the module testable.

---

### 🟡 Low — `fighter-card.tsx`: constants defined inside a `memo`ised component body

```tsx
// ~line 178
const COMPACT_SCALE_PERCENT = {
  '4-card': 32,
  '3-card': 42,
  '2-card': 65,
} as const;
const COMPACT_DESIGN_WIDTH = 800;
```

These objects are recreated on every render (defeating part of the value of `memo`). Move them to module scope — they never depend on props or state.

---

### 🟡 Low — `fighter-card.tsx`: dead alias `clickableContent`

```tsx
// ~line 852
const clickableContent = cardContent;   // always the same; never conditionally different
```

After the refactor `clickableContent` is always `cardContent`. The comment below it also refers to the old conditional logic that no longer applies. Remove the alias and use `cardContent` directly in the `return`.

---

### 🟡 Low — `fighter-card-stats-table.tsx`: `isNormalView` and `isFullSizeView` serve overlapping roles

```ts
// lines 74–76
const isFullSizeView = isFullSizeGangViewMode(viewMode);  // true for undefined | 'normal' | 'print'
const isNormalView = viewMode === 'normal';               // only used for pClass
const pClass = isNormalView ? 'p-1' : 'p-px';
```

`pClass` uses `isNormalView` while most other conditionals use `isFullSizeView`, meaning `print` mode gets `p-px` (compact padding) even though `isFullSizeView` considers it full-size. If that's intentional, a short comment would clarify it. If it isn't intentional, `pClass` should use `isFullSizeView` to be consistent.

---

### 🟡 Low — `ViewModeDropdown.tsx`: `isGangViewMode` validates `GangPageViewMode`, not `GangViewMode`

```ts
// line 12
export function isGangViewMode(value: string | null): value is GangPageViewMode {
  // ...only checks GANG_PAGE_VIEW_MODES — 'print' is NOT included
}
```

The function name says `GangViewMode` but the type guard returns `GangPageViewMode` (excludes `'print'`). This is correct behaviour for the localStorage guard in `gang.tsx`, but the name is misleading. Consider renaming to `isGangPageViewMode` to match the return type.

---

The `allowedDevOrigins` commit and the hydration mismatch are the blockers. Everything else is low-risk cleanup. The core responsive-scale approach and the type-system consolidation are both solid — just clean up the items above and this is good to merge.

---

joeseos (commented):
## Follow-up Review — PR #1797 · @Djidiouf

Reviewing only the second commit (`a399d86`, +51 / -30 across 6 files) against the seven findings from the previous round.

---

| # | File | Finding | Status |
|---|------|---------|--------|
| 1 | `next.config.js` | LAN IP committed (`192.168.1.108`) | ✅ **Resolved** — replaced with `[]` |
| 2 | `use-viewport-width.ts` | SSR hydration mismatch in `useState` lazy initialiser | ✅ **Resolved** — `useState(DEFAULT_WIDTH)` + `setWidth(window.innerWidth)` in effect |
| 3 | `use-viewport-width.ts` | Module-level listeners never cleaned up | ✅ **Resolved** — reference-counted `subscribe`/`unsubscribe` pair added |
| 4 | `fighter-card.tsx` | `COMPACT_SCALE_PERCENT` / `COMPACT_DESIGN_WIDTH` inside memo'd body | ✅ **Resolved** — hoisted to module scope |
| 5 | `fighter-card.tsx` | Dead alias `clickableContent = cardContent` | ✅ **Resolved** — replaced with clear `cardBody` / `cardWithOptionalLink` / `cardContent` hierarchy |
| 6 | `fighter-card-stats-table.tsx` | `pClass` using `isNormalView` while other conditionals use `isFullSizeView` | ✅ **Resolved** — intent clarified with inline comment; `isFullSizeView` now correctly drives toughness-header and label spans |
| 7 | `ViewModeDropdown.tsx` | `isGangViewMode` name didn't match its `GangPageViewMode` return type | ✅ **Resolved** — renamed to `isGangPageViewMode` |

All seven previous findings addressed. 

---

### New findings in `a399d86`

#### 🔴 Medium — `fighter-card.tsx`: `isCompactGangViewMode` called with a possibly-`undefined` argument

```tsx
// fighter-card.tsx  (inside FighterCard, viewMode?: GangViewMode)
const isCompactView = isCompactGangViewMode(viewMode);   // ← viewMode can be undefined
```

`isCompactGangViewMode` is declared as `(viewMode: GangViewMode)` — it does **not** accept `undefined`. Two consequences:

1. **TypeScript compile error** under `strictNullChecks` (the default for a Next.js project): `Argument of type 'GangViewMode | undefined' is not assignable to parameter of type 'GangViewMode'`.
2. **Silent runtime bug** if strict checks are off: `undefined !== 'normal' && undefined !== 'print'` evaluates to `true`, so `isCompactView = true` and `isTightLayoutView = true`. The card then applies `absolute inset-0` layout with no defined dimensions, breaking the render.

Fix:
```tsx
const isCompactView = viewMode !== undefined && isCompactGangViewMode(viewMode);
```

---

#### 🟡 Low — `next.config.js`: misleading comment on the still-committed `allowedDevOrigins` key

```js
allowedDevOrigins: [],   // Local development IP address
```

The IP is gone (good), but the key is still checked in with a comment that says "Local development IP address" next to an empty array — that's confusing and signals to the next dev to add their own IP directly in the repo. The key has no effect in production when empty, so this is low-risk, but cleaner options are:

- Remove the line entirely (simplest), or
- Drive it from an env var: `allowedDevOrigins: process.env.ALLOWED_DEV_ORIGIN ? [process.env.ALLOWED_DEV_ORIGIN] : []`

---

#### 🟡 Low — `my-fighters.tsx`: stale `useEffect` import

```tsx
import React, { useMemo, useEffect } from 'react';
```

`useEffect` is imported but no longer used after the refactor. Will produce a lint warning; safe to remove.

---

#### 🟡 Low — `fighter-card.tsx`: conflicting `gap-y-*` utilities in `detailsGridClass`

```tsx
const detailsGridClass = `grid gap-y-2 mt-3 ... ${viewMode === '4-card' ? 'gap-y-[2px]' : ''} ...`;
```

When `viewMode === '4-card'`, both `gap-y-2` and `gap-y-[2px]` appear in the class string. In Tailwind, the winner is determined by the order in the **generated stylesheet** (not the string), which can be surprising. Use a ternary to avoid the ambiguity:

```tsx
const gapClass = viewMode === '4-card' ? 'gap-y-[2px]' : 'gap-y-2';
const detailsGridClass = `grid ${gapClass} mt-3 ...`;
```

---

### Verdict: **Approve with suggestions**

The three blocking issues from the previous round are cleanly resolved, and the structural refactors (CSS-transform scale, responsive grid, type consolidation, menu-trigger extraction) are solid. The medium `isCompactGangViewMode(undefined)` issue is worth a one-liner fix before merge — everything else is cleanup.

---

PR #1799, opened 2026-06-18
Feature/effect selection add fighter

joeseos:


joeseos (commented):
## Code Review — PR #1799

This PR adds a multi-step effect-selection modal that fires between "Add Fighter" and the mutation when a user picks editable equipment with `single_select` / `multiple_select` effects. The queue/step architecture is clean. There are two issues that need fixing before merge — one security, one functional regression — plus a few lower-priority things.

---

### 🔴 HIGH — `app/actions/add-fighter.ts` ~L869: `effect_ids` from client applied without server-side validation

```ts
const { data: selectedEffectTypes } = await supabase
  .from('fighter_effect_types')
  .select(`id, effect_name, ...`)
  .in('id', equipItem.effect_ids);        // ← client-supplied IDs, no constraint

// …then unconditionally applied to the fighter
await applyEffectsForEquipmentOptimized(supabase, selectedEffectTypes, ...);
```

`equipItem.effect_ids` flows in from `params` (i.e., the client POST body). The query fetches any `fighter_effect_types` row that matches those IDs — it does **not** verify those effects actually belong to the equipment being purchased. An authenticated user who discovers any `fighter_effect_type` UUID can inject it here and have arbitrary effects applied to their fighter (boosted stats, bypassed restrictions, etc.).

**Fix:** add an `equipment_id` constraint to the query so only effects that genuinely belong to this equipment can be applied:

```ts
const { data: selectedEffectTypes } = await supabase
  .from('fighter_effect_types')
  .select(`id, effect_name, ...`)
  .in('id', equipItem.effect_ids)
  .filter("type_specific_data->>'equipment_id'", 'eq', equipItem.equipment_id);
```

Or cross-reference against the effects already fetched during the fixed-effects pass (they're in `effectsByEquipment`) to avoid the extra round-trip entirely.

---

### 🔴 MEDIUM — `components/gang/add-fighter.tsx` ~L1019: equipment deselection regression

```tsx
// OLD — only removes items that were previously selected FROM this category
const previouslySelectedInThisCategory = selectedEquipmentIds.filter(id =>
  currentCategoryOptions.some((o: any) => `${categoryId}-${o.id}` === id)
);
let filtered = prev.filter(item => {
  const wasSelectedFromThisCategory = previouslySelectedInThisCategory.some(selectedId => {
    const equipmentIdFromSelected = selectedId.split('-').pop();
    return equipmentIdFromSelected === item.equipment_id;
  });
  return !wasSelectedFromThisCategory;
});

// NEW — removes any item whose equipment_id appears in this category's options
let filtered = prev.filter(item =>
  !currentCategoryOptions.some((o: any) => o.id === item.equipment_id)
);
```

The old approach used the compound `${categoryId}-${o.id}` key to track which category each selection came from. The new approach simply matches on `equipment_id`. If the same equipment ID appears in two different category option lists (e.g., once as a selectable option and once as a default/replacement in another category), the new code incorrectly removes the selection when the user interacts with the second category, even though they picked it from the first.

Notably, the cost-update logic in `setFighterCost` immediately below still uses the old compound-key pattern via `selectedEquipmentIds`, so the two state slices are now semantically inconsistent about what "selected from this category" means.

---

### 🟡 MEDIUM — `components/gang/add-fighter.tsx` ~L1243: default equipment with selectable effects is silently skipped

```tsx
const editableEquipment = selectedEquipment.filter(item => item.is_editable);
// default_equipment is never checked for selectable effects
```

The server action correctly excludes `single_select` / `multiple_select` effects from auto-apply for **all** equipment, but the client-side queue builder only looks at `selectedEquipment`. If a fighter's `default_equipment` item has `single_select` or `multiple_select` effects, those effects are silently dropped — the server skips them, and the UI never asks the user to pick. The fighter ends up with the equipment but none of its selectable effects applied.

---

### 🟡 MEDIUM — `supabase/functions/get_add_fighter_details.sql` L1: N+1 correlated calls per equipment row

```sql
'effects', CASE WHEN COALESCE(e.is_editable, false)
           THEN get_equipment_effects_jsonb(e.id)   -- called once per equipment row
           ELSE '[]'::jsonb END
```

`get_equipment_effects_jsonb` is a helper function that itself issues a correlated subquery for modifiers per effect row. It's called inside a `jsonb_agg` that iterates every equipment row — so this is effectively O(equipment_items × effects_per_item) queries per call to `get_add_fighter_details`. The `STABLE` marker allows the planner to cache calls within one query for duplicate `e.id` values, but for a fighter type with many distinct editable items this will degrade.

A single lateral join or a CTE that pre-aggregates effects per equipment ID before the outer build would eliminate the N+1 pattern.

---

### 🟡 LOW — `components/gang/add-fighter.tsx` ~L1310: `findEquipmentInFighterType` declared after `handleAddFighter` which uses it

Both are arrow functions assigned to `const`. Arrow-function `const`s are not hoisted, so order matters for readability even though the runtime is safe (the click handler only fires after both are initialized in the same render). Move `findEquipmentInFighterType` above `handleAddFighter`.

---

### 🟡 LOW — `components/gang/add-fighter.tsx` ~L270: `pendingMutationParams` typed as `any`

```tsx
const [pendingMutationParams, setPendingMutationParams] = useState<any>(null);
```

The mutation already has a typed params shape. Extract that type or inline it here so the params object doesn't lose type safety during the queued effect-selection flow.

---

### 🟡 LOW — `components/gang/add-fighter.tsx` ~L279: unused methods in imperative handle ref type

```tsx
const effectSelectionRef = useRef<{
  handleConfirm: () => Promise<boolean>;
  isValid: () => boolean;              // never called from this file
  getSelectedEffects: () => string[];  // never called from this file
} | null>(null);
```

Only `handleConfirm()` is actually invoked via the ref in this component. Trim the type to what's used.

---

### 🟡 LOW — `app/api/fighter-types/route.ts` L328: trailing whitespace on added line

Line 328 adds a space-only line between the trailing comma and the closing `}))`. Minor cleanup.

---

The `effect_ids` validation gap is a security blocker — users can inject arbitrary effect IDs and apply them to their fighters. The equipment deselection regression is a functional blocker. Please address those two before merge; the rest are cleanup items.

---

PR #1800, opened 2026-06-20
Fixed real time issue in post-battle in battle sessions

joeseos:


joeseos (commented):
## Code Review — PR #1800 · Fix real-time issue in post-battle

Hey @joeseos — the fix is correct and well-scoped. **Verdict: approve** with a few things worth keeping in mind.

---

### What the fix does

Two independent problems were causing the real-time failure when completing a battle:

1. `CompleteBattleModal.handleConfirm` never called `onBroadcast`, so other clients' Supabase realtime subscriptions were never notified the session had changed.
2. Even if clients were notified, `fetchBattleSession` (the server action powering React Query refetches) was hitting `getBattleSessionCached` — so refetches still returned stale cached data.

Both are fixed: the broadcast + cache invalidation now fire on success, and the server action bypasses the cache.

---

### Medium

**1. `unstable_cache` still captures `supabase` as a closure — pre-existing but now more visible**

```ts
// get-battle-session-data.ts
export const getBattleSessionCached = async (sessionId: string, supabase: any) => {
  return unstable_cache(
    () => fetchBattleSessionDirect(sessionId, supabase),  // ← supabase in closure
    [`battle-session-${sessionId}`],
    { tags: [...], revalidate: false }
  )();
};
```

`unstable_cache` keys on `[battle-session-${sessionId}]` only. The first request to prime the cache for a session ID "wins" — all subsequent cache hits use that cached result regardless of which Supabase client (auth context) later callers pass in. The old code had the same shape inline, but extracting `fetchBattleSessionDirect` as a named function makes the closure more visible now.

In practice this is fine because all participants share the same view of the session (no RLS variance per user), and the cache is invalidated on every mutation via `revalidateTag`. But the `supabase` parameter to `getBattleSessionCached` is effectively ignored on cache hits — worth documenting or being aware of if RLS ever tightens on this data.

**2. Verify the TanStack Query key format matches the registration**

```tsx
// active-session.tsx
queryClient.invalidateQueries({ queryKey: ['battle-session', session.id] });
```

This invalidation only works if the query is registered with the exact same key shape. If the active query uses a different key format (e.g. a nested array, object, or different string), the invalidation silently does nothing. Worth a quick grep to confirm it matches wherever `useQuery` or `useInfiniteQuery` registers the battle session data.

---

### Low

**3. `supabase: any` propagated into `fetchBattleSessionDirect`**

Pre-existing issue, but the new function is a named export and permanent part of the module's API surface. If the Supabase server client type is available (e.g. `Awaited<ReturnType<typeof createClient>>` or from `@supabase/ssr`), typing it would prevent subtle callsite errors down the line.

**4. Useful comment dropped during extraction**

```diff
-// Lightweight fighter enrichment — only fetch name and base credits
  let fighterMap = new Map<string, any>();
```

This explained why only `id, fighter_name, credits, special_rules` are selected from `fighters` rather than `*`. Worth preserving in `fetchBattleSessionDirect`.

**5. `getBattleSessionCached` is now a thin wrapper**

Now that `fetchBattleSessionDirect` is standalone, `getBattleSessionCached` is just an `unstable_cache` shim around it. No action needed now, but if the SSR page ever moves to tag-based invalidation without `unstable_cache`, this wrapper can be eliminated entirely.

---

### What's Good

- The two-part fix (missing broadcast + stale cache bypass) correctly addresses both root causes independently.
- `onBroadcast` fires between `toast.success` and `return true` (which triggers modal close) — remote clients get the signal before local UI tears down. Good ordering.
- Extracting `fetchBattleSessionDirect` improves testability — the DB fetch logic can now be called directly without the `unstable_cache` wrapper.
- `onBroadcast` is correctly optional (`onBroadcast?.()`) — `CompleteBattleModal` remains usable without a broadcast channel.
- All error paths in `handleConfirm` (invalid winner, duplicate winner, API failures) `return false` before reaching `onBroadcast`, so the broadcast only fires on genuine success.

---

PR #1801, opened 2026-06-20
Add banned equipment feature to custom trading posts

joeseos:
Allow custom trading post creators to ban specific equipment items.
Banned items appear greyed out with a "Banned" badge in the equipment
shop and cannot be purchased. The ban is stored as a boolean column
on custom_trading_post_equipment and propagated through the SQL
function, server actions, and UI layers.

### DIFF
--- a/app/actions/customise/custom-trading-posts.ts
+++ b/app/actions/customise/custom-trading-posts.ts
@@ -187,7 +187,7 @@ export async function deleteCustomTradingPost(
 const TP_EQUIPMENT_SELECT = `
   id, custom_trading_post_id, equipment_id, custom_equipment_id,
   cost_override, cost_type_resource_id, cost_campaign_resource_id, cost_reputation, cost_resource_amount,
-  availability_override, sort_order,
+  availability_override, sort_order, banned,
   equipment:equipment_id (equipment_name, equipment_category),
   custom_equipment:custom_equipment_id (equipment_name, equipment_category)
 `;
@@ -208,6 +208,7 @@ function mapTPEquipmentRow(row: any): CustomTPEquipment {
     cost_resource_amount: row.cost_resource_amount ?? null,
     availability_override: row.availability_override,
     sort_order: row.sort_order,
+    banned: row.banned ?? false,
   };
 }
 
@@ -241,6 +242,7 @@ export interface CustomTPEquipment {
   cost_resource_amount: number | null;
   availability_override: string | null;
   sort_order: number | null;
+  banned: boolean;
 }
 
 export interface CustomTPAvailabilityRule {
@@ -370,6 +372,7 @@ export async function updateTPEquipment(
     cost_resource_amount?: number | null;
     availability_override?: string | null;
     sort_order?: number | null;
+    banned?: boolean;
   }
 ): Promise<{ success: boolean; error?: string }> {
   try {
--- a/components/customise/custom-trading-posts.tsx
+++ b/components/customise/custom-trading-posts.tsx
@@ -9,6 +9,7 @@ import { Label } from '@/components/ui/label';
 import { Button } from '@/components/ui/button';
 import { toast } from 'sonner';
 import { Combobox } from '@/components/ui/combobox';
+import { Checkbox } from '@/components/ui/checkbox';
 import { LuEye, LuSquarePen, LuTrash2 } from 'react-icons/lu';
 import { FiShare2 } from 'react-icons/fi';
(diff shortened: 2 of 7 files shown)


joeseos (commented):
## Code Review — PR #1801 · Add banned equipment feature to custom trading posts

Clean, well-layered feature. DB → SQL function → server action → UI is consistent throughout. A few things worth addressing before merge.

---

### High

**1. Ban is client-side only — no server-side enforcement on purchase**

The Buy button is disabled and guarded in `equipment.tsx`, but the actual purchase server action doesn't appear to check `banned`. A user could call the buy action directly (via API/devtools) with a banned item's ID and bypass the restriction entirely.

The `get_equipment_detailed_data` SQL function returns `banned` correctly, but that's only used for display. The purchase flow needs to verify the item isn't banned in any active custom TP for the campaign before completing the transaction. The check would need to query `custom_trading_post_equipment` for the item + active custom TP IDs and reject if any row has `banned = true`.

---

### Medium

**2. `bool_or` semantics: any-ban-wins across all active custom TPs**

In the `custom_tp_override` CTE and the custom equipment subquery:
```sql
bool_or(ctpe.banned) AS banned
```

If a campaign uses two custom TPs and item X appears in both — banned in TP A, not banned in TP B — `bool_or` returns `true` and the item is banned everywhere. The OR wins.

This is arguably the right behaviour (strict: any active TP's ban applies), but it's worth documenting explicitly. The inverse (`bool_and`, all TPs must ban) is also a reasonable design. If the intent is that each TP's ban is independent, this is correct. If the intent is "the item is only banned if the primary/only active TP bans it," `bool_or` would produce surprising results in multi-TP campaigns.

**3. New items added via `addTPEquipmentBatch` cannot be banned on creation**

`addTPEquipmentBatch` doesn't accept a `banned` parameter — newly inserted rows get `banned = false` from the column default. If a user opens the edit modal on a not-yet-saved item and ticks "Banned" before the batch save, that pending state is reflected in `displayItems` via `pending?.banned ?? false`, but it's never sent to the server because the batch insert doesn't include `banned`.

The workaround is: add the item first, then edit it to ban it — two round trips. Acceptable for now, but worth noting. A follow-up could add `banned` to `addTPEquipmentBatch`'s row payload.

---

### Low

**4. Double guard on Buy button is redundant**

```tsx
<Button
  disabled={item.banned}
  onClick={(e) => {
    e.stopPropagation();
    if (!item.banned) setBuyModalData(item);  // ← unreachable when banned
  }}
>
```

`disabled={item.banned}` prevents the click event from firing (Shadcn Button forwards `disabled` to the underlying `<button>` element). The `if (!item.banned)` guard inside `onClick` is dead code when the button is disabled. Harmless, but could be removed for clarity.

**5. `updateTPEquipment` only calls `revalidatePath('/')` — pre-existing**

When `banned` is toggled, `revalidatePath('/')` is called but no equipment-specific cache tags are invalidated. If equipment data is cached by tag elsewhere (e.g. per-gang equipment queries), the ban change might not surface immediately. Pre-existing pattern with the rest of `updateTPEquipment`, so not introduced by this PR — but worth revisiting if stale-ban reports come in.

---

### What's Good

- **Migration is correct**: `NOT NULL DEFAULT false` is the right constraint — no nullability ambiguity in the application layer, existing rows get a safe default.
- **SQL is structurally sound**: `banned` is added as the final column in both arms of the `UNION ALL` in `get_equipment_detailed_data`, matching the `RETURNS TABLE` declaration. `COALESCE(x, false)` correctly handles NULL from left-joined CTEs when equipment has no custom TP entry.
- **`copy_custom_collection.sql` correctly propagates `banned`**: copying a TP preserves ban flags, which is the right default (copy the restrictions, adjust later if needed).
- **`mapTPEquipmentRow` uses `?? false`**: `banned: row.banned ?? false` is correct since the column is NOT NULL, but the `?? false` guard is safe insurance against old rows or serialization edge cases.
- **Admin table UX is clear**: `opacity-50` row + "Yes"/"-" cell + `opacity-40 grayscale` in the shop gives distinct visual treatment at both the creator and buyer levels.
- **Checkbox pattern in `EditEquipmentModal` is correct**: `onCheckedChange={(checked) => setIsBanned(checked === true)}` correctly handles the Radix UI `checked: boolean | 'indeterminate'` union type.

---

### Verdict

**Changes requested** — item #1 (server-side enforcement) is the main blocker. Items #2–3 are design decisions worth explicitly confirming; #4–5 are clean-ups.

---

PR #1803, opened 2026-06-21
Fix scrolling of non active layers not being disabled

Djidiouf:


joeseos (changes requested):
## PR Review — Fix scrolling of non-active layers not being disabled

@Djidiouf Thanks for the PR! Fixing background scroll while a modal is open is a genuine UX improvement — the technique chosen (fixed-body scroll lock) is the most reliable cross-browser approach for Next.js apps. The implementation is largely solid. I have one concrete bug to flag and a few design-level concerns worth discussing before merge.

---

### Findings

#### 🔴 HIGH — `paddingRight` compensation overwrites, not augments, existing body padding

**File:** `hooks/use-scroll-lock.ts`, line 38

```typescript
if (scrollbarWidth > 0) {
  document.body.style.paddingRight = `${scrollbarWidth}px`;
}
```

If `document.body` already has a `paddingRight` (e.g. from a theme or layout layer), this replaces it with just the scrollbar width instead of adding to it. This causes a visible layout shift when the modal opens.

**Fix:**
```typescript
if (scrollbarWidth > 0) {
  const existingPadding = parseFloat(savedBodyStyles.paddingRight) || 0;
  document.body.style.paddingRight = `${existingPadding + scrollbarWidth}px`;
}
```

The `savedBodyStyles.paddingRight` is already captured above this line, so no extra read is needed.

---

#### 🟡 MEDIUM — `position: fixed` on body can break portaled `Combobox` dropdowns in some environments

**File:** `hooks/use-scroll-lock.ts`, lines 33–36

The scroll-lock sets `document.body.style.position = 'fixed'`. Across the codebase, many modals include `<Combobox>` components (fighter-edit-modal, campaign-edit-modal, campaign-territory-edit-modal, add-vehicle, etc.). `Combobox` portals its dropdown directly to `document.body` with `position: fixed` and positions it via `getBoundingClientRect()` (viewport-relative coords).

In standard CSS, a `position: fixed` child uses the **viewport** as its containing block, so this should work correctly on desktop Chrome/Firefox. However, on **iOS Safari**, the relationship between the layout viewport and visual viewport when `body.position` is `fixed` is non-standard and can cause portaled fixed-position elements to be misaligned (particularly when the software keyboard is shown, or at non-zero `window.scrollY`).

**Recommendation:** Explicitly test on iOS Safari with a modal that contains a Combobox (e.g. the fighter edit modal) at a scrolled position before merging.

---

#### 🟡 MEDIUM — Module-level mutable singleton is a maintenance foothold

**File:** `hooks/use-scroll-lock.ts`, lines 6–16

```typescript
let lockCount = 0;
let savedScrollY = 0;
let savedBodyStyles: SavedStyles | null = null;
```

Module-level state persists for the lifetime of the JS bundle. The reference-counting logic (`lockCount`) handles nested modals correctly, but it makes the state invisible to React DevTools and creates a shared implicit dependency across all modal instances. If `unlockScroll` somehow fails to fire (e.g. an exception in cleanup, though unlikely), `lockCount` stays elevated and no subsequent modal can ever re-acquire the lock correctly — the page would be permanently scrollable even with a modal open.

A common mitigation is to reset `lockCount` to 0 in `lockScroll` when `savedBodyStyles` is null (i.e. the styles aren't actually locked right now), acting as a self-healing sentinel:

```typescript
function lockScroll() {
  if (typeof document === 'undefined') return;
  // Self-heal if state got out of sync
  if (lockCount > 0 && savedBodyStyles === null) lockCount = 0;
  
  lockCount += 1;
  if (lockCount > 1) return;
  // ...rest unchanged
}
```

---

#### 🟡 MEDIUM — `ImageEditModal` nesting: confirm scroll lock fires only once

**File:** `components/ui/image-edit-modal.tsx`

`ImageEditModal` wraps `<Modal>`, which now unconditionally calls `useScrollLock()`. Since `ImageEditModal` does `if (!isOpen) return null` before rendering `<Modal>`, the hook correctly only runs when the modal is visible — no double-fire. ✅ Just confirming this was considered.

---

#### 🔵 LOW — `enabled` parameter creates an unused API surface from `Modal`

**File:** `hooks/use-scroll-lock.ts`, line 57 / `components/ui/modal.tsx`, line 40

`useScrollLock(enabled = true)` accepts an `enabled` flag, but `Modal` always calls it as `useScrollLock()` with no way for callers to opt-out. The parameter is useful if this hook is reused elsewhere (e.g. a drawer), but as-is it's dead weight in the exported API. No action needed, just worth noting if the intent was to expose opt-out to `Modal` callers.

---

### Checklist

| Area | Status |
|---|---|
| Correctness (desktop) | ✅ Logic is correct, order of style restore then `scrollTo` is right |
| Edge cases / null handling | ⚠️ `paddingRight` bug above; SSR guard (`typeof document`) is redundant but harmless |
| Security | ✅ No concerns |
| Performance | ✅ `lockCount` avoids redundant DOM writes for nested modals |
| Readability | ✅ Clear, well-structured |
| Project conventions | ✅ `'use client'`, `useEffect`, hook naming all match existing hooks |
| Cross-browser / iOS | ⚠️ Needs explicit testing on iOS Safari |
| DRY / YAGNI | ✅ Single hook, no prop-threading |

---

### Verdict

**Changes requested** — the `paddingRight` overwrite is a real (if low-frequency) correctness bug. The iOS concern should be verified before merge. Everything else is advisory. The core approach is correct and addresses a genuine UX problem.

> *Reviewed by Claude Code (automated senior engineer review)*

---

joeseos:
## Follow-up Review — Fix scrolling of non-active layers not being disabled

@Djidiouf Thanks for addressing the previous feedback. Here's the status on each prior finding and a look at the new changes.

---

|---|---|---|---|
| 1 | 🔴 HIGH | `paddingRight` overwrites instead of augmenting existing body padding | ✅ **Resolved** |
| 2 | 🟡 MEDIUM | `position: fixed` on body may break portaled Combobox dropdowns on iOS Safari | ⚠️ **Outstanding** (advisory — needs manual testing) |
| 3 | 🟡 MEDIUM | Module-level `lockCount` has no self-healing sentinel if state drifts | ✅ **Resolved** |
| 4 | 🟡 MEDIUM | `ImageEditModal` nesting — double-fire risk | ✅ **Confirmed non-issue** (unchanged) |
| 5 | 🔵 LOW | `enabled` parameter in `useScrollLock` was dead weight from `Modal` | ✅ **Resolved** |

---

### New Changes — Assessment

**`hooks/use-scroll-lock.ts`**

The `paddingRight` fix is correct:
```typescript
const existingPadding = parseFloat(savedBodyStyles.paddingRight) || 0;
document.body.style.paddingRight = `${existingPadding + scrollbarWidth}px`;
```
`savedBodyStyles` is assigned immediately above this block, so the access is safe. On unlock, `savedBodyStyles.paddingRight` correctly restores the original value. ✅

The self-healing sentinel is exactly as suggested:
```typescript
if (lockCount > 0 && savedBodyStyles === null) lockCount = 0;
```
Covers the drift scenario without affecting the normal path. ✅

**`components/ui/modal.tsx`**

The `lockBackgroundScroll` prop (default `true`) is a clean addition. Since `Modal` has no `isOpen` prop and always renders the overlay, callers are responsible for conditional mounting — `useScrollLock` firing on mount is the correct behaviour. The new prop gives callers an explicit opt-out where needed. ✅

---

### New Finding

**`hooks/use-scroll-lock.ts` · LOW — `parseFloat` assumes `px` units for existing body padding**

```typescript
const existingPadding = parseFloat(savedBodyStyles.paddingRight) || 0;
```

`parseFloat("1.5rem")` returns `1.5`, which would be treated as `1.5px` — wrong if the body happens to carry padding in `rem`, `%`, or `em`. In practice this is uncommon on `document.body` and the new code is strictly better than before (which ignored existing padding entirely). No action required, but worth knowing if the project ever adds rem-based body padding.

---

### Verdict

**Approve.** The blocking HIGH issue is fixed correctly, the sentinel addresses the MEDIUM reliability concern, and the new `lockBackgroundScroll` prop turns the previous LOW note into a clean feature. The only remaining item (iOS Safari with portaled Comboboxes) is advisory and requires manual device testing — it cannot be resolved in code. Everything else looks good.

> *Reviewed by Claude Code (automated senior engineer review)*

---

joeseos (approved):
All blocking findings from the previous review have been addressed. The `paddingRight` augmentation fix is correct, the self-healing sentinel is correctly placed, and the `lockBackgroundScroll` opt-out prop is a clean addition. One low-severity note on `parseFloat` unit assumptions left in the issue comment. Approving.

---

joeseos (approved):
## Follow-up Review — Commit 623c929 (`combobox.tsx` positioning rework)

@Djidiouf The third commit directly addresses the outstanding iOS Safari advisory from the first review cycle. Here's the full status.

---

### Previous Findings — Final Status

|---|---|---|---|
| 1 | 🔴 HIGH | `paddingRight` overwrites existing body padding | ✅ **Resolved** (2nd commit) |
| 2 | 🟡 MEDIUM | `position: fixed` on body breaks portaled Combobox on iOS Safari | ✅ **Addressed** (this commit) |
| 3 | 🟡 MEDIUM | Module-level `lockCount` has no self-healing sentinel | ✅ **Resolved** (2nd commit) |
| 4 | 🟡 MEDIUM | `ImageEditModal` nesting — double-fire risk | ✅ **Non-issue** (confirmed) |
| 5 | 🔵 LOW | `enabled` parameter in `useScrollLock` was dead weight | ✅ **Resolved** (2nd commit, `lockBackgroundScroll` prop added) |

---

### New Commit — Assessment (`components/ui/combobox.tsx`)

**`getVisualViewportRect()` helper**

Clean fallback chain: uses `window.visualViewport` when available, otherwise falls back to `window.innerHeight`. Safe in SSR contexts because `updatePosition` already guards on `open && inputRef.current`. ✅

**Coordinate conversion math**

The key fix: `getBoundingClientRect()` returns coordinates relative to the **visual viewport**, but `position: fixed` elements are placed relative to the **layout viewport** (initial containing block). Adding `viewport.offsetTop` / `viewport.offsetLeft` correctly converts between the two. Spot-checked both `up` and `down` paths:

- `down`: `top = rect.bottom + offsetTop` — correct
- `up`: `top = rect.top + offsetTop`, then `transform: translateY(-100%)` — correct; the dropdown's top is anchored at the trigger's top in layout-viewport space, then the whole element shifts up by its own height

`spaceAbove = rect.top - viewportTop` correctly shrinks when the iOS keyboard pushes `offsetTop` up (less visual space above), which feeds into `maxHeight` — accurate. ✅

**`position.bottom` removal**

Switching from `bottom: Npx` to `top + translateY(-100%)` is a cleaner model: one layout property for all cases, direction-switching via transform. No regressions expected. ✅

**RAF debouncing via `schedulePositionUpdate`**

The guard (`if (positionFrameRef.current !== null) return`) correctly collapses bursts of scroll/resize events into a single frame. After the frame fires, the ref resets to `null` so the next burst can be scheduled. Cleanup in both effects (`cancelAnimationFrame` + `clearTimeout`) is thorough. ✅

**`visualViewport` event listeners**

Listening to both `visualViewport.resize` and `visualViewport.scroll` is the correct approach for handling iOS soft-keyboard appearance and panning. ✅

**Two-shot open effect (150ms + 350ms)**

Pragmatic workaround for the iOS keyboard animation timeline. Both timers clean up correctly on `open → false` or unmount. ✅

---

### New Finding

**`components/ui/combobox.tsx` · LOW — redundant `offsetTop` variable**

Inside `updatePosition`, `viewportTop` and `offsetTop` both hold `viewport.offsetTop`:

```typescript
const viewportTop = viewport.offsetTop       // used for spaceAbove
// ...
const offsetTop = viewport.offsetTop         // used for top positioning — same value
```

Not a correctness issue — both names refer to identical values — but a minor cleanup opportunity: `viewportTop` could be used everywhere, dropping `offsetTop`.

---

### Verdict

**Approve.** The `visualViewport` API approach is the right fix for iOS Safari, the coordinate math is correct, and the RAF debouncing and cleanup are solid. The only note is the redundant variable above — too minor to block.

Manual smoke-test recommendation (unchanged from before): open a fighter-edit modal at a scrolled page position on a real iOS device, tap a Combobox, verify the dropdown appears correctly when the software keyboard is visible.

> *Reviewed by Claude Code (automated senior engineer review)*

---

PR #1804, opened 2026-06-21
Fixed so that custom weapon profiles can't be assigned to wargear

joeseos:


joeseos (commented):
## PR Review — Fixed so that custom weapon profiles can't be assigned to wargear

@joeseos Reviewed 4 changed files. The core fix is correct and the direction of the refactoring (server-side filtering, eliminating `any` casts, unifying the type) is good. A few issues to address before merge.

---

### What the fix does

The root bug: when creating/editing **ammunition** custom equipment, the "target weapon" picker (`availableWeapons` query in `custom-equipment.tsx`) was fetching all equipment (weapons + wargear) from `/api/equipment`. Custom equipment with `equipment_type = 'wargear'` could therefore be selected as the attach-target for a custom weapon profile, which is invalid.

The fix adds `?equipment_type=weapon&core_equipment=false` to that specific fetch, filtering server-side to weapons only. The API route is extended to support these two query params. **This is the right approach.**

---

### Findings

#### 🟡 MEDIUM — React Query cache key doesn't include filter params (`custom-equipment.tsx`)

**File:** `components/customise/custom-equipment.tsx`, around line 101

```typescript
const { data: availableWeapons = [] } = useQuery<AvailableWeapon[]>({
  queryKey: ['availableWeapons'],   // ← no params in key
  queryFn: async () => {
    const response = await fetch('/api/equipment?equipment_type=weapon&core_equipment=false');
```

The query key `['availableWeapons']` doesn't encode the URL parameters. If this key is ever used in another query hook (in a different component loaded in the same SPA session) with a different URL, both would share the same cached result and serve the wrong data. Currently there's only one usage, but this is a latent correctness bug.

**Fix:** Include the params in the key:
```typescript
queryKey: ['availableWeapons', 'weapon', 'non-core'],
```

---

#### 🟡 MEDIUM — `existingEquipment` silently drops `equipment_type` in `custom-fighters.tsx`

**File:** `components/customise/custom-fighters.tsx`, diff line ~764

```diff
-  equipment_type: 'wargear' as 'wargear' | 'weapon',
+  is_custom: false,
```

The old code hardcoded `equipment_type: 'wargear'` (which was also wrong for any weapons in the default equipment). The new code omits `equipment_type` entirely — `undefined` in practice since `EquipmentListItem.equipment_type` is optional.

This matters because `setEquipment(existingEquipment)` pre-populates the equipment state when a fighter is loaded for editing, and the `if (equipment.length === 0)` guard further below means the API fetch is **skipped** for fighters that already have default equipment. Those items will have `equipment_type: undefined` for the duration of the edit session.

If any rendering or filtering downstream keys on `equipment_type`, those items would behave as uncategorised. Worth verifying the UI handles `undefined` gracefully here, or restoring a type value from `eq` if `fighter.default_equipment` carries that field.

---

#### 🟡 MEDIUM — `EquipmentListItem` exported from an API route file

**File:** `app/api/equipment/route.ts`, lines 5–13

```typescript
export interface EquipmentListItem { ... }
```

This file imports `createClient` from `@/utils/supabase/server` (a server-only module). Exporting a shared type from here and importing it in three client components works today because all three use `import type`, which TypeScript erases before the bundler sees it. But it's fragile:

- A future developer importing a value (not just a type) from this file in a client component would hit a server/client boundary error.
- The type is a data-contract concern, not a routing concern — it shouldn't live in a route handler.

**Suggested fix:** Move `EquipmentListItem` to `types/equipment.ts` and import from there in both the route and the components.

---

#### 🔵 LOW — Non-null assertion on `item.original_id` (`custom-equipment.tsx`)

**File:** `components/customise/custom-equipment.tsx`, line ~106

```typescript
id: item.is_custom ? item.original_id! : item.id,
```

The `!` is safe because the API always sets `original_id` when `is_custom` is true. But `EquipmentListItem.original_id` is typed as `string | undefined`, so TypeScript can't enforce this guarantee. If the API ever changes, this becomes a silent `undefined` being used as an ID. A runtime guard or a discriminated union type would be more robust, but this is low risk given the controlled API surface.

---

### Checklist

| Area | Status |
|---|---|
| Core bug fix correctness | ✅ Server-side `equipment_type=weapon` filter is correct |
| `core_equipment` filter parity | ✅ `or('core_equipment.is.null,core_equipment.eq.false')` matches old client-side `!e.core_equipment` |
| `custom_equipment.equipment_type` column existence | ✅ Confirmed — `createCustomEquipment` action writes this column |
| All three `/api/equipment` callers updated | ✅ `custom-equipment`, `custom-fighters`, `custom-trading-posts` all updated |
| No other callers of `/api/equipment` missed | ✅ Confirmed by code search (`total_count: 3`) |
| Removal of `any` casts | ✅ `any` replaced with typed `EquipmentListItem[]` throughout |
| `EquipmentOption` local type replaced cleanly | ✅ `type EquipmentOption = EquipmentListItem` is fine as a transitional alias |
| React Query key correctness | ⚠️ Key doesn't include filter params (see above) |
| Security (user auth on API route) | ✅ Auth check unchanged and in place |
| Breaking API shape change | ℹ️ `core_equipment` field removed from response, `equipment_type` is now the actual DB value instead of `'regular'`/`'custom'` — no unupdated consumers exist |

---

### Verdict

**Approve with suggestions** — the fix is correct and safe to ship. The React Query key issue (#1) is the only one I'd want addressed before merge since it could cause confusing cache-sharing bugs as the codebase grows. The `EquipmentListItem` location (#3) is a cleanup item that can be done separately.

> *Reviewed by Claude Code (automated senior engineer review)*

---

PR #1805, opened 2026-06-22
Upgrade React Tooltip to major version necessiting refactoring of imp…

Djidiouf:
…lementation

### DIFF
--- a/components/campaigns/[id]/campaign-add-territory-list.tsx
+++ b/components/campaigns/[id]/campaign-add-territory-list.tsx
@@ -181,9 +181,7 @@ export default function TerritoryList({
           <span
             className="relative cursor-pointer text-muted-foreground hover:text-foreground"
             data-tooltip-id="territories-types-tooltip"
-            data-tooltip-html={
-              'Select the campaign types that you want to add territories for.'
-            }
+            data-tooltip-content="Select the campaign types that you want to add territories for."
           >
             <ImInfo />
           </span>
--- a/components/campaigns/[id]/campaign-edit-modal.tsx
+++ b/components/campaigns/[id]/campaign-edit-modal.tsx
@@ -11,6 +11,7 @@ import { useRouter } from "next/navigation";
 import { deleteCampaign } from "@/app/actions/campaigns/[id]/campaign-settings";
 import { ImInfo } from "react-icons/im";
 import { Tooltip } from 'react-tooltip';
+import { renderDescriptionTooltip } from '@/components/ui/tooltip-renderers';
 import { tradingPostRank } from "@/utils/tradingPostRank";
 import CampaignAllegiancesActions from "@/components/campaigns/[id]/campaign-allegiances-actions";
 import CampaignResourcesActions from "@/components/campaigns/[id]/campaign-resources-actions";
@@ -315,9 +316,7 @@ export default function CampaignEditModal({
                   <span
                     className="relative cursor-pointer text-muted-foreground hover:text-foreground"
                     data-tooltip-id="resources-tooltip"
-                    data-tooltip-html={
-                      'Resources are campaign-specific currencies that gangs can accumulate. Predefined resources come from the campaign type (e.g., Exploration Points for Underhells, Meat and Scavenging Rolls for Uprising). Campaign owners and arbitrators can also add custom resources.'
-                    }
+                    data-tooltip-description="Resources are campaign-specific currencies that gangs can accumulate. Predefined resources come from the campaign type (e.g., Exploration Points for Underhells, Meat and Scavenging Rolls for Uprising). Campaign owners and arbitrators can also add custom resources."
                   >
                     <ImInfo />
                   </span>
@@ -356,9 +355,7 @@ export default function CampaignEditModal({
                   <span
                     className="relative cursor-pointer text-muted-foreground hover:text-foreground"
                     data-tooltip-id="resources-tooltip"
-                    data-tooltip-html={
-                      'Only selected Trading Posts are available for gangs taking part in this campaign when buying equipment. However, this does not prevent players to access the Unrestricted list options.'
-                    }
+                    data-tooltip-description="Only selected Trading Posts are available for gangs taking part in this campaign when buying equipment. However, this does not prevent players to access the Unrestricted list options."
                   >
                     <ImInfo />
                   </span>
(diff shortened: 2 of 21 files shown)


joeseos (changes requested):
## PR Review — Upgrade React Tooltip v5 → v6

@Djidiouf

This is a solid, well-scoped migration. The approach of replacing `data-tooltip-html` with three patterns (static `data-tooltip-content`, shared `<Tooltip>` with JSX children, and custom `render` + `data-*` attributes) maps cleanly to the v6 API. The XSS risk from hand-rolled `escapeHtml` + string interpolation is eliminated across the board. Good work overall.

---

### Findings

---

#### [MEDIUM] Dead export: `EquipmentTooltip` returns `null` but is still rendered in 3 places

**Files:**
- `components/equipment/equipment-tooltip.tsx:207–209`
- `components/fighter/fighter-equipment-list.tsx:1054`
- `components/fighter/vehicle-equipment-list.tsx:1027`
- `components/equipment/equipment.tsx:890`

`EquipmentTooltip` now unconditionally returns `null` — it is a no-op stub kept for backward-compat. The three call sites still import and render it, and `EQUIPMENT_TOOLTIP_ID` is still exported but used nowhere functional.

**Fix:** Remove the `EquipmentTooltip` export, the `EQUIPMENT_TOOLTIP_ID` export, and the three `<EquipmentTooltip />` render calls. This cleans up dead symbols and avoids the next reader wondering what it does.

---

#### [MEDIUM] Per-item `<Tooltip>` in `EquipmentTooltipTrigger` — potential DOM overhead on large lists

**File:** `components/equipment/equipment-tooltip.tsx:173–731`

Each `EquipmentTooltipTrigger` now renders its own `<Tooltip>` (via a React portal to `<body>`). In v5 a single shared instance (keyed by `EQUIPMENT_TOOLTIP_ID`) was used, updated dynamically per hover. The new approach is correct for JSX-rich content in v6, but on pages with many equipment items (full trading post browser, equipment list) you'll mount N portal nodes simultaneously.

React Tooltip v6 does still support the shared-singleton + `render` prop pattern (as demonstrated by `renderDescriptionTooltip` throughout this PR). An alternative would be to encode the needed display data into `data-*` attributes and use a single shared `<Tooltip render={...} />`. For `EquipmentTooltipContent` that's harder because it needs the full `Equipment` object — passing a JSON string through a `data-` attribute is ugly. The per-item approach is a pragmatic call, but keep an eye on performance in the trading post panel.

**Suggestion:** Consider lazy-mounting — render the `<Tooltip>` only when the trigger is hovered (via local `useState`). Not a blocker, but worth a follow-up if users report jank on heavy lists.

---

#### [MEDIUM] `getClassRank` defined at component scope, omitted from `useMemo` deps

**File:** `components/gang/campaign-tab.tsx` (around the `ooaBreakdown` / `deathsBreakdown` memos)

```ts
const getClassRank = (fighterClass: string) =>
  fighterClassRank[fighterClass.toLowerCase().trim()] ?? 99;

const ooaBreakdown = useMemo(() => {
  ...
  return [...breakdown].sort(
    (a, b) => getClassRank(a.fighter_class) - getClassRank(b.fighter_class)
  );
}, [fighterStats?.ooa_breakdown]); // getClassRank missing from deps
```

`getClassRank` is recreated every render but never changes in practice (it reads `fighterClassRank`, a module-level import). So this is not a runtime bug, but `react-hooks/exhaustive-deps` will warn. Either move `getClassRank` outside the component (it has no dependencies on props/state), or add it to the dep arrays. Moving it to module scope is cleaner.

---

#### [LOW] Non-unique keys in OOA / Deaths breakdown lists

**File:** `components/gang/campaign-tab.tsx` (OOA and Deaths `useMemo` render loops)

```tsx
key={`${fighter.fighter_name}-${fighter.fighter_type}-${fighter.fighter_class}`}
```

If the same fighter\_name + fighter\_type + fighter\_class appears more than once (e.g., two fighters with identical names), keys collide. The old HTML-string code had no keys at all, so this is not a regression, but since you're converting to JSX now is a good time to use a stable unique id if one exists on the fighter record (e.g., `fighter.id`).

---

#### [LOW] `EQUIPMENT_TOOLTIP_ID` still exported as dead code

**File:** `components/equipment/equipment-tooltip.tsx:7`

```ts
export const EQUIPMENT_TOOLTIP_ID = 'equipment-tooltip';
```

Nothing outside this file uses it after this PR. Remove alongside the `EquipmentTooltip` stub cleanup.

---

### What's working well

- **XSS risk eliminated**: All hand-rolled `escapeHtml` + `innerHTML`-style string interpolation is gone from React components. React's JSX auto-escaping now handles all user data.
- **`renderDescriptionTooltip` pattern**: The shared `<Tooltip render={fn}>` + custom `data-tooltip-title` / `data-tooltip-description` attributes is a clean and reusable abstraction.
- **`EquipmentTooltipContent` as a React component**: Converting the HTML-string builder into a proper component with Tailwind classes is a significant readability and maintainability improvement.
- **`campaign-map-canvas.tsx` correctly left alone**: It uses `escapeHtml` for Leaflet HTML strings (not react-tooltip), so the intentional carve-out is correct.
- **Migration is complete**: No `data-tooltip-html` attributes remain anywhere in the codebase. ✓

---

### Verdict

**Changes requested** — the two medium issues (dead `EquipmentTooltip` stub + `getClassRank` scope) are straightforward cleanup items, and the dead export is a broken-window risk for future readers. The per-item tooltip concern is worth noting but not a blocker. Once those are addressed this is ready to merge.

---

joeseos (approved):
## Follow-up Review — Upgrade React Tooltip v5 → v6

@Djidiouf — all required changes from the previous review have been addressed. Here's the per-finding breakdown:

---

### Previous Findings

| Finding | Status |
|---|---|
| [MEDIUM] Dead `EquipmentTooltip` stub rendered in 3 places | ✅ Resolved — stub removed from `equipment-tooltip.tsx`; all three `<EquipmentTooltip />` render calls and imports removed from `equipment.tsx`, `fighter-equipment-list.tsx`, `vehicle-equipment-list.tsx` |
| [MEDIUM] `getClassRank` at component scope, missing from `useMemo` deps | ✅ Resolved — moved to module scope as a named function; no longer needs to appear in dep arrays |
| [MEDIUM] Per-item `<Tooltip>` DOM overhead on large lists | ⏭ Outstanding (was explicitly a non-blocker; no action required) |
| [LOW] Non-unique keys in OOA / Deaths breakdown lists | ✅ Resolved — `breakdownRowKey` helper (also at module scope) appends the list index, preventing key collisions |
| [LOW] `EQUIPMENT_TOOLTIP_ID` dead export | ✅ Resolved — export removed |

---

### New Commit (`7b9bc40`) — No New Issues

The 37-line diff is clean:

- **`breakdownRowKey` with index suffix** — using `index` as a tiebreaker is correct here. The list is sorted deterministically by `getClassRank` and is non-interactive (no user reordering), so index-based keys are stable and will not cause reconciliation artifacts.
- **`getClassRank` at module scope** — reads only `fighterClassRank` (a module-level import), so moving it out of the component is strictly correct and removes the eslint warning without touching the dep arrays.
- All dead imports and render calls are gone. No regressions.

---

### Verdict

**Approved** — clean and complete. Ready to merge.

---

PR #1806, opened 2026-06-22
Refactored to manage permissions using a custom jwt

joeseos:


joeseos (commented):
## PR Review — JWT-based Permission Refactor

@joeseos

This is a well-structured architectural shift. Moving permissions out of per-request DB queries and into JWT claims is a real performance win, and the overall layering (hook → `getClaims` → permission functions) is clean. The `Suspense` fixes for `useSearchParams` pages are also correct and independent wins.

That said, this touches auth and RLS policy evaluation, so there are a few correctness and security issues worth addressing before merge.

---

### [HIGH] Bug: `Array.find()` picks the first campaign role, not the highest

**File:** `app/lib/user-permissions.ts` — `isCampaignArbitrator` and `getCampaignPermissions`

```ts
const match = claims.campaignRoles.find(cr => cr.id === campaignId);
return match?.role === 'OWNER' || match?.role === 'ARBITRATOR';
```

The JWT hook inserts **all** `campaign_members` rows for a user without deduplication:

```sql
SELECT COALESCE(jsonb_agg(
  jsonb_build_object('id', cm.campaign_id, 'role', cm.role)
), '[]'::jsonb)
...
FROM public.campaign_members cm WHERE cm.user_id = ...
```

A user with two gangs in the same campaign gets two `campaign_roles` entries for the same campaign ID (e.g. `[{id:"X", role:"MEMBER"}, {id:"X", role:"OWNER"}]`). The old code explicitly handled this:

```ts
// Old getCampaignRole — highest role wins:
if (roles.includes('OWNER')) return 'OWNER';
else if (roles.includes('ARBITRATOR')) return 'ARBITRATOR';
else if (roles.includes('MEMBER')) return 'MEMBER';
```

`Array.find()` returns the first match by insertion order. A campaign owner whose first DB row is `MEMBER` loses OWNER permissions silently.

**Fix (option A — recommended):** Deduplicate to highest role per campaign in the SQL hook.

**Fix (option B):** Replace `find()` with a filter + role-hierarchy scan in both `isCampaignArbitrator` and `getCampaignPermissions`:
```ts
const matches = claims.campaignRoles.filter(cr => cr.id === campaignId);
const roles = matches.map(cr => cr.role);
const campaignRole = roles.includes('OWNER') ? 'OWNER'
  : roles.includes('ARBITRATOR') ? 'ARBITRATOR'
  : roles.includes('MEMBER') ? 'MEMBER'
  : null;
```

---

### [HIGH] RLS breakage on deploy: `is_admin()` has no fallback for old JWTs

**File:** `supabase/functions/is_admin.sql` (and migration)

```sql
SELECT coalesce(auth.jwt()->'user_profile'->>'user_role' = 'admin', false);
```

On deployment day, JWTs issued **before** the hook is enabled have no `user_profile` claim. For those users, this expression evaluates to `false`. Every admin whose token hasn't naturally refreshed (~1hr window) loses all admin-gated RLS access.

`is_arb()` correctly handles this with a DB fallback when `campaign_roles IS NULL`. `is_admin()` needs the same treatment:

```sql
CREATE OR REPLACE FUNCTION private.is_admin()
RETURNS boolean LANGUAGE plpgsql ...
AS $$
BEGIN
  IF auth.jwt()->'user_profile' IS NOT NULL THEN
    RETURN coalesce(auth.jwt()->'user_profile'->>'user_role' = 'admin', false);
  END IF;
  -- fallback for old JWTs issued before the hook was enabled
  RETURN EXISTS (
    SELECT 1 FROM profiles WHERE id = auth.uid() AND user_role = 'admin'
  );
END;
$$;
```

---

### [MEDIUM] Duplicate hook SQL file is missing grants — will silently break local/fresh environments

**Files:** `supabase/functions/custom_access_token_hook.sql` vs `supabase/migrations/20260622120000_custom_access_token_hook.sql`

The migration correctly includes:
```sql
GRANT EXECUTE ON FUNCTION public.custom_access_token_hook(jsonb) TO supabase_auth_admin;
GRANT SELECT ON TABLE public.profiles TO supabase_auth_admin;
GRANT SELECT ON TABLE public.campaign_members TO supabase_auth_admin;
REVOKE EXECUTE ON FUNCTION public.custom_access_token_hook(jsonb) FROM public, anon, authenticated;
```

The standalone `supabase/functions/custom_access_token_hook.sql` has none of these. If a developer runs it directly (local setup, fresh env), `supabase_auth_admin` won't have SELECT on the needed tables. The hook will silently fail, returning an unmodified JWT — every user appears as non-admin with no campaign roles.

**Fix:** Either delete the standalone file (the migration is the canonical source), or add a prominent warning comment and the matching grants to it.

---

### [MEDIUM] `HeaderAuth` causes layout shift (CLS) on every authenticated page load

**File:** `components/header-auth.tsx`

`useClaims()` initializes with `loading: true`. During that window the SettingsModal is not rendered. On every SSR → hydration cycle, logged-in users see the header without the settings icon for ~100–200ms, then it pops in. The old server-component layout rendered the correct header on the first paint with no shift.

This is a measurable CLS regression. The tradeoff (no DB query in root layout) is real and may be worth accepting — but consider seeding the initial render by passing `userId`/`isAdmin` as a prop from the server layout to `HeaderAuth`, so the first render is correct.

---

### [LOW] `isCampaignArbitrator` / `computeGangPermissions` accept `userId` but never use it for the JWT checks

```ts
export async function isCampaignArbitrator(
  userId: string,      // accepted but never used for the JWT claim lookup
  campaignId: string | null
): Promise<boolean> { ... }
```

Admin/arb status is read from the **current caller's** JWT regardless of what `userId` is passed. A future caller could reasonably assume they can check another user's permissions — they'd silently get their own JWT results. Either remove the unused `userId` param from functions that don't need it, or add a clarifying comment.

---

### [LOW] Each permission function creates its own Supabase client

`computeGangPermissions`, `isCampaignArbitrator`, and `getCampaignPermissions` each call `createClient()` internally. A page calling multiple permission functions creates multiple clients. Not a bug — Next.js caches the server client within a request — but worth noting if this proves expensive in profiling.

---

### What's working well

- **`getClaims()` as the single source of truth** — clean; all auth paths funnel through it.
- **`extractCustomClaims()` shared** between server and client paths — good cohesion.
- **`computeGangPermissions` explicit signature** — callers supply `gangOwnerId` + `gangCampaignIds`; the function only decides, doesn't fetch. Easy to test.
- **`is_arb()` DB fallback** is the right pattern; the same just needs to be applied to `is_admin()`.
- **Parallel `gangBasic` + `campaigns` fetch** in gang/print pages is a nice improvement.
- **README auth section** is clear and honestly documents the stale-claim tradeoff.
- **`Suspense` wrappers on auth pages** are a correct, independent fix for the `useSearchParams` SSR warnings.

---

**Summary:** Two issues (#1 role-resolution bug, #2 `is_admin()` fallback) are correctness/safety issues that should be fixed before deploy. The rest are polish. The architectural direction is solid.

---

PR #1807, opened 2026-06-23
Use service-role Supabase client and add error handling in `advanceRound`

joeseos:
### Motivation

- Use a privileged server-side Supabase client when reading/updating session and fighter records to avoid RLS/permission issues and ensure updates succeed. 
- Surface database errors from reads/updates so failures don't silently proceed.
- Ensure fighter activation updates are applied atomically and any activation write errors are detected.

### Description

- Added an import for `createServiceRoleClient` and instantiate `serviceSupabase` in `advanceRound` while keeping the regular `supabase` client for authentication calls.
- Switched session, fighters, fighter-details reads and the session/fighter updates to use `serviceSupabase` instead of the request-scoped client. 
- Added explicit error checks for `sessionError`, `fightersError`, and `fighterDetailsError` and return early with the DB error message when present. 
- Aggregated the results of the activation update `Promise.all` and return the first activation error if any update failed.

### Testing

- Ran the automated test suite via `npm test` and the backend integration tests covering battle session flows, and they passed.
- Verified the `advanceRound` code path for forward and back rounds with sample session and fighter data using the integration tests, and no failures were reported.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3a37b095008332ab9608e3dbaffd4f)

### DIFF
--- a/app/actions/battle-sessions.ts
+++ b/app/actions/battle-sessions.ts
@@ -438,44 +438,63 @@ export async function advanceRound(
       .select('id, fighter_id, session_record')
       .eq('battle_session_id', sessionId);
 
-    if (fightersError) return { success: false, error: fightersError.message };
-
-    if (fighters && fighters.length > 0) {
+    if (fightersError) {
+      console.error('[advanceRound] Failed to load fighters after round advanced:', {
+        sessionId,
+        newRound: nextRound,
+        error: fightersError.message,
+      });
+    } else if (fighters && fighters.length > 0) {
       const fighterIds = fighters.map((f) => f.fighter_id);
       const { data: fighterDetails, error: fighterDetailsError } = await serviceSupabase
         .from('fighters')
         .select('id, special_rules')
         .in('id', fighterIds);
 
-      if (fighterDetailsError) return { success: false, error: fighterDetailsError.message };
-      const rulesMap = new Map(
-        (fighterDetails ?? []).map((f: any) => [f.id, f.special_rules as string[] | null])
-      );
-      const DUAL_ACTIVATION_RULES = ['Spyre Hunter', 'Aranthian Beauty Plating'];
+      if (fighterDetailsError) {
+        console.error('[advanceRound] Failed to load fighter details after round advanced:', {
+          sessionId,
+          newRound: nextRound,
+          error: fighterDetailsError.message,
+        });
+      } else {
+        const rulesMap = new Map(
+          (fighterDetails ?? []).map((f: any) => [f.id, f.special_rules as string[] | null])
+        );
+        const DUAL_ACTIVATION_RULES = ['Spyre Hunter', 'Aranthian Beauty Plating'];
 
-      const activationResults = await Promise.all(
-        fighters.map((f) => {
-          const rules = rulesMap.get(f.fighter_id);
-          const isDual = rules?.some((r) => DUAL_ACTIVATION_RULES.includes(r)) ?? false;
-          // Injured fighters are out of action and get no fresh activation;
-          // players can still grant one manually via updateActivations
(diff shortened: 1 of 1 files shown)


joeseos (commented):
## PR Review — Service Role Client + Error Handling in `advanceRound`

@joeseos

Small, focused change. The intent — surface silent failures and fix a permissions gap for arbitrators — is right. Two issues are worth fixing before merge.

---

### [HIGH] Round increment is committed before error checks: double-advance risk

**File:** `app/actions/battle-sessions.ts`

The operation sequence is:

1. Auth check — `verifySessionParticipant` ✓
2. Read session data
3. **Write** `round: nextRound` — committed to DB ✓
4. Read `battle_session_fighters` → if `fightersError`: `return { success: false }`
5. Read fighter details → if `fighterDetailsError`: `return { success: false }`
6. Update activations → if `activationError`: `return { success: false }`

Steps 4–6 can now return `{ success: false }` **after** step 3 has already committed the round increment. The old code always returned `{ success: true }` regardless of activation errors, which at least kept the client and DB in sync (both believed the round advanced).

The new behavior creates a genuine regression: the UI sees "failure, try again" → user clicks again → round is incremented twice. For a battle session tracking tool this is a real data-integrity issue.

The root fix is to move the entire sequence into a Supabase RPC so the round update and activation resets are atomic. A lighter short-term fix is to acknowledge that the round update is the "commit point" and that activation errors are best-effort, and return `{ success: true, newRound }` (with the error logged server-side) when the round update succeeds but the activation step fails:

```ts
if (activationError) {
  console.error('[advanceRound] Activation update failed:', activationError.message);
  // Round already advanced — surface newRound so client state stays correct
}
revalidateTag(...);
return { success: true, newRound: nextRound };
```

The same logic applies to `fightersError` and `fighterDetailsError` — a read failure here doesn't undo the already-committed round increment, so surfacing it as `{ success: false }` is misleading to the caller.

---

### [MEDIUM] Service role bypasses RLS without fixing the underlying policy gap

**File:** `app/actions/battle-sessions.ts`

The PR description says the service role is needed to "avoid RLS/permission issues." The likely root cause: `verifySessionParticipant` succeeds for arbitrators who have no row in `battle_session_participants` (the "no-participantId form" path), but the RLS policies on `battle_sessions` and `battle_session_fighters` only grant UPDATE to users who have a matching participant row — so arbitrators without a gang in the session can authorize via the app but are blocked by the DB.

Using the service role key is consistent with how the rest of the codebase handles this (campaign-members, stats, patreon sync all do the same), so this isn't a new anti-pattern here. But it's worth filing a follow-up to tighten the RLS policies on `battle_sessions`/`battle_session_fighters` to grant UPDATE to campaign OWNER/ARBITRATOR directly, which would remove the need for the service role bypass here.

---

### [LOW] Only the first activation error is reported; others are silently dropped

```ts
const activationError = activationResults.find((result) => result.error)?.error;
```

If 5 out of 10 fighter updates fail, only one error is reported. Not a correctness issue given the current return-on-first-error behavior, but if you add server-side logging here, log all failures rather than just the first.

---

### What's working well

- The auth check stays on the request-scoped client (`verifySessionParticipant`) — service role is only used after the user is confirmed authorized. The privilege boundary is correct.
- Adding explicit `sessionError` / `fightersError` / `fighterDetailsError` error checks is a clear improvement over the old silent-proceed behavior.
- `activationResults` capture and inspection is cleaner than the fire-and-forget `await Promise.all`.
- `createServiceRoleClient()` call is correctly placed after the auth gate (not before).

---

**Verdict: Changes requested.** The high-priority issue (double-advance risk from returning `{ success: false }` after a committed round write) needs to be addressed. The simplest fix is to treat the round write as the commit point and degrade activation errors to server-side log + continue rather than propagating them as hard failures to the client.

---

PR #1808, opened 2026-06-23
Removed fetch, get username from jwt

joeseos:


joeseos (commented):
## PR Review — #1808: Removed fetch, get username from JWT

---

### Summary

This PR adds a dedicated `hooks/use-claims.ts`, removes the old `useSession()` hook, migrates `app/layout.tsx` to a sync Server Component, and injects `user_profile` into the JWT via a `custom_access_token_hook`. The direction is solid. There are a few issues worth addressing before merge.

---

### Findings

#### 🔴 HIGH — Stale auth state on non-graceful session expiry
**File:** `hooks/use-claims.ts` — `onAuthStateChange` handler

```ts
if (event === 'SIGNED_OUT') {
  setState({ userId: null, email: null, profile: null, loading: false });
  return;
}
if (!session) return;  // ← BUG: silently returns without clearing state
```

If the session expires or is invalidated by the server in a way that fires `onAuthStateChange` with a non-`SIGNED_OUT` event and a null session (e.g. `TOKEN_REFRESHED` failure, `USER_UPDATED` race, network error), the hook leaves the previous `userId`/`profile` in state. The user appears authenticated client-side until a hard reload.

**Fix:** Replace the early return with an explicit state clear:
```ts
if (!session) {
  setState({ userId: null, email: null, profile: null, loading: false });
  return;
}
```

---

#### 🟡 MEDIUM — `checkAdmin()` still queries the DB — contradicts PR goal
**File:** `utils/auth.ts`

`checkAdmin()` is unchanged — it still does a `supabase.from('user_profiles').select('user_role')` DB round-trip. The PR description says "get username from JWT" and the new `extractProfileClaims()` helper is right there. Either migrate `checkAdmin()` to read `claims.profile?.user_role` or document explicitly why it still needs the DB.

---

#### 🟡 MEDIUM — Migration timestamp conflict with PR #1806
**File:** `supabase/migrations/20260623120000_custom_access_token_hook.sql`

PR #1806 (`feature/jwt-claims-permission-refactor`) has migration `20260622…` that also creates `custom_access_token_hook`, but its version injects **both** `user_profile` and `campaign_roles`. This PR's version (`20260623…`) only injects `user_profile`. Because migrations run in timestamp order, if both land on `main`, this PR's migration runs last and silently drops `campaign_roles` from the JWT — breaking every RLS policy and permission check that depends on it.

These two PRs need to be coordinated: one migration should be the canonical hook definition. If PR #1806 lands first, this PR's migration file should be removed or rebased on top with the combined injection.

---

#### 🟡 MEDIUM — `HeaderAuth` FOUC / CLS on initial render
**File:** `components/header-auth.tsx`

`useClaims()` starts with `loading: true` and `userId: null`. On first render the header shows an unauthenticated state, then flips once claims resolve. On SSR-heavy pages this causes a visible flash and a layout shift (CLS). Consider rendering a neutral skeleton/placeholder while `loading` is true rather than the unauthenticated UI.

---

#### 🟢 LOW — Missing newline at end of file
**File:** `hooks/use-claims.ts`

The file is missing a trailing newline (`\ No newline at end of file` in the diff). Minor, but fix it to keep linters/git diffs clean.

---

### What's good

- Removing the `useSession()` → `getSession()` path is correct — `getSession()` doesn't validate the JWT server-side and is deprecated for auth checks.
- `Suspense` wrappers on `sign-in`, `reset-password`, `reset-password/update`, `confirm-email` are the right fix for `useSearchParams()` in Next.js 13+.
- `custom_access_token_hook` in this PR includes the necessary `GRANT EXECUTE` statements that the `supabase/functions/` version in PR #1806 was missing.
- Passing `userId` as a prop to `SkillModal` instead of calling `useSession` inside it is cleaner — the auth check stays at the boundary.

---

### Priority order

1. Fix the stale-state bug in `onAuthStateChange` (HIGH).
2. Coordinate migration with PR #1806 to avoid silently dropping `campaign_roles` (MEDIUM — likely needs a team decision on merge order).
3. Migrate `checkAdmin()` to JWT or document the intentional DB fallback (MEDIUM).


---

PR #1814, opened 2026-06-25
Added fighters and add xp to post-battle in battle sessions

joeseos:


joeseos (commented):
## PR Review — #1814: Added fighters and add XP to post-battle in battle sessions

---

### Summary

This PR makes the fighters table visible in post-battle mode and adds two XP-awarding flows: a per-fighter "Add XP" button (opens `FighterXpModal`) and a bulk "Award Battle Participation" button (new `BattleParticipationModal` with +1 XP for all selected fighters). The post-battle table integration is clean; the bulk award has a correctness issue worth fixing before merge.

---

### Findings

#### 🔴 HIGH — Partial-write rollback permanently de-syncs UI from DB

**File:** `components/battle-session/participant-card.tsx` — `showParticipationModal` `onConfirm` handler

```ts
Promise.all(
  selectedIds.flatMap((id) => {
    // 2 promises per fighter: updateSessionXp + updateFighterXp
  })
).then((results) => {
  const failed = results.filter((r) => !r.success);
  if (failed.length > 0) {
    setLocalFighters(snapshot);            // ← reverts ALL fighters
    toast.error('Failed to award participation XP');
    // ← no onBroadcast call: no re-sync from server
  }
});
```

`Promise.all` fires up to `2 × N` parallel writes (one `updateSessionXp` + one `updateFighterXp` per fighter). Both actions catch errors internally and return `{ success: false }` rather than throwing, so `Promise.all` always resolves. If any result has `!r.success`, the UI is rolled back to `snapshot` (captured before the optimistic update). But the DB writes that already committed are NOT reversed.

**Concrete failure**: 5 fighters selected. Writes for fighters 1–4 succeed; fighter 5's `updateFighterXp` fails (e.g. permission denied). The UI rolls back all 5 to their original XP. Fighters 1–4 now show no XP gain in the UI but have +1 XP committed in the DB. No `onBroadcast?.()` is called on failure, so the component never re-fetches server state. If the user re-opens the modal and awards participation again, fighters 1–4 get double-awarded.

**Fix**: On partial failure, call `onBroadcast?.()` to re-fetch server state instead of reverting to a potentially stale snapshot. This ensures the UI reflects what actually committed.

```ts
}).then((results) => {
  const failed = results.filter((r) => !r.success);
  if (failed.length > 0) {
    toast.error('Failed to award participation XP for some fighters');
    onBroadcast?.();   // re-fetch to show actual server state
  } else {
    toast.success(`Awarded +1 XP to ${selectedIds.length} fighter${selectedIds.length !== 1 ? 's' : ''}`);
    onBroadcast?.();
  }
});
```

---

#### 🟡 MEDIUM — Hand-rolled toggle switch should use the existing `<Switch />` component

**File:** `components/battle-session/participant-card.tsx` — `BattleParticipationModal`

```tsx
<button
  type="button"
  role="switch"
  aria-checked={selected[f.id]}
  onClick={() => toggle(f.id)}
  className={`relative inline-flex h-6 w-11 ... ${selected[f.id] ? 'bg-primary' : 'bg-input'}`}
>
  <span className={`... ${selected[f.id] ? 'translate-x-5' : 'translate-x-0'}`} />
</button>
```

`components/ui/switch.tsx` already exists and wraps Radix UI `SwitchPrimitives` — producing byte-for-byte the same visual result (`h-[24px] w-[44px]`, thumb `h-5 w-5 translate-x-5/0`, `bg-primary`/`bg-input`). The Radix component adds keyboard activation (Space/Enter), `focus-visible:ring-2`, and correct ARIA state management. The hand-rolled version misses all three: pressing Space on a focused toggle does nothing, there is no focus ring, and ARIA `checked` state is not re-signalled for some screen readers.

**Fix**:
```tsx
import { Switch } from '@/components/ui/switch';

<Switch
  checked={selected[f.id]}
  onCheckedChange={() => toggle(f.id)}
/>
```

---

#### 🟢 LOW — `showDirectXp` + `directXpData` two-state pattern leaves stale data in state

**File:** `components/battle-session/participant-card.tsx` — `FighterRow` state (lines 605–606)

```ts
const [showDirectXp, setShowDirectXp] = useState(false);
const [directXpData, setDirectXpData] = useState<{ xp: number; kills: number; kill_count: number } | null>(null);
```

The `onClose` callback sets `setShowDirectXp(false)` but never clears `directXpData`, so stale fighter XP data persists in the component's state indefinitely after the modal is closed. A single nullable state is simpler and self-consistent:

```ts
// Remove showDirectXp entirely. Use:
const [directXpData, setDirectXpData] = useState<{ xp: number; kills: number; kill_count: number } | null>(null);

// open:  setDirectXpData(data)
// close: setDirectXpData(null)
// guard: {directXpData && createPortal(<FighterXpModal ... onClose={() => setDirectXpData(null)} />, document.body)}
```

---

#### 🟢 LOW — `onXpUpdated` ignores kills/kill_count from per-fighter XP modal

**File:** `components/battle-session/participant-card.tsx` — new `FighterXpModal` in post-battle mode

`FighterXpModal.onXpUpdated` signature: `(newXp, newTotalXp, newKills, newKillCount?) => void`

The new callback only uses `newXp`:
```ts
onXpUpdated={(newXp) => {
  const delta = newXp - (directXpData?.xp ?? 0);
  onXpChanged(delta);
  setShowDirectXp(false);
}}
```

If a user awards "Cause OOA" via this modal, `updateFighterXpWithOoa` correctly writes the kill increment to the DB, but `newKills` and `newKillCount` are silently ignored. The fighter's kill count in the session UI stays at its pre-modal value until a full page reload. This is consistent with the pre-existing `openXpModal` pattern (line 325) so it's not a regression, but worth addressing since the post-battle context makes OOA entry more prominent.

---

### What's good

- Removing `{!isPostBattle && (` around the fighters table is the right approach — controlling per-column rendering via `canInteract`/`battleActive`/`canPostBattle` flags is cleaner and more composable.
- `postBattleInteract` prop on `FighterRow` is properly typed and threaded through correctly.
- The optimistic update + snapshot rollback pattern is sound in principle; the issue is rollback granularity on partial failure, not the pattern itself.
- `BattleParticipationModal` initialises all fighters as selected (`Object.fromEntries(fighters.map((f) => [f.id, true]))`), which is the right default for a "participation" award.


---

joeseos (commented):
## Follow-up Review — #1814: Added fighters and add XP to post-battle in battle sessions

@joeseos — all required findings from the previous review are resolved. One small new issue below, and a design-change note worth a quick sanity-check.

---

### Previous Findings

| Finding | Status |
|---|---|
| 🔴 HIGH — Partial-write rollback permanently de-syncs UI from DB | ✅ Resolved — `57322bb` removes the snapshot + `setLocalFighters(snapshot)` rollback entirely and makes `onBroadcast?.()` unconditional, so the UI always re-syncs from server state regardless of success or failure |
| 🟡 MEDIUM — Hand-rolled toggle switch missing keyboard/focus/ARIA | ✅ Resolved — `f9c8cd7` replaced the `<button role="switch">` with `+` / `−` stepper buttons; the toggle pattern is gone |
| 🟢 LOW — `showDirectXp` + `directXpData` two-state stale data | ✅ Resolved — both states removed entirely in `f9c8cd7` |
| 🟢 LOW — `onXpUpdated` ignores kills/kill_count from `FighterXpModal` | ✅ Resolved — the per-fighter `FighterXpModal` in post-battle mode removed entirely |

---

### New Finding

#### 🟢 LOW — Success toast removed; no user feedback on successful XP award

**File:** `components/battle-session/participant-card.tsx` — `onConfirm` handler (`57322bb`)

The "Fixes" commit restructured the `if/else` branch and dropped the `toast.success` in the process:

```ts
// After 57322bb:
if (failed.length > 0) {
  toast.error('Failed to update participation XP');
}
onBroadcast?.();
```

There is no confirmation toast when XP saves successfully. `onBroadcast` will refresh the table values — implicit feedback — but a silent success after a modal confirm is jarring and users may re-submit thinking it didn't take.

**Fix:**
```ts
if (failed.length > 0) {
  toast.error('Failed to update participation XP');
} else {
  toast.success(`Participation XP updated for ${entries.length} fighter${entries.length !== 1 ? 's' : ''}`);
}
onBroadcast?.();
```

---

### Design Change — Per-fighter "Add XP" in post-battle mode removed

`f9c8cd7` removes the `postBattleInteract` prop, the per-row "Add XP" button, and the `FighterXpModal` portal that were introduced in the first commit. The PR title still says _"add xp to post-battle"_ — the only remaining XP path in post-battle mode is the bulk "Participation XP" modal. Just confirming this is intentional scope reduction and not an accidental drop.

---

### What's good in the new commits

- `uniqueFighters` correctly sorts crew fighters (initialised to 1 XP) to the top of the list, making the most relevant fighters immediately visible.
- `confirmDisabled={totalXp === 0}` cleanly prevents an empty submission.
- Removing the snapshot rollback in favour of an unconditional `onBroadcast` re-fetch is the right call — replaying a stale in-memory copy after a partial write is always worse than pulling from the source of truth.

---

### Verdict

**Ready to merge** — the single new LOW (missing success toast) is a one-liner restore and not a correctness issue. The core data-correctness fix landed cleanly.

---

PR #1817, opened 2026-06-25
Added TP roll to post-battle

joeseos:


joeseos (commented):
## PR Review — #1817: Added TP roll to post-battle

---

### Summary

This PR adds a Trading Post roll widget (2D6 + optional modifier) to the post-battle section and logs each roll via `createGangLog`. The dice mechanic is correct for Necromunda (`rollNd6Outcome(2)` = 2D6, as opposed to the D66 used for injury tables). Two small but user-facing issues are worth fixing.

---

### Findings

#### 🟡 MEDIUM — Modifier not reset after rolling; old value silently reapplied on next click

**File:** `components/battle-session/participant-card.tsx` — Roll button `onClick`

After clicking Roll, `tradingPostModifier` is never cleared. If a user rolls with modifier `+3`, then clicks Roll again intending a plain 2D6 roll, the input still shows `3`, the modifier is reapplied, and the log entry records `+3` again with no indication. Every comparable input in this file is cleared after its action completes (e.g. `creditsDelta`, `repDelta` reset on save).

**Fix**: Add `setTradingPostModifier('')` at the end of the onClick handler, or clear it only when the field is empty on roll (leave populated if the user intentionally keeps a recurring modifier — a deliberate UX choice that should be explicit either way).

---

#### 🟡 MEDIUM — `inputMode="numeric"` diverges from file-wide `inputMode="url"` pattern; blocks negative modifier entry on mobile

**File:** `components/battle-session/participant-card.tsx` — modifier `<input>`

```tsx
// New input (PR)
<input type="tel" inputMode="numeric" ... />

// All four existing numeric inputs in this file
<input type="tel" inputMode="url" ... />   // e.g. line 1551 (Credits), 1586 (Rep), 1610 (Resource)
```

`inputMode="url"` is used throughout this file to trigger a full alphanumeric keyboard on iOS (which includes the `-` sign, needed for negative modifiers). `inputMode="numeric"` shows a digit-only number pad on iOS with no minus key — a user who needs to enter `-2` can't do it on mobile.

**Fix**: Use `inputMode="url"` to match the existing pattern.

---

#### 🟢 LOW — Roll button is a raw `<button>` instead of `<Button>` from the design system

**File:** `components/battle-session/participant-card.tsx` — Roll button

```tsx
// New button (PR)
<button
  type="button"
  className="px-3 py-2 bg-neutral-900 text-white rounded-sm hover:bg-gray-800 shrink-0"
>

// All other post-battle action buttons
<Button variant="default" ...>
```

Using the `<Button>` component would inherit focus styles, dark-mode theming, and any future design updates centrally. The manual Tailwind variant will drift if the design system evolves.

---

#### 🟢 LOW — `createGangLog` fire-and-forget: log failures are invisible

**File:** `components/battle-session/participant-card.tsx` — Roll button `onClick`

`createGangLog` is called without `await` or a check on the return value. Since `createGangLog` catches all errors internally and returns `{ success: bool }`, there is no unhandled rejection risk — but if the log write fails (network error, auth issue), the roll result displays locally as if fully recorded while the gang log is never written. This is the same intentional pattern used elsewhere in this file, so it's not a regression — worth noting for visibility.

---

### What's good

- `rollNd6Outcome(2)` is the correct dice mechanic for a Necromunda Trading Post roll (2D6, not D66).
- The modifier display correctly handles sign: positive shows explicit `+`, negative inherits `-` from the number itself.
- Adding `'trading_post_roll'` to `LOG_TYPE_LABELS` in `log-types.ts` ensures the gang log panel displays a readable label rather than the raw action key.
- `tradingPostRoll` is stored as `{ dice, total, modifier, grandTotal }` — keeping the dice breakdown alongside the total is useful for players who want to verify the roll.


---

PR #1818, opened 2026-06-25
Refactored permissions

joeseos:


joeseos (commented):
## PR #1818 — Refactored Permissions Review

@joeseos — automated review across all 28 changed files. This is a well-structured cleanup: removing 393-line `PermissionService` class, flattening to plain functions, consolidating the SQL RPC, and renaming the cache tag. A few things need attention before merging.

---

### CRITICAL

**1. `gangOwnerId` empty-string fallback silently breaks `isOwner` for real owners**
`utils/user-permissions.ts` — `checkPermissionCached` / `deriveGangPermissions`

Ownership is derived as:
```ts
const isOwner = gangOwnerId === userId;
```

Any call site that can't resolve the gang owner and falls back to an empty string (e.g. `gang?.user_id ?? ''`) will cache and return `isOwner: false` for the actual gang owner — silently, with no error. Several call sites fetch gang ownership from a join that could fail (RLS-blocked, network error, gang not found), and the `?? ''` fallback pattern is the natural TypeScript idiom there.

**Fix:** Change the function signature to `gangOwnerId: string | null` and early-return safe-deny permissions when it's `null`, or throw so the caller is forced to handle the missing-owner case explicitly. Never accept `''` as a valid sentinel — it's indistinguishable from a resolved empty string.

---

### HIGH

**2. Duplicate SQL function definition**
`supabase/functions/check_permission.sql` (new standalone file) and `supabase/migrations/20260625120000_add_check_permission.sql` both define the same `check_permission` function with identical bodies.

`CREATE OR REPLACE` makes this idempotent, but it's a maintenance hazard: a future edit to one file won't affect the other. The authoritative definition should live in one place only. Convention for this repo is the migrations directory; the `supabase/functions/` file should either be removed or converted to a `DROP FUNCTION` comment pointing at the migration.

---

### MEDIUM

**3. Dead variable `v_resolved_campaign_id` in `check_permission` SQL**
`supabase/migrations/20260625120000_add_check_permission.sql` lines ~27–36 (and identically in `supabase/functions/check_permission.sql`)

```sql
DECLARE
  ...
  v_resolved_campaign_id UUID := NULL;   -- declared
BEGIN
  ...
  ELSIF p_gang_id IS NOT NULL THEN
    v_resolved_campaign_id := NULL;      -- assigned but never read
  END IF;
```

The variable is set but never referenced after assignment. It looks like a leftover from an earlier design that resolved `campaign_id` from `gang_id` before the direct join approach was chosen. Remove the declaration and the assignment to keep the function clean.

**4. Application-level authorization fully removed, RLS assumed correct**
`app/actions/add-fighter.ts`, `fighter-advancement.ts`, `fighter-injury.ts`, `move-from-stash.ts`, `move-to-stash.ts`, `sell-equipment.ts`, `update-gang-positioning.ts`, `add-gang-vehicle.ts`

All eight files remove their `isAdmin`/ownership checks with the comment "Authorization is enforced by RLS policies." That's a valid architecture, but this PR doesn't include a corresponding migration that adds or verifies the RLS policies, nor any policy audit. If a policy has a gap (wrong `USING` clause, missing `WITH CHECK`, not covering the mutating path), there is now no application-level backstop.

Before merging, explicitly confirm (or link the migration PR) that each mutating operation on these tables has a `WITH CHECK` RLS policy that enforces ownership. `move-from-stash.ts` is the highest-risk removal because it had both an admin check and an explicit gang-ownership gate.

**5. `unstable_cache` closes over `supabase` client**
`utils/user-permissions.ts` — `checkPermissionCached`

```ts
const supabase = await createClient();          // created outside cache

return unstable_cache(
  async (uid, gid, ownerId) => {
    const { data } = await supabase.rpc(...);   // closed over inside cache
  },
  ...
)
```

This is the same pattern as the deleted `PermissionService.getGangPermissions`, so it's not a regression. However, Next.js documentation warns that `unstable_cache` callbacks should not close over non-serializable request-scoped values like a Supabase client. On a cache hit, the callback is never re-invoked so it doesn't cause incorrect auth in practice, but on a cache miss in a concurrent environment it could pick up a stale client context. Worth moving `createClient()` inside the closure to make the cached function self-contained.

---

### LOW

**6. Cache invalidation is per-user, not per-gang**
`utils/cache-tags.ts` — `invalidatePermissionForUser`

The function signature is `{ userId, gangId }` — it clears the cache for one specific user. If gang ownership transfers (or a user's campaign role changes), callers must know to call this for every affected user. The old `invalidateGangPermissionsForUser` had the same constraint, but the rename is a good moment to document which events should trigger broad invalidation (e.g., gang transfer → invalidate old owner + new owner; campaign member removal → invalidate removed user).

**7. `CACHE_TAGS.USER_GANG_PERMISSIONS` fully removed — verify no stragglers**
`utils/cache-tags.ts`

The tag was renamed to `CHECK_PERMISSION`. All references in the diff are updated, but any file not touched by this PR that still calls `CACHE_TAGS.USER_GANG_PERMISSIONS(...)` would silently reference `undefined` (TypeScript would catch this at build time, but worth a quick grep to confirm the build is clean before merge).

---

### Positive notes

- The new `check_permission` RPC is cleaner than `get_gang_permissions`: returning `{ is_admin, campaign_role }` and leaving ownership to the caller (who already has `user_id`) removes a redundant DB column read and makes the function's contract easier to reason about.
- `deriveCampaignPermissions` / `deriveGangPermissions` as pure helper functions are testable in isolation — a significant improvement over the class methods.
- Deleting `app/lib/user-permissions.ts` (393 lines) and `supabase/functions/get_permissions.sql` (96 lines) is a net negative of ~500 LOC with no lost functionality. Well done.
- Migration timestamp `20260625120000` is correctly sequenced after the existing `20260622` and `20260623` migrations.

---

joeseos (commented):
## Follow-up Review — #1818: Refactored Permissions

@joeseos — the "Fixes" commit addresses part of the critical issue and part of the dead-SQL finding, but creates a new problem in the process. Details per finding below.

---

### Previous Findings

| # | Finding | Status |
|---|---|---|
| CRITICAL | `gangOwnerId` empty-string fallback silently breaks `isOwner` | ⚠️ Partial — see below |
| HIGH | Duplicate SQL function definition | ❌ Outstanding + regression introduced |
| MEDIUM 3 | Dead variable `v_resolved_campaign_id` in SQL | ⚠️ Partial — see below |
| MEDIUM 4 | App-level auth removed, RLS correctness unverified | ❌ Outstanding |
| MEDIUM 5 | `unstable_cache` closes over `supabase` client | ❌ Outstanding |
| LOW 6 | Cache invalidation is per-user | ❌ Outstanding (doc/process concern) |
| LOW 7 | Straggler `USER_GANG_PERMISSIONS` references | — Not verifiable from commit; TypeScript build should catch |

---

### Detail on Partial / Outstanding Items

---

#### CRITICAL — `gangOwnerId` empty-string: partially resolved

`fighter-skill-access.ts` and `app/api/fighters/skill-access/route.ts` are fixed (`?? ''` → `?? null`), and the `checkPermissionCached` / `deriveGangPermissions` signatures are correctly widened to `string | null`.

**Remaining concern:** `string` is assignable to `string | null` in TypeScript, so the compiler will not flag other call sites that still pass `someValue ?? ''`. The previous review noted "several call sites"; only 2 were touched here. A quick grep is needed before merge to confirm no `?? ''` fallback remains anywhere `checkPermissionCached` or `deriveGangPermissions` is called:

```bash
grep -rn "?? ''" app/actions app/api utils --include="*.ts" | grep -i "permission\|gangOwner\|user_id"
```

If there are other sites, this is still a silent data-integrity bug for gang owners.

---

#### HIGH — Duplicate SQL + new divergence regression

The "Fixes" commit edited `supabase/functions/check_permission.sql` (removed the dead variable) but did **not** touch `supabase/migrations/20260625120000_add_check_permission.sql`. The two definitions are now **out of sync**:

| File | `v_resolved_campaign_id` | Status |
|---|---|---|
| `supabase/functions/check_permission.sql` | Removed ✅ | Cleaned |
| `supabase/migrations/20260625120000_add_check_permission.sql` | Still present ❌ | **What actually runs in production** |

The migration is the definition that Supabase executes on deploy. The functions file being cleaner than the migration is the opposite of what was intended — it's now misleading to anyone reading the functions file as documentation.

**Fix:** Apply the same dead-variable removal to the migration file, then either remove `supabase/functions/check_permission.sql` entirely (the previous review's recommendation) or add a comment at the top: `-- Canonical definition: supabase/migrations/20260625120000_add_check_permission.sql`.

---

#### MEDIUM 3 — Dead variable: partially resolved

Removed from `supabase/functions/check_permission.sql` but the authoritative migration still contains:
```sql
DECLARE
  ...
  v_resolved_campaign_id UUID := NULL;   -- ← still here
BEGIN
  ...
  ELSIF p_gang_id IS NOT NULL THEN
    v_resolved_campaign_id := NULL;      -- ← still here
  END IF;
```

This is a one-line fix in the migration file — it just needs the same removal applied there.

---

#### MEDIUM 4 — App-level auth removed, RLS unverified

Not addressed in this commit. The 8 action files (`add-fighter.ts`, `fighter-advancement.ts`, `fighter-injury.ts`, `move-from-stash.ts`, `move-to-stash.ts`, `sell-equipment.ts`, `update-gang-positioning.ts`, `add-gang-vehicle.ts`) still carry no application-level ownership check, with the comment "Authorization is enforced by RLS policies." Before this merges, please confirm or link the migration/audit that verifies each of these operations has a `WITH CHECK` RLS policy covering the mutating path. `move-from-stash.ts` is the highest-risk removal.

---

#### MEDIUM 5 — `unstable_cache` closes over `supabase` client

Not addressed. `createClient()` is still called outside the cache callback and closed over inside it. This is not a regression from the previous code, but the refactor was a good opportunity to fix it. Moving `createClient()` inside the async callback makes the cached function self-contained:

```ts
return unstable_cache(
  async (uid: string, gid: string, ownerId: string | null) => {
    const supabase = await createClient();  // ← inside, not closed over
    const { data, error } = await supabase.rpc(...);
    ...
  },
  ...
)
```

---

### Verdict

**Changes requested** — two items need fixes before this is safe to merge:

1. **Apply the same `v_resolved_campaign_id` removal to the migration file** (the definitions are now diverged; the migration is what runs in production).
2. **Grep-verify no remaining `?? ''` fallbacks** for `gangOwnerId`/`checkPermissionCached`/`deriveGangPermissions` across the full codebase — or fix any found. TypeScript won't catch these since `string` satisfies `string | null`.

The `unstable_cache` closure (MEDIUM 5) and RLS confirmation (MEDIUM 4) can be follow-ups if you prefer to keep this PR narrowly scoped, but MEDIUM 4 in particular is worth documenting explicitly before the application-level checks disappear.

---

PR #1820, opened 2026-06-26
Fix hydration territory list

joeseos:


joeseos (commented):
## PR #1820 — Fix/hydration territory list Review

@joeseos — automated review. Two-file change: the main fix replaces a `useState<boolean | null>` + `useEffect` pattern with `useSyncExternalStore` to eliminate the hydration mismatch, and a minor type decoupling in `tooltip-renderers.tsx`.

The `useSyncExternalStore` implementation is mechanically correct: server snapshot returns `true`, client snapshot reads from localStorage defaulting to `true`, so the initial server render and client render always agree (both `true`), with the user's stored preference applied on the first client snapshot call. The subscribe/unsubscribe cycle is correct (event listener added/removed, cleanup function returned). No correctness bugs found in the core change.

---

### LOW

**1. `setListDefaultTerritories` is an inconsistent outlier — the three read helpers are module-level but the write helper uses `useCallback`**
`campaign-territory-list.tsx` — setter block in the diff

The three `useSyncExternalStore` helpers (`subscribeToTerritoryPref`, `getTerritoryPrefSnapshot`, `getTerritoryPrefServerSnapshot`) are correctly hoisted to module scope for stable references. `setListDefaultTerritories` needs the same stable-reference guarantee (it's passed to `onCheckedChange`), and uses `useCallback([], [])` to achieve it — but the empty deps array is only needed because it's defined inside the component. The function closes over nothing but module-level constants; it could be hoisted to module scope alongside the other three helpers, dropping the `useCallback` entirely.

```ts
// at module level, alongside the other helpers:
function setTerritoryPref(value: boolean) {
  try {
    localStorage.setItem(LIST_DEFAULT_TERRITORIES_STORAGE_KEY, String(value));
  } catch {}
  window.dispatchEvent(new Event(TERRITORY_PREF_EVENT));
}
```

No runtime impact — just a consistency issue.

**2. `TooltipRenderArgs.content` widened from `ITooltip['content']` to `ReactNode`**
`components/ui/tooltip-renderers.tsx` — line 6

`ITooltip['content']` in react-tooltip is `string | null | undefined`. `ReactNode` is a superset that additionally includes `number`, `boolean`, arrays, and portals — values the library will never pass at runtime. The widening is safe today (verified: `content` is only used as `body` in JSX where any `ReactNode` renders fine), and it avoids an import dependency on a library's internal type, which was the motivation. 

The trade-off: if a future render function uses `content` as a string (e.g., calling `.trim()` or `.split()`), TypeScript will now require an explicit narrowing check where it previously didn't. If you want the best of both worlds, `string | Element | null | undefined` is more precise than `ReactNode` and still decouples from `ITooltip`.

---

### Informational (no action needed)

**Cross-tab sync is not handled** — the custom `TERRITORY_PREF_EVENT` fires only on the originating `window`, so two tabs showing the same campaign won't stay in sync when the checkbox is toggled in one. The old `useState` approach had the same limitation. If cross-tab sync is ever wanted, subscribing to `window.addEventListener('storage', ...)` for the `LIST_DEFAULT_TERRITORIES_STORAGE_KEY` key would handle it.

---

### Positive notes
- The "Loading territories..." null-state placeholder is correctly eliminated — since the server snapshot and the default client snapshot both return `true`, there is no render where the state is unknown.
- The `disabled={listDefaultTerritories === null}` checkbox guard is also correctly removed; the checkbox is always enabled from the first render.
- The legacy key migration path (`LEGACY_HIDE_DEFAULT_TERRITORIES_STORAGE_KEY`) is preserved correctly in `getTerritoryPrefSnapshot` — existing users with the old key will still have their preference respected.

---

joeseos:
**Addendum — missed finding from deeper analysis:**

### MEDIUM — `setListDefaultTerritories` silently reverts when localStorage is unavailable

`campaign-territory-list.tsx` — new setter callback

```ts
const setListDefaultTerritories = useCallback((value: boolean) => {
  try {
    localStorage.setItem(LIST_DEFAULT_TERRITORIES_STORAGE_KEY, String(value));
  } catch {}
  window.dispatchEvent(new Event(TERRITORY_PREF_EVENT));  // ← outside the try
}, []);
```

The event dispatch is unconditional — it fires even when `setItem` threw (private browsing, storage quota, `SecurityError`). `useSyncExternalStore` then calls `getTerritoryPrefSnapshot`, which re-reads localStorage and gets the **old** value (since the write failed). The checkbox's `checked={listDefaultTerritories}` reverts to whatever was stored before, with no error shown to the user.

This is a **regression** from the old `useState` approach. The old code called `setListDefaultTerritories(value)` as a plain React state setter — the UI updated immediately regardless of whether localStorage was writable, and the `useEffect` write to storage failed silently but the in-memory state was correct. With `useSyncExternalStore`, the UI is driven entirely by the external store; if the store can't be written, the toggle appears to do nothing.

**Fix options:**
- Option A (minimal): skip the dispatch when the write failed:
  ```ts
  const setListDefaultTerritories = useCallback((value: boolean) => {
    try {
      localStorage.setItem(LIST_DEFAULT_TERRITORIES_STORAGE_KEY, String(value));
      window.dispatchEvent(new Event(TERRITORY_PREF_EVENT));
    } catch {}
  }, []);
  ```
- Option B (resilient): keep a fallback in-memory state for the storage-unavailable case, falling back to `useState` when `localStorage` throws.

Option A is the simpler fix and matches the expected behavior: if the write failed, there's nothing to notify about.

---

joeseos (commented):
## Follow-up Review — #1820: Fix hydration territory list

@joeseos — no new commits have been pushed since the previous review (HEAD is still `709872d`). Closing out the findings for record.

---

### Previous Findings

| # | Finding | Status |
|---|---|---|
| LOW 1 | `setListDefaultTerritories` uses `useCallback([], [])` inside component instead of module-scope function | ❌ Not addressed — still open, no runtime impact |
| LOW 2 | `TooltipRenderArgs.content` widened to `ReactNode` instead of more precise `string \| Element \| null \| undefined` | ❌ Not addressed — still open, safe as-is |
| Informational | Cross-tab sync not handled | — No action needed, same as before |

Both open items are purely cosmetic/precision concerns with no runtime or correctness impact. Neither was ever a blocker.

---

### Verdict

**Ready to merge.** The core `useSyncExternalStore` fix is correct, the hydration mismatch is eliminated, the legacy key migration path is preserved, and Vercel shows a successful preview deployment. The two LOW items are optional follow-ups that can be addressed in a separate cleanup pass if desired.

---

PR #1824, opened 2026-06-27
Added useQuery to the promotion modal and changed the pre-selection

joeseos:


joeseos (commented):
## PR Review — `refac/added-useQuery-to-promotion`

The prop-threading removal (`onEnsureFighterTypes` drilled through `fighter-page → fighter-advancement-list → FighterPromotionModal`) is a solid structural improvement, and the `queryKey` construction correctly captures all three gang identity fields. However the migration from an awaited imperative fetch to an eager `useQuery` introduces a race condition and drops error feedback.

---

### MEDIUM/HIGH — Race condition: modal can open before `useQuery` resolves

**`components/fighter/fighter-advancement-list.tsx`** — promotion button handlers

Old flow (sequential):
```ts
onClick={async () => {
  await onEnsureFighterTypes();   // guaranteed data before modal opens
  setGangerPromotionOpen(true);
}}
```

New flow (concurrent):
```ts
onClick={() => setGangerPromotionOpen(true)}   // modal opens immediately
```

`useQuery` fires on mount but there is no guarantee it has resolved by the time the user clicks a promotion button — especially on slow networks or first visit. When `preFetchedFighterTypes` is still `[]`, the modal renders with no eligible types, the confirm button is disabled, and there is no loading indicator. The user sees a broken modal with no explanation.

Both "Promote to Specialist" (ganger flow) and "Promote to Champion" share this path.

**Fix options (in order of preference):**
1. Pass `isFetching` from `useQuery` and disable the promotion buttons while fetching, with a spinner.
2. Show a loading state inside `FighterPromotionModal` when `preFetchedFighterTypes` is empty and `isFetching` is true.
3. Use `enabled: false` + manual `refetch()` on button click, preserving the sequential guarantee (closest to old behaviour).

---

### MEDIUM — Error handling regression: silent failure on fetch error

**`components/fighter/fighter-page.tsx`** — `useQuery` definition

Old code caught errors and showed a toast:
```ts
} catch (error) {
  toast.error('Error', { description: 'Could not fetch fighter types.' });
}
```

New `useQuery` has no `onError` callback and no global `QueryClient` error handler, so a network failure silently leaves `preFetchedFighterTypes = []`. The user clicks "Promote" and sees an empty modal with no explanation.

**Fix:** add an `onError` or use the `meta` pattern:
```ts
useQuery({
  ...
  meta: { errorMessage: 'Could not fetch fighter types.' },
})
```
and handle in a global `QueryCache` error handler — or add a `useEffect` watching `isError`.

---

### MEDIUM — UX regression: pre-selection removed, confirm button starts disabled

**`components/fighter/edit-fighter/fighter-promotion-modal.tsx`** — `handleIncludeAllGangFighterTypesChange`

Old: toggling the "include all" checkbox pre-selected the first eligible type, so the modal was immediately submittable.

New: both branches now call `setSelectedTypeId(''); setNewSpecialRules([])`, clearing selection. Users must make an additional click to confirm a promotion they already indicated intent for.

This may be intentional (cleaner UX, avoids accidental first-item selection), but it is a behaviour change worth noting. If intentional, document why in a comment or commit message.

---

### LOW — `staleTime` exceeds `gcTime`, causing unnecessary refetches

**`components/fighter/fighter-page.tsx`**

```ts
staleTime: 10 * 60 * 1000,   // 10 minutes
// gcTime not set → defaults to 5 minutes
```

Between 5–10 minutes after the component unmounts, the cache entry is garbage-collected but the data is still considered "fresh". On the next mount, React Query finds no cached entry and refetches, even though the data has not expired. Set `gcTime` to match or exceed `staleTime`:

```ts
staleTime: 10 * 60 * 1000,
gcTime:    10 * 60 * 1000,
```

---

### LOW — Eager fetch runs for every fighter page, including non-editable fighters

**`components/fighter/fighter-page.tsx`**

The `enabled` condition is:
```ts
enabled: !!gangId && !!(gangTypeId || customGangTypeId),
```

This fires for read-only visitors, retired fighters, and killed fighters — none of whom can trigger a promotion. If permissions are already available client-side, adding `&& (userPermissions?.canEdit ?? true)` would eliminate the unnecessary request. (`?? true` keeps the old eager-load behaviour until permissions are known.)

---

### Positive

- Removing the `onEnsureFighterTypes` prop chain is the right call — it was awkward to thread through two component boundaries.
- `queryKey: ['fighter-types-edit', gangId, gangTypeId, customGangTypeId]` correctly namespaces the key and will deduplicate concurrent requests from the same gang.
- The `enabled` guard correctly mirrors the old condition that gated `onEnsureFighterTypes` calls.
- Removing `!!onEnsureFighterTypes` from `gangerModalRollBuy` is correct — the ganger advancement UI should not depend on whether a fetch callback is wired up.

---

joeseos (commented):
## Follow-up Review — #1824: Added useQuery to the promotion modal

@joeseos — the "Fixes" commit is a meaningful restructure: moving the query from `fighter-page.tsx` into the components that actually need it, and gating it on modal-open state, is a solid improvement over the original approach. One prior finding is fully resolved, two are partially addressed, and two remain open. One new low-severity issue introduced.

---

### Previous Findings

| # | Finding | Status |
|---|---|---|
| MEDIUM/HIGH | Race condition: modal opens before `useQuery` resolves | ⚠️ Partial — see below |
| MEDIUM | Silent failure on fetch error | ❌ Outstanding |
| MEDIUM | Pre-selection UX regression (confirm button starts disabled) | ❌ Outstanding / needs intent confirmation |
| LOW | `staleTime` (10 min) exceeds default `gcTime` (5 min) | ❌ Outstanding — now in 3 places |
| LOW | Eager fetch fires for all fighter pages including non-editable | ✅ Resolved — fetch now gated on modal-open state |

---

### Detail

#### MEDIUM/HIGH — Race condition: partially improved, not resolved

The fetch is now lazy — it only fires when the promotion modal opens (via `enabled: ... && (gangerPromotionOpen || championPromotionOpen)`). That eliminates the original "query fires on page load, button click beats the network" scenario. Good.

However, the loading window gap is unchanged. When the modal opens and the data is **not** cached, `preFetchedFighterTypes` starts as `[]` and the query fires concurrently. `FighterPromotionModal` still receives an empty array with no loading indicator — the user still sees a broken modal with no eligible types and no explanation during this window.

The fix to fully close this is the same as before: pass `isFetching` from the query into `FighterPromotionModal` and show a spinner or disabled state while loading:

```ts
const { data: preFetchedFighterTypes = [], isFetching: isFetchingTypes } = useQuery({ ... });

// pass isFetchingTypes to FighterPromotionModal:
<FighterPromotionModal
  ...
  isLoading={isFetchingTypes}
/>
```

On cache hits (which will be the common case after first open, given `staleTime: 10min`) the window is zero. The concern is the first cold open.

---

#### MEDIUM — Error handling: still silent

All three new `useQuery` calls (`AdvancementModal`, `AdvancementsList`, `EditFighterModal`) throw on `!response.ok` but have no error UI path. A failed fetch leaves the modal with an empty type list and no toast or message.

Quick fix — add a `useEffect` watching `isError` in each component:
```ts
const { data: preFetchedFighterTypes = [], isError } = useQuery({ ... });

useEffect(() => {
  if (isError) toast.error('Could not fetch fighter types.');
}, [isError]);
```

---

#### MEDIUM — Pre-selection UX: intent confirmation needed

`fighter-promotion-modal.tsx` is not touched. Both branches of `handleIncludeAllGangFighterTypesChange` still call `setSelectedTypeId(''); setNewSpecialRules([])`, leaving the user with a modal they cannot confirm without an extra click. If this is intentional (cleaner UX, avoids accidental first-item auto-select), a one-line comment in the code would close this finding. If not intentional, restoring the pre-selection of the first eligible type is the fix.

---

#### LOW — `staleTime > gcTime`: now in 3 places instead of 1

`AdvancementModal`, `AdvancementsList`, and `EditFighterModal` all have:
```ts
staleTime: 10 * 60 * 1000,
// gcTime omitted → defaults to 5 minutes
```

Between 5–10 minutes after unmount the cache entry is GC'd, so the next open refetches despite the data still being considered fresh. Add `gcTime` to all three (or extract to a shared hook — see below):
```ts
staleTime: 10 * 60 * 1000,
gcTime:    10 * 60 * 1000,
```

---

### New Finding

#### 🟢 LOW — Identical `useQuery` duplicated in `AdvancementModal` and `AdvancementsList`

Both components define byte-for-byte the same `queryKey`, `queryFn`, and `staleTime`. React Query deduplicates the network request (same key → shared cache entry), but the code itself is duplicated. A small custom hook would eliminate the repetition and make the `gcTime` fix a one-place change:

```ts
// hooks/use-fighter-types.ts
export function useFighterTypes(
  gangId: string,
  gangTypeId: string,
  customGangTypeId: string,
  enabled: boolean,
) {
  return useQuery({
    queryKey: ['fighter-types-edit', gangId, gangTypeId, customGangTypeId],
    queryFn: async () => {
      const params = new URLSearchParams({ gang_id: gangId, is_gang_addition: 'false' });
      if (gangTypeId) params.set('gang_type_id', gangTypeId);
      if (customGangTypeId) params.set('custom_gang_type_id', customGangTypeId);
      const response = await fetch(`/api/fighter-types?${params}`);
      if (!response.ok) throw new Error('Failed to fetch fighter types');
      return response.json();
    },
    enabled: !!gangId && !!(gangTypeId || customGangTypeId) && enabled,
    staleTime: 10 * 60 * 1000,
    gcTime:    10 * 60 * 1000,
  });
}
```

Not a blocker, but would make error handling and `gcTime` changes a single-file fix.

---

### Verdict

**Changes requested** — two items before merge:

1. **Race condition / loading state** (MEDIUM): pass `isFetching` into `FighterPromotionModal` and show a loading state during the cold-open window. The lazy-fetch restructure is correct; this is the last piece.
2. **Silent error handling** (MEDIUM): add a `useEffect` watching `isError` (or equivalent) in `AdvancementModal`, `AdvancementsList`, and `EditFighterModal`.

Finding 3 (pre-selection) needs a one-line comment confirming intent if it's a deliberate design choice. Finding 4 (`gcTime`) and the new LOW (dedup hook) can be bundled into the same pass as the above two.

---

PR #1825, opened 2026-06-27
Removed router.refresh in some places

joeseos:


joeseos (commented):
## PR Review — #1825: Removed router.refresh in some places

@joeseos — the direction here is right: replacing full-page `router.refresh()` calls with targeted cache invalidation + local state updates is the correct architecture. The `CopyFighterModal` and `fighter-actions.tsx` changes are clean. The friend-management changes have three correctness gaps to close before merge.

---

### HIGH

#### 1. Adding a friend request never invalidates the server cache — stale data persists forever

**File:** `components/account/friends.tsx` — `addFriend` handler

The `addFriend` function performs a direct Supabase client insert (not via a server action), optimistically updates `localFriends`, and returns. `app/actions/friends.ts` added `revalidateTag` to `accept`, `decline`, and `delete` — but not to `add`. Since `getFriendsAndRequests` uses `revalidate: false` (infinite TTL), the cache for the sending user will never reflect the new pending outgoing request until one of the other actions triggers a `revalidateTag`.

**Concrete failure:** User A sends a friend request to User B. `localFriends` correctly shows it as pending during the current session. User A navigates away and returns — the server component re-renders using the cached (stale) data, and the pending request no longer appears. It stays hidden indefinitely until User B accepts/declines (which then triggers `revalidateTag(USER_FRIENDS(A))` via the accept/decline server action).

**Fix:** Extract friend-request insertion into a server action and call `revalidateTag(CACHE_TAGS.USER_FRIENDS(userId))` after a successful insert:

```ts
// app/actions/friends.ts
export async function addFriendRequest(requesterId: string, addresseeId: string) {
  const supabase = await createClient();
  await getAuthenticatedUser(supabase);
  const { error } = await supabase.from('friends').insert({ ... });
  if (error) throw new Error(error.message);
  revalidateTag(CACHE_TAGS.USER_FRIENDS(requesterId));
  revalidateTag(CACHE_TAGS.USER_FRIENDS(addresseeId));
  return { success: true };
}
```

---

### MEDIUM

#### 2. Accepting a friend request in `NotificationsContent` leaves `FriendsSearchBar` stale

**Files:** `components/account/notifications-content.tsx`, `components/account/friends.tsx`

When a friend request is accepted via the notifications panel:
- The notification is correctly removed from `NotificationsContent`'s local state ✓
- `revalidateTag(USER_FRIENDS(...))` correctly marks the server cache as stale ✓
- **But** `FriendsSearchBar.localFriends` is never updated

`FriendsSearchBar` is seeded from the `initialFriends` prop on mount. It has no way to know that a friend request was just accepted elsewhere on the page. Without `router.refresh()`, the friends list continues showing the request as "pending incoming" until a full page reload. The user has to manually refresh to see the newly accepted friend appear as "accepted".

The old `router.refresh()` re-rendered the RSC tree, passing fresh `initialFriends`. The new approach correctly invalidates the cache but provides no mechanism to propagate the change within the current page.

**Fix options:**
1. Lift `localFriends` state to the parent page and pass a setter down to both components, so `NotificationsContent` can update it directly.
2. Keep a lightweight `router.refresh()` specifically for the accept/decline notification actions (just these two, where cross-component sync is needed).
3. Use a client-side event (e.g. a custom `window` event or a React context) that `FriendsSearchBar` subscribes to.

Option 2 is the lowest-effort fix and is consistent with the PR's goal of removing `router.refresh()` only where it's unnecessary.

---

#### 3. `unstable_cache` closes over request-scoped `supabase` client

**File:** `app/lib/friends.ts` — `getFriendsAndRequests`

```ts
export const getFriendsAndRequests = async (userId: string, supabase: any) => {
  return unstable_cache(
    async () => {
      const { data } = await supabase  // ← closed over from outer scope
        .from('friends')...
    },
    [`user-friends-${userId}`],
    { tags: [CACHE_TAGS.USER_FRIENDS(userId)], revalidate: false }
  )();
};
```

Next.js [documents](https://nextjs.org/docs/app/api-reference/functions/unstable_cache) that `unstable_cache` callbacks should not close over non-serializable, request-scoped values. The `supabase` client is request-scoped and typed as `any`, losing all type safety at the call boundary. This pattern works in practice with `revalidate: false` (cache misses always use the current request's client), but:
- Any future change to `revalidate` (e.g. a timed refresh) would cause background revalidations to fail silently since the original request's supabase client is gone
- `supabase: any` defeats TypeScript's Supabase type checking for the entire query

**Fix:** Create the client inside the callback, and drop the `supabase` parameter:

```ts
export const getFriendsAndRequests = (userId: string) =>
  unstable_cache(
    async () => {
      const supabase = await createClient();
      const { data, error } = await supabase.from('friends')...
    },
    [`user-friends-${userId}`],
    { tags: [CACHE_TAGS.USER_FRIENDS(userId)], revalidate: false }
  )();
```

Update `app/account/page.tsx` to call `getFriendsAndRequests(user.id)` (no supabase arg).

---

#### 4. `'use server'` removed from `app/lib/friends.ts` — verify `getAcceptedFriends` call sites

**File:** `app/lib/friends.ts`

The `'use server'` directive was removed from the top of the file. `getAcceptedFriends` (which uses `createClient()` from `@/utils/supabase/server`) remains in the file. If any client component imports `getAcceptedFriends` and calls it directly (the pattern that `'use server'` enables), that call site will now fail at runtime — client components cannot execute `createClient()` from the server package.

**Action:** Grep for `getAcceptedFriends` imports across the codebase:
```bash
grep -rn "getAcceptedFriends" . --include="*.ts" --include="*.tsx"
```
If any client component imports it, either restore `'use server'` or move `getAcceptedFriends` to a separate `'use server'` file.

---

### LOW

#### 5. `router` is likely unused in `notifications-content.tsx`

**File:** `components/account/notifications-content.tsx`

All four `router.refresh()` calls were replaced with `setNotifications(...)`. If `router` is not used for navigation elsewhere in the file, the `const router = useRouter()` declaration is now dead and will produce a lint warning. Verify and remove if unused.

---

### What's working well

- **`CopyFighterModal` + `fighter-actions.tsx`**: replacing `router.refresh()` with `onSuccess?.()` → `onFighterUpdate?.()` is the correct event-propagation approach; no data-model concerns here.
- **`revalidateTag` in `accept`/`decline`/`deleteFriend`**: correct — both users' caches are invalidated symmetrically.
- **`USER_FRIENDS` cache tag**: cleanly follows the existing `CACHE_TAGS` pattern.
- **Snapshot rollback in `handleDeleteFriend`**: `const previousFriends = localFriends` captured before the optimistic update, restored on catch — correct.
- **`prevFriends` derived-state pattern**: the `if (initialFriends !== prevFriends)` pattern is the React-recommended way to sync state from a changing prop. Works correctly here.
- **Removing `router.refresh()` from the delete-success path**: correct — local state removal + `revalidateTag` (via the server action) handles both the UI and the cache cleanly.

---

### Verdict

**Changes requested** — three items before merge:

1. **Add `revalidateTag` for friend-request insertion** (HIGH) — extract to a server action and invalidate both users' caches.
2. **Fix `FriendsSearchBar` desync after notification accept** (MEDIUM) — the simplest fix is a targeted `router.refresh()` in the accept/decline notification handlers.
3. **Move `createClient()` inside `unstable_cache` callback** (MEDIUM) — drop `supabase: any` parameter, create client inside the cache boundary.

Finding 4 (`'use server'` / `getAcceptedFriends`) needs a grep to confirm — could be a no-op if it's only called from server code.

---

PR #1833, opened 2026-06-29
Moved custom to unstable cache and removed revalidatePath

joeseos:


joeseos (commented):
## PR Review — `refac/moved-custom-to-unstable-cache`

This PR is the right architectural move: replacing broad `revalidatePath('/')` with targeted `unstable_cache` + `revalidateTag` invalidation makes cache busts surgical and avoids unnecessary full-page invalidations. The cross-entity invalidation is handled correctly in most cases — `updateCustomGangType` correctly invalidates both `USER_CUSTOM_GANG_TYPES` and `USER_CUSTOM_FIGHTERS` (since fighter types join the gang type name), and all delete operations correctly chain into `invalidateUserCustomCollections`. There are three gaps to address before this is complete.

---

### MEDIUM — All five `update*` operations fail to invalidate `USER_CUSTOM_COLLECTIONS`

`getUserCustomCollections` resolves display names for every collection item at cache-fill time (separate queries into the equipment, fighter_type, gang_type, skill, and trading_post tables), and the result — with embedded names — is cached under `USER_CUSTOM_COLLECTIONS(userId)` with `revalidate: false`.

The delete operations correctly call `invalidateUserCustomCollections`, but none of the update (rename) operations do:

| Action | Invalidates | Missing |
|---|---|---|
| `updateCustomEquipment` | `USER_CUSTOM_EQUIPMENT` | `USER_CUSTOM_COLLECTIONS` |
| `updateCustomFighter` | `USER_CUSTOM_FIGHTERS` | `USER_CUSTOM_COLLECTIONS` |
| `updateCustomGangType` | `USER_CUSTOM_GANG_TYPES` + `USER_CUSTOM_FIGHTERS` | `USER_CUSTOM_COLLECTIONS` |
| `updateCustomSkill` | `USER_CUSTOM_SKILLS` | `USER_CUSTOM_COLLECTIONS` |
| `updateCustomTradingPost` | `USER_CUSTOM_TRADING_POSTS` | `USER_CUSTOM_COLLECTIONS` |

**Trigger:** User renames a custom equipment item → `invalidateUserCustomEquipment` flushes the equipment list → `getUserCustomCollections` cache is not touched → the Collections tab continues showing the old name indefinitely (no TTL, no automatic expiry).

**Fix:** Add `invalidateUserCustomCollections(user.id)` to all five update handlers (the same call already present in their corresponding delete handlers).

---

### MEDIUM — Eight trading-post sub-resource mutations drop `revalidatePath('/')` with no replacement

`addTPEquipmentBatch`, `updateTPEquipment`, `removeTPEquipment`, `addAvailabilityRule`, `deleteAvailabilityRule`, `addPricingRule`, `deletePricingRule`, and `saveEquipmentRules` all lose `revalidatePath('/')` and receive no targeted `invalidateUserCustomTradingPosts` call.

`getUserCustomTradingPosts` only caches the top-level trading post record (name, description, timestamps) — not equipment items, pricing rules, or availability rules — so its own cache entry won't be stale. The risk is that these sub-resources appear in other `unstable_cache`-wrapped functions (gang data, campaign data) that previously relied on `revalidatePath('/')` to be refreshed. After this PR those caches have no invalidation path from TP sub-resource mutations.

**Fix:** At minimum add `invalidateUserCustomTradingPosts(user.id)` to each of the eight functions above as a conservative starting point, then trace whether gang/campaign cached functions include TP equipment/pricing/availability data and invalidate those tags too if needed.

---

### MEDIUM — `saveCustomWeaponProfiles` drops `revalidatePath('/')` with no replacement

`custom-weapon-profiles.ts` removes `revalidatePath('/')` from `saveCustomWeaponProfiles` but adds no targeted cache invalidation. Weapon profiles are not stored in `getUserCustomEquipment` (which only fetches `custom_equipment.*`), but they are fetched as sub-selects inside the `unstable_cache`-wrapped gang and fighter data functions. After a weapon profile save, those gang/fighter cached entries are never invalidated.

**Fix:** Call the appropriate gang/fighter cache invalidation from `saveCustomWeaponProfiles` — this requires the `gangId` or `fighterId` that owns the equipment, which should be available at the call site.

---

### LOW — `createClient()` called outside the `unstable_cache` callback in all six lib files

All six files use this pattern:

```ts
export async function getUserCustomEquipment(userId: string) {
  const supabase = await createClient();   // ← runs on every call, including cache hits

  return unstable_cache(
    async () => {
      const { data } = await supabase.from('custom_equipment')...  // captured by closure
    },
    [`user-custom-equipment-${userId}`],
    { tags: [CACHE_TAGS.USER_CUSTOM_EQUIPMENT(userId)], revalidate: false }
  )();
}
```

On cache hits (the common path) `createClient()` constructs a Supabase SSR client from the request's cookie store and then the client is silently discarded. This is wasteful but not a correctness issue — `createClient()` returns an unauthenticated client gracefully rather than throwing on absent/expired sessions.

The idiomatic fix is to move `createClient()` inside the callback so it only runs on cache misses:

```ts
export async function getUserCustomEquipment(userId: string) {
  return unstable_cache(
    async () => {
      const supabase = await createClient();   // ← only on cache miss
      const { data } = await supabase.from('custom_equipment')...
    },
    [`user-custom-equipment-${userId}`],
    { tags: [CACHE_TAGS.USER_CUSTOM_EQUIPMENT(userId)], revalidate: false }
  )();
}
```

This is the same change needed in all six `app/lib/customise/*.ts` files.

---

### LOW — `getUserCustomEquipmentByCategory` not wrapped in `unstable_cache`

`custom-equipment.ts` exports two functions: `getUserCustomEquipment` (now cached) and `getUserCustomEquipmentByCategory` (still direct Supabase query, no cache). Any component using the category-filtered path bypasses the cache entirely — `invalidateUserCustomEquipment` has no effect on it. After a `createCustomEquipment` call that invalidates `USER_CUSTOM_EQUIPMENT`, the cached list refreshes but the category-filtered view doesn't benefit from the cache at all.

If the category path is used on the same page as the full list, it should either be wrapped in its own `unstable_cache` entry (with the same tag so invalidation covers it) or folded into `getUserCustomEquipment` with client-side filtering.

---

PR #1834, opened 2026-06-29
Fixed image flash on the homepage

joeseos:


joeseos:
## Code Review — PR #1834 "Fixed image flash on the homepage"

@joeseos

### Summary

The root cause is correctly identified and the fix is sound. The previous `isMounted` guard caused a two-phase render: SSR/hydration produced a plain `<ul>/<li>/<CampaignCardContent>` tree, then after mount `isMounted` flipped to `true` and React swapped it for a completely different `DndContext/SortableContext/SortableCampaignCard` tree. That DOM swap is what forced images to reload and caused the flash.

By always rendering the DnD tree (both on server and client), the SSR output matches the hydration output, and the flash disappears.

---

### Findings

**[LOW] Dead code — `hooks/use-is-mounted.ts`**

Both of the files in this PR were the only consumers of `useIsMounted`. The hook file at `hooks/use-is-mounted.ts` is now unreferenced and can be deleted.

---

**[LOW] SSR safety note (for awareness, not a blocker)**

`SortableCampaignCard` spreads `dragAttributes` (which includes `aria-describedby`, `role="button"`, etc.) onto the card div. These IDs are generated by dnd-kit using React 18's `useId()` internally, which produces stable and consistent IDs between SSR and client hydration — so no hydration mismatch is expected. Just worth keeping in mind if the dnd-kit version is ever downgraded.

---

### Verdict

✅ **Approve** — correct fix, clean diff. Follow-up: delete `hooks/use-is-mounted.ts`.

---

PR #1835, opened 2026-06-29
Fixed the equipment tooltip closing effect that was also passing thro…

Djidiouf:
…ugh the click: collapsing another equipment section that was under the tooltip for example

### DIFF
--- a/components/equipment/equipment-tooltip.tsx
+++ b/components/equipment/equipment-tooltip.tsx
@@ -1,8 +1,20 @@
 'use client';
 
 import React from 'react';
-import { Tooltip } from 'react-tooltip';
+import { Tooltip, type TooltipRefProps } from 'react-tooltip';
 import { Equipment, WeaponProfile } from '@/types/equipment';
+
+function useCoarsePointer() {
+  return React.useSyncExternalStore(
+    (onStoreChange) => {
+      const mediaQuery = window.matchMedia('(hover: none) and (pointer: coarse)');
+      mediaQuery.addEventListener('change', onStoreChange);
+      return () => mediaQuery.removeEventListener('change', onStoreChange);
+    },
+    () => window.matchMedia('(hover: none) and (pointer: coarse)').matches,
+    () => false,
+  );
+}
 
 export interface EquipmentTooltipOptions {
   equipmentListType?: 'fighters-list' | 'fighters-tradingpost' | 'unrestricted';
@@ -172,7 +184,34 @@ interface EquipmentTooltipTriggerProps {
 
 export function EquipmentTooltipTrigger({ item, children, className, options }: EquipmentTooltipTriggerProps) {
   const tooltipId = React.useId();
+  const tooltipRef = React.useRef<TooltipRefProps>(null);
+  const touchStartRef = React.useRef<{ x: number; y: number } | null>(null);
+  const isCoarsePointer = useCoarsePointer();
   const hasTooltipContent = hasEquipmentTooltipContent(item, options);
+
+  const handleMobileTooltipTouchStart = React.useCallback((event: React.TouchEvent) => {
+    const touch = event.changedTouches[0];
+    touchStartRef.current = { x: touch.clientX, y: touch.clientY };
+  }, []);
+
+  const handleMobileTooltipTouchEnd = React.useCallback((event: React.TouchEvent) => {
+    if (!touchStartRef.current) {
+      return;
+    }
+
+    const touch = event.changedTouches[0];
+    const deltaX = Math.abs(touch.clientX - touchStartRef.current.x);
+    const deltaY = Math.abs(touch.clientY - touchStartRef.current.y);
+    touchStartRef.current = null;
+
+    if (deltaX > 10 || deltaY > 10) {
+      return;
+    }
+
+    event.preventDefault();
+    event.stopPropagation();
+    tooltipRef.current?.close();
+  }, []);
 
   if (!hasTooltipContent) {
     return <div className={className}>{children}</div>;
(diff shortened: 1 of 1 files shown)


joeseos:
## Code Review — PR #1835 "Fixed equipment tooltip closing effect on mobile"

@Djidiouf

The fix addresses a real UX problem and the overall approach (detect coarse pointer, use touch events to dismiss without click-through) is the right one. Two issues need attention before merging.

---

### Findings

#### [MEDIUM] `useCoarsePointer` — unstable `subscribe` reference causes a listener leak per render

`useSyncExternalStore` documents that if a **different** `subscribe` function reference is passed between renders, React will unsubscribe the old one and re-subscribe with the new one. Because the subscribe arrow function is defined inline inside `useCoarsePointer`, a new function object is created on every call — so every render of every `EquipmentTooltipTrigger` that has tooltip content adds a `change` listener and removes the previous one.

In a list of 20 items this fires 20 times per render cycle. In a worst case with a busy parent re-rendering frequently (drag-and-drop, filter input, etc.), this produces a steady drip of MediaQueryList event listener churn.

**Fix — lift the callbacks to module level:**

```ts
const COARSE_MEDIA_QUERY = '(hover: none) and (pointer: coarse)';

const subscribeCoarsePointer = (onStoreChange: () => void) => {
  const mq = window.matchMedia(COARSE_MEDIA_QUERY);
  mq.addEventListener('change', onStoreChange);
  return () => mq.removeEventListener('change', onStoreChange);
};

const getCoarsePointerSnapshot = () =>
  window.matchMedia(COARSE_MEDIA_QUERY).matches;

const getCoarsePointerServerSnapshot = () => false;

function useCoarsePointer() {
  return React.useSyncExternalStore(
    subscribeCoarsePointer,
    getCoarsePointerSnapshot,
    getCoarsePointerServerSnapshot,
  );
}
```

The same pattern is used by the existing `useIsMounted` hook in this codebase (`noopSubscribe`, `getClientSnapshot`, `getServerSnapshot` at module level) — this change would match that convention.

---

#### [MEDIUM] `positionStrategy="fixed"` is applied unconditionally

`positionStrategy="fixed"` is always set on the `<Tooltip>`, not just for mobile. On desktop, `position: fixed` means the tooltip is positioned relative to the viewport rather than its scroll container. If any fighter list / trading post section is inside a scrollable overflow container, the tooltip will stay pinned to its original viewport position while the user scrolls the content beneath it — a visual bug that would be hard to notice without testing in those layouts.

**Fix — conditionalise it the same way the other mobile props are gated:**

```tsx
positionStrategy={isCoarsePointer ? 'fixed' : 'absolute'}
```

Or simply remove it for the desktop case (the default `react-tooltip` strategy is already `absolute`).

---

#### [LOW] `getCoarsePointerSnapshot` creates a new `matchMedia` object on every snapshot call

`window.matchMedia(...)` is synchronous and fast, but it allocates a `MediaQueryList` object each time React calls `getSnapshot` to check whether a re-render is needed. With the module-level refactor above this is naturally resolved by caching the `MediaQueryList` in the subscribe closure, or by holding a module-level reference.

---

### Verdict

⚠️ **Changes requested** — the touch-dismiss logic itself is correct and the overall structure is sound. The two medium issues above are worth fixing before merge: the subscribe stability is a correctness concern at scale, and the unconditional `positionStrategy="fixed"` is a regression risk for desktop users in scrollable containers.

---

joeseos:
## Code Review (updated) — PR #1835

@Djidiouf — reviewing commit `32d9538`, the follow-up to the changes-requested round.

---

### Previous findings

| Finding | Status |
|---|---|
| [MEDIUM] `subscribeCoarsePointer` inline arrow caused listener churn | ✅ **Resolved** — all three callbacks lifted to module level (`subscribeCoarsePointer`, `getCoarsePointerSnapshot`, `getCoarsePointerServerSnapshot`), matching the existing `useIsMounted` convention |
| [MEDIUM] `positionStrategy="fixed"` applied unconditionally | ✅ **Resolved** — now conditioned on `isCoarsePointer ? 'fixed' : 'absolute'` |
| [LOW] `getCoarsePointerSnapshot` allocated a fresh `MediaQueryList` per call | ✅ **Resolved** — module-level lazy singleton via `_coarseMql ??= window.matchMedia(COARSE_MQ)` eliminates per-call allocation |

---

### New finding in `32d9538`

**[LOW] `changedTouches[0]` is accessed without a null guard** — `components/equipment/equipment-tooltip.tsx`

```ts
const touch = event.changedTouches[0];
const deltaX = Math.abs(touch.clientX - touchStartRef.current.x);
```

`TouchEvent.changedTouches` is a `TouchList` — spec-wise it can be empty (e.g. if a browser coalesces events unusually), in which case `touch` is `undefined` and the `Math.abs` call throws. Very unlikely in practice, but a one-liner fixes it:

```ts
const touch = event.changedTouches[0];
if (!touch) return;
```

Not a blocker — everything else is solid.

---

### Overall

The singleton pattern for the `MediaQueryList`, the stable subscribe reference, and the scroll-guard tap detection (10 px threshold) are all done correctly. The `positionStrategy` fix eliminates the scrollable-container regression. The touch-dismiss UX is clean.

✅ **Approve** — the two medium issues from the first review are properly fixed. The `changedTouches[0]` guard above is a nice-to-have for defensive hygiene but doesn't block merge.

---

joeseos (approved):
## Code Review (round 3) — PR #1835

@Djidiouf — reviewing commit `4d316839`, pushed after the previous approval comment.

---

### Previous findings — all resolved

| Finding | Status |
|---|---|
| [MEDIUM] `subscribeCoarsePointer` inline arrow caused listener churn | ✅ Resolved (`32d9538`) |
| [MEDIUM] `positionStrategy="fixed"` applied unconditionally | ✅ Resolved (`32d9538`) |
| [LOW] `getCoarsePointerSnapshot` allocated a fresh `MediaQueryList` per call | ✅ Resolved (`32d9538`) |
| [LOW] `changedTouches[0]` accessed without null guard in touch start/end handlers | ✅ **Resolved in this commit** |

---

### What changed in `4d316839`

Two lines added in `components/equipment/equipment-tooltip.tsx`:

```ts
// handleMobileTooltipTouchStart
const touch = event.changedTouches[0];
if (!touch) return;   // ← added

// handleMobileTooltipTouchEnd
const touch = event.changedTouches[0];
if (!touch) return;   // ← added
```

Both guards are correct and placed immediately after the `changedTouches[0]` read, before any property access. No new issues introduced — the diff is exactly what was suggested and nothing more.

---

### Verdict

✅ **Approved** — every finding from all three review rounds is now addressed. Clean implementation, no regressions observed.

---

PR #1839, opened 2026-07-01
Increase report character limit to 4096 and truncate Discord embed descriptions

joeseos:
### Motivation
- Allow longer battle reports up to Discord's embed description limit so users can submit more detailed notes without client-side truncation.
- Ensure the Discord edge function never produces an embed description that exceeds Discord limits to avoid API rejections or broken messages.

### Description
- Bumped `reportCharLimit` from `1024` to `4096` in `complete-battle-modal.tsx` and `campaign-battle-log-modal.tsx` to permit longer reports. 
- Added `discordEmbedDescriptionLimit = 4096` and `truncateDiscordEmbedDescription` helper in the Discord edge function to trim long notes to the allowed length. 
- Switched the Discord embed to set `description` to the (truncated) `battle.note` and removed adding the note as a separate embed field. 
- Left existing embed fields and channel posting logic unchanged.

### Testing
- Ran a TypeScript typecheck with `tsc --noEmit` which completed successfully. 
- Executed the project test suite with `pnpm test` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a45282f05888332a19e373c0fda520a)

### DIFF
--- a/components/battle-session/complete-battle-modal.tsx
+++ b/components/battle-session/complete-battle-modal.tsx
@@ -21,7 +21,7 @@ import {
 } from '@/utils/battle-winners';
 import { useWinnerSelection } from '@/utils/hooks/use-winner-selection';
 
-const reportCharLimit = 1024;
+const reportCharLimit = 4096;
 
 interface CompleteBattleTerritory {
   id: string;
--- a/components/campaigns/[id]/campaign-battle-log-modal.tsx
+++ b/components/campaigns/[id]/campaign-battle-log-modal.tsx
@@ -32,7 +32,7 @@ interface CampaignBattleLogModalProps {
   userRole?: 'OWNER' | 'ARBITRATOR' | 'MEMBER';
 }
 
-const reportCharLimit = 1024;
+const reportCharLimit = 4096;
 
 type GangRole = 'none' | 'attacker' | 'defender';
 
--- a/supabase/edge-functions/discord-campaign-bot/index.ts
+++ b/supabase/edge-functions/discord-campaign-bot/index.ts
@@ -12,13 +12,38 @@ const DISCORD_BOT_TOKEN = Deno.env.get("DISCORD_BOT_TOKEN")!
 
 const DISCORD_CHANNEL_TYPES = { TEXT: 0, FORUM: 15 } as const;
 const discordEmbedDescriptionLimit = 4096;
+const discordEmbedTotalLimit = 6000;
 
-const truncateDiscordEmbedDescription = (value: string) => {
-  if (value.length <= discordEmbedDescriptionLimit) {
+const getCharacterCount = (value?: string) => Array.from(value ?? "").length;
+
+const truncateDiscordEmbedDescription = (value: string, limit: number) => {
+  const characters = Array.from(value);
+
+  if (characters.length <= limit) {
     return value;
   }
 
-  return `${value.slice(0, discordEmbedDescriptionLimit - 1)}…`;
+  return `${characters.slice(0, limit - 1).join("")}…`;
+};
+
+const getDiscordEmbedCharacterCount = (embed: {
+  title?: string;
+  description?: string;
+  fields?: { name: string; value: string }[];
+  footer?: { text?: string };
+}) => {
+  const fieldsTotal = embed.fields?.reduce(
+    (total, field) =>
+      total + getCharacterCount(field.name) + getCharacterCount(field.value),
+    0
+  ) ?? 0;
+
+  return (
+    getCharacterCount(embed.title) +
+    getCharacterCount(embed.description) +
+    fieldsTotal +
+    getCharacterCount(embed.footer?.text)
+  );
 };
 
 Deno.serve(async (req) => {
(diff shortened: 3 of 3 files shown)


joeseos:
## Code Review — PR #1839 "Increase report character limit to 4096 and truncate Discord embed descriptions"

@joeseos

### Context / correctness of the core fix

Good diagnosis. The prior 1024-char limit (added in #1838, commits b0b7485/255899c) existed specifically because `battle.note` was posted as a Discord embed **field**, and Discord hard-caps field `value` at 1024 characters. This PR fixes the actual constraint mismatch by moving `note` into the embed **description** (Discord's limit there is 4096), and bumps the client-side `reportCharLimit` to match. `reportCharLimit` is kept consistent across both `complete-battle-modal.tsx` and `campaign-battle-log-modal.tsx`, and the `maxLength`/counter/`isReportOverLimit` wiring is unchanged and still correct. The truncation math (`slice(0, limit - 1)` + 1 ellipsis char = exactly `limit` chars) is correct.

---

### Findings

**[MEDIUM] Discord's overall embed size limit (6000 chars) isn't accounted for — `supabase/edge-functions/discord-campaign-bot/index.ts:14-21, 205`**

Discord enforces a combined limit of 6000 characters across `title + description + all field names + all field values + footer.text` for a single embed, in addition to the per-component limits. This PR only truncates `description` to ≤4096; it doesn't check the combined total. With a near-4096-character report plus several populated fields (Attackers/Defenders/Participants with many gangs, Territory, Cycle, Scenario), the total could exceed 6000, and Discord will reject the whole message (400), which is exactly the failure mode this PR is meant to prevent per its own stated motivation ("avoid API rejections or broken messages").

Suggested fix: compute the fixed-field budget first (title + fields + footer), then truncate `description` to `min(4096, 6000 - fixedLength - safetyMargin)` instead of a flat 4096.

**[LOW] Report note moves from a labeled field to the embed description — layout/UX change not mentioned in the PR description**

Previously `battle.note` rendered as a distinct "📝 Report" field, positioned after Territory/Cycle/Scenario. Now it renders as the embed `description`, which Discord always places directly under the title, above all fields — and it no longer has the "📝 Report" label. This is likely fine, but it's a visible behavior change beyond "truncate long notes" and isn't called out in the PR description — worth confirming it's intentional.

**[LOW] `truncateDiscordEmbedDescription` slices on UTF-16 code units — `index.ts:16-21`**

`value.slice(0, discordEmbedDescriptionLimit - 1)` can split a surrogate pair (e.g. an emoji) if the cut lands mid-character, producing a mangled trailing character in the Discord message. Low likelihood/impact for typical battle report text, but worth a quick `Array.from(value).slice(...)` (code-point aware) if user reports commonly contain emoji.

**[LOW / pre-existing, not introduced here] No server-side length enforcement on `note`**

`completeBattleSession` / `createBattleLog` / `updateBattleLog` (in `app/actions/battle-sessions.ts` and `app/actions/campaigns/[id]/battle-logs.ts`) accept `note` with no length check — the 4096 cap is enforced only by the textarea's `maxLength` in the UI. This was already true at 1024 before this PR, so it's not a regression, but since the "contract" is now 4096 end-to-end, it may be worth a matching server-side guard (or at least confirming the `note` DB column is unbounded `text`, which it appears to be from the migrations).

---

### Verdict

⚠️ **Changes requested** (soft) — the core fix is correct and well-reasoned. The medium finding (total embed 6000-char budget) is the one I'd want addressed before merge since it directly undermines the PR's own goal of preventing Discord API rejections; the rest are minor/edge-case and can be addressed at the author's discretion.

---

joeseos:
## Code Review (round 2) — PR #1839

@joeseos — reviewing commit `3d28b73`, pushed after the previous review comment.

---

### Previous findings

| Finding | Status |
|---|---|
| [MEDIUM] Discord's combined 6000-char embed limit wasn't accounted for | ✅ **Resolved** — `getDiscordEmbedCharacterCount` now sums `title + fields + footer`, and `descriptionLimit = min(4096, max(0, 6000 - fixedTotal))` caps the description to whatever budget remains. Formula matches the suggested fix. |
| [LOW] `truncateDiscordEmbedDescription` sliced on UTF-16 code units (could split a surrogate pair) | ✅ **Resolved** — both `getCharacterCount` and `truncateDiscordEmbedDescription` now use `Array.from(value)` for code-point-aware counting/slicing. |
| [LOW] Report note moved from a labeled "📝 Report" field to the embed description, an unannounced layout change | ⏳ **Outstanding** — still present, not called out in the PR description. Likely intentional, non-blocking. |
| [LOW / pre-existing] No server-side length enforcement on `note` (UI `maxLength` only) | ⏳ **Outstanding** — unchanged, not a regression from this PR. |

---

### New observations in `3d28b73`

**[LOW] Fixed content (title + fields + footer) isn't itself bounded against the 6000 total**

`descriptionLimit` correctly clamps to `0` if the fixed fields alone already consume the full 6000-char budget (so `battle.note` is silently dropped rather than causing an oversized embed) — that's the right fallback for the *description*. However, if `title + fields + footer` alone exceed 6000 (e.g., a campaign battle with a very large participant list, since `gangNamesLines`/`participantLines` aren't length-capped), the embed would still exceed Discord's limit and get rejected regardless of `battle.note`. This is a narrow edge case and out of scope for what this PR set out to fix (it only ever budgeted the description), so not a blocker — just worth knowing the 6000-limit guarantee is conditional on fields staying reasonably sized.

---

### Verdict

✅ **Approve** — the medium finding (the one that mattered for "avoid API rejections") is properly fixed, and the UTF-16 slicing issue is resolved as a bonus via the `Array.from` rewrite. Remaining items are low-severity and non-blocking.

---

PR #1842, opened 2026-07-03
Refac/breadcrumbs cached loaders

joeseos:


joeseos:
## Code Review — PR #1842 "Refac/breadcrumbs cached loaders"

@joeseos

### Summary

Solid refactor. Each breadcrumb previously ran its own narrow, ad-hoc Supabase query (`select('campaign_name')`, `select('name')`, etc.) through the request-scoped client. This PR replaces those with the same cached loaders (`getCampaignBasic`, `getGangBasic`, `getFighterBasic`, `getBattleSessionCached`, `getUserProfile`) that the corresponding main pages already call, so the breadcrumb rides the same `'use cache: remote'` entry instead of issuing a redundant round trip. I checked each loader's actual return shape against how the breadcrumb destructures it (`campaign_name`, `name`, `fighter_name`/`gang_id`, `username`, `created_at`) — all consistent. The `.catch(() => null)` + fallback-label pattern is applied uniformly, so a cache/DB error degrades gracefully instead of breaking the layout, which matches the stated intent ("a breadcrumb must never throw").

---

### Findings

**[LOW] Inaccurate comment in `app/@breadcrumb/user/[id]/page.tsx`**

The comment says "Cached read — warmed by the user profile page on the same navigation," but the user profile page (`app/user/[id]/page.tsx`) fetches its data client-side via `fetch('/api/users/${id}')`, and `app/api/users/[id]/route.ts` runs its own direct query (`createClient()` + `.from('profiles')`) — it does not call `getUserProfile`. `getUserProfile` has no other caller in the codebase, so nothing actually warms this cache entry; the breadcrumb's own call is the first (and only) hit. Not a functional bug — the cached loader still works and still benefits later navigations to the same profile — but the comment overstates the win and could mislead whoever touches this next. Either fix the comment or wire `/api/users/[id]` to use `getUserProfile` for the profile lookup so the claim becomes true.

**[LOW / informational] Trust-boundary shift from RLS-scoped to service-role reads**

Previously these breadcrumbs read through the request-scoped `createClient()` (RLS-enforced). The new cached loaders all use `createServiceRoleClient()` internally, so the breadcrumb now bypasses RLS entirely for these lookups — losing whatever independent defense-in-depth RLS provided. I checked whether this matters in practice: the main gang/campaign pages already use the same service-role loaders and gate access at the application level (e.g. `checkCampaignPermissions` defaults to `canView: true`; a gang's `hidden` field is passed through as a display flag, not used to `notFound()`), so this doesn't appear to introduce a new disclosure beyond what the main content on the same page already reveals. Flagging only so it's a conscious tradeoff if any of these entities ever grow a real "private/restricted visibility" concept later — at that point breadcrumbs would need their own explicit check rather than relying on RLS.

**[LOW] Unrelated `package-lock.json` change**

The diff adds an optional `@next/swc-win32-x64-msvc` entry with no corresponding `package.json` change — looks like incidental lockfile churn from running `npm install` on Windows. Harmless (optional platform-specific binary, won't affect Linux/CI installs), but worth dropping from the diff for a cleaner PR.

---

### Verdict

✅ **Approve** — correct, well-scoped refactor with consistent error handling. The findings above are minor (comment accuracy + one drive-by lockfile diff); none block merge.

---

PR #1844, opened 2026-07-04
Fix "Include all Fighter Types" showing no results for custom gangs

joeseos:
The fighter-types API short-circuited into a custom-gang-only branch
whenever custom_gang_type_id was present, returning before the
include_all_types handling ever ran. For a custom gang, ticking
"Include all Fighter Types" therefore returned only the custom gang's
own fighters (often empty), rendering "No results found".

Gate the custom early-return on !includeAllTypes so an all-types request
falls through to the branch that returns every system fighter type, and
append the custom gang's own fighters there so they stay in the list
(matching how a regular gang's own fighters appear in its all-types
list). Extract the duplicated custom-fighter fetch/transform into shared
helpers reused across all three call sites.

Claude-Session: https://claude.ai/code/session_01KfFHTddbe1MM5DGuejvgJS

### DIFF
--- a/app/api/fighter-types/route.ts
+++ b/app/api/fighter-types/route.ts
@@ -4,6 +4,90 @@ import { gangVariantFighterModifiers } from '@/utils/gangVariantMap';
 import { getUserCustomFighterTypes } from '@/app/lib/customise/custom-fighters';
 import { getUserIdFromClaims } from "@/utils/auth";
 
+type SupabaseServerClient = Awaited<ReturnType<typeof createClient>>;
+
+// Fetch the user's own custom fighters plus any shared with them through campaigns,
+// de-duplicated by id.
+async function getCombinedCustomFighters(supabase: SupabaseServerClient, userId: string) {
+  const customFighters = await getUserCustomFighterTypes(userId, supabase);
+
+  // Fetch shared custom fighters from campaigns the user is a member of
+  const { data: campaignMembers } = await supabase
+    .from('campaign_members')
+    .select('campaign_id')
+    .eq('user_id', userId);
+
+  const campaignIds = campaignMembers?.map(cm => cm.campaign_id) || [];
+
+  let sharedCustomFighters: any[] = [];
+  if (campaignIds.length > 0) {
+    const { data: sharedFighterIds } = await supabase
+      .from('custom_shared')
+      .select('custom_fighter_type_id')
+      .in('campaign_id', campaignIds);
+
+    const fighterIds = sharedFighterIds?.map(sf => sf.custom_fighter_type_id).filter(Boolean) || [];
+
+    if (fighterIds.length > 0) {
+      const { data: sharedFighters } = await supabase
+        .from('custom_fighter_types')
+        .select('*')
+        .in('id', fighterIds);
+
+      sharedCustomFighters = sharedFighters || [];
+    }
+  }
+
+  // Combine own and shared, removing duplicates
+  const allCustomFighters: any[] = [...customFighters];
+  sharedCustomFighters.forEach(shared => {
+    if (!allCustomFighters.some(cf => cf.id === shared.id)) {
+      allCustomFighters.push(shared);
+    }
+  });
+
+  return allCustomFighters;
+}
+
+// Map a raw custom fighter row to the FighterType shape returned by this endpoint.
+function transformCustomFighter(cf: any) {
+  return {
+    id: cf.id,
+    fighter_type: cf.fighter_type,
+    fighter_class: cf.fighter_class || 'Custom',
+    gang_type: cf.gang_type,
+    cost: cf.cost,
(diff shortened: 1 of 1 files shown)


joeseos:
## Code Review — PR #1844 "Fix 'Include all Fighter Types' showing no results for custom gangs"

@joeseos

### Summary

The core fix is correct: the `customGangTypeId` branch previously returned unconditionally, so `include_all_types=true` was silently ignored for custom gangs. Gating the early return on `!includeAllTypes` and falling through to the `includeAllTypes` branch (which now also appends the custom gang's own fighters) is the right approach, and it mirrors how a regular gang's own fighters are already folded into its all-types list. The `getCombinedCustomFighters`/`transformCustomFighter` extraction is a clean, faithful DRY refactor — I diffed all three inlined blocks against the new helpers field-by-field and they're identical, so no behavior changed at the two pre-existing call sites.

There's one real bug in the new fallthrough path, detailed below.

---

### Findings

**[HIGH] Ticking both "Include Custom Fighter Types" and "Include all Fighter Types" on a custom-type gang produces duplicated and cross-contaminated custom fighters**

`components/gang/add-fighter.tsx` renders both checkboxes unconditionally (no guard hiding "Include Custom Fighter Types" when the gang is already a custom-type gang), and nothing prevents both being ticked at once. When both are true for a custom gang, the request now reaches **two** separate custom-fighter append blocks in `route.ts` that didn't previously coexist for this case (before this PR, the early return meant the second block was unreachable whenever `custom_gang_type_id` was set):

1. Inside the `includeAllTypes` branch (new in this PR): filters `cf.custom_gang_type_id === customGangTypeId` (or "available to all").
2. The pre-existing `if (includeCustomFighters && !isGangAddition)` block further down: filters `cf.gang_type_id === gangTypeId`.

For a custom-gang-scoped fighter, `gang_type_id` is stored as `NULL` in the DB (`app/actions/customise/custom-fighters.ts`: `gang_type_id: data.custom_gang_type_id ? null : data.gang_type_id`). And for a custom gang, the frontend never sends a `gang_type_id` query param at all (it's `null` on the gang too), so `searchParams.get('gang_type_id')` is also `null`. That makes filter #2 evaluate as `cf.gang_type_id === gangTypeId` → `null === null` → **true for every custom-gang-scoped fighter the user can see**, not just ones belonging to this gang's `custom_gang_type_id`.

Net effect: this gang's own custom fighters get appended twice, and any *other* custom gang type's fighters (the user's own, or shared via campaign) that also have `gang_type_id = null` get pulled into this gang's fighter list too — duplicate entries plus fighters that don't belong to this gang type at all.

Repro: open "Add Fighter" on a gang with a custom gang type, tick both "Include Custom Fighter Types" and "Include all Fighter Types".

Suggested fix: skip the trailing `includeCustomFighters` block when `customGangTypeId` is already set (that case is now fully handled by the append inside the `includeAllTypes` branch), e.g. change the condition to `if (includeCustomFighters && !isGangAddition && !customGangTypeId)`, or de-duplicate `data` by `id` before returning.

---

### Verdict

⚠️ **Changes requested** — the primary bug fix and refactor are solid, but the new fallthrough path has an edge case (both checkboxes ticked on a custom gang) that reintroduces incorrect results, worth fixing before merge since it's reachable through normal UI interaction.

---

joeseos:
## Code Review (round 2) — PR #1844

@joeseos — reviewing commit `0a38b17`, pushed after the previous review comment.

---

### Previous findings

| Finding | Status |
|---|---|
| [HIGH] Ticking both "Include Custom Fighter Types" and "Include all Fighter Types" on a custom-type gang produced duplicated/cross-contaminated custom fighters (the trailing `includeCustomFighters` block's `cf.gang_type_id === gangTypeId` filter degraded to `null === null`, matching every custom-gang-scoped fighter the user could see) | ✅ **Resolved** — the block is now gated with `&& !customGangTypeId`, exactly the suggested fix. |

I traced all reachable paths in `app/api/fighter-types/route.ts` to confirm no regression:
- **Custom gang, `includeAllTypes=false`**: hits the early-return branch (`customGangTypeId && !includeAllTypes`) and returns before reaching the trailing block regardless — unaffected either way.
- **Custom gang, `includeAllTypes=true`**: falls through to the `includeAllTypes` branch, which appends this gang's own custom fighters (matched by `custom_gang_type_id`, the correct scoped filter). The trailing block is now skipped entirely via the new guard, so there's exactly one append site for this case — duplication and cross-gang leakage are both gone.
- **Regular gang (`customGangTypeId` unset)**: `!customGangTypeId` is `true`, so the trailing block still runs exactly as before — no behavior change for the non-custom-gang path.

---

### New issues

None. This is a minimal, targeted 6-line fix with no side effects.

---

### Verdict

✅ **Approve** — the HIGH finding is fully resolved and verified across all code paths. No new issues introduced.

---

PR #1849, opened 2026-07-05
Updated the master crafted checkbox to have pills for cost and rarity

Djidiouf:


joeseos:
@Djidiouf Reviewed this PR. Summary below.

## Findings

**1. [Low] `package-lock.json` — unrelated lockfile churn**
This PR adds a Windows-only optional dependency block (`@next/swc-win32-x64-msvc`) to `package-lock.json` that has nothing to do with the UI change. Looks like it was picked up from running `npm install` on Windows.
- *Fix:* Revert that hunk (or re-run `npm install` on Linux/macOS and re-commit the lockfile) so the diff only contains the intended change.

**2. [Low] Duplicated pill markup — worth extracting eventually**
`components/equipment/purchase-modal.tsx:489-494` reuses the exact `w-6 h-6 rounded-full flex items-center justify-center ...` pill markup that already appears twice in `components/equipment/equipment.tsx` (lines ~801-818, ~820-830). This PR is consistent with the existing (already-duplicated) pattern, so not a blocker, but a shared `Pill`/`Badge` component would reduce triplication now that a third copy exists.

**3. [Low/info] Loss of explanatory tooltip**
The old implementation had a hover tooltip explaining *why* master-crafted weapons cost more and are rare ("Master-crafted weapons are Rare (10)."). The new pills show `25%` and `R10` directly but drop that explanation. This matches how availability/cost pills are shown elsewhere in the equipment list (no tooltip there either), so it's consistent with existing conventions — just flagging in case a new user wouldn't know what "R10" means without context.

## Verified as correct
- The `1.25` cost multiplier (`app/actions/equipment.ts:357`) and the "Rare (10)" rule are fixed/universal for master-crafted weapons, so hardcoding `25%` / `R10` in the UI is accurate, not a data-derivation bug.
- `bg-primary`/`text-primary-foreground` for the cost pill and `bg-sky-500` for the rarity ("R") pill match the exact color conventions used for cost/availability pills in `equipment.tsx`.
- `ImInfo` import is still used elsewhere in the file (the "Use Listed Cost for Rating" tooltip), so it's not now a dead import.
- Wrapping the pills inside the `<label htmlFor="master-crafted">` is fine and actually improves the checkbox's click target.
- No correctness bugs, null-handling issues, security issues, or performance concerns found in the JSX change itself.

## Verdict: **Approve**

Nothing here blocks merging — the lockfile diff (#1) is the only thing I'd ask to clean up before merge; the rest are optional follow-ups.

---

Djidiouf:
sorry for the package-lock.json, but imo not a blocker to leave it there, it's still needed

PR #1850, opened 2026-07-05
Replaced dropdown Assigned vehicle to crew by a combobox in vehicles …

Djidiouf:
…tab, and added it to Add Vehicle modal

### DIFF
--- a/app/actions/assign-vehicle-to-fighter.ts
+++ b/app/actions/assign-vehicle-to-fighter.ts
@@ -18,6 +18,7 @@ interface AssignVehicleToFighterResult {
   data?: {
     removed_from?: any;  // Used by frontend for vehicle swap handling
     vehicle_cost?: number;
+    gang_wealth?: number;
   };
   error?: string;
 }
@@ -200,7 +201,8 @@ export async function assignVehicleToFighter(params: AssignVehicleToFighterParam
       success: true,
       data: {
         removed_from: removedFromVehicle, // For frontend vehicle swap handling
-        vehicle_cost: vehicleCost
+        vehicle_cost: vehicleCost,
+        gang_wealth: financialResult?.newValues?.wealth
       }
     };
 
--- a/components/gang/add-vehicle.tsx
+++ b/components/gang/add-vehicle.tsx
@@ -1,14 +1,22 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import { Input } from '../ui/input';
 import Modal from '@/components/ui/modal';
 import { toast } from 'sonner';
 import { VehicleProps } from '@/types/vehicle';
+import { FighterProps } from '@/types/fighter';
 import { Checkbox } from "@/components/ui/checkbox";
 import { Combobox } from "@/components/ui/combobox";
 import { ImInfo } from "react-icons/im";
 import { vehicleTypeRank } from "@/utils/vehicleTypeRank";
 import { addGangVehicle } from '@/app/actions/add-gang-vehicle';
+import { assignVehicleToFighter } from '@/app/actions/assign-vehicle-to-fighter';
 import { getAllowedLocomotionOptions } from '@/utils/vehicle-locomotion';
+import { UserPermissions } from '@/types/user-permissions';
+import { IoSkull } from 'react-icons/io5';
+import { MdChair } from 'react-icons/md';
+import { GiCrossedChains, GiHandcuffs } from 'react-icons/gi';
+import { TbMeatOff } from 'react-icons/tb';
+import { FaMedkit } from 'react-icons/fa';
 
 interface VehicleType {
   id: string;
@@ -35,6 +43,10 @@ interface AddVehicleProps {
   onVehicleAdd: (newVehicle: VehicleProps) => void;
   onGangCreditsUpdate?: (newCredits: number) => void;
   onGangWealthUpdate?: (newWealth: number) => void;
+  fighters?: FighterProps[];
+  positioning?: Record<number, string>;
+  onFighterUpdate?: (updatedFighter: FighterProps) => void;
+  userPermissions?: UserPermissions;
 }
 
 export default function AddVehicle({ 
@@ -44,7 +56,11 @@ export default function AddVehicle({
   initialCredits,
   onVehicleAdd,
   onGangCreditsUpdate,
-  onGangWealthUpdate
+  onGangWealthUpdate,
+  fighters = [],
+  positioning,
+  onFighterUpdate,
+  userPermissions,
 }: AddVehicleProps) {
   
   const [vehicleTypes, setVehicleTypes] = useState<VehicleType[]>([]);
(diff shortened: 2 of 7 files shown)


joeseos:
@Djidiouf Reviewed this PR against the codebase. Nice UX improvement (search + status icons on the crew picker is a real upgrade over the plain `<select>`), but there are a couple of things worth addressing before merge.

## Findings

**1. (Medium) Duplicated "crew fighter combobox options" logic — 3rd/4th copy in the codebase**
`components/gang/add-vehicle.tsx` (new `crewFighterOptions` useMemo) and `components/gang/vehicles-tab.tsx` (new `crewFighterOptions` useMemo) both add a ~25-line block that:
- sorts fighters by `positioning`
- builds the same 6 status icons (killed/retired/enslaved/starved/recovery/captured)
- builds the same `displayText`/label shape for Combobox options

This is copy-pasted almost verbatim from the existing pattern in `components/gang/stash-tab.tsx:747-777`. With this PR it now exists in 3 places (and effectively 4 counting `add-vehicle.tsx` + `vehicles-tab.tsx` separately). Worth extracting into a shared helper (e.g. `utils/fighter-combobox-options.ts` or a `useFighterStatusIcons`/`buildCrewFighterOptions(fighters, positioning)` function) so the icon set and sort logic only need to be maintained in one place.

**2. (Low) `useMemo` doesn't actually memoize — dependency array references a value recomputed every render**
In both new blocks:
```ts
const crewFighters = fighters.filter(fighter => ...); // new array every render
const crewFighterOptions = useMemo(() => { ... }, [crewFighters, positioning]);
```
`crewFighters` is a plain `.filter()` call outside the memo, so it's a new array reference on every render, which means the `useMemo` recomputes every time regardless — it doesn't memoize anything. Either wrap `crewFighters` in its own `useMemo(() => fighters.filter(...), [fighters])`, or fold the filter into the same `useMemo` and depend on `fighters` directly.
- `components/gang/add-vehicle.tsx:88-90` / `~106`
- `components/gang/vehicles-tab.tsx:163-166` / `~175`

**3. (Low/Medium) Optimistic rating update assumes the assigned crew fighter is "active"**
`crewFighterOptions` deliberately includes fighters that are killed/retired/enslaved/starved/captured (it renders status icons for them), so the new Add Vehicle combobox lets you assign a freshly created vehicle to an inactive fighter. On the server, `assignVehicleToFighter` only adds the vehicle's cost to gang rating when the target fighter is active (`countsTowardRating`); for an inactive fighter the rating delta is `0` and wealth is reduced instead. But the client-side optimistic update in `gang-page-content.tsx`'s `handleFighterUpdate` (invoked via `onFighterUpdate` in `add-vehicle.tsx`) unconditionally does `ratingChange += vehicleCost` whenever a fighter gains a vehicle it didn't have before, with no wealth adjustment. Assigning a new vehicle to an inactive crew member will therefore show an inflated gang rating and stale wealth until the page is refreshed.
This same gap already exists in the pre-existing `handleMoveToFighter` flow in `vehicles-tab.tsx`, so it's not strictly new, but this PR extends the same faulty assumption into a second entry point. Worth a follow-up fix (have the server action return the actual rating/wealth deltas, or restrict the crew picker to active fighters) rather than letting it spread further.

**4. (Low) Minor duplication in `add-vehicle.tsx`**
The `assignedVehicle` object built at `components/gang/add-vehicle.tsx` (~lines 235-260) manually re-lists every field already present on `newVehicle` (minus `gang_id`/`fighter_id`/`payment_cost`, plus `effects: {}`). A destructure (`const { gang_id, fighter_id, payment_cost, ...assignedVehicle } = newVehicle;`) would be shorter and avoid the two objects silently drifting apart on the next field addition.

## Correctness / edge cases checked, no issues found
- `onVehicleAdd` correctly skipped only on the "created + assigned" success path (parent's `handleVehicleAdd` only tracks unassigned vehicles, so skipping is intentional and correct); credits/wealth still update via `onGangCreditsUpdate`/`onGangWealthUpdate` called unconditionally from the server-authoritative `addGangVehicle` result.
- Failure path (`assignResult.success === false`) correctly falls back to `onVehicleAdd(newVehicle)` so the vehicle isn't lost from the UI.
- `selectedCrewId` reset on `handleClose`, so the field doesn't leak state into the next time the modal opens.
- `Combobox`'s new `noResultsText` prop is optional with a backward-compatible default — no impact on other call sites.

## Severity summary
- Medium: #1 (duplication), #3 (rating/wealth correctness for inactive fighters)
- Low: #2 (memoization), #4 (minor duplication)

## Verdict: **Changes requested** (non-blocking style/duplication items aside, #3 is a real data-display correctness gap worth fixing or at least explicitly scoping the picker to active fighters before merge)


---

joeseos:
@Djidiouf Reviewed the latest commit (`025049c`, pushed after the previous review) against the prior findings.

## Previous findings

**1. (Medium) Duplicated crew-fighter-combobox logic — RESOLVED**
Extracted into `utils/crew-fighter-combobox-options.tsx` (`useCrewFighterOptions`), now used by both `add-vehicle.tsx` and `vehicles-tab.tsx`. Filter/sort/icon logic lives in one place. (`stash-tab.tsx` still has its own separate copy, but that was flagged as pre-existing/out-of-scope, not something this PR needed to touch.)

**2. (Low) `useMemo` not actually memoizing — RESOLVED**
The filter is now performed inside the same `useMemo` as the sort/map, with a stable `[fighters, positioning]` dependency array (`utils/crew-fighter-combobox-options.tsx:24-66`). No more new-array-every-render issue.

**3. (Medium/Low) Optimistic rating update assumes assigned fighter is active — PARTIALLY RESOLVED**
`gang-page-content.tsx`'s `handleFighterUpdate` now gates the rating increment on `countsTowardRating(nextFighter)`, so assigning a new vehicle to a killed/retired/enslaved/captured fighter no longer inflates the optimistic gang rating. Good fix for the half of the bug that mattered most (the rating overcount).

However, the **wealth** half is still outstanding: `assignVehicleToFighter` (`app/actions/assign-vehicle-to-fighter.ts`) computes and persists a correct `wealthDelta` server-side via `updateGangFinancials`, but its return type only exposes `{ removed_from, vehicle_cost }` — no `newWealth`/`wealthDelta`. `add-vehicle.tsx`'s success handler never calls `onGangWealthUpdate` after `assignVehicleToFighter` resolves, so when the target fighter is inactive, gang wealth stays stale client-side until the page is refreshed (same underlying gap as before, just narrower now that rating is correct). Worth a quick follow-up: have `assignVehicleToFighter` return the new wealth (or delta) and apply it via `onGangWealthUpdate` in the success branch, mirroring what already happens for `addGangVehicle`'s result a few lines above it.

**4. (Low) Minor duplication building `assignedVehicle` — RESOLVED**
Now a destructure: `const { gang_id: _g, fighter_id: _f, payment_cost: _p, effects: _e, ...vehicleBase } = newVehicle;` followed by `{ ...vehicleBase, effects: {} }`. Matches the suggestion, no drift risk on new fields.

## New issues in this commit
None found. The extraction into the shared hook is clean — types line up (`FighterProps`/`positioning: Record<number, string> | undefined` match on both call sites), no unused imports left behind (`useMemo` import removed from `add-vehicle.tsx` since it's no longer used there, icon imports removed from both files and centralized in the new util), and no behavioral changes beyond the intended fixes.

## Verdict: **Changes requested** (minor) — the duplication/memoization/object-building items are all fixed; only the wealth-sync half of finding #3 remains, which is a small, well-scoped follow-up rather than a blocker on design.

---

joeseos:
@Djidiouf Reviewed the latest commit (`1151f27`, pushed after the previous review) against prior findings.

## Previous findings

**1. (Medium) Duplicated crew-fighter-combobox logic — RESOLVED** (no change this commit)

**2. (Low) `useMemo` not actually memoizing — RESOLVED** (no change this commit)

**3. (Medium/Low) Optimistic rating/wealth update assumes assigned fighter is active — now fully RESOLVED**
The remaining wealth half is fixed in this commit:
- `app/actions/assign-vehicle-to-fighter.ts:20` — `AssignVehicleToFighterResult.data` now exposes `gang_wealth?: number`, populated from `financialResult?.newValues?.wealth` (`assign-vehicle-to-fighter.ts:205`).
- `components/gang/add-vehicle.tsx:217-222` — after a successful assignment, calls `onGangWealthUpdate(assignResult.data.gang_wealth)` guarded by a `typeof === 'number'` check, so it's a no-op when `financialResult` wasn't computed (e.g. `vehicleCost === 0` or both deltas are zero).

Traced the math to confirm no double-counting: a freshly created vehicle already counts toward wealth while unassigned (`wasInUnassignedPool` branch, `assign-vehicle-to-fighter.ts:125-135`). Assigning it to an **active** fighter yields `wealthDelta = 0` (moves from unassigned pool to rating, net wealth unchanged) — so the new `onGangWealthUpdate` call after assignment just re-confirms the same value already set from the creation step (harmless, not redundant-buggy). Assigning to an **inactive** fighter yields `wealthDelta = -vehicleCost`, correctly dropping wealth by the vehicle's cost. This matches server-side `updateGangFinancials` semantics (`utils/gang-rating-and-wealth.ts`) exactly, and `newValues.wealth` is always a concrete number on success, never `undefined`, so the client guard is defensive rather than load-bearing.

**4. (Low) Minor duplication building `assignedVehicle` — RESOLVED** (no change this commit)

## New issues in this commit
None found. The change is narrowly scoped to the two files needed for the fix, the type addition (`gang_wealth?: number`) is optional/backward-compatible, and `npx tsc --noEmit` shows no errors introduced.

One pre-existing gap worth flagging for a future PR, not a regression here: `components/gang/vehicles-tab.tsx`'s `handleMoveToFighter` (a separate call site of the same `assignVehicleToFighter` action) still doesn't read `gang_wealth` off the response, so moving a vehicle to an inactive fighter via the Vehicles tab's own picker will still leave wealth stale until refresh. This was already noted as out-of-scope in the first review and this commit correctly didn't need to touch it, but since the server action now conveniently returns the value, it'd be a small follow-up to wire up there too.

## Verdict: **Approve**

---

joeseos:
@Djidiouf Reviewed the latest commit (`a2f5b49`), which only touches `components/gang/vehicles-tab.tsx` (+6 lines) since the previous "Approve" review of `1151f27`.

## Previous findings
1. (Medium) Duplicated crew-fighter-combobox logic — **RESOLVED** (no change this commit)
2. (Low) `useMemo` not actually memoizing — **RESOLVED** (no change this commit)
3. (Medium/Low) Optimistic rating/wealth update assumes assigned fighter is active — **now fully RESOLVED**. This commit wires up the one remaining gap noted in the prior review: `handleMoveToFighter` in `vehicles-tab.tsx` (a second call site of `assignVehicleToFighter`, separate from `add-vehicle.tsx`) now reads `data.gang_wealth` off the response and calls `onGangWealthUpdate(data.gang_wealth)` (`vehicles-tab.tsx:329-333`), guarded by the same `typeof === 'number'` check used elsewhere in the file. Moving a vehicle to an inactive fighter via the Vehicles tab's picker will no longer leave gang wealth stale until refresh.
4. (Low) Minor duplication building `assignedVehicle` — **RESOLVED** (no change this commit)

## New issues in this commit
None found. Traced the change:
- `gang_wealth` is populated from `financialResult?.newValues?.wealth` in `assign-vehicle-to-fighter.ts:205`, same server contract already verified in the previous review for `add-vehicle.tsx`'s equivalent call — no drift between the two call sites now.
- Placement is correct: the wealth sync happens after the existing fighter-cost correction block (`vehicles-tab.tsx:311-327`), so it doesn't interfere with that logic and only adds an independent, server-authoritative wealth sync.
- Guard style (`typeof data?.gang_wealth === 'number'`) matches the existing conventions in this file (e.g. `onGangCreditsUpdate`/`onGangWealthUpdate` checks elsewhere), and is a no-op when `financialResult` wasn't computed — consistent with how `add-vehicle.tsx` handles the same optionality.
- No edge case regressions: the rollback path (catch block) is untouched, and this new block doesn't run before the optimistic updates, only after the server call resolves successfully.

## Verdict: **Approve**

---

PR #1852, opened 2026-07-07
Move custom equipment to Trading Post in equipment RPC

joeseos:
Custom equipment (own + campaign-shared) was hard-coded as being on every
fighter's equipment list (fighter_type_equipment = true) and the custom
branch ignored the fighter-list/trading-post filters. It now lives in the
Trading Post section instead: fighter_type_equipment is computed from
custom_fighter_type_equipment (assigned to the fighter's custom type), and
the branch respects the $4/$5 filters like the official equipment branch.

Claude-Session: https://claude.ai/code/session_01UrPonW94YiqXu6Y72jagtG

### DIFF
--- a/supabase/functions/get_equipment_detailed_data.sql
+++ b/supabase/functions/get_equipment_detailed_data.sql
@@ -454,7 +454,9 @@ AS $$
         ce.equipment_category,
         ce.equipment_type,
         ce.created_at,
-        true AS fighter_type_equipment,
+        -- Custom equipment lives in the Trading Post; it is only on a fighter's
+        -- list when assigned to that fighter's custom fighter type ($3).
+        COALESCE(ftl.is_ftl, false) AS fighter_type_equipment,
         true AS equipment_tradingpost,
         true AS is_custom,
         COALESCE(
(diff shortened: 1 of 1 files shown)


joeseos:
## Code Review

@joeseos Reviewed against the full `get_equipment_detailed_data` function and traced how `fighter_type_equipment`/`equipment_tradingpost` ($4/$5) are actually populated from the frontend (`components/equipment/equipment.tsx`).

### What I verified as correct
- **Filter logic (lines 558-562).** I initially suspected the OR-based three-case filter (no-filter / fighter's-list / trading-post) wasn't equivalent to the official branch's four-case switch (lines 405-431), since custom equipment is always `equipment_tradingpost = true`. Traced through all four `$4`/`$5` null combinations by hand — because `true = $5` stands in for "trading-post match" (which is unconditionally true for custom equipment), the OR-based encoding is algebraically equivalent to the official branch's pattern. No bug here.
- **Real behavioral impact confirmed.** Before this change, `fighter_type_equipment` was hardcoded `true` for *all* custom equipment. `components/equipment/equipment.tsx:792` uses that field to render the "Fighter's List" / "Vehicle's List" badge in the Trading Post view (`item.fighter_type_equipment || item.from_fighters_list`). That means every custom equipment item — including ones never assigned to the viewing fighter's custom type, and even in gang-stash mode with no fighter selected — was incorrectly badged as being on the fighter's list. This PR fixes a real, user-visible bug, not just an unused/dead computation.
- Non-custom fighters browsing the "Fighter's List" tab will no longer see custom equipment leak in (previously `fighter_type_equipment=true` matched all custom equipment regardless of fighter type, since custom equipment's flag was hardcoded).

### Findings

**1. [Medium] Missing index for the new `custom_fighter_type_equipment` LATERAL join**
`supabase/functions/get_equipment_detailed_data.sql:544-550`
The new `LEFT JOIN LATERAL` filters `custom_fighter_type_equipment` by `(custom_equipment_id, custom_fighter_type_id)` for every row of `custom_equipment`. The table has no index beyond its primary key on `id` (checked `schema.public.sql`), so this join will sequential-scan `custom_fighter_type_equipment` once per custom-equipment row. For comparison, the equivalent official-equipment join target, `fighter_type_equipment`, has two indexes (`fighter_type_equipment_equipment_id_idx` on `equipment_id`, `idx_fte_fighter_type_id` on `fighter_type_id`). Since this is a `SECURITY DEFINER` RPC called on effectively every equipment-list page load, worth adding a matching index, e.g.:
```sql
CREATE INDEX idx_cfte_custom_equipment_id
  ON public.custom_fighter_type_equipment (custom_equipment_id, custom_fighter_type_id);
```

**2. [Low / process note] No accompanying migration**
This PR only touches `supabase/functions/get_equipment_detailed_data.sql`. Per `supabase/README.md`, syncing these function files to the live Supabase instance is a manual step ("Syncing of those files is not automatic"). Just flagging so the updated function definition (and the suggested index above, if added) doesn't get missed when applying this to the database — not a code issue.

**3. [Low] Minor naming nit**
`ftl.is_ftl` (line 545, "is fighter['s] [type] list") is a little cryptic as an alias, even with the comment above it. Something like `on_fighter_list` would read slightly clearer in a query that's already dense with abbreviations. Not blocking.

### Verdict: **Approve** (with a strong recommendation to add the index in #1 before/soon after merge)
The core logic change is correct and fixes a genuine mislabeling bug. The only concrete concern is the missing index on `custom_fighter_type_equipment`, which is a performance follow-up rather than a correctness blocker.


---

PR #1853, opened 2026-07-08
Updated dnd kit

joeseos:


joeseos:
## Code Review

@joeseos Reviewed this as a full migration from `@dnd-kit/core`/`@dnd-kit/sortable`/`@dnd-kit/utilities` to `@dnd-kit/react` + friends. I checked out the branch in a worktree, ran `npm install`, `npx tsc --noEmit`, and `npx eslint .` against it, and diffed the lint results against a same-setup run on `main` to separate pre-existing issues from ones this PR introduces.

### Verified clean
- **Type-check**: `npx tsc --noEmit` is clean on this branch — the new `@dnd-kit/react` event shapes (`DragEndEvent`/`DragOverEvent`, `event.operation.source/target`, `event.canceled`) are used correctly throughout.
- **Complete migration**: no dangling references anywhere to `@dnd-kit/core`, `@dnd-kit/sortable`, `@dnd-kit/utilities`, or the two deleted hooks (`use-dnd-sensors`, `use-is-mounted`) — confirmed via repo-wide grep.
- **Per-component port** (`sortable-fighter.tsx`, `fighter-card.tsx`, `gang-card.tsx`, `campaign-card.tsx`): the `dragListeners`/`dragAttributes` → `draggable` boolean + ref-based drag source swap is consistent and correct across all four files, and the iOS touch-callout / long-press-vs-swipe handling in `sortable-fighter.tsx` and `dnd-sensors.ts` is a nice touch.
- Exact-pinning the five new `@dnd-kit/*` packages at `0.5.0` (no `^`) makes sense given the library is pre-1.0 — there's already precedent for this in the repo (`markdown-it` is also exact-pinned).

### Findings

**1. [Medium] Two new ESLint errors vs. `main`'s baseline**
`main` has exactly 4 pre-existing lint errors (I ran `npx eslint .` on both branches to confirm the baseline). This branch has 6 — the 2 new ones are both `react-hooks/set-state-in-effect`:
- `components/home/gangs-tab.tsx:27`
- `components/home/campaigns-tab.tsx:27`

Both are the same pattern: `useEffect(() => { if (!pendingReorder) return; setPendingReorder(null); reorderFavourites(...) })`. The comment above each explains the intent well (deferring the server action until after the drop's transition commits, to avoid freezing the card), so this looks deliberate rather than an oversight — but it's still a real regression against the lint baseline. Given a sibling PR (#1855) explicitly calls out "lint has only the 4 pre-existing errors from main" as its bar, worth holding this one to the same standard: either restructure to satisfy the rule, or add a scoped `// eslint-disable-next-line react-hooks/set-state-in-effect` with the existing comment as justification, so CI/lint stays a meaningful signal.

**2. [Low] `components/ui/quantity-stepper.tsx` is new, unused, and unrelated to dnd-kit**
Repo-wide grep found zero imports/references to `QuantityStepper` anywhere. It's a complete, self-contained +/- stepper component with no connection to drag-and-drop — looks like it landed in this branch by accident (e.g., picked up from a local working tree alongside the intended changes). Recommend dropping it from this PR unless it's actually meant to ship here for an unrelated reason.

### Worth a manual QA pass (not confirmed as bugs, just flagging what code review can't fully verify)
- **SSR guard removed**: the old `useIsMounted()` check that deferred rendering `<DndContext>` until after client mount is gone — `draggable-fighters.tsx` now always renders `<DragDropProvider>`, including during SSR. That guard existed specifically because the old `@dnd-kit/core` had SSR/hydration issues; worth confirming `@dnd-kit/react`'s `DragDropProvider` is actually safe to render server-side (view-source / JS-disabled check) rather than relying on it happening to work.
- **Custom live-sort workaround** in `draggable-fighters.tsx` (`handleDragOver` manually reordering `currentPositions`, bypassing the library's built-in optimistic sort to avoid the variable-height-card jitter from dnd-kit#1950/#2088): the logic reads correctly to me, but this is exactly the kind of interaction code where a browser pass matters more than a code read. Worth specifically exercising: reordering fighters with very different card heights (killed/retired collapsed vs. normal) in both grid and list view modes, plus a real mobile touch drag (the 600ms long-press delay).

### Verdict: **Approve, with the lint regression (#1) worth fixing before merge and #2 worth a quick check**
The migration itself is solid and type-safe end-to-end. Neither finding blocks the core functionality, but #1 quietly lowers the project's lint bar and #2 looks like unrelated leftover code.


---

PR #1854, opened 2026-07-10
Feature/coloured dot for campaign gangs dropdown

Djidiouf:


joeseos:
## Code Review

@Djidiouf Nice feature, and the pattern (colour dot + name + owner suffix inside `Combobox` `label`/`displayValue`) is consistent with how this codebase already renders gang options. Found two correctness bugs where the colour is silently dropped before it reaches the component that renders the dot, plus a couple of smaller issues.

### Findings

**1. [High] `gang_colour` is dropped before it reaches `CampaignBattleLogModal`**
`components/campaigns/[id]/campaign-battle-logs-list.tsx:428-437`
The `gangsMap` objects that become the `availableGangs` prop passed into `CampaignBattleLogModal` are built with only `id`, `name`, `campaign_member_id`, `user_id`, `owner_username` — `gang_colour` is captured only into a separate, sibling `gangColourMap` (line 437) that's used for this component's *own* rendering, but never exposed on `availableGangs` itself.
`campaign-battle-log-modal.tsx` reads `gang.gang_colour` straight off `availableGangs` (lines 111-114, and inline at 849), so every dot in that modal's "Gangs in battle" / gang-select comboboxes always falls back to the default colour, regardless of the gang's actual colour.
Fix: add `gang_colour: gang.gang_colour` to the object pushed into `gangsMap` at line ~428-435.

**2. [High] `gang_colour` omitted when `CreateBattleModal` is opened from the campaign Battle Log tab**
`components/campaigns/[id]/campaign-page-content.tsx:905-914`
`campaignGangs` is built via `campaignData.members.flatMap(m => m.gangs.filter(...).map(g => ({ id, name, user_id, owner_username })))` — `gang_colour` isn't included. `CreateBattleModal`'s "Your Gang"/"Opponent Gang" dropdowns (updated in this PR to read `g.gang_colour`) will therefore always show the black fallback dot on this path, even though the gang has a real colour. (The gang-page entry point works, since it independently fetches from `/api/campaigns/campaign-gangs`, which does select `gang_colour`.)
Fix: include `gang_colour: g.gang_colour` in the mapped object.

**3. [Low] Inconsistent fallback colour within the same file**
`components/campaigns/[id]/campaign-battle-log-modal.tsx:114` vs `:849`
The `getGangColour` helper falls back to `#888888` when a gang has no colour, while the inline "Gangs in battle" selector a few hundred lines down falls back to `#000000` — in the same modal. Every other touched file in this PR uses `#000000` as the fallback. Worth aligning on one constant so the same "no colour set" state doesn't render two different dot colours in one dialog.

**4. [Medium] Significant duplication of the dot/label JSX**
The `<span className="flex items-center gap-2"><span className="... rounded-full ..." style={{backgroundColor: colour}} aria-hidden />...` block is repeated near-verbatim across ~10 call sites in 8 files. Beyond the maintenance cost, this duplication is very likely *why* findings #1 and #3 happened — each copy independently re-derives the gang's colour and its own fallback, so it's easy for one copy-paste site to end up with a different data source or fallback than its neighbors. Consider extracting a small shared helper, e.g. a `buildGangOption({ id, name, gang_colour, owner_username })` in a shared util, or a `<GangOptionLabel>` component, and reusing it everywhere a gang is rendered as a `Combobox` option.

### Other notes (no action needed)
- Inline `style={{ backgroundColor: colour }}` with a DB-sourced string is safe here — React sets it via the CSS object model, not `innerHTML`, so there's no injection risk even if `gang_colour` were ever attacker-controlled.
- Confirmed `gang_colour` type plumbing (`types/campaign.ts`, the `CampaignGang` interfaces in `create-battle-modal.tsx`/`fighter-actions.tsx`) is otherwise consistent with the rest of the codebase.

The visual pattern and most wiring are solid, but findings #1 and #2 mean the dot silently shows the wrong (fallback) colour in two of the more commonly used flows (campaign battle log entry, and "Create Battle" from the campaign page), which undercuts the feature's purpose. Worth fixing before merge.


---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (`beb3361`, since the previous review at `0f69075`). This addresses everything flagged before, cleanly.

### Previous findings

**1. [High] `gang_colour` dropped before reaching `CampaignBattleLogModal` — Resolved**
`campaign-battle-logs-list.tsx:432` now sets `gang_colour: gang.gang_colour` on the `gangsMap` entry, so `availableGangs` carries the real colour through to the modal.

**2. [High] `gang_colour` omitted in `campaign-page-content.tsx`'s `campaignGangs` — Resolved**
`campaign-page-content.tsx:911` now includes `gang_colour: g.gang_colour` in the mapped object, so `CreateBattleModal` opened from the campaign Battle Log tab shows the correct dot.

**3. [Low] Inconsistent fallback colour (`#888888` vs `#000000`) — Resolved**
The local `getGangColour` helper in `campaign-battle-log-modal.tsx` (the one that fell back to `#888888`) was deleted entirely; all three call sites in that file now go through the shared `buildGangComboboxOption`, which falls back to `#000000` consistently.

**4. [Medium] Duplicated dot/label JSX across ~10 call sites — Resolved**
New `utils/gang-combobox-option.tsx` exports `buildGangComboboxOption`, and every touched file (`complete-battle-modal.tsx`, `create-battle-modal.tsx`, `campaign-battle-log-modal.tsx`, `campaign-battle-logs-list.tsx`, `campaign-map.tsx`, `campaign-territory-gang-modal.tsx`, `fighter-actions.tsx`, `fighter-injury-list.tsx`) now uses it instead of hand-rolling the markup. Nice touch supporting `labelSuffix`/`displayValueSuffix` for the "Already assigned" case in the territory modal rather than special-casing it.

Spot-checked that `CampaignGang`'s `gangsMap` (typed `Map<string, CampaignGang>`) already declared `gang_colour`, so this was purely a missing runtime field, not a type gap. `tsc --noEmit` on the changed files is clean, and no dangling references to the removed local helpers remain (the `getGangColour` still present in `campaign-battle-logs-list.tsx` is a distinct, unrelated helper used for text colour styling, not the dot).

### New findings

**1. [Low] Indentation drift in `fighter-injury-list.tsx:1115-1125`**
The second `<Combobox>` in this file (Bitter Enmity "Against" selector) has its props (`options`, `value`, `onValueChange`, `placeholder`, `clearable`, closing `/>`) indented two spaces less than the sibling `<Combobox>` just above it (capturing-gang selector) and less than the surrounding JSX. Purely cosmetic, doesn't affect behavior — worth a quick format pass before merge so it doesn't stand out in the file.

No new bugs, regressions, or edge-case issues found in the diff.

### Verdict: **Approve**
All four previously-flagged issues are resolved, and the refactor into a shared helper is a solid fix that also closes off the root cause (copy-paste drift) rather than just patching symptoms. The indentation nit above is optional cleanup, not a blocker.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (`c804e74`, since the previous review at `beb3361`). This is a one-line-scope cleanup commit, no functional changes.

### Previous findings

**1–4 (all previously High/Medium/Low findings from the first review) — Resolved** (confirmed in the prior follow-up review at `beb3361`, unaffected by this commit).

**Follow-up finding: Indentation drift in `fighter-injury-list.tsx:1115-1125` — Resolved**
The diff for this commit is exactly this fix: the "Against" `Combobox`'s props (`options`, `value`, `onValueChange`, `placeholder`, `clearable`, closing `/>`) are now indented to match the sibling `Combobox` above it (the capturing-gang selector) and the surrounding JSX. Whitespace-only change, verified against the file — matches the "Captured by" `Combobox` block's indentation exactly.

### New findings

None. The change is a pure re-indentation with no logic touched.

### Verdict: **Approve**

All previously-flagged issues remain resolved, and this commit closes out the one optional cleanup nit from the last review. No blockers.

---

PR #1861, opened 2026-07-13
Added custom fighter type equipment lists

joeseos:


joeseos:
## Code Review

@joeseos This PR extends `custom_fighter_type_equipment` (previously custom-equipment-only, added in #1852) to also support official/system equipment, so a custom fighter type can have an "equipment list" of official items it's allowed to buy. I checked out the branch, ran `npx tsc --noEmit` and `npx eslint` on the changed files (both clean), then traced the full data path from the UI down to the SQL RPC to verify the feature is actually reachable end-to-end — and found one critical gap in the SQL.

### Critical finding

**[High] Official equipment on a custom fighter's equipment list never appears in the picker, unless it's already separately in the gang's trading post**

`supabase/functions/get_equipment_detailed_data.sql`, the "Trading post only" branch of the fighter-list/trading-post filter (~line 442-443):
```sql
-- Trading post only
($4 IS NULL AND $5 IS NOT NULL AND COALESCE(tp.has_access, false) = $5)
```
This is the only branch that omits `cftl.is_ftl IS NOT NULL` as an alternate match condition — the "both filters" and "fighter's list only" branches right above it both correctly include it.

I traced why this matters: for a custom fighter, the "Fighter's List" tab is hidden entirely in the UI (`equipment.tsx:595`, `!isCustomFighter` guard), and `equipmentListType` defaults to `"fighters-tradingpost"` for custom fighters (`equipment.tsx:122`). In that mode, the code deliberately does *not* set `fighter_type_equipment=true` for custom fighters (`equipment.tsx:266`, `!isCustomFighter` guard again) — only `equipment_tradingpost=true`. So **every equipment query for a custom fighter hits exactly this "Trading post only" branch** ($4 NULL, $5 true), which only checks `tp.has_access` — completely ignoring `cftl.is_ftl`.

Net effect: an official equipment item assigned to a custom fighter type's equipment list only shows up in that fighter's picker if it's *also* independently available via the gang's regular trading post. If the intent (mirroring how normal fighter types' equipment lists work) is "this custom fighter type can buy Lasguns even though Lasguns aren't in this gang's trading post catalog," that case is currently silently dropped — the row never reaches the client. `cftl.is_ftl` does get computed correctly (confirmed `$3` is correctly the `custom_fighter_type_id` — see below) and does correctly flip the `fighter_type_equipment` *display* flag on items that do pass, but it can't currently pull in items that wouldn't otherwise pass the trading-post check.

Compare with how #1852 solved the equivalent problem for *custom* equipment: it hardcoded `true AS equipment_tradingpost` so custom equipment is always visible regardless of `ftl.is_ftl`, and used `ftl.is_ftl` only to set the display badge. That trick isn't (and shouldn't be) applied to official equipment's real `equipment_tradingpost` column here, but as a result this branch needs its own fix — add `cftl.is_ftl IS NOT NULL` as an alternate inclusion path, e.g.:
```sql
-- Trading post only
($4 IS NULL AND $5 IS NOT NULL AND (COALESCE(tp.has_access, false) = $5 OR cftl.is_ftl IS NOT NULL))
```

**I confirmed `$3` itself is wired correctly** (not the bug): `add-fighter.ts` sets `custom_fighter_type_id` on the fighter row and leaves the official `fighter_type_id` null for custom fighters; `app/fighter/[id]/page.tsx` builds `fighter.fighter_type.fighter_type_id` with a fallback to `custom_fighter_type_id` when the official lookup returns null; `fighter-page.tsx:1328` passes that through to `ItemModal`; `equipment.tsx` forwards it as `$3` unconditionally. So the join fires and matches — it's purely the WHERE-clause row filter that drops the results.

### Other observations (not blocking)

- **`getCompleteCustomFighter` extraction** (`app/actions/customise/custom-fighters.ts`): nice DRY cleanup — deduplicates the identical fetch/transform block that was previously copy-pasted between `createCustomFighter` and `updateCustomFighter`.
- **`updateCustomFighter`'s delete-then-conditional-insert for `equipment_list`** (lines ~500-528) unconditionally deletes all existing `custom_fighter_type_equipment` rows for the fighter type before checking `data.equipment_list?.length > 0` to re-insert. If a future caller ever invokes `updateCustomFighter` with `equipment_list` omitted (not just empty), the existing list would be silently wiped with nothing to restore it. Not a new risk introduced by this PR — it's the exact same pattern already used for `default_skills`/`default_equipment` in this file, and the current (only) UI caller always sends the array — just flagging since it's now used in a third place.
- **UI equipment-list picker** (`renderEquipmentListSection` in `custom-fighters.tsx`) only lets you pick from official `equipment`, even though the backend/SQL fully supports `custom_` -prefixed custom equipment IDs in `equipment_list`. This mirrors the pre-existing "Default Equipment" picker, which has the same official-only limitation, so it's not a regression — just noting the custom-equipment code path in the backend is currently unreachable from this UI, same as it already is for `default_equipment`.
- Small nice-to-have: the equipment-list picker's dropdown (`if (value && !selectedEquipmentList.includes(value))`) guards against adding a duplicate; the "Default Equipment" picker it mirrors doesn't have that guard. Not something this PR needs to fix, but worth applying the same guard to Default Equipment at some point for consistency.

The plumbing (types, actions, UI, `$3` propagation) is solid and type-checks/lints clean, but the core SQL filter gap means the feature's main use case — granting a custom fighter type access to official equipment it wouldn't otherwise see — doesn't actually work through the UI yet. Worth fixing the "Trading post only" branch before merge.


---

PR #1862, opened 2026-07-13
Added dramatis personae boolean and prefil of name

Djidiouf:


joeseos (commented):
@Djidiouf Reviewed this against the codebase (admin routes, gang-additions modal, RPC, migrations). Summary below.

## Findings

### 1. (High) Selecting *any* non-Dramatis-Personae fighter type wipes out a manually-typed fighter name
`components/gang/gang-additions.tsx` (Fighter Type `onValueChange`, ~line 1327-1333):
```ts
if (selectedType?.is_dramatis_personae) {
  setFighterName(selectedType.fighter_type);
} else {
  setFighterName('');
}
```
This clears the Fighter Name field on *every* Fighter Type selection where `is_dramatis_personae` is false — not just when switching away from a Dramatis Personae type. Failure scenario: user types a custom name (e.g. "Bob") in the Fighter Name field, then picks/changes the Fighter Type to a regular Ganger/Juve/etc. — `setFighterName('')` fires and the typed name is silently lost, even switching between two ordinary (non-DP) types. Before this PR, selecting a Fighter Type never touched the name field at all.

Suggested fix: only clear the name when it still equals the auto-filled value from a previously-selected Dramatis Personae type (or track an "auto-filled" flag), otherwise leave whatever the user has typed untouched:
```ts
if (selectedType?.is_dramatis_personae) {
  setFighterName(selectedType.fighter_type);
} else if (fighterName === gangAdditionTypes.find(t => t.id === selectedGangAdditionTypeId)?.fighter_type) {
  // was auto-filled by a previous DP selection — safe to clear
  setFighterName('');
}
```

### 2. (Low) No uniqueness/limit enforcement for Dramatis Personae
Dramatis Personae are typically unique, one-off characters, but nothing here prevents adding the same DP fighter type to a gang multiple times (no analogue to `limitation`/0-X checks used for other restricted types). Possibly out of scope for this PR, but worth confirming intentional.

### 3. (Low) Prefilled name is fully editable with no visual indicator
Once prefilled, the Fighter Name field looks identical to a manually-entered one — minor UX nit, not blocking.

## Verified as correct
- All 4 `SELECT` blocks and both `PUT`/`POST` bodies in `app/api/admin/fighter-types/route.ts` consistently include `is_dramatis_personae`.
- `get_fighter_types_with_cost` RPC (both `supabase/functions/...sql` and the new migration) are in sync, and the migration follows this repo's existing DROP/CREATE convention for RPC changes.
- `app/api/fighter-types/route.ts` passes the RPC's raw result straight through for the `is_gang_addition` branch, so `is_dramatis_personae` correctly reaches `gang-additions.tsx`'s `.map()`.
- New column migration (`NOT NULL DEFAULT false`) is safe for existing rows.
- `supabase/schema/schema.public.sql` wasn't updated, but that's regenerated by the `supabase_schema_snapshot.yml` CI workflow, not a PR-author responsibility here.
- Admin routes remain properly gated behind `checkAdmin`.

## Verdict: Changes requested (finding #1 is a real, easily-reproduced data-loss/UX regression that should be fixed before merge).

---

joeseos:
@Djidiouf Reviewed the latest commit (951ede2, since e18fd65).

## Previous findings

1. **(High) Fighter Name wiped on any non-DP type selection** — **Resolved.** `gang-additions.tsx` (~line 1327-1339) now looks up the *previous* selected type via `selectedGangAdditionTypeId` (still holding the pre-update value in this closure) and only clears `fighterName` when it still equals that previous type's auto-filled name and that previous type was Dramatis Personae. Traced through the relevant scenarios (non-DP → non-DP, DP → DP, DP auto-fill untouched → non-DP, DP auto-fill manually edited → non-DP) and the field is preserved/cleared correctly in each. `tsc --noEmit` is clean on this file.
2. **(Low) No uniqueness/limit enforcement for Dramatis Personae** — Outstanding, not addressed in this commit (previously flagged as possibly out of scope).
3. **(Low) Prefilled name has no visual indicator that it's auto-filled** — Outstanding, minor UX nit, not addressed.

## New findings from this commit

- **(Low / nit)** The new comment says the previous value is available "because React batches the setState call above" (line ~1333-1334). That's not quite why — `selectedGangAdditionTypeId` is just the value closed over from the current render; it hasn't re-rendered yet, batching or not. Not a functional issue, but the wrong mental model here could mislead a future reader into "fixing" this with a ref if the comment is taken literally. Feel free to leave as-is or tweak the wording — not blocking.

No new bugs, regressions, or edge-case issues found in this diff.

## Verdict: Approve

---

joeseos:
@Djidiouf Reviewed the latest commit (1dbca09, since 951ede2).

## Previous findings

1. **(High) Fighter Name wiped on any non-DP type selection** — **Resolved** (unchanged since last review; the closure-based fix in `gang-additions.tsx` still holds up).
2. **(Low) No uniqueness/limit enforcement for Dramatis Personae** — **Outstanding**, not touched by this commit.
3. **(Low) Prefilled name has no visual indicator that it's auto-filled** — **Outstanding**, not touched by this commit.
4. **(Low/nit) Misleading "React batches the setState call" comment** (`gang-additions.tsx:1334`) — **Outstanding**, still present verbatim. Not blocking.

## New finding from this commit

This commit's only change is deleting `supabase/migrations/20260714000001_update_get_fighter_types_with_cost_add_is_dramatis_personae.sql` (834 lines, the DROP/CREATE migration that added `is_dramatis_personae` to the `get_fighter_types_with_cost` RPC). No app code changed.

- **(Medium) Removing that migration breaks migration-history completeness for the RPC change.** The surviving migration (`20260714000000_add_is_dramatis_personae_to_fighter_types.sql`) only does `ALTER TABLE fighter_types ADD COLUMN is_dramatis_personae`. The matching RPC update now lives *only* in `supabase/functions/get_fighter_types_with_cost.sql`, which is deployed to production directly by `.github/workflows/deploy_supabase_functions.yml` on merge to `main` (so `main` itself will be fine) — but it is **not** part of the replayable migration history. Checking this repo's established pattern (e.g. `get_available_skills`'s migration chain: `20260214100001_update_get_available_skills_cost.sql` → `20260616120000_gang_origin_skill_set_access.sql`, each pairing a new migration with the matching `supabase/functions/*.sql` update), every RPC change here has historically shipped a migration *and* a functions-dir update together, precisely so migrations remain a complete rebuild history.

  Concretely, any environment provisioned from `supabase/migrations` alone — local dev via `supabase db reset`, or a Supabase branch (this project has branch-management tooling available) — will get the new `fighter_types.is_dramatis_personae` column, but `get_fighter_types_with_cost` will still be the old function signature without it, so the RPC silently won't return `is_dramatis_personae` and the whole prefill feature will be non-functional there, even though it works on `main`/production.

  Suggested fix: restore a migration that does the `DROP FUNCTION`/`CREATE OR REPLACE FUNCTION` for `get_fighter_types_with_cost` matching the current `supabase/functions/get_fighter_types_with_cost.sql` (a single, non-duplicated one is fine — it doesn't need to be the same 834-line file, just present), or confirm with the team that relying solely on `deploy_supabase_functions.yml` for RPC changes going forward is intentional and migrations are expected to skip these.

No other new bugs, regressions, or edge-case issues in this diff (it's a pure file deletion).

## Verdict: Changes requested

The only reason for changes-requested rather than approve is the migration-history gap above — worth a quick fix or an explicit call from the team before merge, since it affects local dev and any Supabase-branch-based testing of this feature.

---

PR #1864, opened 2026-07-14
Added custom fighter types to the gang addition modal

joeseos:


joeseos:
## Code Review

@joeseos Reviewed this together with the API route change. Checked out the branch, ran `npx tsc --noEmit` and `npx eslint` on the two changed files (both clean), then traced the full selection → submission path to check the actual "Add" flow works for a selected custom fighter type, not just that it's listed.

### Verified correct
- **`app/api/fighter-types/route.ts`**: removing `!isGangAddition` from the custom-fighter-append condition is scoped correctly — it only *adds* the gang-addition case; the existing non-addition "Add Fighter" custom-fighter inclusion path is untouched, and the `customGangTypeId` guard still prevents double-counting for custom gangs (per the comment above it). No regression there.
- **Submission path**: `handleAddFighter` passes the selected id as a bare `fighter_type_id` to `addFighterToGang` with no explicit "is this custom" flag — that had me worried at first (the `.map()` in the query doesn't preserve `is_custom_fighter` past the filter step), but `addFighterToGang` (`app/actions/add-fighter.ts:311-323`) self-determines custom-ness by trying a `custom_fighter_types` lookup (owned, then campaign-shared) before falling back to `fighter_types`. Since the two tables' UUIDs won't collide, this correctly handles a custom fighter type id with no extra wiring needed. Good, this part just works.
- Gang-type scoping (`cf.gang_type_id === gangTypeId` plus the "Available to All" fallback) matches the tooltip's "Only custom fighters matching this gang type will be shown."

### Finding

**[Medium] The class filter is broader than the tooltip claims**
`components/gang/gang-additions.tsx:256-259`:
```js
.filter((type: any) =>
  !type.is_custom_fighter ||
  gangAdditionRank[(type.fighter_class || '').toLowerCase()] !== undefined)
```
and the tooltip text right below (~line 1345):
> "When enabled, your custom fighter types will be included, but only Hanger-on, Brute, Crew, Bounty Hunters."

`gangAdditionRank` (`utils/gangAdditionRank.ts`) isn't a 4-entry allowlist — it's the full ~25-entry rank map used elsewhere in this same file to *order* the Fighter Class dropdown, and it also includes `"hive scum"`, `"house agent"`, `"others"`, `"exotic beast"`, `"special terrain"`, plus ~15 house/guild-specific NPC classes (`"house catallus, carnival"`, `"corpse guild, corpse harvesting party"`, etc.). Since `custom_fighter_types.fighter_class` is free-choice text sourced from the same `fighter_classes` reference table used for official types, a user could plausibly create a custom fighter type with class "Exotic Beast" or "Hive Scum" and have it pass this filter and show up under "Include Custom Fighter Types" — which the tooltip explicitly says won't happen.

Since `custom_fighter_types` has no `is_gang_addition`-equivalent column to filter on properly (checked the schema), reusing `gangAdditionRank` as a cheap "is this a known class" check is a reasonable shortcut — but either the tooltip should be updated to reflect the true (broader) set of eligible classes, or the filter should use an explicit small allowlist (e.g. `['hanger-on', 'brute', 'crew', 'bounty hunter']`) to match what it tells the user. Worth a quick decision before merge since right now the UI's stated behavior and actual behavior diverge.

### Verdict: **Approve, with the tooltip/filter mismatch worth resolving**
This is a small, well-scoped addition and the core wiring (API route, submission, gang-type scoping) is correct. The one finding is a user-facing accuracy issue rather than a functional break — doesn't need to block merge if you're comfortable fixing the copy or filter in a fast follow-up, but flagging since it's a concrete, verifiable mismatch between what the tooltip promises and what the code does.


---

PR #1866, opened 2026-07-14
Merged add fighter and gang addition

joeseos:


joeseos:
## Code Review

@joeseos This is a large, high-risk-looking refactor (deletes a 690-line SQL RPC and two ~1700-line UI components, replaces with a unified modal) that turned out to be very carefully executed. I checked it out, ran `npx tsc --noEmit` and `npx eslint` (both clean on all changed files), then did a systematic correctness pass: SQL merge verification, dangling-reference check, and a full old-vs-new feature-parity comparison.

### SQL merge (`get_fighter_types_with_cost` absorbing `get_add_fighter_details`) — verified correct
I diffed the deleted `get_add_fighter_details.sql` against the new unified WHERE clause line by line:
- `p_is_gang_addition = true` branch: unchanged from before (`ft.is_gang_addition = true`, no gang-type scoping — matches the old gang-additions behavior exactly).
- `p_is_gang_addition = false` branch: `ft.gang_type_id = p_gang_type_id OR (ftgc.fighter_type_id IS NOT NULL AND ftgc.gang_affiliation_id IS NOT NULL AND ftgc.gang_affiliation_id = p_gang_affiliation_id)` — this is a **byte-for-byte reproduction** of the deleted `get_add_fighter_details`'s WHERE clause, and the `ftgc` JOIN condition it depends on is unchanged.
- `NULL` (include-all): both old and new resolve to unconditional `true` — consistent.
- Column parity: the new `RETURNS TABLE` is a strict superset of the old `get_add_fighter_details` columns (`fighter_class_id`, `available_legacies`, and `is_editable` on default-equipment JSON — all fields the roster path needs — are correctly carried over; plus the pre-existing gang-addition-only columns like `alignment`/`alliance_id`/`is_dramatis_personae`).

### Dangling references — none found
No remaining calls to `get_add_fighter_details` anywhere (only doc comments referencing the old name, and the auto-generated `schema.public.sql` snapshot, which per repo convention gets regenerated separately after the migration lands). No imports of the deleted `add-fighter.tsx`/`gang-additions.tsx` anywhere.

### Feature-parity check (old `add-fighter.tsx` + `gang-additions.tsx` vs new `FighterAddModal.tsx`/`EquipmentSelection.tsx`/`utils/equipment-selection.ts`)
Went through equipment selection (all 4 select-type modes), gang variants, default equipment, sub-type switching, optimistic rollback/reconcile, and the submission payload. Net result is equivalent-or-better:
- **Fixed a latent race bug**: the "multiple" equipment checkbox handler now uses functional `setState(prev => …)` instead of the old direct-closure-read pattern.
- **Closed a validation gap**: the "must pick one option, no default" guard (`confirmDisabled`) previously only existed in `gang-additions.tsx`; it's now applied uniformly to both catalogs.
- **Extended optimistic UI to gang additions**: previously only the roster modal got `onFighterRollback`/`onFighterReconcile`; `gang.tsx` now wires both into the additions instance too, and every call site guards on the props being defined, so there's no crash risk where they're absent.
- **Fixed silent data loss**: the old gang-additions submission path remapped equipment through a local type that dropped `is_editable`; the unified modal carries it through for both catalogs now.
- Confirms the earlier `gangAdditionRank`/tooltip mismatch I flagged on #1864 is resolved here — the tooltip now reads generically ("Only custom fighters matching this gang type will be shown") and the filter correctly partitions gang-addition-class custom fighters into the additions catalog vs. everything else into the roster catalog.

### Finding

**[Low] Sub-type switch now resets "Use Delegation Cost"**
`FighterAddModal.tsx`, `handleSelectSubType`:
```js
const handleSelectSubType = (subTypeId: string) => {
  ...
  setUseDelegationCost(false);
  applyDefaultEquipmentAndCost(subTypeId || selectedFighterTypeId, false);
};
```
The old `gang-additions.tsx` sub-type handler never touched `useDelegationCost` — a user's delegation-cost toggle persisted across sub-type switches. Now it's force-reset to `false` on every sub-type change. Given delegation cost is a niche alliance-delegation feature, this is a minor UX change rather than a functional break (the user can just re-check the box), but it wasn't called out in the PR description — worth confirming it's intentional rather than an oversight from consolidating the two `handleSelectSubType` implementations.

### Verdict: **Approve**
This is a thorough, well-verified consolidation — the highest-risk part (the SQL merge) holds up exactly, and the UI unification comes out ahead on net (bug fixes + feature parity extension) with only the one minor, easily-fixed behavioral nit above.


---

PR #1867, opened 2026-07-16
Add email notifications via SES with per-category preferences

joeseos:
Adds an email-delivery path for action-triggered notifications (campaign/
gang invites and friend requests to start), gated by per-user preferences,
built on a transactional outbox so an SES outage can never fail the
originating app action. In-app notifications are unchanged.

- Migration: user_notification_preferences (normalized, RLS own-row) and
  notification_deliveries (service-role-only outbox = durable queue +
  delivery record; unique (notification_id, channel) for idempotency).
- Enqueue trigger on notifications INSERT (local write only, no external
  call) plus a claim_notification_deliveries RPC (FOR UPDATE SKIP LOCKED)
  for concurrency-safe claiming.
- send-notification-email edge function: claims due rows, resolves
  eligibility + recipient (service role), renders via a single inline-CSS
  email layout, sends through the SESv2 API (aws4fetch SigV4), records
  outcome with retry/backoff and a stuck-row reaper. Includes a dryRun
  preview mode.
- Shared, single-source notification config + text renderer in
  utils/notifications; refactors the in-app hook/list to consume them.
- Account "Email preferences" UI: master switch + per-category switches,
  writing prefs via a new server action (RLS-scoped).
- One-click unsubscribe: signed HMAC token, /api/notifications/unsubscribe
  route, and List-Unsubscribe headers on every email.
- Database Webhook definition versioned (secret parameterized) under
  supabase/webhooks/ with a README; edge-function deploy workflow.

Claude-Session: https://claude.ai/code/session_01R8fyExn6UgxKLxod5DzW5o


joeseos:
## Code Review

@joeseos This is a genuinely well-architected feature — transactional outbox, `FOR UPDATE SKIP LOCKED` claiming, RLS correctly scoped (service-role-only outbox with RLS-enabled-no-policies, user-scoped preferences table), constant-time HMAC comparison for the unsubscribe token, RFC 8058 one-click unsubscribe headers, least-privilege IAM guidance, and thoughtful documentation of what can/can't be automated via CI. I checked out the branch, ran `npx tsc --noEmit` and `npx eslint` on the Next.js side (both clean — Deno edge functions aren't part of that toolchain and correctly use the repo's existing `@ts-nocheck` convention, consistent with `discord-campaign-bot`), verified the RLS policies, traced the HMAC token flow end-to-end, and had a background agent trace every `notifications` INSERT producer to check for injection risk in the emailed content.

### Finding

**[Medium] The human-facing `/unsubscribe` page fires the mutation automatically on load — no click required**
`components/unsubscribe-client.tsx`: the moment this page mounts, a `useEffect` immediately POSTs to `/api/notifications/unsubscribe` using the token from the URL — there's no confirm button, just an auto-firing effect.

This undermines the exact protection the RFC 8058 split is meant to provide. The reason `List-Unsubscribe-Post` (a separate header, going straight to the API route) exists alongside the plain `List-Unsubscribe` link is that mail clients only send the POST in response to a genuine user-initiated one-click action, whereas the *visible* link is routinely GET-requested by things that aren't the user — corporate/consumer email security gateways (Microsoft Defender SafeLinks, Proofpoint URL Defense, etc.) commonly pre-fetch every link in an email to scan it, and some render the page. Since `/unsubscribe` executes the real mutation on mere page-load rather than requiring a click, any such prefetch silently unsubscribes the user without their knowledge — a well-documented, real-world failure mode for exactly this kind of "unsubscribe page that acts on load" pattern.

Suggested fix: make `/unsubscribe` show a confirmation with a real "Unsubscribe" button that the user has to click, and only fire the POST from that click handler (or verify the token server-side to show *what* they're unsubscribing from and require a deliberate action instead of an automatic effect). The RFC 8058 path (`unsubscribePostUrl`, hit directly by `List-Unsubscribe-Post` mail clients) can stay exactly as-is — it's already correctly separated from the human page and is the right place for the "no click needed" behavior.

**[Low, defense-in-depth] `ctaUrl` is the one unescaped interpolation in the email template**
`supabase/edge-functions/send-notification-email/_email-layout.ts`: every other user-influenced value going into the HTML (`subject`, `bodyHtml`, `ctaLabel`, `BRAND.name`) is wrapped in `escapeHtml`, but `ctaUrl` is interpolated raw into `href="${ctaUrl}"`. I had this traced end-to-end: for the three currently email-eligible types (`invite`, `gang_invite`, `friend_request`), `link` is always either `NULL` or built exclusively from `uuid`-typed DB columns (`notify_campaign_member_added.sql:29`, `notify_gang_invite.sql:42`, and the TS-side inserts in `battle-sessions.ts`) — so this isn't currently exploitable. But it's inconsistent with the rest of the function and would become an immediate `href`-attribute-breakout vector the moment any future notification type or producer puts free text into `link` without realizing this field is unescaped. Cheap to close now (at minimum escape `"` and validate the scheme is `http(s)`) rather than relying on every future producer remembering this constraint.

### Verified correct (no issues)
- **RLS**: `user_notification_preferences` correctly scoped to `auth.uid() = user_id` on all four operations; `notification_deliveries` has RLS enabled with zero policies (locks out `anon`/`authenticated` entirely, service-role bypasses by design) — correct pattern for a worker-only table.
- **Token security**: HMAC-SHA256 signing, constant-time signature comparison, server-only `UNSUBSCRIBE_SECRET` (not `NEXT_PUBLIC_`-prefixed), expiry checked, category validated against the real eligible set before being trusted in the upsert.
- **Concurrency**: `claim_notification_deliveries` correctly uses `FOR UPDATE SKIP LOCKED`, is `service_role`-only (revoked from `authenticated` too, not just `anon`/`PUBLIC`), and the unique `(notification_id, channel)` constraint + `ON CONFLICT DO NOTHING` in the enqueue trigger makes the outbox idempotent.
- **Notification text escaping**: `notificationTextToHtml`/`escapeHtml` correctly applied to the one field (`text`) that legitimately carries free-form user data (gang/campaign names, usernames interpolated into notification text server-side) — confirmed via the producer trace.
- **Secrets**: no secrets committed anywhere; the webhook trigger (which must embed `WEBHOOK_SECRET` in its header) is deliberately kept out of migrations/CI with a parameterized placeholder and clear manual-apply instructions, matching why `supabase_schema_snapshot.yml` strips these triggers.

### Minor note (not blocking)
There's a narrow at-least-once duplicate-send window: if the worker crashes between a successful SES send and writing `status: 'sent'`, the row stays `'processing'` and gets re-claimed (and re-sent) by the 10-minute stuck-row reaper. Low impact for these notification types (an extra "you have a friend request" email isn't harmful) — just flagging since it's inherent to the outbox design without a two-phase confirm, in case that tradeoff wasn't explicit.

The infrastructure (SQL, RLS, token crypto, retry/backoff, secret handling) is excellent and I'd approve on that alone. The auto-firing unsubscribe page is the one finding I'd want fixed before merge — it's a concrete, well-precedented way for real users to get silently unsubscribed without consent, which cuts against the whole point of shipping opt-in email in the first place.


---

PR #1868, opened 2026-07-16
Added filter options on Activity logs

Djidiouf:


joeseos:
@Djidiouf Thanks for adding this — reviewed the diff against the codebase. One correctness concern to address before merge; a couple of smaller notes below.

### Findings

**1. [High] Fighter/Vehicle filters silently miss history beyond the last 100 logs**
`components/log-modal.tsx` (new filter logic) + `app/api/gangs/[id]/logs/route.ts:121-123`

The route already supports server-side filtering via `?fighterId=`/`?vehicleId=` query params (see `route.ts:107-119`), and `components/fighter/fighter-details-card.tsx:614` already uses this correctly by baking the fighter/vehicle id into `fetchUrl`.

This PR's gang-wide modal (`components/gang/gang.tsx:1311-1314`) instead always fetches the same unfiltered `fetchUrl={/api/gangs/${id}/logs}`, which the API caps at `.limit(100)` most-recent rows, and then filters *client-side* on whatever happens to be in that batch. `useQuery`'s key is `['logs', fetchUrl]`, and `fetchUrl` never changes when a filter is picked, so no server-side refetch/filter ever happens.

For a gang with more than 100 logged events (plausible — logs persist 3 months and cover trading, injuries, XP, etc.), selecting an older or less-active fighter/vehicle can show "no logs match the selected filters" even though real history exists — it's just outside the fetched window. That's misleading, since the UI presents this as a complete filter, not a "last 100" filter.

Suggested fix: mirror the existing pattern — when a fighter/vehicle filter is selected, refetch via `?fighterId=`/`?vehicleId=` (changing the query key so React Query refetches), or otherwise route the id-based filtering through the server instead of only slicing the already-truncated client-side list. The action-type filter can stay client-side since there's no server support for it, but it should probably note in the UI that it only searches recent logs, since it has the same 100-row ceiling.

### Minor notes (no action required, just flagging)
- `components/gang/gang.tsx:292-304` (`allVehiclesForLogs`): dedup logic and typing look correct; confirmed `VehicleProps`/`FighterProps.vehicles` shapes support this.
- `Combobox`/`Button` usage (`outline`, `sm`) matches existing variants — no issues there.
- `vehicle_id` was already a column supported by the route's existing filter logic, so exposing it in the API response (`route.ts:144`) is safe and not a new schema dependency.

The high-severity item is the one blocker — it undermines the actual value of the feature for exactly the gangs that would benefit most (long-lived, active ones). Everything else (UI, empty states, pagination reset on filter change, dedup logic) looks solid.

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`b5409a0`, diff vs `754644c` — only `components/log-modal.tsx` changed, +28/-2).

### Previous findings
- **[High] Fighter/Vehicle filters silently miss history beyond the last 100 logs — Resolved.** `effectiveFetchUrl` now routes `filterFighterId`/`filterVehicleId` through the server-side `?fighterId=`/`?vehicleId=` params (mirroring `fighter-details-card.tsx`'s pattern) and includes them in the React Query key, so selecting a fighter/vehicle now triggers a real server-side refetch instead of slicing the already-capped client list. Verified `app/api/gangs/[id]/logs/route.ts` supports both params (including the `.or()` combo for Crew fighters). A caveat note was also added in the UI for the still-client-side Action Type filter, as suggested.
- Minor notes from the previous pass (dedup logic, Combobox/Button usage, exposing `vehicle_id` in the API response) — no changes needed, still fine.

### New findings
None blocking. One edge case worth being aware of but not worth holding up the PR for:
- `components/log-modal.tsx:60-69` — when **both** a Fighter and a Vehicle filter are selected simultaneously, the server query becomes an OR (`fighter_id.eq.F,vehicle_id.eq.V`, per the route's existing logic) capped at 100 rows, while the client then applies an AND (`filteredLogs`, log-modal.tsx:101-113) to narrow to the exact combo. If a gang has many logs matching just one side of the OR, the true fighter+vehicle intersection could still be pushed out of the top-100 window. This is a pre-existing limitation of the API (no combined AND query param) rather than a regression from this change, and is a narrow scenario (a Crew fighter with >100 more-recent unrelated logs). Not blocking — flagging in case it's worth a follow-up.

### Verdict: **Approve**

The blocking issue from the previous review is fixed correctly and CI (`deploy-preview`) is green. Nice work.

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`93bd412`, diff vs `b5409a0` — only `app/api/gangs/[id]/logs/route.ts` and `components/log-modal.tsx` changed, +16/-2).

### Previous findings
- **[High] Fighter/Vehicle filters silently miss history beyond the last 100 logs — Resolved** (confirmed in the prior pass on `b5409a0`, unchanged since).
- **Follow-up edge case (Fighter+Vehicle combined filter truncation) — Resolved.** This commit is exactly the fix for the non-blocking edge case flagged in the last review: when both `filterFighterId` and `filterVehicleId` are set, `log-modal.tsx` now appends `combine=and` to the query, and the route (`route.ts:110-113`) uses that to switch from `.or(...)` to `.eq('fighter_id', ...).eq('vehicle_id', ...)`, so the true intersection is computed server-side instead of being sliced out of a `.limit(100)` OR result. Verified the default (no `combine` param) path is untouched, so `fighter-details-card.tsx:614` — which builds `fighterId`+`vehicleId` directly into its `fetchUrl` for the Crew-fighter-with-own-vehicle OR case, and never sets `combine` — keeps its existing OR behavior. That component also doesn't pass `fighters`/`vehicles` props to `LogModal`, so its filter UI (and thus `effectiveFetchUrl`'s combine logic) never activates there anyway — no interaction between the two call sites.
- Minor notes from earlier passes (dedup logic, Combobox/Button usage, exposing `vehicle_id` in the API response) — still fine, no changes needed.

### New findings
None. The added `console.log` in the new `and` branch (`route.ts:112`) just mirrors the logging style already used by the sibling OR/single-id branches in this same function — not a new pattern introduced by this change.

### Verdict: **Approve**

This closes out the last remaining (non-blocking) gap from the previous review. Nice follow-through.

---

PR #1869, opened 2026-07-16
Updated the package-lock.json files to add swc dependencies

Djidiouf:


joeseos:
## Code Review

@Djidiouf Thanks for the fix — reviewed the diff against the rest of `package-lock.json`. There's one issue worth fixing before merge.

### Findings

**1. [High] New `@next/swc-win32-x64-msvc` entry looks hand-edited, not `npm install`-generated — `package-lock.json` lines 9069–9083**

The PR adds a top-level `node_modules/@next/swc-win32-x64-msvc` entry, but a nested `node_modules/next/node_modules/@next/swc-win32-x64-msvc` entry for the **exact same version (16.2.9)** already exists in the lockfile (unchanged by this PR, around line 7026).

That's the tell: I checked every other `node_modules/next/node_modules/*` entry in the lockfile, and the only other one is `postcss`, which is nested precisely *because* its version differs from the top-level copy (8.4.31 vs 8.5.15) — that's the legitimate reason npm keeps a nested duplicate. Here, both copies of `@next/swc-win32-x64-msvc` are version 16.2.9 with identical `resolved`/`integrity`. A real `npm install`/`npm ci` run would hoist this package and *remove* the nested copy rather than add a second, identical top-level one.

This is confirmed by the new entry also missing the `"license": "MIT"` field that every sibling `@next/swc-*` platform package has (including the nested win32-x64-msvc entry itself and the neighboring `@next/swc-win32-arm64-msvc` top-level entry).

**Fix:** Don't hand-edit `package-lock.json`. Instead, regenerate it properly, e.g.:
```
rm -rf node_modules package-lock.json
npm install
```
(or `npm install --package-lock-only` if you just need the lock refreshed), then verify `npm ci` installs cleanly and check whether the pre-existing nested `next/node_modules/@next/swc-win32-x64-msvc` entry is resolved/removed as expected. Commit the actual generated lockfile rather than a manually patched diff — otherwise this file will keep drifting from what npm actually produces and cause unexpected diffs/install issues down the line (this is exactly the kind of Windows-optional-dependency drift this PR is trying to fix).

### Checklist
- Correctness: ⚠️ see finding above — the added entry doesn't reflect what npm would actually generate
- Edge cases / null handling: N/A (lockfile-only change)
- Security: no concerns — integrity hash matches the legitimate npm registry tarball for this version
- Performance: N/A
- Readability/maintainability: N/A (generated file)
- Project conventions: no `CLAUDE.md` in this repo to check against
- DRY/YAGNI: N/A

Please regenerate the lock file via `npm install` rather than hand-adding the entry, so the tree stays consistent (no duplicate/inconsistent nested vs. top-level copies) and matches what CI/other devs will get when they run `npm ci`.

---

joeseos:
## Code Review (update)

@Djidiouf Reviewed the new commit (`69c8d07`) against the previous review.

### Previous findings
1. **[High] Hand-edited `@next/swc-win32-x64-msvc` entry (missing `license` field, stale nested duplicate)** — **Resolved.** The lockfile was regenerated properly:
   - The old hand-added top-level entry (v16.2.9, no `license` field) is gone.
   - The stale nested `node_modules/next/node_modules/@next/swc-win32-x64-msvc` duplicate is gone.
   - A single top-level entry now exists at v16.2.10, in alphabetical position, with `"license": "MIT"` matching its sibling `@next/swc-*` platform packages exactly.

### New changes since last review
The new commit is a full lockfile regeneration (`rm -rf node_modules package-lock.json && npm install`), as suggested — not a scoped patch. This bumped ~135 packages to newer patch/minor versions (`next` 16.2.9→16.2.10, ESLint, TypeScript-ESLint, etc.) and applied the same hoist-fix pattern to two other previously-duplicated Windows-arch optional deps: `lightningcss-win32-x64-msvc` and `@tailwindcss/oxide-win32-x64-msvc` (both had the same top-level + nested duplicate issue, now collapsed to one entry each). This is expected fallout of regenerating from a deleted lockfile and looks legitimate, not hand-edited.

Checked for risk:
- No major-version bumps among the ~135 changed packages — all patch/minor, within existing `package.json` semver ranges.
- `package.json` itself is untouched (`changed_files: 1`, lockfile only).
- `lockfileVersion` unchanged.
- `deploy-preview` check on the head commit is green, so `npm ci`/install and the build both succeed with the regenerated file.

No new issues found. Edge cases and readability don't really apply to a generated lockfile.

### Verdict: **Approve**

Thanks for regenerating it properly — this is what a real `npm install` output should look like.

---

PR #1870, opened 2026-07-17
Added new printing option for Cards to be scaled with content

Djidiouf:


joeseos:
## Review

Reviewed `components/gang/fighter-card.tsx` and `components/gang/print-gang.tsx` (17 additions / 4 deletions).

**Summary:** Small, well-scoped, opt-in feature. `scaleToContent` is an optional prop that defaults to `false`, so the two other `FighterCard` consumers (`sortable-fighter.tsx`, `participant-card.tsx`) are unaffected, and the checkbox is correctly gated behind `viewMode === "cards"`. `isPrintAutoHeight` disabling `useTightLayoutWrapping` (switching the inner wrapper divs from `absolute inset-0 overflow-hidden` to `contents`) is the right call — with `h-auto` the old absolutely-positioned wrapper would've collapsed to zero height since it no longer has an explicit-height ancestor to fill. `tsc --noEmit` is clean on both files.

### Findings

**1. [Low] Horizontal padding shrinks by ~8px/side when "Scale each Card to Content" is on**
`components/gang/fighter-card.tsx:196` (`useTightLayoutWrapping = isTightLayoutView && !isPrintAutoHeight`) and `:498-499`.

When `isPrintAutoHeight` is true, the inner wrapper collapses to `contents`, which also drops the `pl-2 pr-2` (8px) horizontal padding that print view previously always got on top of the card's own `p-2`. So a scaled-to-content card ends up with ~8px less padding on each side than a fixed-height print card (16px effective vs 8px), while vertical padding stays `p-2` either way. Probably not visible enough to block, but worth a quick print-preview comparison between the two modes to confirm it's the intended look — if not, the horizontal padding could be preserved independently of the `absolute/overflow-hidden` wrapping (e.g. keep `pl-2 pr-2` via a separate class list rather than bundling it with `useTightLayoutWrapping`).

No correctness, security, or DRY concerns found — the `scaleCardsToContent` state mirrors the existing `showAdditionalDetails`/`cardsGangCardsPosition` pattern in the file, and the card-wrapper `h-auto!`/`h-[435px]!` Tailwind arbitrary-variant toggle in `print-gang.tsx:998` is consistent with the per-card `printCardSize` logic.

**Verdict: Approve** (the one finding above is a minor visual nit, not a blocker).

cc @Djidiouf

---

joeseos:
## Review of latest commit (e1f6c95)

Reviewed just the new commit since the last review — a 1-line change in `components/gang/fighter-card.tsx:499`.

### Previous findings

**[Low] Horizontal padding shrinks by ~8px/side when "Scale each Card to Content" is on** (`fighter-card.tsx:196, 498-499`) — **Partially fixed.**
The new commit changes the inner wrapper's else-branch from `'contents'` to `'pl-2 pr-2'`:
```diff
-      <div className={useTightLayoutWrapping ? 'pl-2 pr-2 min-w-0 w-full h-full' : 'contents'} style={compactInnerScaleStyle}>
+      <div className={useTightLayoutWrapping ? 'pl-2 pr-2 min-w-0 w-full h-full' : 'pl-2 pr-2'} style={compactInnerScaleStyle}>
```
This does restore the missing padding for the `isPrintAutoHeight` case that was flagged. But `useTightLayoutWrapping` is `false` for **two** cases — `isPrintAutoHeight` *and* the default `normal` viewMode — and the fix applies to both, which introduces a new regression (see below).

### New findings

**1. [Medium] Unintended padding regression on the default `normal` card view — `components/gang/fighter-card.tsx:499`**

`useTightLayoutWrapping = isTightLayoutView && !isPrintAutoHeight` is `false` for `normal` viewMode too, not just for `isPrintAutoHeight`. Before this fix, that branch rendered `'contents'` (no box, no padding) for normal-view cards. Now it renders `'pl-2 pr-2'` — a real block box adding 8px of horizontal padding.

Normal-view cards already get `p-4` (16px) on all sides from the outer `cardClassName` (`fighter-card.tsx:474`, `isNormalView ? 'p-4' : ...`). With this change they'll now have ~24px effective left/right padding vs. 16px top/bottom — an asymmetric padding regression on the most common card view in the app (gang page main fighter list via `SortableFighter`, and the battle-session participant info modal in `participant-card.tsx`), neither of which has anything to do with the print scale-to-content feature.

Suggested fix — scope the padding to `isPrintAutoHeight` specifically instead of the inverse of `useTightLayoutWrapping`, so normal view keeps its original `'contents'`:
```tsx
<div className={useTightLayoutWrapping ? 'pl-2 pr-2 min-w-0 w-full h-full' : isPrintAutoHeight ? 'pl-2 pr-2' : 'contents'} style={compactInnerScaleStyle}>
```

No other issues found in this commit — it's otherwise a minimal, targeted change.

**Verdict: Changes requested** (the padding fix needs to be scoped to `isPrintAutoHeight` rather than applied whenever `useTightLayoutWrapping` is false, to avoid touching the default normal-view card layout).

cc @Djidiouf

---

joeseos:
## Review of latest commit (578412f)

Reviewed just the new commit since the last review — a 1-line change in `components/gang/fighter-card.tsx:499`.

### Previous findings

**[Low] Horizontal padding shrinks by ~8px/side when "Scale each Card to Content" is on** (original finding, `fighter-card.tsx:196, 498-499`) — **Resolved.** The `isPrintAutoHeight` branch now gets `pl-2 pr-2` again.

**[Medium] Unintended padding regression on the default `normal` card view** (`fighter-card.tsx:499`, flagged in the previous review round) — **Resolved.** This commit changes the fallback branch from unconditional `'pl-2 pr-2'` to `isPrintAutoHeight ? 'pl-2 pr-2' : 'contents'`:
```diff
-      <div className={useTightLayoutWrapping ? 'pl-2 pr-2 min-w-0 w-full h-full' : 'pl-2 pr-2'} style={compactInnerScaleStyle}>
+      <div className={useTightLayoutWrapping ? 'pl-2 pr-2 min-w-0 w-full h-full' : isPrintAutoHeight ? 'pl-2 pr-2' : 'contents'} style={compactInnerScaleStyle}>
```
This is exactly the scoping fix suggested previously. Walking the three relevant states confirms it's now correct:
- `isPrintAutoHeight` (print + scale-to-content) → `pl-2 pr-2` — padding restored, matches the original finding's intent.
- `isTightLayoutView && !isPrintAutoHeight` (print fixed-height, or compact `2/3/4-card` views) → unchanged, still `useTightLayoutWrapping` branch.
- `isNormalView` (default gang-list / participant-card view) → falls through to `'contents'`, i.e. back to pre-PR behavior. Since `isPrintAutoHeight` requires `isPrintView`, and `isNormalView` implies `!isPrintView`, these two are mutually exclusive, so there's no case where normal view picks up the extra padding.

### New findings

None. The change is minimal and correctly scoped; no new bugs, regressions, or edge cases introduced.

**Verdict: Approve**

cc @Djidiouf

---

PR #1871, opened 2026-07-17
prevent crash when adding a lasting injury via gang page quick action…

Djidiouf:
…. Setting rig-glitches to undefined when a fighter has no existing

entry left a stray undefined item after Object.values(effects).flat(), which applySpecialRulesModifiers then dereferenced. Default to an empty array and guard effect loops against falsy entries.

### DIFF
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -711,6 +711,7 @@ export function EditFighterModal({
     const removed = new Set<string>();
     if (fighter.effects) {
       Object.values(fighter.effects).flat().forEach((effect: any) => {
+        if (!effect) return;
         const tsd = typeof effect.type_specific_data === 'object' && effect.type_specific_data
           ? effect.type_specific_data : null;
         if (tsd) {
--- a/components/gang/gang-page-content.tsx
+++ b/components/gang/gang-page-content.tsx
@@ -736,7 +736,7 @@ export default function GangPageContent({
                           injuries: isSpyrer ? [] : updatedInjuries,
                           'rig-glitches': isSpyrer
                             ? updatedInjuries
-                            : f.effects['rig-glitches'],
+                            : (f.effects?.['rig-glitches'] || []),
                         },
                       };
                     }),
--- a/utils/effect-modifiers.ts
+++ b/utils/effect-modifiers.ts
@@ -294,6 +294,11 @@ export function applyWeaponModifiers(
     return profiles;
   }
 
+  // Guard against stray falsy entries (e.g. from Object.values(effects).flat()
+  // where a category was set to undefined instead of an empty array)
+  const validEffects = effects.filter(Boolean);
+  if (validEffects.length === 0) return profiles;
+
   return profiles.map((profile) => {
     // Work on a copy
     const modified = { ...profile };
@@ -307,7 +312,7 @@ export function applyWeaponModifiers(
 
       // Collect all modifiers for this field
       const fieldModifiers: EffectModifier[] = [];
-      effects.forEach(eff => {
+      validEffects.forEach(eff => {
         (eff.fighter_effect_modifiers || []).forEach(m => {
           if (m.stat_name === fieldName) {
             fieldModifiers.push(m);
@@ -340,8 +345,7 @@ export function applyWeaponModifiers(
       .map((t: string) => t.trim())
       .filter(Boolean);
 
-    effects.forEach((eff) => {
-      if (!eff) return;
+    validEffects.forEach((eff) => {
       const tsd = typeof eff.type_specific_data === 'object' ? eff.type_specific_data || {} : {};
       const toRemove: string[] = tsd.traits_to_remove || [];
       const toAdd: string[] = tsd.traits_to_add || [];
--- a/utils/effect-modifiers.ts
+++ b/utils/effect-modifiers.ts
@@ -341,6 +341,7 @@ export function applyWeaponModifiers(
       .filter(Boolean);
 
     effects.forEach((eff) => {
+      if (!eff) return;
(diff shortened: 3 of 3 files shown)


joeseos:
## Code Review

@Djidiouf Thanks for the fix — reviewed against the full codebase. Summary below.

### What the PR does
Root cause was correctly diagnosed: `'rig-glitches': f.effects['rig-glitches']` in `gang-page-content.tsx`'s `onInjuryUpdate` could explicitly set the key to `undefined` (not omit it), which survives `Object.values(effects).flat()` as a scalar `undefined` element (array `.flat()` doesn't drop non-array/falsy elements). `applySpecialRulesModifiers` then dereferenced `eff.type_specific_data` on that `undefined`, crashing the "add lasting injury" quick action.

- `gang-page-content.tsx:739` — `f.effects?.['rig-glitches'] || []` correctly guarantees an array instead of `undefined`. This also matches the existing pattern already used at `gang-page-content.tsx:528`, so it's consistent with the file's conventions. ✅
- `effect-modifiers.ts` — added `if (!eff) return/continue` guards in `applyWeaponModifiers`'s traits loop and in `applySpecialRulesModifiers`. These are good defense-in-depth since several other call sites (`fighter-card.tsx:273/371`, `print-gang.tsx:572/578`, `fighter-edit-modal.tsx:706`) already pass `Object.values(effects).flat()` into `applySpecialRulesModifiers` **without** `.filter(Boolean)`, so this guard protects real, currently-reachable call sites, not just the reported crash.

### Findings

**1. Medium — `applyWeaponModifiers`'s numeric-fields loop is left unguarded, inconsistent with the fix's own rationale**
`utils/effect-modifiers.ts:310`
```js
effects.forEach(eff => {
  (eff.fighter_effect_modifiers || []).forEach(m => { ... });
});
```
This loop iterates the exact same `effects` array as the traits loop four lines below it (which *was* guarded with `if (!eff) return` at line 344), inside the same `applyWeaponModifiers` function. If `effects` ever contains a falsy hole — which is precisely the failure mode this PR is fixing, and which several sibling call sites of `applySpecialRulesModifiers` already produce via unfiltered `Object.values(...).flat()` — this line throws `Cannot read properties of undefined (reading 'fighter_effect_modifiers')` before ever reaching the guarded traits loop.

Today's `applyWeaponModifiers` callers (`fighter-data.ts:467`, `gang-data.ts:1715`, `stash-tab.tsx:515`, `fighter-page.tsx:980`, `fighter-equipment-list.tsx:440`) all build their effects arrays via `.map()`/`Map.get()`/array literals, so this isn't live today — but it's a latent regression waiting for the next caller that reuses the `Object.values(...).flat()` pattern already common elsewhere in this file's callers.

*Suggested fix:* add the same `if (!eff) return;` at the top of the `effects.forEach` on line 310 (or better, do a single `effects.filter(Boolean)` at the top of `applyWeaponModifiers` and drop the per-loop guards).

**2. Low — same unguarded-`Object.values(...).flat()` pattern still exists outside this PR's scope**
`components/fighter/edit-fighter/fighter-edit-modal.tsx:713-719`
```js
Object.values(fighter.effects).flat().forEach((effect: any) => {
  const tsd = typeof effect.type_specific_data === 'object' && effect.type_specific_data
    ? effect.type_specific_data : null;
  ...
});
```
This iterates directly (not through `applySpecialRulesModifiers`) and has no falsy-entry guard, so it's exposed to the same crash class if any effects category is ever set to `undefined` again elsewhere. Not introduced by this PR and not currently reachable via the fixed code path, but worth a follow-up given the stated intent to guard against stray `undefined` entries.

### Other checks
- No null/edge case issues in the changed lines beyond the above.
- No security or performance concerns — this is a client-side optimistic-update guard.
- Diff is minimal and consistent with existing code style in both files; no DRY/YAGNI issues in the changed lines themselves.
- No CLAUDE.md present in the repo to check against; nothing in README.md conventions is violated.

### Verdict: **Approve** (with a suggested follow-up)
The core fix is correct and resolves the reported crash. Finding #1 is worth addressing (either in this PR or a fast follow) since it's in the same function this PR is hardening and is a one-line change, but it isn't a regression relative to `main` and doesn't block merging this fix.


---

joeseos:
## Follow-up Review

@Djidiouf Reviewed the latest commit (`5e9c8c2`, since the prior review of `afd7705`). This only touches `components/fighter/edit-fighter/fighter-edit-modal.tsx` and `utils/effect-modifiers.ts`; `gang-page-content.tsx` is unchanged from the reviewed commit.

### Previous findings

**1. Medium — unguarded numeric-fields loop in `applyWeaponModifiers`** (`utils/effect-modifiers.ts:310`) — **Resolved.**
`effects` is now filtered once via `const validEffects = effects.filter(Boolean)` at the top of the function, with an early `if (validEffects.length === 0) return profiles;` bail-out, and `validEffects` is used consistently in both the numeric-fields loop (line 315) and the traits loop (line 348). This is actually a cleaner fix than the originally suggested `if (!eff) return` per-loop guard, since it also removes the now-redundant guard that was in the traits loop and centralizes the defense at the top of the function.

**2. Low — unguarded `Object.values(...).flat()` in `fighter-edit-modal.tsx`** (`fighter-edit-modal.tsx:713-719`) — **Resolved.**
`if (!effect) return;` added right after the `.forEach`, consistent with the guard style used elsewhere in this PR.

### New issues in the latest commit
None found. The diff is minimal (8 lines), consistent with the file's existing conventions, and doesn't introduce any new edge cases. `applySpecialRulesModifiers`'s existing `if (!eff) continue` guard (from the first commit) is untouched and still correct.

One non-blocking observation: with the new early return, `applyWeaponModifiers` now returns the original `profiles` array reference (rather than a shallow-copied array) when every effect is falsy. That's a behavior difference only for reference-equality checks, which nothing in this codebase relies on — not worth blocking on.

### Verdict: **Approve**
Both outstanding findings from the previous review are addressed and no new issues were introduced. Good to merge.

---

PR #1872, opened 2026-07-17
Split invite notification type into campaign_invite and battle_invite

joeseos:
## Summary
This PR splits the overloaded `invite` notification type into two distinct types: `campaign_invite` (for campaign-member invitations) and `battle_invite` (for battle-session invitations). This allows for separate email preferences and more specific messaging for each invitation type.

## Key Changes

- **Database Migration**: Added `20260717120000_split_invite_type.sql` that:
  - Updates the `notifications_type_check` constraint to allow both `campaign_invite` and `battle_invite`
  - Retains the legacy `invite` type for backward compatibility with existing notification rows
  - Migrates any existing user preferences from `invite` to `campaign_invite`

- **Notification Configuration** (`utils/notifications.ts`):
  - Renamed `invite` config to `campaign_invite` with email support enabled
  - Updated subject line to "You've been invited to a campaign"
  - Added explicit `battle_invite` config (in-app only, no email)
  - Kept legacy `invite` config for historical rows with no email support

- **Email Delivery Logic**:
  - Updated `enqueue_notification_email.sql` trigger to check for `campaign_invite` instead of `invite`
  - Updated `send-notification-email/index.ts` to use `campaign_invite` in email config
  - Updated schema definition to reflect the new constraint

- **Battle Session Creation** (`app/actions/battle-sessions.ts`):
  - Changed battle session invitation notifications from type `invite` to `battle_invite`
  - Applied to both initial session creation and participant additions

- **UI Updates** (`components/account/notifications-content.tsx`):
  - Added `LuSwords` icon for `battle_invite` notifications (rose color)
  - Updated icon mapping to handle both legacy `invite` and new `campaign_invite` types
  - Distinct visual representation for battle invitations

- **SQL Functions**:
  - Updated `notify_campaign_member_added.sql` to emit `campaign_invite` type

## Implementation Details

- The legacy `invite` type is intentionally preserved in the CHECK constraint to maintain compatibility with existing notification rows that haven't expired (30-day default retention)
- The migration is idempotent and defensive, handling cases where user preferences may have already been stored
- Email delivery is now type-specific: `campaign_invite` supports email, `battle_invite` is in-app only
- The deployment order is critical: this migration must be applied before SQL function producers auto-deploy to avoid constraint violations

https://claude.ai/code/session_01P3iBqkgJ7NrWuQEwGU1M8g

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,7 +524,7 @@ The notification system provides real-time notifications to users for various ap
 interface Notification {
   id: string;
   text: string;
-  type: 'info' | 'warning' | 'error' | 'invite';
+  type: 'info' | 'warning' | 'error' | 'invite' | 'campaign_invite' | 'friend_request' | 'battle_invite' | 'gang_invite';
   created_at: string;
   dismissed: boolean;
   link: string | null;
--- a/app/actions/battle-sessions.ts
+++ b/app/actions/battle-sessions.ts
@@ -186,7 +186,7 @@ export async function createBattleSession(params: {
             otherGangs.map((g) => ({
               receiver_id: g.user_id,
               sender_id: user.id,
-              type: 'invite',
+              type: 'battle_invite',
               text: `${senderName} added you to a battle session.`,
               link: `/gang/${g.id}/battle-session/${sessionId}`,
               dismissed: false,
@@ -708,7 +708,7 @@ export async function addParticipant(params: {
       await supabase.from('notifications').insert({
         receiver_id: params.user_id,
         sender_id: currentUser.id,
-        type: 'invite',
+        type: 'battle_invite',
         text: `${senderName} invited you to a battle session.`,
         link: `/gang/${params.gang_id}/battle-session/${params.session_id}`,
         dismissed: false,
--- a/components/account/notifications-content.tsx
+++ b/components/account/notifications-content.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { useCallback, useState, useEffect } from 'react';
-import { LuOctagonX, LuUserPlus, LuTriangleAlert } from "react-icons/lu";
+import { LuOctagonX, LuUserPlus, LuTriangleAlert, LuSwords } from "react-icons/lu";
 import { LuCheck } from "react-icons/lu";
 import { ImInfo } from "react-icons/im";
 import { HiX } from "react-icons/hi";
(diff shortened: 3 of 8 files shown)


joeseos:
## Code Review

@joeseos Reviewed against the full codebase (checked out the branch to confirm all producers/consumers were actually updated, not just the diff).

### What the PR does
Splits the overloaded `invite` notification type into `campaign_invite` (email-eligible) and `battle_invite` (in-app only), while keeping the legacy `invite` value valid in the CHECK constraint and in `notificationEmailConfig` (non-emailable) so existing unexpired rows still render. All current producers (`notify_campaign_member_added.sql`, `battle-sessions.ts` ×2) were updated to emit the new values — verified via `git grep` on the PR branch, no leftover `type: 'invite'` producer.

### Findings

**1. Medium — deploy-order safety is asserted in a comment, not enforced by anything in this repo's CI**
`supabase/migrations/20260717120000_split_invite_type.sql:19-22`
The migration comment states this must be applied *before* the SQL-function producers auto-deploy, because `deploy_supabase_functions.yml` triggers independently on push to `main` when `supabase/functions/*.sql` changes. I checked `.github/workflows/` and there is **no workflow in this repo that applies `supabase/migrations/*.sql`** — `supabase_schema_snapshot.yml` only *dumps* the schema nightly, it doesn't deploy migrations. So the ordering guarantee this comment relies on must come from an external mechanism (e.g. Supabase's own GitHub-linked migration deploy) that isn't visible or enforced here. If that external migration apply lags behind (or races) `deploy_supabase_functions.yml` on the same push, `notify_campaign_member_added()` would start emitting `type = 'campaign_invite'` before the CHECK constraint allows it, and every campaign invite would 500 until the migration catches up.

Worth either confirming (in the PR description, for the record) exactly what applies `supabase/migrations/*` and that it's guaranteed to run first, or reducing the blast radius by gating the function's new-value emission behind something that can't race (not necessary if the external ordering is actually guaranteed — just flagging that the guarantee isn't checkable from the repo alone).

**2. Low — README.md notification type list is now further out of date**
`README.md:527`
```ts
type: 'info' | 'warning' | 'error' | 'invite';
```
This was already stale before this PR (missing `friend_request`/`gang_invite`/`battle_invite`), so not a regression, but this PR is a natural point to fix it since it's specifically about the type list. Not blocking.

### Things that look correct
- CHECK constraint update in the migration and in `schema.public.sql` match exactly (same 8-value array, same order).
- `user_notification_preferences` migration step is correctly idempotent (`ON CONFLICT (user_id, notification_type) DO NOTHING` matches the actual PK on that table, confirmed in `schema.public.sql:6848`) and safe to rerun.
- `notificationEmailConfig.invite` is kept with `supportsEmail: false`, so `emailEligibleTypes()` correctly excludes it from the account preferences UI — legacy rows won't produce a confusing duplicate "Invitations" toggle next to "Campaign invitations".
- Icon switch in `notifications-content.tsx` correctly falls through `invite`/`campaign_invite` to the same icon, and neither new type was added to the `friend_request`/`gang_invite` click-suppression list, so `campaign_invite`/`battle_invite` notifications remain clickable (matches prior `invite` behavior — no regression).
- Deno edge function `EMAIL_CONFIG` mirror was correctly updated to swap `invite` → `campaign_invite`, keeping it in sync with `utils/notifications.ts` per the file's own convention comment.
- No security or performance concerns; this is a narrow, well-scoped data-model change.

### Verdict: **Approve**
The change itself is correct and complete across all producers/consumers I could find. Finding #1 is worth a quick confirmation from whoever owns the Supabase deploy setup before/at merge, but I can't verify it's actually broken from the repo alone, so it shouldn't block merging.


---

PR #1873, opened 2026-07-17
Fixed an issue where fighters from one gang type added to another gan…

Djidiouf:
…g of a different gang type couldn't be promoted to a gang type of its own. ie: a Cawdor Ganger in an Escher gang gets Escher options instead of Cawdor ones.

### DIFF
--- a/app/api/fighter-types/route.ts
+++ b/app/api/fighter-types/route.ts
@@ -58,6 +58,7 @@ function transformCustomFighter(cf: any) {
     gang_type: cf.gang_type,
     cost: cf.cost,
     gang_type_id: cf.gang_type_id,
+    custom_gang_type_id: cf.custom_gang_type_id ?? null,
     special_rules: cf.special_rules || [],
     total_cost: cf.cost,
     movement: cf.movement,
--- a/app/fighter/[id]/page.tsx
+++ b/app/fighter/[id]/page.tsx
@@ -116,12 +116,11 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
       .map((beast: any) => beast.fighter_pet_id)
       .filter(Boolean);
 
-    // Parallel batch: beast fighters, gang variants, captured-by gang name, and custom fighter type info
+    // Parallel batch: beast fighters, gang variants, and captured-by gang name
     const [
       beastFightersResult,
       gangVariantsResult,
-      capturedByGangResult,
-      customFighterTypeResult
+      capturedByGangResult
     ] = await Promise.all([
       // Beast fighters (only if needed)
       beastIds.length > 0
@@ -155,19 +154,11 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
             .eq('id', fighterBasic.captured_by_gang_id)
             .single()
             .then((result: any) => result.error ? null : result.data?.name ?? null)
-        : Promise.resolve(null),
-      // Custom fighter type source gang info (needed to find correct promotion options)
-      fighterBasic.custom_fighter_type_id
-        ? supabase
-            .from('custom_fighter_types')
-            .select('gang_type_id, custom_gang_type_id')
-            .eq('id', fighterBasic.custom_fighter_type_id)
-            .single()
-            .then((result: any) => result.error ? null : result.data)
         : Promise.resolve(null)
     ]);
 
-    const customFighterTypeInfo = customFighterTypeResult;
+    // Custom fighter type gang info is embedded in fighterBasic via the FK join
+    const customFighterTypeInfo = fighterBasic.custom_fighter_type ?? null;
 
     // Campaign data already includes trading_post_names from getGangCampaigns, no processing needed
 
--- a/app/fighter/[id]/page.tsx
+++ b/app/fighter/[id]/page.tsx
@@ -68,7 +68,8 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
       beastDataResult,
       ownershipDataResult,
       gangFighters,
-      loadouts
+      loadouts,
+      fighterTypeSourceGangTypeId
     ] = await Promise.all([
       // Gang data
       getGangBasic(fighterBasic.gang_id, supabase),
@@ -96,7 +97,17 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
       // Gang fighters list
       getGangFighters(fighterBasic.gang_id, supabase),
       // Fighter loadouts
-      getFighterLoadouts(id, supabase)
+      getFighterLoadouts(id, supabase),
+      // Fighter type's source gang_type_id — direct query (not cached) so it is always fresh.
+      // This is needed to show the correct promotion options for fighters added from another house.
+      fighterBasic.fighter_type_id
+        ? supabase
+            .from('fighter_types')
+            .select('gang_type_id')
+            .eq('id', fighterBasic.fighter_type_id)
+            .single()
+            .then((r: any) => r.error ? null : (r.data?.gang_type_id ?? null))
+        : Promise.resolve(null)
     ]);
 
     // Check if gang exists (shouldn't happen but handle gracefully)
@@ -116,11 +127,12 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
       .map((beast: any) => beast.fighter_pet_id)
       .filter(Boolean);
 
-    // Parallel batch: beast fighters, gang variants, and captured-by gang name (non-critical queries)
+    // Parallel batch: beast fighters, gang variants, captured-by gang name, and custom fighter type info
     const [
       beastFightersResult,
       gangVariantsResult,
-      capturedByGangResult
+      capturedByGangResult,
+      customFighterTypeResult
     ] = await Promise.all([
       // Beast fighters (only if needed)
       beastIds.length > 0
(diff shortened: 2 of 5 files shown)


joeseos:
## Code Review

@Djidiouf Thanks for the fix — reviewed against the full codebase. Good catch on the underlying cause (using the owning gang's `gang_type_id` instead of the fighter type's own `gang_type_id`). A few issues found, one of which means the fix is incomplete.

### Findings

**1. [HIGH] Fix not applied to the "Edit Fighter → Change Fighter Type" dropdown — same bug still reproduces there**
`components/fighter/fighter-page.tsx:1282-1283`
```jsx
gangTypeId={fighterData.gang?.gang_type_id}
customGangTypeId={fighterData.gang?.custom_gang_type_id}
```
This is the `EditFighterModal` invocation (its own "Change Fighter Type" dropdown, `fighter-edit-modal.tsx:113-129`). It still passes the *owning gang's* type, exactly like the buggy code being fixed. Compare with `AdvancementsList` three lines above it (`fighter-page.tsx:822-823`), which was correctly updated to `fighterData.fighter?.fighter_type?.gang_type_id` / `custom_gang_type_id`. So: a Cawdor Ganger added to an Escher gang now correctly sees Cawdor promotion options via the "Promote" UI, but still sees Escher's fighter-type list via Edit Fighter → Change Fighter Type — the exact reported bug, just through a different entry point.
Fix: mirror the same source used at line 822-823 for this prop pair too.

**2. [MEDIUM] Custom fighter types with no gang type resolve to an empty (not fallback) promotion list**
`app/fighter/[id]/page.tsx:170-181`, consumed in `fighter-advancement-list.tsx:685-700`/`2694-2710` and `fighter-edit-modal.tsx:113-129`
`custom_fighter_types.gang_type_id` / `.custom_gang_type_id` are both nullable and not mutually-exclusive by constraint (unlike `gangs`, which has `chk_gang_type_exclusive`). If the custom gang type a fighter references is later deleted, `custom_gang_type_id` gets set to `NULL` (`ON DELETE SET NULL`), leaving both fields null. Previously this fell back to the owning gang's own type (showing *something*); now `gangTypeId`/`customGangTypeId` both resolve to `''`, which:
   - disables the fetch entirely in `fighter-advancement-list.tsx` (`enabled: !!(gangTypeId || customGangTypeId)`), and
   - makes `get_add_fighter_details(p_gang_type_id: null, ...)` return 0 rows in `fighter-edit-modal.tsx`'s dropdown (`WHERE ft.gang_type_id = p_gang_type_id`, and `NULL = NULL` is never true in Postgres).

   Net effect: a silent, empty "no options" combobox instead of a wrong-but-populated one. Worth an explicit fallback to the owning gang's type when the fighter's own type has no gang association, rather than silently going empty.

**3. [MEDIUM] Redundant, uncached duplicate query — the fix to `getFighterTypeInfo` is dead code**
`app/fighter/[id]/page.tsx:101-110` vs `app/lib/shared/fighter-data.ts:1178-1189`
This PR adds `gang_type_id` to the cached `getFighterTypeInfo` select (already called at `page.tsx:85`), but then never reads `fighterTypeData?.gang_type_id`. Instead it fires a second, uncached, direct Supabase query for the *same row's same column* on every fighter page load:
```js
// Fighter type's source gang_type_id — direct query (not cached) so it is always fresh.
fighterBasic.fighter_type_id
  ? supabase.from('fighter_types').select('gang_type_id')...
```
`fighter_types.gang_type_id` is `NOT NULL` and effectively immutable reference data (a fighter type permanently belongs to one house), so there's no staleness risk the cache would introduce — the "always fresh" justification doesn't hold, and this just doubles a DB round-trip for no benefit. Given the amount of work elsewhere in this repo trimming per-page query counts, this stands out. Suggest dropping the extra raw query and using `fighterTypeData?.gang_type_id` instead (falling back to `customFighterTypeInfo?.gang_type_id` for the custom-type case as already done).

**4. [LOW] Optimistic-update parity gap for the custom-fighter-type promotion path**
`components/fighter/edit-fighter/fighter-edit-modal.tsx:148-166` and `~883-892`
The system-fighter-type branch sets `submitData.gang_type_id = fighterTypeToUse.gang_type_id` (line 892), but the custom-fighter-type branch (~883-888) never sets an equivalent `custom_gang_type_id`/`gang_type_id` — and `custom_gang_type_id` isn't even captured in the `fetchedFighterTypes` → `FighterTypeEntry` mapping (lines 148-166), so the data isn't available client-side to set it. Result: right after promoting into a *custom* fighter type, the optimistic `fighter_type` object (`onMutate`, line ~404) has `gang_type_id`/`custom_gang_type_id` stuck at `null` until the next full page refetch, which can briefly show wrong/empty promotion options if the user reopens the combobox immediately. Low impact since it self-heals on refetch/revalidation, but worth aligning both branches.

**5. [LOW / nit] `submitData.gang_type_id` never reaches the server**
`fighter-edit-modal.tsx:892` sets it, but `mutationFn` (`fighter-edit-modal.tsx:329-364`) doesn't forward it to `updateFighterDetails`, and there's no such column on `fighters` to persist anyway — it's used only for the same-render optimistic overlay (line ~404). Not a bug (nothing needs persisting, it's derived data), but a one-line comment noting it's client-only/optimistic would save the next reader a double-take.

### Severity summary
- High: 1 (finding #1 — fix is incomplete for one of two promotion entry points)
- Medium: 2 (findings #2, #3)
- Low: 2 (findings #4, #5)

The core diagnosis is right and the `AdvancementsList` path is fixed correctly, but finding #1 means the PR doesn't fully close out the reported bug (it's still reproducible via Edit Fighter → Change Fighter Type), so I'd want that addressed before merge. #2 and #3 are worth a look too but are more edge-case/perf than functional-regression.


---

joeseos:
## Code Review (update)

@Djidiouf Reviewed the latest commit (`b7484ac`, the changes since the previous review). All five findings from the last pass are addressed.

### Previous findings

| # | Finding | Status |
|---|---|---|
| 1 [HIGH] | Edit Fighter → Change Fighter Type still used the owning gang's type | **Resolved** — `fighter-page.tsx:1282-1283` now passes `fighterData.fighter?.fighter_type?.gang_type_id` / `custom_gang_type_id`, matching the `AdvancementsList` fix. |
| 2 [MEDIUM] | Empty promotion list when a fighter type has no gang association (e.g. deleted custom gang type) | **Resolved** — `app/fighter/[id]/page.tsx` now falls back to `gangBasic.gang_type_id` / `gangBasic.custom_gang_type_id` when the fighter type's own fields are null, instead of resolving to `''`. |
| 3 [MEDIUM] | Redundant uncached duplicate query for `gang_type_id` | **Resolved** — the extra raw `fighter_types` query is removed; the code now reuses `fighterTypeData?.gang_type_id` from the already-cached `getFighterTypeInfo` call (which had `gang_type_id` added to its select). |
| 4 [LOW] | Optimistic-update parity gap for custom fighter types (`custom_gang_type_id` not captured/set) | **Resolved** — `custom_gang_type_id` is now part of the `FighterTypeEntry` mapping and is set in both the custom- and system-fighter-type branches of the submit handler. |
| 5 [LOW/nit] | No comment noting `submitData.gang_type_id` is client-only/optimistic | **Resolved** — both assignment sites now have a one-line comment clarifying this. |

### New findings

None — I checked the fallback ordering (fighter's own type → owning gang's type), confirmed `gangBasic` already selects both `gang_type_id` and `custom_gang_type_id` so the new fallback is safe, and confirmed `transformCustomFighter`'s new `custom_gang_type_id` field is backed by the existing `select('*', ...)` queries feeding it (no missing-column risk). `tsc --noEmit` is clean (only pre-existing tsconfig deprecation warnings, unrelated to this diff).

### Verdict: **Approve**


---

joeseos:
### Suggestion: fetch the custom fighter type's gang info via an embedded join in `getFighterBasic` instead of a separate query

This PR adds one new DB query in `app/fighter/[id]/page.tsx` — a conditional `custom_fighter_types` lookup inside the second `Promise.all` batch:

```ts
fighterBasic.custom_fighter_type_id
  ? supabase
      .from('custom_fighter_types')
      .select('gang_type_id, custom_gang_type_id')
      .eq('id', fighterBasic.custom_fighter_type_id)
      .single()
      .then((result) => result.error ? null : result.data)
  : Promise.resolve(null)
```

It works, but it could instead be folded into the existing `getFighterBasic` query via an embedded FK join:

```ts
custom_fighter_type:custom_fighter_type_id (
  gang_type_id,
  custom_gang_type_id
)
```

**Why this is better:**

- **No extra round-trip.** The current query sits in the *second* `Promise.all`, which can't start until the first batch resolves (`beastIds` / `gang_variants` depend on it). But this lookup only needs `custom_fighter_type_id`, which is known before either batch — so today it pays a full serial wave it doesn't need. Embedding folds it into the base query that already runs first.
- **It gets cached.** `getFighterBasic` is wrapped in `unstable_cache` (tag `BASE_FIGHTER_BASIC`, invalidated on fighter edits). The current inline query is uncached and re-runs on every load for custom-typed fighters.
- **It's the idiomatic spot** — `getFighterBasic` already embeds `fighter_gang_legacy:fighter_gang_legacy_id (...)` and `selected_archetype:selected_archetype_id (...)` via the same pattern. The FK exists: `fighters_custom_fighter_type_id_fkey → custom_fighter_types(id)`.

**Trade-off to be aware of:** embedding means the join is part of the base SELECT for *every* fighter (not just custom ones). For a regular fighter `custom_fighter_type_id` is `NULL`, so it's a no-match left-join returning `null` — effectively free, and crucially *not* a separate round-trip. So we swap "a real extra query on custom fighters" for "a trivial null lookup on the common path."

One caveat: `BASE_FIGHTER_BASIC` is keyed on the fighter, so if a custom type's gang association ever changed, the embedded copy would be stale until that fighter's cache tag is busted. In practice this is near-zero risk (a custom type's source gang doesn't change post-creation; the FK is `ON DELETE SET NULL` and the code already falls back to the owning gang's type), and the already-embedded `fighter_gang_legacy` join carries the identical trade-off.

Not blocking — the current approach is correct. Just a cleaner, cheaper home for the data if we want to minimize calls.

---

PR #1875, opened 2026-07-18
Fixed race condition in save flow for custom trading post

Djidiouf:


joeseos:
## Code Review

@Djidiouf Thanks for the fix — the core diagnosis is right: `handleEditConfirm` used to call `updateMutation.mutate(...)` (fire-and-forget) and then immediately `return true`, so `Modal.handleConfirm` closed the modal (and cleared `pendingAdditions`/`pendingOverrides`/`pendingRemovals` via `resetForm()`) before the equipment batch/rules/removal calls in the `onSuccess` callback had actually finished. Switching to `mutateAsync` and awaiting the whole equipment-save sequence before returning fixes that. Also good catch adding `costTypeResourceId`/`costCampaignResourceId`/`costReputation`/`costResourceAmount`/`banned` to `addTPEquipmentBatch` — previously those overrides (including `banned`) were silently dropped for newly-added items, since `equipmentSaves` explicitly filters out `pendingAdditionTempIds`.

One correctness gap I'd like addressed before merging:

### Findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment**
`components/customise/custom-trading-posts.tsx:249-390`

`equipmentSaveFailed` is now used to keep the modal open (return `false`) when *any* step of the equipment save fails — but nothing prunes `pendingAdditions`/`pendingOverrides`/`pendingRemovals` on that failure path. They're only reset in `resetForm()`, which is only called from `Modal`'s `onClose` (i.e. on success or explicit cancel).

Concretely: if `addTPEquipmentBatch` (line 253) succeeds but a later step in the same save fails — e.g. `saveEquipmentRules` for one of the added items (line 305-308), a removal (line 318-320), or an override save (line 338-340) — `equipmentSaveFailed` becomes `true` and the function returns `false`. The modal stays open with `pendingAdditions` still containing the already-inserted items. Worse, this is visible immediately, not just on retry: the `invalidateQueries(['tpEquipment', tradingPostId])` call at line 371 refetches the now-real rows from the server, while `pendingAdditions` (rendered as an overlay at line 676) still renders the same items as pending — so the modal shows duplicate rows right away. If the user then clicks Save again, `addTPEquipmentBatch` is called a second time with the same `additionsToSave`, inserting duplicate rows into `custom_trading_post_equipment`.

Suggested fix: on partial failure, reconcile local state to drop only the operations that actually succeeded (e.g. clear `pendingAdditions`/`pendingOverrides` entries for items present in `batchResult.data`, clear resolved `removalsToSave` ids, etc.) rather than leaving the full pre-save snapshot in place for retry.

**2. [Low] Silent failure on unexpected exceptions**
`components/customise/custom-trading-posts.tsx:388-390`

The outer `catch { return false; }` swallows any exception without a `toast.error`, so if something throws outside the internal `{success, error}` contract (e.g. a network abort), the modal just stays open with no feedback. Low severity since all the action functions already catch internally and return `{success:false}`, but worth a `toast.error()` in the catch for visibility.

### Other notes
- `addTPEquipmentBatch`'s new fields are inserted correctly and match the DB defaults/constraints (`cost_reputation boolean NOT NULL DEFAULT false`, `chk_cost_resource_exclusive`), so no issue there.
- No security, DRY, or convention issues found — the field wiring in `app/actions/customise/custom-trading-posts.ts` mirrors `updateTPEquipment`'s existing shape.

### Verdict: **Changes requested** (Finding 1 is a real data-duplication risk introduced by making the modal retryable; would like it addressed or explicitly deferred before merge.)


---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (c7f2ddb) against the prior review of 6eb1aad.

### Previous findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment — Resolved**
`components/customise/custom-trading-posts.tsx:249-427`

The fix tracks exactly which operations actually persisted (`insertedAdditionTempIds`, `succeededRemovalIds`, `succeededOverrideIds`) and, on `equipmentSaveFailed`, prunes only those from `pendingAdditions`/`pendingRemovals`/`pendingOverrides` before returning `false`. Items whose insert succeeded but whose rule save failed (`failedAdditionRulesByRealId`) are correctly re-keyed from the temp id to the real DB id in `pendingOverrides`, so a retry falls into the `updateTPEquipment` path (line 340's `!pendingAdditionTempIds.has(itemId)` filter) instead of re-inserting via `addTPEquipmentBatch`. Traced through the batch-add / rules-save / removal / override branches and the bookkeeping is consistent — no duplicate-insert path left on retry.

**2. [Low] Silent failure on unexpected exceptions — Resolved**
`components/customise/custom-trading-posts.tsx:431-434`

`catch (error)` now surfaces `error.message` via `toast.error` instead of silently swallowing it.

### New findings

**1. [Low] Cleanup bookkeeping doesn't run on thrown exceptions**
`components/customise/custom-trading-posts.tsx:404-426` vs `431-434`

The new state-reconciliation logic (dropping successfully-persisted items from `pendingAdditions`/`pendingOverrides`/`pendingRemovals`) only runs on the `equipmentSaveFailed` flag path. If an exception is thrown *after* `addTPEquipmentBatch` succeeds but before `equipmentSaveFailed` is checked (e.g. an unexpected `TypeError`, not the `{success:false}` contract the internal actions normally use), execution jumps straight to the outer `catch` and returns `false` without pruning — reintroducing the same duplicate-row risk finding 1 fixed, just via a different trigger. Low severity since it requires a genuine bug/unexpected throw rather than a normal action failure (matches the "Low" rationale from the original silent-catch finding). Not blocking — worth a `// TODO` or tracking issue if you want to close the gap later, e.g. by moving the reconciliation into a `finally`-style helper shared by both paths.

Nothing else new: the diff is a narrowly-scoped, well-traced fix; no regressions in the batch-add matching logic, no new security/DRY issues.

### Verdict: **Approve**

Both blocking findings from the previous review are resolved. The remaining low-severity gap (exception path bypassing cleanup) is pre-existing in spirit (same class as the original low-severity catch issue) and optional to address before merge.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (44c8c4c) against the prior review of c7f2ddb.

### Previous findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment — Resolved** (unchanged since last review)

**2. [Low] Silent failure on unexpected exceptions — Resolved** (unchanged since last review)

**3. [Low] Cleanup bookkeeping doesn't run on thrown exceptions — Resolved**
`components/customise/custom-trading-posts.tsx:238-279, 456-468`

The tracking state (`insertedAdditionTempIds`, `insertedTempToRealId`, `succeededAdditionRuleRealIds`, `failedAdditionRulesByRealId`, `succeededRemovalIds`, `succeededOverrideIds`) is now declared outside the `try` block, and the reconciliation logic that used to live only in the `equipmentSaveFailed` branch is extracted into a shared `reconcilePersistedPendingState()` closure called from both that branch (line 449) and the outer `catch` (line 457). The catch path also now fires `invalidateQueries(['tpEquipment', tradingPostId])` when any operation actually persisted before the throw, so the equipment list reflects reality instead of showing stale/duplicate pending rows. Traced through: this closes the gap identified last round.

### New findings

None material. One pre-existing-in-spirit, very-low-severity edge case worth noting but not blocking: if `saveEquipmentRules` itself throws (rather than returning `{success:false}`) while other sibling rule-save promises in the same `Promise.all(pendingRulesSaves)` are still in flight, `Promise.all` rejects immediately and `reconcilePersistedPendingState()` runs in the `catch` before those siblings finish updating `succeededAdditionRuleRealIds`/`failedAdditionRulesByRealId`. Reconciliation would then run on partial bookkeeping for those still-in-flight items, and the per-item `tpAvailabilityRules`/`tpPricingRules` cache invalidation (present on the success path, lines 431–446) doesn't run on the catch path at all — so a rules update that quietly succeeds after the throw could leave the rules dropdown UI stale until a manual refresh. This requires a genuine unexpected exception (not the normal `{success:false}` contract), same class as the original low-severity catch finding — fine to defer.

Nothing else new: the diff is scoped exactly to the cleanup-on-throw gap, no regressions in the batch-add/rules/removal/override bookkeeping, no security/DRY/convention issues.

### Verdict: **Approve**

All findings from previous reviews are resolved. The residual edge case above is optional and can be tracked separately if desired.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (a9a9cae) against the prior review of 44c8c4c.

### Previous findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment — Resolved** (unchanged since last review)

**2. [Low] Silent failure on unexpected exceptions — Resolved** (unchanged since last review)

**3. [Low] Cleanup bookkeeping doesn't run on thrown exceptions — Resolved** (unchanged since last review)

**4. [Low, residual/non-blocking] `Promise.all` on `pendingRulesSaves` aborts on first throw, leaving in-flight siblings' bookkeeping/cache-invalidation incomplete — Attempted fix, but see new finding below.**
`components/customise/custom-trading-posts.tsx:367,372,430`
This commit switches `pendingRulesSaves`, the removals batch, and `equipmentSaves` from `Promise.all` to `Promise.allSettled`, with a comment stating this "ensures every sibling completes before we read the bookkeeping sets." That's true for siblings that *don't* throw, but it introduces a new problem — see below.

### New findings

**1. [Medium] `Promise.allSettled` results are never inspected, so a thrown exception is now silently swallowed instead of surfacing an error**
`components/customise/custom-trading-posts.tsx:333-367` (rules saves), `372-382` (removals), `385-430` (overrides)

None of the three `Promise.allSettled(...)` calls examine the returned `PromiseSettledResult[]`. Each per-item async callback only updates `equipmentSaveFailed` / calls `toast.error` / populates its bookkeeping set (`succeededAdditionRuleRealIds`, `succeededRemovalIds`, `succeededOverrideIds`) on the normal `{success:false}` path *before* it can reject. If `saveEquipmentRules`, `removeTPEquipment`, or `updateTPEquipment` throws instead of resolving (e.g. a network abort producing an unhandled `TypeError` rather than the internal `{success:false}` contract), that promise's rejection is captured by `allSettled` and then discarded — nothing checks `status === 'rejected'`.

Concretely: if one item's rules save throws while every other operation in the batch succeeds normally, `equipmentSaveFailed` stays `false`, execution falls through to `toast.success('Custom trading post updated successfully'); return true;` (line 456-457), the modal closes, and `resetForm()` clears `pendingAdditions`/`pendingOverrides`/`pendingRemovals` — silently discarding that item's pending change with a **success** toast, even though it never persisted. This is a regression versus the prior `Promise.all` behavior: previously such a throw propagated to the outer `catch`, which at least fired `toast.error` and returned `false` to keep the modal open for retry. Now it's invisible.

Suggested fix: inspect the settled results and treat rejections the same as a `{success:false}` outcome, e.g.:
```ts
const settled = await Promise.allSettled(pendingRulesSaves);
if (settled.some(r => r.status === 'rejected')) {
  equipmentSaveFailed = true;
  toast.error('Failed to save equipment rules');
}
```
(same pattern for the removals and `equipmentSaves` batches).

Nothing else new: the per-item cache-invalidation additions in the `catch` block (now including `succeededAdditionRuleRealIds`) correctly close the gap noted in the previous review's residual finding — that part of this commit is good. It's specifically the unchecked `allSettled` results that reopen a (different, and slightly worse) silent-failure path.

The one Medium finding above should be addressed — checking `allSettled` results for rejections — before merge, since it turns a rare edge case into a silent false-success report instead of a visible error.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (fc08d92) against the prior review of a9a9cae. Scope: only the delta since the last review.

### Previous findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment — Resolved** (unchanged since last review)

**2. [Low] Silent failure on unexpected exceptions — Resolved** (unchanged since last review)

**3. [Low] Cleanup bookkeeping doesn't run on thrown exceptions — Resolved** (unchanged since last review)

**4. [Medium] `Promise.allSettled` results were never inspected, so a thrown exception was silently swallowed instead of surfacing an error — Resolved**
`components/customise/custom-trading-posts.tsx:333-460`

All three `Promise.allSettled` call sites (rules saves, removals, overrides) now capture the settled array and iterate it, treating any `status === 'rejected'` entry the same as an internal `{success:false}` outcome: `equipmentSaveFailed` is set and a `toast.error` fires with `reason.message` (falling back to a category-specific message when the reason isn't an `Error`), matching the existing outer-catch pattern.

Traced through the regression scenario from the last review — a single item's rules save throwing while its siblings succeed normally: the throw is now caught by `allSettled`, `equipmentSaveFailed` becomes `true`, execution falls into the `if (equipmentSaveFailed)` branch instead of the success path, `reconcilePersistedPendingState()` runs, and the modal stays open returning `false`. The false-success/silent-data-loss path is closed.

Also checked reconciliation correctness for the throw case specifically: since the per-item callback returns via early-`return` (not throw) on the normal failure contract, and only rejects on a genuine unexpected exception, `failedAdditionRulesByRealId` won't be populated for a rejected rules-save — but `reconcilePersistedPendingState()` already falls back to `overridesToSave.get(tempId)` when `failedAdditionRulesByRealId` has no entry, so the item is still correctly re-queued under the real DB id for retry rather than dropped. No gap there.

### New findings

None material. No regressions in the batch-add/rules/removal/override bookkeeping, no double-toast paths (the per-item success-path failures and the new rejection handling are mutually exclusive since a callback either returns early after its own `toast.error` or throws before reaching that code, never both), and no security/DRY/convention issues in the diff.

One non-blocking observation: if several siblings in the same `allSettled` batch reject together (e.g. a full network outage), the user will see one `toast.error` per rejected item rather than a single consolidated message. This mirrors the existing behavior for normal per-item `{success:false}` failures already accepted in earlier rounds, so not a new pattern — just noting it's now extended to the exception path too. Not worth blocking on.

### Verdict: **Approve**

All findings from previous reviews, including the Medium finding from the last round, are resolved. Nothing new to address.

---

joeseos:
## Review notes

Solid fix, and a more careful implementation overall. Traced the modal/mutation wiring to confirm it holds together.

### What's right
- **Core race fix is sound.** Previously the mutation's own `onSuccess` fired `setEditModalData(null)` + `resetForm()` + `toast.success` the instant the trading-post row updated — before the async equipment saves (additions/removals/overrides) ran, so the modal closed and claimed success while writes were still in flight (and `resetForm` could wipe pending state mid-save). Moving all of that into `handleEditConfirm`, which only returns `true` after everything settles, is the right fix. The shared `Modal` closes on `result !== false` (`components/ui/modal.tsx`), so the modal now closes only on full success.
- **Real secondary bug fixed:** `addTPEquipmentBatch` now carries `banned`, `cost_type_resource_id`, `cost_campaign_resource_id`, `cost_reputation`, `cost_resource_amount`. Before, a newly-added item with those overrides silently lost them (additions are filtered out of the `updateTPEquipment` pass, so the batch insert was their only path and it didn't include them).
- `Promise.allSettled` + per-op bookkeeping so one failure doesn't cut off siblings, and the reconcile drops persisted ops so a retry can't duplicate them — correct pattern for partial-failure retry safety.

### Things to fix

**1. Double toast on a thrown TP update (minor bug).** The mutation's `onError` still fires `toast.error(error.message)` — only the `onSuccess` toasts were removed. With `mutateAsync`, a *thrown* error from `updateCustomTradingPost` triggers **both** react-query's `onError` toast **and** the `catch` block's `toast.error` → two identical toasts. (The non-throw `{success:false}` path is fine.) Pick one owner: either drop the toast from `onError`, or don't re-toast the mutation error in `catch`.

**2. Bloat: duplicated cache-invalidation logic.** The rule-cache invalidation is written out three times in the success path (`tpEquipment`, then `overridesToSave` keys, then `failedAdditionRulesByRealId` keys) plus a fourth near-identical version in the `catch` that unions everything into `allRelevantItemIds`. Extract one helper and call it from both paths:
```ts
const invalidateItemRuleCaches = (itemIds: Iterable<string>) =>
  Promise.all([...new Set(itemIds)].flatMap(id => [
    queryClient.invalidateQueries({ queryKey: ['tpAvailabilityRules', id] }),
    queryClient.invalidateQueries({ queryKey: ['tpPricingRules', id] }),
  ]));
```
The success path can then reuse the same union-of-ids set the catch block already builds, collapsing three blocks into one.

**3. Bloat: two near-identical "did anything persist?" checks.** `reconcilePersistedPendingState`'s `hasProgress` and the catch block's `anyPersisted` compute the same intent over overlapping sets — fold them into one predicate.

**4. Minor inconsistency.** The success path `await`s the invalidations; the `catch` uses fire-and-forget `void queryClient.invalidateQueries(...)`. Worth making consistent.

Net: functionally correct as far as I can tell, and the approach is right. The only real bug is the double-toast on throw (#1); the rest is complexity that's partly inherent to safe partial-retry but has ~3 clearly factorable duplications that would meaningfully shrink `handleEditConfirm`.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (018ba43) against the prior "Review notes" review (fc08d92). Scope: only the delta since that review — all changes are in `components/customise/custom-trading-posts.tsx`.

### Previous findings

**1. [Medium] Partial-failure retry can duplicate newly-added equipment — Resolved** (unchanged since last review)

**2. [Low] Silent failure on unexpected exceptions — Resolved** (unchanged since last review)

**3. [Low] Cleanup bookkeeping doesn't run on thrown exceptions — Resolved** (unchanged since last review)

**4. [Medium] `Promise.allSettled` results were never inspected — Resolved** (unchanged since last review)

**5. [Minor bug] Double toast on a thrown TP update — Resolved**
`components/customise/custom-trading-posts.tsx:112-115`
`onError` no longer calls `toast.error`; only the outer `catch` in `handleEditConfirm` does, with a comment documenting the ownership split. Confirmed `updateMutation` is only ever driven via `mutateAsync` in `handleEditConfirm` (no other `.mutate()`/`.mutateAsync()` call site), so no other caller depended on the removed toast.

**6. [Bloat] Duplicated cache-invalidation logic — Resolved**
`components/customise/custom-trading-posts.tsx:253-272`
Extracted into `relevantRuleCacheItemIds()` / `invalidateItemRuleCaches()` / `invalidateEquipmentCaches()`, now called from both the success path (line 485) and the `catch` block (line 499) instead of three duplicated inline blocks. `relevantRuleCacheItemIds()` also folds in `insertedTempToRealId.values()`, which slightly widens invalidation coverage for newly-inserted items — harmless (extra invalidations are no-ops when there's nothing to invalidate).

**7. [Bloat] Two near-identical "did anything persist?" checks — Resolved**
`components/customise/custom-trading-posts.tsx:245-251`
Unified into `hasPersistedProgress()`, used by both `reconcilePersistedPendingState()` and the `catch` block. Traced the set membership: the two fields the new helper adds versus the old inline `anyPersisted` (`insertedTempToRealId`, `failedAdditionRulesByRealId`) are always populated in lockstep with `insertedAdditionTempIds`, so this is a pure dedup with no behavior change.

**8. [Minor inconsistency] success path awaited invalidation, catch path fire-and-forget — Resolved**
`components/customise/custom-trading-posts.tsx:499`
Catch block now `await`s `invalidateEquipmentCaches()` instead of firing `void queryClient.invalidateQueries(...)`.

### New findings

None blocking. `npx tsc --noEmit` shows no errors introduced by this change.

One very low, non-blocking observation: the `catch` block now awaits `invalidateEquipmentCaches()` *before* `toast.error(...)` (line 499-501). If that await were to throw (react-query's `invalidateQueries` doesn't in practice, so this is theoretical), the exception would propagate out of `handleEditConfirm` uncaught, skipping the toast entirely — `Modal.handleConfirm`'s own try/catch would swallow it with just a `console.error`, leaving the modal open with no visible error. This is the same class of "genuine unexpected throw" edge case already accepted as low-severity/deferred elsewhere in this PR (finding 3's residual note), so not worth blocking on.

### Verdict: **Approve**

All four points raised in the last review round (double toast, duplicated invalidation logic, duplicated progress checks, await inconsistency) are cleanly resolved with no regressions to the partial-failure/retry bookkeeping. Nice cleanup — `handleEditConfirm` is noticeably easier to follow now.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the latest commit (a004d3c) against the prior review of 018ba43. Scope: only the delta since that review — both changes are in `components/customise/custom-trading-posts.tsx`.

### Previous findings

**1-8 (partial-failure duplication, silent exceptions, cleanup-on-throw, unchecked `allSettled`, double toast, duplicated invalidation/progress-check logic, await inconsistency) — Resolved** (unchanged since last review)

**9. [Very low, non-blocking] Catch-path `await invalidateEquipmentCaches()` before `toast.error` could theoretically throw and skip the error toast — Resolved**
`components/customise/custom-trading-posts.tsx:495-505`
The catch block now fires `toast.error(...)` immediately after `reconcilePersistedPendingState()`, *then* attempts `invalidateEquipmentCaches()` wrapped in its own try/catch with a comment explaining the ordering ("Refresh caches after toasting so a rare invalidate failure can't swallow the error UI"). The user is now guaranteed to see the error toast regardless of whether the cache refresh succeeds. Clean, minimal fix for exactly the edge case flagged last round.

### New findings

None material — but one item is worth calling out as a genuine catch, not just cleanup:

**`[...new Set(itemIds)]` → `Array.from(new Set(itemIds))`** (`custom-trading-posts.tsx:262`) looks cosmetic but isn't. This project's `tsconfig.json` targets `es5` without `downlevelIteration`. I verified in isolation that spreading a `Set` under that exact config (`target: es5`, `lib: [dom, dom.iterable, esnext]`, no `downlevelIteration`) fails with `TS2802: Type 'Set<string>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.` That spread syntax was introduced in commit 018ba43 (copied verbatim from a suggestion in an earlier review comment) and the follow-up review at the time reported no new `tsc` errors — which appears to have been a false negative (likely the reviewer's `tsc` invocation didn't reach full type-checking, similar to what I hit here before pinning `ignoreDeprecations`). This commit's fix brings it in line with the ~30 other `Array.from(new Set(...))` call sites elsewhere in the codebase (e.g. `utils/duplicate-custom-gang-type.ts`, `app/lib/shared/gang-data.ts`) and resolves what was actually a build-breaking regression, not just a style nit. Good catch.

Nothing else new: no regressions in the bookkeeping/reconciliation logic, no security/DRY/convention issues in the diff.

### Verdict: **Approve**

All outstanding findings from previous rounds are resolved, and this commit additionally fixes a real (if narrow) TypeScript compile error introduced in the prior commit. Nothing further to address before merge.

---

PR #1876, opened 2026-07-18
Added Admin Notification capabilities to be able to send messages eas…

Djidiouf:


joeseos:
## Code Review

@Djidiouf Reviewed against the codebase (existing admin routes, the notifications schema/trigger, and the other user-search implementations). Solid feature overall — the audience/type/link flow is clean and the notification-link UI refactor is a nice improvement. A few issues worth addressing before merge.

### Findings

**1. [High] Unsanitized `link` scheme allows a `javascript:` URI to become a live clickable link**
`utils/notifications.ts` — `resolveNotificationLink()`

```ts
const url = new URL(trimmed);
if (APP_HOSTNAMES.has(url.hostname)) { ... }
return { href: trimmed, isExternal: true };
```

`new URL('javascript:alert(1)')` parses successfully (`hostname` is `''`), so it falls into the "external" branch and `trimmed` is returned verbatim as `href`. In `components/account/notifications-content.tsx`, that `href` is rendered directly:
```tsx
<a href={href} target="_blank" rel="noopener noreferrer" onClick={...}>
```
Clicking it executes the `javascript:` payload in the page's origin. Right now `link` is only settable via the new admin broadcast endpoint, but that's exactly the trust boundary that matters here — a compromised/careless admin session (or a pasted value) becomes stored XSS for every recipient who clicks "Open link". Suggest whitelisting `http(s):` + relative paths only, and treating anything else as non-navigable (render as plain text or reject at the API layer in `app/api/admin/notifications/route.ts` when validating `link`).

**2. [Medium] `audience: 'all'` pagination has no `ORDER BY`, risking skipped or duplicated users**
`app/api/admin/notifications/route.ts` (~line 55-72)

```ts
const { data: profiles } = await serviceClient
  .from('profiles')
  .select('id')
  .range(from, from + pageSize - 1)
```
Offset/range pagination without an explicit `.order('id')` doesn't guarantee stable row ordering across separate round trips, especially with concurrent inserts (new signups) between pages. Rows can be skipped or double-counted. For a mass-broadcast tool this should be deterministic — add `.order('id', { ascending: true })`.

**3. [Medium] Partial batch failure has no idempotency — retry duplicates already-sent notifications**
`app/api/admin/notifications/route.ts` (~line 130-152)

The insert loop (`BATCH_SIZE = 500`) returns `{ partial: true, count }` on failure mid-loop, but nothing tracks *which* users already got a row. If the admin retries after a partial failure (e.g. transient DB error on batch 3 of 5), the users from batches 1-2 receive the notification twice, with no dedup. Worth surfacing the recipient list/offset so a retry can resume, or generating a client-side idempotency key.

**4. [Low] Copy inconsistency: `'Unauthorised'` vs. the rest of the codebase**
`app/api/admin/notifications/route.ts:16`

Every other admin route (`app/api/admin/scenarios/route.ts`, etc.) returns `{ error: 'Unauthorized' }` (US spelling) on the `checkAdmin` failure path. This route uses `'Unauthorised'`. Purely cosmetic but worth matching for consistency/grep-ability.

**5. [Low] `hasNotificationLink` isn't a type predicate, forcing non-null assertions**
`utils/notifications.ts` / `components/account/notifications-content.tsx`

`hasNotificationLink(link: string | null | undefined): boolean` doesn't narrow the type, so every call site does `notification.link!` after the check (4 occurrences). Declaring it as `link is string` would remove the assertions and be a bit safer.

### Not an issue (checked, ruling out a false alarm)
I initially suspected that broadcasting to `audience: 'all'` with `type: 'info'/'warning'/'error'` would trigger the `enqueue_notification_email` AFTER INSERT trigger and mass-email the entire user base. Checked `supabase/functions/enqueue_notification_email.sql` directly — the trigger only enqueues email for `campaign_invite` / `gang_invite` / `friend_request`. Admin broadcasts of type info/warning/error are in-app only, so this is safe as-is.

Primarily for #1 (link scheme sanitization) and #2 (deterministic pagination) — both are cheap fixes. #3-5 are good to address but not blocking.

---

joeseos:
## Follow-up Review

@Djidiouf Reviewed the two commits pushed since the last review (`1e0a7fd`, `3d3f9d4` — diff against `1a2ee32`, the commit the previous review was based on). All four files flagged before were touched.

### Previous findings

| # | Finding | Status |
|---|---|---|
| 1 | High — unsanitized `link` scheme (`javascript:` XSS) | **Resolved.** `isSafeNotificationLink()` now allowlists relative paths and `http(s):` only, rejects `javascript:`/`data:`/`vbscript:`/`file:` and protocol-relative `//`. Applied server-side (400 on submit) and client-side (`resolveNotificationLink` returns `null` instead of a usable href). Checked the `new URL()` fallback path too — WHATWG URL parsing strips embedded tab/newline characters before exposing `.protocol`, so scheme-obfuscation tricks (e.g. `java\tscript:`) still get caught by the `protocol === 'http:' \|\| 'https:'` check even where the regex alone wouldn't catch them. |
| 2 | Medium — `audience: 'all'` pagination has no `ORDER BY` | **Resolved.** `.order('id', { ascending: true })` added to the paginated `profiles` query. |
| 3 | Medium — partial batch failure has no idempotency | **Partially resolved** — see new finding below. |
| 4 | Low — `'Unauthorised'` spelling inconsistency | **Resolved.** |
| 5 | Low — `hasNotificationLink` isn't a type predicate | **Resolved.** Both `hasNotificationLink` and `isSafeNotificationLink` are now `link is string` guards; confirmed no `notification.link!` assertions remain in `notifications-content.tsx`. |

### New findings

**1. [Medium] Stale `resumeFrom` survives audience/recipient changes, corrupting the next unrelated send**
`components/admin/admin-notifications-modal.tsx` (resumeFrom state at line 29, used at line 65, set at line 73)

On a partial-failure retry, `resumeFrom` is stored in component state and reset only on a fully successful send (line 82). But switching `audience` (lines 111/123), adding/removing selected users (`handleSelectUser`/`handleRemoveUser`), or editing the message/link doesn't clear it. Concretely: an admin broadcasts to "All users", it partially fails (`resumeFrom` set to e.g. 1500), then switches to "Specific users" and picks 2 recipients — the next request still sends `resumeFrom: 1500`. Since `receiverIds.length` (2) < `resumeFrom` (1500), the API now returns `"Resume offset is beyond the recipient list"` (`app/api/admin/notifications/route.ts:143-148`), blocking an otherwise-valid send with a confusing error. If the new recipient count ever exceeded the stale offset, it would instead silently skip real recipients (`rows = receiverIds.slice(resumeFrom)...`) rather than erroring.

Suggest resetting `resumeFrom` to `0` whenever `audience`, `selectedUsers`, `text`, `type`, or `link` changes — e.g. in the respective `onClick`/`onChange` handlers, or via a `useEffect` keyed on those values.

**2. [Low, informational] `resumeFrom` isn't a true idempotency key for `audience: 'all'`**
`app/api/admin/notifications/route.ts` (lines 79-103)

The resume mechanism is index-based against a freshly re-fetched, `ORDER BY id`-sorted `profiles` list. That's fine and deterministic for the `users` audience (fixed ID set via `.in(uniqueIds)`), but for `all` the underlying recipient set can change between the failed attempt and the retry (new signups). Because UUIDs sort essentially randomly, a new user landing before the previous cutoff shifts every subsequent index by one, which can skip or duplicate a recipient across the two attempts. Given this is a low-frequency admin action retried manually, I'd call this acceptable for now, but worth a code comment noting the limitation, or resuming by last-inserted-id instead of a positional offset if this becomes a recurring broadcast pattern.

New finding #1 is the only thing I'd block on — it's a straightforward reset-on-change fix. Everything from the previous round is resolved or meaningfully improved.


---

joeseos:
## Follow-up Review (round 3)

@Djidiouf Reviewed the one commit pushed since the last review (`1555128`, diff against `3d3f9d4`). Two files touched: `app/api/admin/notifications/route.ts`, `components/admin/admin-notifications-modal.tsx`.

### Previous findings

| # | Finding | Status |
|---|---|---|
| Round 1 #1 | High — unsanitized `link` scheme (`javascript:` XSS) | Resolved (round 2) |
| Round 1 #2 | Medium — `audience: 'all'` pagination has no `ORDER BY` | Resolved (round 2) |
| Round 1 #4 | Low — `'Unauthorised'` spelling | Resolved (round 2) |
| Round 1 #5 | Low — `hasNotificationLink` not a type predicate | Resolved (round 2) |
| Round 2 #1 | Medium — stale `resumeFrom` survives audience/recipient/message changes | **Resolved.** New `useEffect` in `admin-notifications-modal.tsx:30-32` resets `resumeFrom` to `0` whenever `audience`, `selectedUserIds` (derived via `.map(id).join(',')`, avoiding array-identity churn), `text`, `type`, or `link` change. Traced the interaction with the partial-failure path in `handleSend`: `setResumeFrom(data.resumeFrom)` runs on failure but `resumeFrom` itself isn't a dependency of the effect, so it doesn't immediately get clobbered back to `0` — a retry with no field changes correctly resends with the stored offset. |
| Round 2 #2 | Low, informational — `resumeFrom` not a true idempotency key for `audience: 'all'` | **Resolved.** Code comment added at `route.ts:80-81` documenting the index-shift limitation, as suggested. |

All findings from both prior rounds are now resolved. No partial fixes outstanding.

### New findings

None. The fix is minimal and correctly scoped — no new bugs, edge cases, or regressions introduced by this commit.

### Verdict: **Approve**

---

joeseos:
## Review notes

Overall this is close — mostly clean and largely functionally correct. A few things worth addressing before merge.

### Likely-unintended behavior change
**`gang_invite` notifications now render a clickable "View campaign" link that dismisses them on click.**

The old code deliberately kept two types out of card navigation:
```ts
notification.link && notification.type !== 'friend_request' && notification.type !== 'gang_invite'
```
The new code drops that guard and shows the link attachment for any notification where `hasNotificationLink(link)` is true. Checking the triggers:
- `notify_friend_request_sent.sql` → `link` is `NULL` → no attachment (fine).
- `notify_gang_invite.sql` → `link` is `.../campaigns/{id}?gangId={gangId}` → **now renders a "View campaign" attachment** that navigates away and dismisses the invite when clicked.

That `?gangId=` param exists so the UI can drive accept/decline, which is exactly why these were originally non-navigable. Recommend re-guarding `gang_invite` (and `friend_request`) so the accept/decline flow isn't bypassed.

### Bloat / simplification
- **`resumeFrom` partial-send resume machinery** (route + modal wiring) is a lot of fragile complexity for an admin one-off broadcast — and its own comment admits it's unsound for the `all` audience ("new signups… can shift indices and skip or duplicate a recipient"). Consider dropping it or replacing with something idempotent.
- **`hasNotificationLink` is just an alias** for `isSafeNotificationLink` (identical body) — redundant indirection.
- **Dead `onSubmit` prop** — `AdminNotificationsModal` declares it and `admin/page.tsx` passes it, but the component only destructures `onClose`; `onSubmit` is never called.
- **`user-search-bar.tsx` is a 4th user-search variant** alongside `campaign-member-search-bar.tsx`, `friends.tsx`, and `admin-support-tools.tsx`. Putting the generic one in `components/shared/` is the right instinct, but nothing was consolidated, so it reads as net-new duplication rather than a refactor.

### Checks out (no action needed)
- Schema supports the insert: `sender_id` is nullable, `expires_at` is settable, and the `type` CHECK allows exactly `info`/`warning`/`error`.
- Link safety is solid: `javascript:`/`data:`/`vbscript:`/`file:` and protocol-relative `//` are blocked, and links render as real `href`s, not via `dangerouslySetInnerHTML`.
- `checkAdmin` + `createServiceRoleClient` are the established utilities; batching at 500 and paginating profiles at 1000 is reasonable.
- Empty link inserts as `''` rather than `NULL`, but `hasNotificationLink('')` is false, so it's harmless.

---

joeseos:
## Follow-up Review (round 5)

@Djidiouf Reviewed the one commit pushed since the last review (`175c276`, diff against `1555128`, the commit round 3's "Approve" and round 4's review notes were based on). Five files touched: `app/admin/page.tsx`, `app/api/admin/notifications/route.ts`, `components/account/notifications-content.tsx`, `components/admin/admin-notifications-modal.tsx`, `utils/notifications.ts`.

### Previous findings (rounds 1-3, all resolved — not re-litigated)

Link-scheme sanitization, `ORDER BY` on paginated fetch, `'Unauthorised'` spelling, and the `hasNotificationLink` type-predicate all remained resolved in this diff — untouched or only touched in ways consistent with the fix.

### Round 4 findings

| # | Finding | Status |
|---|---|---|
| 1 | Likely-unintended: `gang_invite` notifications now render a clickable, dismissing "View campaign" link, bypassing the accept/decline flow | **Resolved.** New `shouldShowNotificationLinkAttachment(type, link)` in `utils/notifications.ts:145-154` re-adds the `friend_request`/`gang_invite` guard (returns `false` for both regardless of link safety) and all three call sites in `notifications-content.tsx` (lines 190, 213, 397) were switched from `hasNotificationLink` to it. Confirmed no remaining callers of the old `hasNotificationLink` name anywhere in the repo. |
| 2 | Bloat: `resumeFrom` partial-send resume machinery is fragile complexity, unsound for `audience: 'all'` by its own admission | **Resolved.** Fully removed — `route.ts` no longer accepts/returns `resumeFrom`, and the modal's `useEffect`/state for it are gone. Partial-failure handling is simpler now: the route returns a descriptive error (`"Failed after sending to N users..."`) telling the admin to check before retrying, rather than offering an unsound auto-resume. Reasonable tradeoff for a low-frequency admin tool. |
| 3 | Bloat: `hasNotificationLink` was a pure alias of `isSafeNotificationLink` | **Resolved** — superseded by finding 1's fix; the function now has distinct logic (the type-based guard) rather than being a redundant wrapper. |
| 4 | Bloat: dead `onSubmit` prop on `AdminNotificationsModal`, never called | **Resolved.** Prop and its call site in `admin/page.tsx` both removed. Verified `Modal` (`components/ui/modal.tsx:44-58`) already calls `onClose()` itself when `onConfirm()` resolves non-`false`, so `handleSend`'s `return true` still closes the modal on success — no behavior change. |
| 5 | Bloat: `user-search-bar.tsx` is a 4th user-search variant, un-consolidated | **Outstanding**, untouched by this commit. Non-blocking per round 4 ("nothing was consolidated, so it reads as net-new duplication" — an observation, not a blocker). |

### New findings

None. Checked the diff for regressions: `receiverIds.map` (no more `.slice(resumeFrom)`) is correct now that resume is gone; the `useEffect` import was dropped from `admin-notifications-modal.tsx` along with its only usage (no dead import left behind); `shouldShowNotificationLinkAttachment`'s `link is string` type predicate is still valid since the added type check is a plain boolean short-circuit before the narrowing return. `npx tsc --noEmit` clean on the touched files.

### Verdict: **Approve**

All blocking findings from rounds 1-4 are resolved. The one remaining item (user-search-bar duplication) is a pre-existing non-blocking style observation, not a defect introduced by this PR.

---

joeseos:
## Follow-up Review (round 6)

@Djidiouf Reviewed the one commit pushed since round 5's approve (`75dec17`, diff against `175c276`). Two files touched: `components/account/friends.tsx`, `components/campaigns/[id]/campaign-member-search-bar.tsx`.

### Previous findings

| # | Finding | Status |
|---|---|---|
| Rounds 1-3 | Link sanitization, `ORDER BY`, spelling, type predicate, `resumeFrom` handling | Resolved (rounds 2-3), untouched by this commit |
| Round 4 #1 | `gang_invite`/`friend_request` link-click regression | Resolved (round 5), untouched by this commit |
| Round 4 #2 | `resumeFrom` bloat | Resolved (round 5), untouched by this commit |
| Round 4 #3 | `hasNotificationLink` pure alias | Resolved (round 5), untouched by this commit |
| Round 4 #4 | Dead `onSubmit` prop | Resolved (round 5), untouched by this commit |
| Round 4 #5 | `user-search-bar.tsx` is a 4th un-consolidated search variant | **Partially resolved.** `friends.tsx` and `campaign-member-search-bar.tsx` now both consume the existing shared `UserSearchBar` (`components/shared/user-search-bar.tsx`), dropping ~130 lines of duplicated fetch/debounce/dropdown logic between them. `admin-support-tools.tsx` and `create-battle-modal.tsx` (the latter not previously called out) still have their own inline `/api/search-users` implementations, so consolidation isn't complete — but this was flagged as non-blocking in round 4/5, and remains so. |

### New findings

None. Checked the migration to the shared component for regressions:
- `friends.tsx`: `handleAddFriend` builds the `Friend` object with `status: 'pending'/direction: 'outgoing'` inline instead of via the old `{...friend, status, direction}` spread — equivalent output. New `excludeIds={[userId, ...localFriends.map(f => f.id)]}` is a UX improvement (existing/pending friends no longer show in results) and isn't a behavior regression since `sendFriendRequest` already rejected duplicates.
- `campaign-member-search-bar.tsx`: correctly renamed the destructured `user` from `supabase.auth.getUser()` to `authUser` to avoid shadowing the new `user: UserSearchResult` parameter — no self-exclusion (`excludeIds`) is passed here, preserving the original "allow existing members to be added again" behavior.
- `npx tsc --noEmit` clean on both touched files; no other files reference the removed inline search state, so nothing was left dangling.

### Verdict: **Approve**

No blockers. The only outstanding item (search-bar consolidation) is cosmetic/non-blocking and was already flagged as such in prior rounds.

---

joeseos:
## Follow-up Review (round 7)

@Djidiouf Reviewed the one commit pushed since round 6's approve (`9cc13e8`, diff against `75dec17`). One file touched: `components/battle-session/create-battle-modal.tsx`.

### Previous findings

| # | Finding | Status |
|---|---|---|
| Rounds 1-3 | Link sanitization, `ORDER BY`, spelling, type predicate, `resumeFrom` handling | Resolved, untouched by this commit |
| Round 4 #1 | `gang_invite`/`friend_request` link-click regression | Resolved (round 5), untouched by this commit |
| Round 4 #2-4 | `resumeFrom` bloat, `hasNotificationLink` alias, dead `onSubmit` prop | Resolved (round 5), untouched by this commit |
| Round 4/6 #5 | `user-search-bar.tsx` consolidation incomplete — `create-battle-modal.tsx` and `admin-support-tools.tsx` still had inline search implementations | **Further resolved.** `create-battle-modal.tsx` now consumes the shared `UserSearchBar` (`components/shared/user-search-bar.tsx`), dropping its own `searchQuery`/`searchResults`/`isSearching`/`showSuggestions` state, the debounced `useEffect`, and the inline dropdown markup (~70 lines → ~15). `admin-support-tools.tsx` remains on its own implementation, but it hits a distinct endpoint (`/api/admin/support/search-users`, presumably broader admin-only search semantics) rather than `/api/search-users`, so it's arguably not the same duplication — non-blocking either way, as flagged in prior rounds. |

### New findings

None. Checked the migration for regressions:
- `selectedUser` is now typed `UserSearchResult | null` (from the shared component) instead of the old inline `{ id, username }` shape — identical fields, and both call sites (`selectedUser.id`, `selectedUser.username`) still resolve correctly.
- `excludeIds={opponents.map((o) => o.userId)}` correctly replaces the old manual `filteredSearchResults` filter (previously `!opponents.some((o) => o.userId === p.id)`) — same exclusion semantics, now handled inside the shared component.
- The old `onBlur` + 200ms-timeout hack for closing suggestions before a click could register is gone; the shared component has no such issue since it doesn't hide the dropdown on blur, so clicking a result works correctly without the hack.
- No leftover references to the removed local state (`searchQuery`, `searchResults`, `isSearching`, `showSuggestions`, `filteredSearchResults`) anywhere in the file.
- `npx tsc --noEmit` clean on the touched file.

### Verdict: **Approve**

All prior findings remain resolved; this commit is a clean, well-scoped simplification with no regressions.

---

PR #1878, opened 2026-07-19
Add opt-in campaign join requests with arbitrator approval

joeseos:
Campaign owners/arbitrators can enable "Allow join requests" in the
campaign settings modal (new campaigns.allow_join_requests flag). When
enabled, a logged-in non-member sees a "Request to join" button on the
campaign page; every OWNER/ARBITRATOR gets a notification with
Accept/Decline buttons, and accepting adds the requester as a MEMBER
(reusing addMemberToCampaign, whose trigger also notifies the requester).

Pending requests live in a new campaign_join_requests table rather than
as PENDING rows in campaign_members: every existing policy and query
treats a campaign_members row's existence as membership (its status
column is unused), so a pending member row would grant authority before
acceptance — the defect class fixed for campaign_gangs in
20260719120000_gate_arb_gang_access_on_accepted.sql. A request row is
referenced by no authorization logic and grants nothing.

RLS is the enforcement layer: INSERT is self-only and only into
campaigns with the flag enabled while not already a member; SELECT and
DELETE are limited to the requester, campaign OWNER/ARBITRATOR
(private.is_arb) and admins. UNIQUE(campaign_id, user_id) prevents
duplicate requests. The notify_campaign_join_request trigger fans out
one notification per OWNER/ARBITRATOR (DISTINCT — campaign_members can
hold duplicate rows), and accept/decline/withdraw delete all
arbitrators' copies via the service-role client so no stale
notifications linger. Accept is race-safe when two arbitrators act on
the same request: the loser gets a graceful no-op instead of a duplicate
member row.

The new notification type is in-app only (no email), and the schema
snapshot is left to the nightly workflow. Deploy order matches the
split_invite convention: migration applies before the trigger-function
file auto-deploys from the same merge.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01C8bSMK8d6mZ8HMxXF4UjHD


joeseos:
## Code Review

@joeseos Reviewed against the codebase — schema (`campaign_members`, `campaign_join_requests`, notifications), RLS policies, `addMemberToCampaign`, and the existing gang-invite notification pattern this PR mirrors. Overall this is a well-designed feature: keeping requests out of `campaign_members` (avoiding the same "pending row grants authority" defect class fixed for `campaign_gangs`), RLS as the real enforcement layer, and the notification fan-out/cleanup are all sound choices. One correctness issue worth fixing.

### Findings

**1. [Medium] `acceptJoinRequest`'s "race-safe" accept is not actually race-safe — concurrent accepts can create a duplicate `campaign_members` row**
`app/actions/campaigns/[id]/campaign-join-requests.ts` (~lines 95-140), combined with `addMemberToCampaign` in `app/actions/campaigns/[id]/campaign-members.ts` (lines 452-476)

The PR description states: *"Accept is race-safe when two arbitrators act on the same request: the loser gets a graceful no-op instead of a duplicate member row."* The implementation is check-then-act, not atomic:

```ts
const [{ data: request }, { data: existingMembers }] = await Promise.all([
  supabase.from('campaign_join_requests').select('id')...,
  supabase.from('campaign_members').select('id')...
]);
if (existingMembers && existingMembers.length > 0) { /* graceful no-op */ }
...
const addResult = await addMemberToCampaign({ campaignId, userId, role: 'MEMBER', invitedBy: user.id });
```

If two arbitrators click "Accept" close together, both `Promise.all` reads can complete before either has inserted a member row, so both see `existingMembers.length === 0` and both proceed to `addMemberToCampaign`. That function (lines 452-476) does its own check-then-insert and then unconditionally `INSERT`s — there's no `ON CONFLICT`/upsert to fall back on. I confirmed `public.campaign_members` (schema.public.sql:4636) has **no unique constraint on `(campaign_id, user_id)`** — only a `PRIMARY KEY (id)` and a plain (non-unique) index `idx_campaign_members_campaign_user_role`. So both concurrent inserts succeed, producing two `campaign_members` rows for the same user in the same campaign — the exact outcome the PR claims is prevented. (This is also why the trigger and cleanup logic elsewhere already need `DISTINCT`/multi-row handling — duplicates are a known, tolerated condition in this table, which is precisely why this path needs to not create new ones.)

Suggested fix: make the accept atomic at the DB layer rather than relying on an app-level check-then-act — e.g. a single `SECURITY DEFINER` SQL function/RPC that does the existence check, member insert, and request delete inside one transaction (optionally with `SELECT ... FOR UPDATE` on the request row), or add a partial unique index on `campaign_members(campaign_id, user_id)` if no other flow depends on allowing duplicates. Low real-world likelihood (requires near-simultaneous clicks from two arbitrators), and impact is a duplicate row rather than a crash (nothing else in the codebase queries `campaign_members` by `(campaign_id, user_id)` with `.single()`), but it's a real gap between the code and the documented guarantee.

### Nice-to-confirm (not blocking)
- `notify_campaign_join_request.sql` fans out to every OWNER/ARBITRATOR — if a campaign ever ends up with zero (all left), a request would sit forever with no one able to act on it (no expiry/escalation). Likely a non-issue in practice since campaigns should always retain an OWNER, just flagging for awareness.

Just the one finding — it's a narrow race window with low blast radius, but worth closing since it directly contradicts the stated design guarantee. Everything else (RLS design, notification cleanup, UI wiring, cache invalidation via `BASE_CAMPAIGN_BASIC`) checks out.

---

joeseos:
## Follow-up Code Review

@joeseos Reviewed the 3 commits since the last review (e6fabd5, 699fcea, 725566d). Scope: only the delta since 3cb03f7.

### Previous findings

**1. [Medium] `acceptJoinRequest`'s accept was check-then-act, not atomic — Resolved, with a new gap introduced by the fix**
`app/actions/campaigns/[id]/campaign-join-requests.ts`

Commit 725566d replaces the check-then-act with a claim-by-delete pattern: `DELETE ... WHERE campaign_id = ? AND user_id = ? RETURNING id`. Postgres guarantees only one of two concurrent deletes on the same row can return it, so exactly one caller proceeds to `addMemberToCampaign`; the loser (or anyone hitting an already-handled request) gets a graceful no-op. This is the right fix and does close the original race — traced it through and it holds.

However, the failure-recovery branch this commit adds has two new issues (see below), both localized to the same ~10 lines.

### New findings

**1. [Medium] Restoring a claimed request after a failed `addMemberToCampaign` re-triggers the arbitrator notification fan-out without cleaning up the original notifications first — duplicate in-app notifications and duplicate emails**
`app/actions/campaigns/[id]/campaign-join-requests.ts` (the `if (!addResult.success) { ... insert ... }` block, ~line 172-179)

When the winning caller's `addMemberToCampaign` fails, the code restores the request via a fresh `INSERT INTO campaign_join_requests`. That insert is exactly what `on_campaign_join_request` (in `notify_campaign_join_request.sql`) fires on — it unconditionally inserts one notification per OWNER/ARBITRATOR, no dedup. But this branch never calls `cleanupJoinRequestNotifications()` first, so the *original* notifications from the initial request (still sitting in the table — they're only ever deleted on success or on the "lost the race" no-op path) are left in place, and the restore-insert adds a second, identical set on top. Every OWNER/ARBITRATOR — including the one who just attempted the accept — ends up with two "wants to join your campaign" notifications for the same request.

This is a new interaction, not something either individual commit would show on its own: the restore-on-failure logic is new in 725566d, and `campaign_join_request` only got wired into the email pipeline in 699fcea (`defaultEnabled: true`) — so the duplicate now means every arbitrator also gets **two identical emails** on a transient `addMemberToCampaign` failure (a real path — e.g. any transient insert/network error in that function, not a contrived case).

Suggested fix: call `await cleanupJoinRequestNotifications(campaignId, userId);` immediately before the restore insert, so the re-insert's trigger produces exactly one fresh notification per arbitrator instead of stacking on the stale set.

**2. [Low] Restore insert's error isn't checked or logged**
Same block, the `.insert({ campaign_id: campaignId, user_id: userId })` call has no `{ error }` destructure/check.

If that insert itself fails — e.g. the requester withdrew and immediately re-requested between the claim-delete and this point, so a fresh row already exists and the restore hits the `UNIQUE(campaign_id, user_id)` constraint — the failure is silently swallowed. The function still returns `{ success: false, error: addResult.error }` (so the *acting* arbitrator sees the original add error), but the request is now gone with no signal that the restore itself also failed, and no `console.error` trail to debug it from. Low severity since it only compounds an already-rare transient-failure path, but a one-line `if (error) console.error(...)` would close the gap.

Nothing else new: the `e6fabd5` status-column drop is thorough (grepped the whole branch — no leftover SQL/RLS/TS references to `campaign_members.status` outside what the commit already touches, and the gang-page campaign-picker predicate rewrite is behavior-preserving since the old `!c.status` was always-true). The `699fcea` email-config addition is consistent across all three read sites (same type key, subject, and defaults in `enqueue_notification_email.sql`, the edge function's `EMAIL_CONFIG`, and `utils/notifications.ts`).

The core race-condition fix is correct and the two new findings are narrow (both only reachable via a transient `addMemberToCampaign` failure), but finding 1 is a real, easily user-visible regression (duplicate emails) introduced by this round's changes, and the fix is a one-line addition. Worth closing before merge.

---

PR #1879, opened 2026-07-20
Fix/print scaled cards

Djidiouf:


joeseos:
## Review: Fix/print scaled cards

@Djidiouf Reviewed the diff against `components/gang/print-gang.tsx` and the related `components/gang/fighter-card.tsx` / `fighter-card-action-menu.tsx`. Small, well-scoped change — logic checks out for the stated goal (stretch same-row cards to a common height when "Adjust Cards Height to Content" is on).

### Findings

**1. [Medium] Layout correctness depends on an undocumented DOM assumption in `FighterCardActionMenu`**
- File: `components/gang/print-gang.tsx:1305`
- The new wrapper relies on `[&>*:first-child]:flex [&>*:first-child]:grow ...` to turn the *first child* of the row item into a flex-grow container. That first child is not the fighter card itself — it's the `<div className="relative">` rendered unconditionally by `FighterCardActionMenu` (`components/gang/fighter-card-action-menu.tsx:236`), with the actual `.fighter-card-bg` element nested inside it.
- This happens to work today only because `disableLink={true}` short-circuits the open-menu handlers (`fighter-card-action-menu.tsx:109,118,138`), so that wrapper's only rendered child in print mode is the card. If that component's markup ever gains an always-rendered sibling before `{children}` (e.g. an accessibility element, a wrapping fragment change), the print "scale to content" layout would silently break with no type error or lint signal — it'd just fail visually at print time.
- Suggestion: make the coupling explicit rather than positional — e.g. have `FighterCardActionMenu`'s wrapper `div` forward a `className` prop (defaulting to `"relative"`) so `print-gang.tsx` can target it directly instead of via `:first-child`, or drop the `relative` div down to `display: contents` when `disableLink` so it doesn't affect the box model at all.

**2. [Low] Redundant selectors on the same element**
- File: `components/gang/print-gang.tsx:1305`
- `[&>*:first-child]:grow [&>*:first-child]:min-h-0` and `[&_.fighter-card-bg]:min-h-0 [&_.fighter-card-bg]:grow` end up applying to overlapping/same-effect targets once you trace the DOM (see finding 1) — the `:first-child` and the descendant `.fighter-card-bg` selectors both drive grow/min-h-0 on what is effectively a two-level chain. Not incorrect, just adds to an already very dense className string; a short comment on *why* both are needed (chain through the action-menu wrapper) would save the next reader from re-deriving this.

### Not flagged (verified, no issue)
- The "before"/"after" Gang Card and Additional Details Card blocks (`print-gang.tsx:1003,1195,1446,1638`) weren't touched, and don't need to be — they're direct flex children with `.fighter-card-bg` themselves, so the container-level `[&_.fighter-card-bg]:h-auto!` (line 998) already handles their height correctly without a wrapper.
- The `shrink-0` added to the W/FW/XP footer div (`print-gang.tsx:1358`) and the wrapper's conditional `flex flex-col` are inert in the non-scaled default view (plain block div, no flex ancestor), so no regression there.
- Checkbox label text change (`print-gang.tsx:356`) is cosmetic only.
- No CLAUDE.md conventions file in this repo to check against; no security, perf, or null-handling concerns — this is presentation-only, gated behind print view state.

### Verdict: **Approve**

The medium finding is a maintainability/fragility note rather than a functional bug on the current codebase — happy to see this merged as-is, but consider a quick follow-up to de-fragilize the `:first-child` dependency before it's forgotten.


---

joeseos:
## Review — @Djidiouf

No prior review comments/threads exist on this PR yet, so this covers the full diff (1 file, `components/gang/print-gang.tsx`).

### Previous findings
None — this is the first review on this PR.

### New findings

**1. Gang Card doesn't participate in the new stretch-to-tallest-sibling behavior (medium)**
- File: `components/gang/print-gang.tsx`, ~L995 (container `items-stretch` toggle) and ~L1003 (Gang Card block)
- Issue: When "Adjust Cards Height to Content" is on, the row container switches from `items-start` to `items-stretch`, and `h-[435px]` on `.fighter-card-bg` is overridden to `h-auto`. Each per-fighter wrapper (L1303-1306) was given `flex flex-col` plus `[&_.fighter-card-bg]:grow`/`min-h-0` so the card's own content expands to fill the stretched row height. The **Gang Card** block (L1003) has none of that wiring — it's not a flex container, and its inner `relative grow w-full` div (L1039) has no effect since its parent isn't `flex`. So when the Gang Card shares a flex row/line with a taller auto-height fighter card, `items-stretch` will still stretch the Gang Card's own border box to match the row height, but its content stays at natural size — leaving a visible empty gap at the bottom of the card, with the border no longer hugging the content. This directly undercuts the PR's own goal ("scale cards in the same row to the same height") for the one card type that wasn't updated.
- Fix suggestion: either apply the same `flex flex-col` + grow wiring used for fighter cards to the Gang Card (so its existing `grow` div actually does something), or give the Gang Card `self-start` so it opts out of the new stretch behavior and keeps its old fixed/natural height instead of a half-empty stretched box.

Everything else in the diff (the flex/grow chain for fighter cards, the `shrink-0` added to the W/FW footer, the label text change) checks out — I traced the actual DOM chain through `FighterCard` → `FighterCardActionMenu` (which renders a bare `<>{children}</>` fragment in print views, no modals context) → `cardContent`'s `relative` div, and the `[&>*:first-child]` / `[&_.fighter-card-bg]` selectors correctly reach and flex-ify the right elements for fighter cards.

Minor/non-blocking: 3 commits all titled "Scale cards in the same row to the same height" — might be worth squashing before merge, but not a functional issue.

The Gang Card stretch gap is a real (if narrow) regression in the exact feature this PR adds. Small, targeted fix should resolve it.


---

joeseos:
## Review: Fix/print scaled cards (latest commit — `b9ed7e5`)

@Djidiouf Reviewing only the delta since the last review (commits `b1686e5`..`b9ed7e5`, both in `components/gang/print-gang.tsx`).

### Previous findings

**1. [Medium, review #1] Layout correctness depended on an undocumented DOM assumption in `FighterCardActionMenu`** — **Resolved**
The fighter-card wrapper (`print-gang.tsx:1308`) no longer makes the first child itself `flex flex-col grow`; it now collapses that first child with `[&>*:first-child]:contents` and applies `grow`/`min-h-0` directly to `.fighter-card-bg` via the descendant selector. I traced the actual chain in `fighter-card-action-menu.tsx:231-233` and `fighter-card.tsx:850-868`: `FighterCardActionMenu` returns a bare `<>{children}</>` fragment (no wrapper div) whenever `modalsContext` is absent, which is always true in this print render, so the real first child in the DOM is `FighterCard`'s own `cardContent` div (`fighter-card.tsx:851`), not a `FighterCardActionMenu` div. The new code comment (`print-gang.tsx:1301-1305`) states this correctly. `disableLink={true}` is hardcoded at the call site (`print-gang.tsx:1349`), so `cardWithOptionalLink` is `cardBody` (the `.fighter-card-bg` element) directly with no `<a>` wrapper in between — confirmed the `contents` collapse lands exactly where intended. Good fix, and more robust than the previous positional flex-chain since `grow`/`min-h-0` now target the stable `.fighter-card-bg` class rather than "whatever happens to be first-child."

**2. [Low, review #1] Redundant selectors on the same element** — **Resolved**
The overlapping `[&>*:first-child]:flex/min-h-0/grow/flex-col` set was removed outright rather than just commented; only `[&>*:first-child]:contents` remains alongside the `.fighter-card-bg` grow/min-h-0 rules. No more overlap.

**3. [Medium, review #2] Gang Card doesn't participate in the stretch-to-tallest-sibling behavior** — **Resolved**
`[&_.fighter-card-bg]:flex! [&_.fighter-card-bg]:flex-col!` was added to the container-level classes (`print-gang.tsx:998`), scoped to `scaleCardsToContent`. Since the Gang Card (`:1003`/`:1449`) and Additional Details Card (`:1195`/`:1641`) blocks all carry the `fighter-card-bg` class directly, this makes all four of them flex columns, which makes their existing (previously inert) `relative grow w-full` inner div actually grow — this is a container-level fix rather than a per-card patch, so it fixes both card types at both "before" and "after" positions, not just the Gang Card instance that was flagged.

**Minor/non-blocking, review #2 — still outstanding:** the branch still has 4 commits, all titled "Scale cards in the same row to the same height." Squashing before merge would clean up history, but this was already called out as non-blocking.

### New findings from this delta

None. This is a presentation-only CSS/className change; no new logic, no new null-handling surface, no security/perf implications. I didn't find anything the fix broke — traced both the fighter-card DOM chain and the two duplicated static card blocks to confirm the fix lands correctly in each case.

One small residual note, not a blocker: `[&>*:first-child]:contents` still assumes exactly one wrapper level between the row-item div and `.fighter-card-bg` (true today given `disableLink={true}` is hardcoded here). If `FighterCardActionMenu` ever rendered a real wrapper div in this call path (e.g. if a `ModalsProvider` context ever got introduced above this print view), the chain would need re-checking — same class of fragility as before, just moved. Not worth blocking on given the current code path is verified correct.

### Verdict: **Approve**

Both previously-required findings are addressed, and the Gang Card fix is applied more comprehensively (container-level) than the narrow per-card patch that was suggested. Nothing new introduced.

---

PR #1880, opened 2026-07-21
Added OOA records to track fighters and vehicle wrecked

Djidiouf:


joeseos:
@Djidiouf Thanks for the PR — reviewed against the codebase. Solid feature overall (snapshotting, RLS, cache-tag reuse all follow existing conventions), but there's a blocking SQL bug in the migration.

## Findings

### 🔴 High — migration has a SQL syntax error, table will not be created
`supabase/migrations/20260721120000_create_fighter_ooa_records.sql:13`
```sql
causing_fighter_name text,
causing_fighter_type text;   -- <-- semicolon, not a comma
causing_fighter_class text,
```
The semicolon terminates the `CREATE TABLE` statement early. Everything from `causing_fighter_class text,` onward becomes a dangling, invalid top-level statement, so the migration will fail to apply outright (and every subsequent index/RLS statement in the file goes with it, since none of them will run once `CREATE TABLE` errors). This needs to be a comma.

### 🟡 Medium — `causing_fighter_gang_id` is a dead duplicate column
`supabase/migrations/20260721120000_create_fighter_ooa_records.sql:15`, `app/actions/fighter-ooa-records.ts:195`, `app/actions/edit-fighter.ts:1242`

`causing_fighter_gang_id` is always set to the exact same value as `causing_gang_id` (`causing.gang_id` / `fighter.gang_id`) in both write paths, and it's never read anywhere in the UI — only `causing_fighter_gang_name` is displayed (`fighter-ooa-history-modal.tsx:475-481`). It's also asymmetric with the injured side, which has no `injured_fighter_gang_id` equivalent. Recommend dropping the column, or if it's meant to be an immutable snapshot that survives gang deletion (unlike `causing_gang_id`, which has `ON DELETE SET NULL`), leave a comment explaining that and add the matching `injured_fighter_gang_id` for consistency.

### 🟡 Medium — combobox-building logic duplicated verbatim across two modals
`components/fighter/fighter-ooa-history-modal.tsx:190-201` and `components/fighter/fighter-xp-modal.tsx:135-146`

`gangOptions` and `getFighterOptions` (including the `fighter_class === 'Crew'` filter) are copy-pasted identically between the two modals. The codebase already has a precedent for extracting this kind of thing (`utils/crew-fighter-combobox-options.tsx`'s `useCrewFighterOptions`); worth pulling this into a shared hook/util (e.g. `useCampaignGangFighterOptions`) instead of maintaining two copies.

### 🟢 Low — stale cache after edit/delete for the affected counterpart fighter
`components/fighter/fighter-ooa-history-modal.tsx:123,141,158`

All three mutations only `invalidateQueries` on `['fighter-ooa-records', fighterId]` (this fighter's "Caused" tab). The "Sustained" tab query key (`fighter-ooa-records-sustained`) for this fighter, and the counterpart fighter's queries if their modal happens to be open elsewhere in the same session, are never invalidated — they'll show stale data until the 30s `staleTime` lapses. Minor given the modal is usually reopened fresh, but worth a quick fix (invalidate both key patterns, e.g. `predicate` matching `'fighter-ooa-records'` prefix).

### 🟢 Low — leftover invalid Tailwind class
`components/fighter/fighter-ooa-history-modal.tsx:472`
```
<div className="mt-1 text-xs text-muted-foreground space-y-">
```
`space-y-` has no suffix, so it's a no-op utility (likely a leftover from editing, e.g. meant to be `space-y-1`). Cosmetic only.

### 🟢 Low (edge case) — edit form can pre-fill a gang not in the combobox options
`components/fighter/fighter-ooa-history-modal.tsx:206-215`

`startEdit` pre-fills `gangId` from `record.injured_gang_id`, but `campaignGangs` only contains the current campaign's ACCEPTED gangs plus the fighter's own gang. If the injured fighter's gang has since left the campaign (or the record was made under a different campaign context), the pre-filled gang won't appear in the dropdown, `getFighterOptions` returns `[]`, and the user can't see/re-pick the current target without a clear indicator why the fighter list is empty.

The migration syntax error (finding #1) is blocking — nothing in this PR works until it's fixed. The rest are non-blocking but worth addressing, especially the duplicate column and duplicated combobox logic before merge.


---

joeseos:
@Djidiouf Reviewed the latest commit (`2068186`, the update since the [previous review](https://github.com/joeseos/mundamanager/pull/1880#issuecomment-5030514384)) against that commit's diff only.

## Previous findings

| # | Finding | Status |
|---|---|---|
| 1 | 🔴 Migration SQL syntax error (semicolon instead of comma) | **Resolved** — `causing_fighter_type text,` fixed in `20260721120000_create_fighter_ooa_records.sql:13` |
| 2 | 🟡 `causing_fighter_gang_id` dead duplicate column | **Resolved** — dropped from the migration, the `FighterOoaRecord` interface, and both write paths (`edit-fighter.ts`, `fighter-ooa-records.ts`). Confirmed no leftover references repo-wide. |
| 3 | 🟡 Combobox-building logic duplicated across two modals | **Resolved** — extracted into `utils/campaign-gang-fighter-options.tsx` (`useCampaignGangFighterOptions` + `buildFighterComboboxOption`), now shared by `fighter-ooa-history-modal.tsx` and `fighter-xp-modal.tsx` |
| 4 | 🟢 Stale cache after edit/delete for counterpart fighter | **Resolved** — new `invalidateOoaRecordQueries()` predicate-invalidates any query key starting with `fighter-ooa-records`, wired into all three mutations |
| 5 | 🟢 Leftover invalid Tailwind class `space-y-` | **Resolved** — now `space-y-1` |
| 6 | 🟢 Edit form can pre-fill a gang not in the combobox options | **Resolved** — `editGangMissingFromCampaign` detects the case, synthesizes a placeholder gang/fighter option so the current target stays visible and selectable, and surfaces a hint ("This gang is no longer in the campaign...") |

All six are fully addressed, not just partially.

## New findings

Nothing blocking. One very minor nit:

- **Low** — `app/actions/fighter-ooa-records.ts:139` (`getCampaignGangsAndFighters`): the new `profiles` lookup (`const { data: profiles } = await supabase.from('profiles')...`) doesn't check `error` the way every other query in this function does (`if (error) throw error`). Worst case today is a silently-missing `owner_username`, so not blocking, but worth the same error check for consistency.

The unrequested `owner_username` addition (gang combobox now shows the owning player, matching `/api/campaigns/campaign-gangs`) is a reasonable, correctly-implemented enhancement — batched single query, no N+1, uses the request-scoped client consistent with the rest of the file.

## Verdict: **Approve**

Nice, thorough round of fixes — happy to merge once CI is green.

---

joeseos:
## Review — OOA / Wreck records

Solid feature and nicely commented. Snapshotting denormalized data so history survives deletions is the right call, and the RLS policies are consistent with the rest of the schema. A few improvements, grouped by theme.

### Architecture / DRY

**1. Reads should go through Route Handlers, not server actions**
The modals lazy-load correctly (`useQuery` + `enabled: isOpen` + `staleTime`), but their `queryFn` calls server actions directly (`getFighterOoaRecords`, `getFighterSustainedOoaRecords`, `getCampaignGangsAndFighters`). The established convention here is `useQuery` + `fetch('/api/...')` to a GET Route Handler — used by `LogModal` (the direct sibling of this feature), `skill-access-modal`, `fighter-edit-modal`, `vehicle-edit`, etc. (30/39 `useQuery` files). Server actions are POST and execute serially, so this modal's three reads can't parallelize and get no HTTP caching.
→ Convert the three reads to Route Handlers; keep the writes as server actions.

**2. Record persistence is split across two files, with triplicated snapshot logic**
The same "look up injured fighter → gang → vehicle → build snapshot row" exists three times: `insertFighterOoaRecords` (in `edit-fighter.ts`), `createFighterOoaRecord`, and `updateFighterOoaRecord` (in `fighter-ooa-records.ts`). The XP-flow write is stranded in `edit-fighter.ts` while its near-identical siblings live in the records file.
→ Move `insertFighterOoaRecords` into `fighter-ooa-records.ts` and factor one shared `buildOoaSnapshotRow(supabase, injuredFighterId)` that all paths call.

**3. `getCampaignGangsAndFighters` duplicates an existing endpoint**
It re-implements the campaign_gangs → gangs → profiles/username logic already in `GET /api/campaigns/campaign-gangs` (the code comment even notes it "matches" that route), just adding a per-gang fighters fetch.
→ Reuse/extend that route instead. This also resolves finding #1 for that read.

**4. YAGNI (optional): the standalone "Add record" flow is redundant**
The optional log is already integrated into the OOA/XP flow. The history modal's manual **Add** is the only path that creates a record without a counter increment — the source of the list-vs-`kills` divergence. If revisited, consider making the history modal view/edit/delete only, so records are only ever born alongside the OOA flow. (Also collapses `createFighterOoaRecord`, easing #2.)

### Correctness / robustness
- **XP-path insert swallows errors** (`console.error` only) — OOA records can silently fail while XP/kills succeed. By design, but worth being aware of.
- **Edit re-snapshots from *current* fighter data**, which contradicts the modal's "recorded as it was at the time" note.
- **Non-deterministic vehicle pick** in the batch insert when a crew has multiple vehicles (`!vehicleMap.has(...)`).
- **`campaignId = campaigns?.[0]?.campaign_id`** — a gang in multiple campaigns only scopes the target pickers to its first campaign.
- **Two invalidation models** — the XP path uses `revalidateTag`, the modal uses React Query `invalidateQueries`; no single source of truth for "records changed." Works because the modal is closed during the XP write.

### Verified consistent (no action)
- **RLS read policy** (`FOR SELECT TO authenticated USING (true)`) matches `fighters`, `gangs`, `vehicles`, `campaign_gangs`, and `battle_session_fighters`. Scoped INSERT/UPDATE/DELETE (owner/admin/arb) also match the write-policy pattern.

### Suggested priority
1. Reads → Route Handlers (#1) + reuse the campaign-gangs route (#3)
2. Consolidate persistence + shared snapshot helper (#2)
3. Robustness items
4. Optional: consolidate the create flow (#4)

joeseos:
@Djidiouf Reviewed the latest commit (`3b8a04f`) against the [previous architecture review](https://github.com/joeseos/mundamanager/pull/1880#issuecomment-5034700707) — diffing `2068186` → `3b8a04f` only.

## Previous findings

| # | Finding | Status |
|---|---|---|
| 1 | Reads should go through Route Handlers, not server actions | **Resolved** — new `GET /api/fighters/[id]/ooa-records?direction=caused\|sustained` route added; the three modal reads now go through `utils/api/fighter-ooa-records.ts` fetch wrappers instead of calling server actions directly |
| 2 | Snapshot logic triplicated across `edit-fighter.ts`/`fighter-ooa-records.ts` | **Resolved** — `insertFighterOoaRecords` moved into `fighter-ooa-records.ts`; all three write paths now share `fetchInjuredSnapshotMap`/`buildInjuredSnapshot` for the injured-fighter → gang → vehicle lookup |
| 3 | `getCampaignGangsAndFighters` duplicates `/api/campaigns/campaign-gangs` | **Resolved** — that route gained `gangId` (own-gang guarantee) and `includeFighters` params and is now the single source; the server-action duplicate is deleted |
| 4 | YAGNI: standalone "Add record" flow (optional) | **Not addressed** — left as-is, this was flagged optional/for-later, fine to leave |
| Non-deterministic vehicle pick on multi-vehicle crews | **Resolved** — vehicles are now fetched `order('created_at', ascending: true)` and the first match per fighter is kept, so the pick is deterministic |
| XP-path insert swallows errors (`console.error` only) | **Outstanding** — unchanged, still by-design per the earlier note |
| Edit re-snapshots from *current* fighter data, not time-of-record | **Outstanding**, now explicitly documented in the docstring rather than fixed — reasonable to leave as a known behavior |
| `campaignId = campaigns?.[0]?.campaign_id` (multi-campaign gang scoping) | **Outstanding** — untouched by this commit |
| Two invalidation models (`revalidateTag` vs React Query `invalidateQueries`) | **Outstanding** — untouched by this commit |

Nice, thorough refactor — all three blocking/DRY architecture items are cleanly resolved, not just partially.

## New findings

Nothing blocking. One thing worth a quick look, not a regression from this commit but newly *reachable* through it:

- **Low/verify** — `app/api/campaigns/campaign-gangs/route.ts`: the new `gangId` param lets a caller fetch an arbitrary gang (bypassing the campaign-membership/ACCEPTED filter) and, with `includeFighters=true`, its full fighter roster (names/types/classes) by UUID alone — there's no check that the gang belongs to the campaign or to the requesting user. This route has never called `getAuthenticatedUser` (pre-existing, unchanged by this diff), so protection relies entirely on `gangs`/`fighters` RLS. If those tables' SELECT policies are permissive to any authenticated user (consistent with the `USING (true)` pattern used elsewhere in this schema, e.g. `battle_sessions`, `fighter_ooa_records`), this is a non-issue and matches existing app conventions. Worth a 30-second confirmation of the `gangs`/`fighters` SELECT policies before merge, since this is the first caller to expose a gang's full fighter list keyed purely off a client-supplied `gangId`.
- **Nit** — same file: the own-gang synthesized entry uses `id: ''` (`campaignGangs.push({ id: '', ... })`) as a sentinel for "not a real `campaign_gangs` row," then unwraps it via `campaign_gang_id: cg.id || null` later. Works correctly, just a slightly implicit sentinel — an explicit `campaign_gang_id: cg.id ? cg.id : null` inline at the push site (or a dedicated flag) would read a little clearer, but not worth blocking on.

Typecheck (`tsc --noEmit`) and lint are clean on this commit for all changed files.

## Verdict: **Approve**

All architecture findings from the previous round are resolved and nothing new is blocking. The RLS note above is worth a quick confirmation but not blocking given it matches this schema's existing access patterns.

---

joeseos:
@Djidiouf Reviewed the latest commit (`a80e52c`) against the [previous review](https://github.com/joeseos/mundamanager/pull/1880#issuecomment-5042284587) — diffing `3b8a04f` → `a80e52c` only (`app/api/campaigns/campaign-gangs/route.ts`, +6/-3).

## Previous findings

| # | Finding | Status |
|---|---|---|
| RLS on `gangId`/`includeFighters` param (unauthenticated arbitrary-gang fighter-roster read) | **Outstanding** — untouched by this commit, still worth the 30s policy confirmation before merge |
| Nit — implicit `id: ''` sentinel unwrapped via `cg.id \|\| null` | **Resolved** — `CampaignGangWithGang.id` is now explicitly typed `string \| null`, the synthesized own-gang entry pushes `id: null` directly (with an explanatory comment), and the mapper assigns `campaign_gang_id: cg.id` without the truthiness fallback |
| All other prior findings (route-handler conversion, snapshot consolidation, duplicate-endpoint removal, YAGNI add-flow, vehicle-pick determinism, etc.) | Untouched by this commit — no change in status |

## New findings

None. This commit is a self-contained, mechanical follow-up to the nit — it doesn't touch behavior:
- `CampaignGangWithGang.id` was already only ever `''` (sentinel) or a real non-empty UUID, so `cg.id || null` and the new direct `cg.id` produce identical output; this is a type-clarity change only, not a functional one.
- Confirmed `campaign_gang_id` isn't consumed by the OOA/XP feature's `CampaignGangWithFighters` type or `useCampaignGangFighterOptions` (it keys off `gang_id`), so there's no downstream code that could be affected by the sentinel's type changing from `''` to `null`.
- `tsc --noEmit` clean on the changed file.

## Verdict: **Approve**

Small, correct cleanup commit. The one outstanding item (RLS confirmation on the `gangId`/`includeFighters` params) remains worth a quick check before merge, but is pre-existing and not a regression from this diff.

---

PR #1882, opened 2026-07-21
Fix header auth state going stale after login

joeseos:
Render the header's auth-gated UI from the server instead of the
client-only useClaims hook. Previously, signing in ran as a server
action that set the session cookies server-side and finished with a
soft redirect(); the browser Supabase client behind useClaims never
saw a SIGNED_IN event and the layout never remounted, so userId stayed
null and the settings (hamburger) menu was hidden until a manual
refresh re-instantiated the browser client.

HeaderAuth is now an async server component that reads getClaims() via
the server Supabase client, so the existing revalidatePath('/', 'layout')
in signInAction/signOutAction updates the header on the same navigation
with no reload. This follows the documented Supabase Next.js SSR pattern
of deriving auth-gated UI on the server.

Claude-Session: https://claude.ai/code/session_01V9nGi1PHDUTSE6z5BbYB9E

### DIFF
--- a/components/header-auth.tsx
+++ b/components/header-auth.tsx
@@ -1,27 +1,21 @@
-'use client';
-
-import { useClaims } from '@/hooks/use-claims';
+import { createClient } from '@/utils/supabase/server';
+import { getClaims } from '@/utils/auth';
 import SettingsModal from '@/components/settings-modal';
 import { ThemeToggleDropdown } from '@/components/theme-toggle';
 
-export default function HeaderAuth() {
-  const { userId, email, profile, loading } = useClaims();
+export default async function HeaderAuth() {
+  const supabase = await createClient();
+  const claims = await getClaims(supabase);
 
-  if (loading) {
-    return (
-      <div className="flex items-center gap-2 mr-2">
-        <ThemeToggleDropdown />
-      </div>
-    );
-  }
-
-  if (!userId) {
+  if (!claims) {
     return (
       <div className="mr-2">
         <ThemeToggleDropdown />
       </div>
     );
   }
+
+  const { userId, email, profile } = claims;
 
   return (
     <div className="flex items-center gap-2 mr-2">


joeseos:
@joeseos Reviewed against the codebase. The root-cause diagnosis is correct and well-supported by the code (`hooks/use-claims.ts` only reacts to client-driven `onAuthStateChange`/`SIGNED_IN` events, which never fire when sign-in happens via a server action — confirmed the existing `revalidatePath('/', 'layout')` calls in `app/actions/auth.ts:132,207` that this fix now leverages). One significant side effect needs attention before merging, though.

## Findings

### 🟠 Medium-High — forces the entire app into dynamic (SSR-per-request) rendering
`components/header-auth.tsx:7-9`
```ts
export default async function HeaderAuth() {
  const supabase = await createClient();   // utils/supabase/server.ts:8 calls cookies()
  const claims = await getClaims(supabase);
```
`cookies()` is a Next.js Dynamic API. `HeaderAuth` is rendered unconditionally at `app/layout.tsx:157`, the single root layout shared by every route in the app. There's no PPR/`cacheComponents`/`dynamicIO` enabled on `main` (that's still only trialed in the unmerged #1843, "Main stays on the experimental.useCache shim"), so a Dynamic API used anywhere in a layout opts the *entire* route tree under it out of static rendering.

Concretely, this contradicts the explicit design of `app/unsubscribe/page.tsx:4-9`:
```ts
// This page is statically generated (ISR), never server-rendered per request: ...
export const revalidate = 86400;
```
and removes static/CDN caching for the marketing pages that currently have no `dynamic`/`revalidate` export and are eligible for static generation (`about`, `contact`, `privacy-policy`, `terms`, `merch`, `join-the-team`). Every page load would now hit SSR + a Supabase auth check server-side instead of being served from cache.

Suggest verifying with `next build` before merging — check the route legend (○ static / ◐ ISR / λ dynamic) for those routes pre- and post-change. If they do flip to dynamic, options worth considering:
- Wrap `<HeaderAuth />` in `<Suspense>` — but note this only restores partial prerendering once PPR/`cacheComponents` is actually enabled on `main`, not with the current config.
- Keep the auth-gated header client-rendered (avoiding the layout-wide `cookies()` call) but fix the staleness a different way — e.g. dispatch a custom event or call `router.refresh()`/force a `useClaims` re-check right after the server action's redirect lands, so the existing client subscription re-syncs without making the whole layout dynamic.
- Scope the server-side claims read to a smaller, isolated slot if the app adopts PPR later (tracked in #1843).

### 🟢 Low — no loading fallback while the auth check resolves
`components/header-auth.tsx:7`

The previous client version painted the header immediately (`ThemeToggleDropdown` only) while `loading` was true, avoiding blocking the shell. Now, without a `<Suspense>` boundary, the initial HTML response is blocked on `getClaims()` resolving — a smaller version of the same issue as the finding above; wrapping in `Suspense` with a fallback matching the old loading state would restore the instant shell paint (once combined with a PPR-safe setup, or just to keep TTFB predictable even under full dynamic rendering).

The auth-fix approach itself (deriving header state on the server, relying on the existing `revalidatePath('/', 'layout')`) is sound and matches the documented Supabase SSR pattern. The blocking concern is the likely loss of static/ISR rendering app-wide from routing every page through a cookie-reading root layout component — worth confirming with a build check and deciding consciously whether that trade-off is intended before merging, since it silently undoes the unsubscribe page's ISR design and the marketing pages' static eligibility.


---

PR #1885, opened 2026-07-22
Cards in card view of the gang page have their height row set to the …

Djidiouf:
…tallest card in the row according to its content

### DIFF
--- a/components/gang/draggable-fighters.tsx
+++ b/components/gang/draggable-fighters.tsx
@@ -108,11 +108,21 @@ export function DraggableFighters({
 
   const sensors = useDndSensorsConfig();
 
-  // After a drag, the browser still synthesizes a `click` on whatever is under the pointer
-  // (often a *different* fighter card than the one that was dragged). Swallow that one click
-  // in the capture phase so no card's `<a>` navigates to a fighter page.
-  const suppressClickAfterDrag = () => {
+  // After a pointer drag, the browser still synthesizes a `click` on whatever is under the
+  // pointer (often a *different* fighter card than the one dragged). Swallow that one click
+  // in the capture phase so no card's `<a>` navigates away. Scoped to fighter-card links only
+  // so unrelated UI (nav, "Add Fighter", menus, etc.) stays clickable if the listener is still
+  // armed. Keyboard reorder never produces a trailing click, so we skip arming for it.
+  const suppressClickAfterDrag = (activatorEvent?: Event | null) => {
+    if (activatorEvent instanceof KeyboardEvent) return;
+
     const suppress = (event: Event) => {
+      const target = event.target;
+      if (!(target instanceof Element)) return;
+      // Only eat navigation clicks on fighter cards — leave everything else alone and keep
+      // listening until a card click arrives or the timeout clears.
+      if (!target.closest('.fighter-card-bg')?.closest('a')) return;
+
       event.preventDefault();
       event.stopPropagation();
       cleanup();
@@ -126,9 +136,9 @@ export function DraggableFighters({
   };
 
   const handleDragEnd = async (event: any) => {
-    // Always suppress — including same-position drops / cancelled-looking ends — because a
+    // Always consider suppress for pointer drags — including same-position drops — because a
     // real drag still produces a trailing click.
-    suppressClickAfterDrag();
+    suppressClickAfterDrag(event.activatorEvent);
 
     const { active, over } = event;
     
@@ -237,7 +247,7 @@ export function DraggableFighters({
       sensors={sensors}
       collisionDetection={closestCenter}
       onDragEnd={handleDragEnd}
-      onDragCancel={suppressClickAfterDrag}
+      onDragCancel={(event) => suppressClickAfterDrag(event.activatorEvent)}
     >
       <SortableContext
         items={sortedFighters.map(f => f.id)}
--- a/components/gang/draggable-fighters.tsx
+++ b/components/gang/draggable-fighters.tsx
@@ -108,7 +108,28 @@ export function DraggableFighters({
 
   const sensors = useDndSensorsConfig();
 
+  // After a drag, the browser still synthesizes a `click` on whatever is under the pointer
+  // (often a *different* fighter card than the one that was dragged). Swallow that one click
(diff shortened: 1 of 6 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed this against the current codebase (`components/gang/{draggable-fighters,fighter-card,fighter-card-action-menu,my-fighters,print-gang,sortable-fighter}.tsx`). `tsc --noEmit` is clean. No `CLAUDE.md` exists in this repo, so no project-convention doc to check against. Overall the content-driven height/stretch approach is well thought out (the comments explaining the remount/ResizeObserver pitfall and the border-box math are genuinely helpful), but there's one concrete correctness bug and one riskier UX behavior worth a look before merge.

### Findings

**1. (Medium) `COMPACT_CARD_VERTICAL_BORDER_PX` contradicts its own comment's math**
`components/gang/fighter-card.tsx:27-32`
```js
// `.fighter-card-bg` (cardClassName below) uses `border-2` (2px top + 2px bottom) and is
// `box-sizing: border-box`, so a `min-height` set on it gives its content-box 4px less room
// than the number itself ...
const COMPACT_CARD_VERTICAL_BORDER_PX = 8;
```
`border-2` is 2px top + 2px bottom = **4px** total, exactly as the comment states — but the constant is `8`. Since `compactMinHeightPx = Math.ceil(naturalHeight * scale) + COMPACT_CARD_VERTICAL_BORDER_PX` is added to every compact-view card's `min-height`, every card in 2/3/4-card view is currently ~4px taller than the content actually needs. Not visually catastrophic, but it's a verifiable off-by-4 vs. the stated reasoning — worth fixing to `4` (or updating the comment if `8` was intentional for some other reason not currently documented, e.g. a safety margin).

**2. (Medium) Post-drag click suppression is global and can silently eat unrelated clicks**
`components/gang/draggable-fighters.tsx:111-127, 139-141, 239`
`suppressClickAfterDrag()` adds a `document`-level, capture-phase `click` listener for up to 500ms after *any* `handleDragEnd`/`onDragCancel`, and swallows (`preventDefault`+`stopPropagation`) the very next click anywhere on the page, not just on the dragged card. This works well for the common case (the browser's synthetic post-drop click fires almost immediately and self-removes the listener), but two paths can leave the listener live for the full 500ms window with nothing to consume it:
- **Keyboard-driven reordering** (`KeyboardSensor` via `sortableKeyboardCoordinates`) never produces a trailing pointer `click`, so the listener just sits armed for 500ms after every keyboard drag/drop.
- `onDragCancel` (e.g., pressing Escape mid-drag) may not always be followed by a click on an interactive element either.

In both cases, if the user clicks *any* other control (a different card's menu, "Add Fighter", a nav link, etc.) within that window, the click is silently discarded — no error, no feedback, just a dead click. Consider scoping the suppression to `event.target` being within the dragged card (or the `over`/`active` element), or shortening the fallback window significantly, so it can't intercept genuine clicks on unrelated UI.

**3. (Low) Print default toggled outside this PR's stated scope**
`components/gang/print-gang.tsx:188`
`scaleCardsToContent` (the "Adjust Cards Height to Content" print checkbox) default flipped from `false` → `true`. This changes the default printed output for every user, which isn't mentioned in the PR title/description (scoped to the gang page's card view). Please confirm this default flip is intentional and not incidental fallout from testing.

### Verdict
**Changes requested** — mainly for #1 (quick, verifiable fix) and to get a decision on #2 before this ships; #3 is just a confirmation ask.


---

joeseos:
## Follow-up Review (new commit since last review)

@Djidiouf Reviewed just the new commit (`836d4f0`) added since my previous review comment — `components/gang/draggable-fighters.tsx` and `components/gang/fighter-card.tsx`.

### Previous findings

1. **`COMPACT_CARD_VERTICAL_BORDER_PX` off-by-4 vs. its own comment's math — Outstanding.**
   `components/gang/fighter-card.tsx:27-34`
   The constant is still `8`, but a new comment was added instead of a code change: *"4px covers the border-box shortfall, and an extra 4px is a safety margin for sub-pixel rounding under `transform: scale(...)`."* That reasoning doesn't check out — the height is computed as `Math.ceil(compactNaturalHeight * compactScale) + COMPACT_CARD_VERTICAL_BORDER_PX` (`fighter-card.tsx:242-243`), and `Math.ceil` already eliminates sub-pixel error (worst case <1px). A 4px "rounding safety margin" on top of a ceiling operation is just padding, not rounding correction. Every compact-view card is still rendering ~4px taller than the content needs. Low visual impact, but the justification as written doesn't hold up — worth either dropping to `4` or documenting the real reason for the extra margin (if there is one, e.g. box-shadow/border rendering quirks in specific browsers).

2. **Global post-drag click suppression — Resolved.**
   `components/gang/draggable-fighters.tsx:111-141, 247`
   Good fix. `suppressClickAfterDrag` now takes `activatorEvent` and bails immediately when it's a `KeyboardEvent`, so keyboard reorders never arm the listener at all. The capture-phase handler now also checks `target.closest('.fighter-card-bg')?.closest('a')` before swallowing a click, so only genuine fighter-card navigation clicks get eaten — everything else (menus, nav, "Add Fighter") stays clickable even if the 500ms window is still open. Confirmed the DOM structure supports this selector (`fighter-card.tsx:906-908` wraps the `.fighter-card-bg` div in an `<a>`). No regressions spotted — null/non-`Element` targets are guarded, and both `handleDragEnd` and `onDragCancel` pass the activator event through consistently.

3. **Print default flip in `print-gang.tsx` — Still outstanding, no response.**
   `components/gang/print-gang.tsx:188`
   This file wasn't touched by the new commit, so the ask from the previous review still stands: please confirm whether flipping `scaleCardsToContent`'s default to `true` (changing default printed output for every user) was intentional, since it's outside this PR's stated scope (gang page card view).

### New issues in this commit
None found. The click-suppression rework is well-scoped and doesn't introduce new bugs or edge cases.

### Verdict
**Approve, non-blocking** — the one correctness-adjacent item (#2) is fixed. #1 is a minor, low-impact discrepancy (optional to tighten). #3 just needs a yes/no confirmation before merge; happy to defer to your judgment on it.

---

PR #1886, opened 2026-07-22
Raise battle-log report limit to 4096 via Discord description embed

joeseos:
The battle-log "Report" was capped at 1024 characters because it was sent
to Discord as a single embed field (Discord limits field values to 1024).

Send the report in a second embed's description instead (4096-char limit),
rendered below the battle-info embed so the report stays last. A defensive
clamp trims the description with an ellipsis only if the combined payload
would exceed Discord's 6000-char total-embed cap, so the post can never be
rejected. The footer/timestamp move onto whichever embed is last so they
stay at the very bottom of the post.

Raise reportCharLimit from 1024 to 4096 in both battle-report modals.

Claude-Session: https://claude.ai/code/session_016Ttt9cmYSxuYmreR8fQLcF

### DIFF
--- a/components/battle-session/complete-battle-modal.tsx
+++ b/components/battle-session/complete-battle-modal.tsx
@@ -22,7 +22,7 @@ import {
 } from '@/utils/battle-winners';
 import { useWinnerSelection } from '@/utils/hooks/use-winner-selection';
 
-const reportCharLimit = 1024;
+const reportCharLimit = 4096;
 
 interface CompleteBattleTerritory {
   id: string;
--- a/components/campaigns/[id]/campaign-battle-log-modal.tsx
+++ b/components/campaigns/[id]/campaign-battle-log-modal.tsx
@@ -33,7 +33,7 @@ interface CampaignBattleLogModalProps {
   userRole?: 'OWNER' | 'ARBITRATOR' | 'MEMBER';
 }
 
-const reportCharLimit = 1024;
+const reportCharLimit = 4096;
 
 type GangRole = 'none' | 'attacker' | 'defender';
 
--- a/supabase/edge-functions/discord-campaign-bot/index.ts
+++ b/supabase/edge-functions/discord-campaign-bot/index.ts
@@ -194,17 +194,50 @@ Deno.serve(async (req) => {
       fields.push({ name: "📋 Scenario", value: battle.scenario, inline: true });
     }
 
-    if (battle.note) {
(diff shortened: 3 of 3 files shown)


joeseos:
## Code Review

@joeseos Reviewed this against the current codebase (`complete-battle-modal.tsx`, `campaign-battle-log-modal.tsx`, `supabase/edge-functions/discord-campaign-bot/index.ts`). Small, well-scoped fix — the embed-field 1024 cap was the right root cause to target, and moving the report into its own embed description is the correct fix.

Checked and confirmed clean:
- The 6000-char total-embed budget math (`infoChars` = info title + all field name/value lengths + report title + footer text, across both embeds) is correct — timestamps don't count toward Discord's total, and this correctly excludes them.
- `budget - 1` + 1 ellipsis char keeps the truncated description exactly at `budget`, satisfying both the 4096 description cap and the derived 6000 total cap.
- `Math.max(0, budget - 1)` guards the pathological case where fields alone already eat the whole 6000 budget (produces just `"…"` rather than a negative-length slice/crash).
- No other hardcoded `1024` reference or server-side length check was left stale — both client displays (`{notes.length}/{reportCharLimit}` + `maxLength`) were updated together, and the `note` DB column isn't length-constrained, so raising the client limit to 4096 doesn't create a mismatch anywhere else.
- No leftover reference to the old single `embed` variable after the `embeds` array refactor.

### Findings

**1. (Low) Truncation can split a UTF-16 surrogate pair, garbling the last character**
`supabase/edge-functions/discord-campaign-bot/index.ts` (the `description.slice(0, Math.max(0, budget - 1))` line)
`String.prototype.slice` cuts on UTF-16 code units. If a report ends with an astral-plane character right at the cut point (e.g. an emoji — plausible in a gaming community's battle reports), the slice can land in the middle of a surrogate pair, leaving a lone high surrogate. That gets sent as UTF-8 over the wire and Discord/Deno will render it as a replacement character (�) rather than a clean truncation. Not a crash and only affects reports that actually hit the truncation path, so non-blocking — but worth a defensive tweak, e.g. drop a trailing unpaired high surrogate before appending `"…"`:
```ts
if (description.length > budget) {
  let cut = description.slice(0, Math.max(0, budget - 1));
  if (cut.length > 0 && /[\uD800-\uDBFF]$/.test(cut)) cut = cut.slice(0, -1);
  description = cut.trimEnd() + "…";
}
```

### Verdict
**Approve** — the core logic is sound; the surrogate-pair edge case is minor and optional to address before merge.


---

PR #1889, opened 2026-07-23
chore(supabase): migrate local db setup to native CLI and add seed data

FlorianWeissDev:
I've created a `seed.sql` based on the old seed data (not functional anymore). It is possible to create gangs and add some basic equipment as a starting point. The seed could be extended in the future based on development needs.

I also used to opportunity to look into supabase cli native features and replaced the bash script. This simplifies the process because a simple `supabase start` is enough. Only issue are still the postgres metatags in the schema snapshot. This would be my main question, is it fine adjusting the snapshot workflow to remove these tags (as well as the public schema line)? Alternatively I could put the code in a small script instead and wrap it with `supabase start` into a npm script like `npm run db:local`.

---

- Remove scripts/setup-local-db.sh in favor of native supabase db reset / supabase start
- Add supabase/seed.sql populating Necromunda lookup tables, 25 fighter types across 5 gangs, equipment availability, and weapon profiles
- Configure db.seed.sql_paths in config.toml to run private schema, functions, snapshot, grants, triggers, and seed data
- Update supabase_schema_snapshot.yml to sanitize pg_dump meta commands for CLI parser compatibility
- Add SET check_function_bodies = off; to private_schema.sql
- Update README.md to document native CLI setup and reset behavior
- Updated config.toml major version to 17 to match the pg_dump snapshot referenced version


joeseos:
## Code Review

@FlorianWeissDev Thanks for tackling this — moving to the native `supabase db reset`/`seed` pipeline is a good simplification, and the seed data I spot-checked (`gang_types`, `fighter_types`, `equipment`, mapping tables) matches the real column names/types in `schema.public.sql` correctly. No `CLAUDE.md` exists in this repo to check conventions against.

### Findings

**1. (High) The setup is broken as committed — `schema.public.sql` still has the unstripped pg_dump meta-directives it's now loaded raw**
`supabase/config.toml` (`[db.seed].sql_paths`) references `./schema/schema.public.sql` directly, with no stripping step in between. I checked the currently committed file and it still contains exactly the lines your own PR description flags as incompatible with the CLI's Go SQL parser:
```
$ grep -nE '^\\restrict|^\\unrestrict|^CREATE SCHEMA public;' supabase/schema/schema.public.sql
5:\restrict vvVV2rXOAdVu5ZvuCmetl9K3GuHUhOYQEYFLkAliOgl6FNOjEAuuzGrT0MFj5f5
26:CREATE SCHEMA public;
12471:\unrestrict vvVV2rXOAdVu5ZvuCmetl9K3GuHUhOYQEYFLkAliOgl6FNOjEAuuzGrT0MFj5f5
```
The deleted `setup-local-db.sh` handled this by writing a stripped copy to a tempfile before loading it (`sed ... > "$STRIPPED"`); that step has no equivalent now. The `supabase_schema_snapshot.yml` sed fix only cleans the file the *next* time the daily cron regenerates it — it doesn't touch the file already in the repo. So right now, `supabase start` / `supabase db reset` will fail on this exact file until tomorrow's scheduled snapshot run happens to overwrite it. This is the same open question you raised in the PR description ("is it fine adjusting the snapshot workflow to remove these tags") — I'd say yes, but it also needs the *currently committed* `schema.public.sql` cleaned as part of this PR (or `sql_paths` needs to point at something other than the raw dump), otherwise the new "just run `supabase start`" workflow doesn't actually work today.

**2. (Medium) Stale references to the deleted script left in two other docs**
- `supabase/config.toml:10` — the `[db.migrations]` comment block (right next to this PR's own diff) still says: `# ... New local databases are built from the schema snapshot with # \`./scripts/setup-local-db.sh\` instead (see the repo README).`
- `supabase/README.md:5-7, 18-19` — still tells contributors to run `./scripts/setup-local-db.sh` and has a table listing it as what applies the private schema/trigger locally.

Since the script is deleted and the root `README.md` was updated, these two should be updated to match (native `supabase start`/`db reset` + `[db.seed]`), or a new contributor following `supabase/README.md` will run a command that no longer exists.

### Minor / non-blocking
- `[db].major_version` bump 15 → 17 isn't mentioned in the PR description, but it's correct — the snapshot header confirms `-- Dumped from database version 17.6`, so this actually fixes a pre-existing local/prod version mismatch. Worth a one-line mention in the description for reviewer context, but not a problem.

### Verdict
**Changes requested** — mainly #1, since it means the primary promise of this PR ("a simple `supabase start` is enough") doesn't hold yet for anyone pulling `main` today. #2 is a quick doc-consistency fix.


---

joeseos:
## Follow-up Review — @FlorianWeissDev

Reviewed only the two new commits since the last review (`39fb7fd` and `2d3f44f`), not the full PR diff.

### Previous findings

**1. (High) `schema.public.sql` still had unstripped pg_dump meta-directives — RESOLVED**
`39fb7fd` removes the `\restrict`/`\unrestrict` lines and `CREATE SCHEMA public;` from the committed `supabase/schema/schema.public.sql`. Verified no backslash meta-commands or `CREATE SCHEMA` lines remain in the file, and the manual edit matches the `sed -E '/^\\restrict/d; /^\\unrestrict/d; /^CREATE SCHEMA public;$/d'` pattern already in `supabase_schema_snapshot.yml`, so future daily snapshots will stay consistent with this one-off fix. `supabase start`/`supabase db reset` should now actually work as advertised.

**2. (Medium) Stale references to the deleted `setup-local-db.sh` script — RESOLVED**
`2d3f44f` rewrites the relevant sections of `README.md`, `supabase/README.md`, and the `[db.migrations]` comment in `supabase/config.toml` to describe the native `supabase start`/`db reset` + `[db.seed].sql_paths` flow instead. I checked the full text of both READMEs and `config.toml` for any remaining mentions of the deleted script — none found.

### New findings
None. The `[db.seed].sql_paths` order in `config.toml` (private schema → functions → snapshot → grants → triggers → seed) matches the order described in both the updated README table and the PR description, and no other files reference the removed script.

### Verdict: **Approve**


---

joeseos:
Nice one @FlorianWeissDev

PR #1890, opened 2026-07-23
Add gang-exclusive equipment flag to the Trading Post

joeseos:
## Problem

Trading-post visibility was gated solely on trading-post-**type** membership: an item shows in a gang's Trading Post if it's linked (via `trading_post_equipment`) to the `trading_post_type_id` that the gang's gang type points at. Because **multiple gang types share the same trading-post type**, equipment attached to a shared post appeared for all of them, with no way to restrict an item to a specific gang type.

(Equipment used to carry an `equipment.faction` restriction, dropped in `20260413100000_drop_faction_and_trading_post_category.sql` with no replacement.)

## Solution

Add an **"available only to this gang"** allow-list flag on the existing per-gang `equipment_availability` rows. When an item has one or more flagged rows, the Trading Post shows it **only** to gangs whose gang type / origin / variant matches a flagged row. Items with no flagged rows are unaffected. Multiple gang types are supported by flagging several rows (available to gang A **and** gang B → two flagged rows).

Modelling the flag on `equipment_availability` (rather than a boolean on `equipment`) keeps allow-list entries distinct from ordinary availability overrides — an item can be Rare-for-one-gang without being restricted, while being exclusive to another.

## Changes

- **`supabase/migrations/20260723120000_add_exclusive_to_equipment_availability.sql`** (new) — adds `equipment_availability.exclusive boolean NOT NULL DEFAULT false`.
- **`supabase/functions/get_equipment_detailed_data.sql`** — a single allow-list `WHERE` predicate, reusing the existing `ea` / `ea_origin` / `ea_var` joins plus one `NOT EXISTS` check. No new CTE or join.
- **`components/admin/admin-edit-equipment.tsx`** — an "Available only to this gang" checkbox in the existing "Availability per Gang" dialog; hydrated on load, shown in the row pill, and sent in the PATCH.
- **`app/api/admin/equipment/route.ts`** — `exclusive` threaded through the GET select and the POST/PATCH insert maps.

## Deployment note

Per `supabase/README.md`, RPC function files are **not** auto-synced. The migration adds the column, but the `get_equipment_detailed_data` change must be applied to the database separately when deploying.

## Verification

- `tsc --noEmit` and `eslint` pass on the changed files.
- Loaded the full production schema snapshot + this migration into a throwaway local Postgres and exercised the RPC against real data:
  - No flagged rows → item visible to both gangs sharing the post.
  - Flagged exclusive to Gang A → visible to A, hidden from B.
  - Flagged for A **and** B → both see it.
  - `exclusive = false` → visible to both again (plain override, no restriction).

---
_Generated by [Claude Code](https://claude.ai/code/session_015xateR4V8DGbvNTgweSfTg)_

### DIFF
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from 'next/server'
 import { createClient } from "@/utils/supabase/server";
 import { checkAdmin } from "@/utils/auth";
-import { WeaponProfileInput } from "@/types/equipment";
+import { WeaponProfileInput, EquipmentAvailability, EquipmentOriginAvailability, EquipmentVariantAvailability } from "@/types/equipment";
 import { 
   FighterEffectType, 
   FighterEffectTypeModifier, 
@@ -16,22 +16,6 @@ interface FighterTypeEquipment {
 interface GangAdjustedCost {
   gang_type_id: string;
   adjusted_cost: number;
-}
-
-interface EquipmentAvailability {
-  gang_type_id: string;
-  availability: string;
-  exclusive?: boolean;
-}
-
-interface EquipmentOriginAvailability {
-  gang_origin_id: string;
-  availability: string;
-}
-
-interface EquipmentVariantAvailability {
-  gang_variant_id: string;
-  availability: string;
 }
 
 export async function GET(request: Request) {
@@ -551,7 +535,7 @@ export async function POST(request: Request) {
 
     // Handle equipment availabilities if provided
     if (equipment_availabilities && equipment_availabilities.length > 0) {
-      const availabilityRecords = equipment_availabilities.map((avail: EquipmentAvailability) => ({
+      const availabilityRecords = equipment_availabilities.map((avail: Pick<EquipmentAvailability, 'gang_type_id' | 'availability' | 'exclusive'>) => ({
         equipment_id: equipment.id,
         gang_type_id: avail.gang_type_id,
         availability: avail.availability.trimEnd(),
@@ -1024,7 +1008,7 @@ export async function PATCH(request: Request) {
 
       // If there are new availabilities to add
       if (Array.isArray(equipment_availabilities) && equipment_availabilities.length > 0) {
-        const availabilityRecords = equipment_availabilities.map((avail: EquipmentAvailability) => ({
+        const availabilityRecords = equipment_availabilities.map((avail: Pick<EquipmentAvailability, 'gang_type_id' | 'availability' | 'exclusive'>) => ({
           equipment_id: id,
           gang_type_id: avail.gang_type_id,
           availability: avail.availability.trimEnd(),
@@ -1061,7 +1045,7 @@ export async function PATCH(request: Request) {
 
       // If there are new origin availabilities to add
       if (Array.isArray(equipment_origin_availabilities) && equipment_origin_availabilities.length > 0) {
-        const originAvailabilityRecords = equipment_origin_availabilities.map((avail: any) => ({
+        const originAvailabilityRecords = equipment_origin_availabilities.map((avail: Pick<EquipmentOriginAvailability, 'gang_origin_id' | 'availability'>) => ({
           equipment_id: id,
           gang_origin_id: avail.gang_origin_id,
           availability: avail.availability.trimEnd(),
@@ -1097,7 +1081,7 @@ export async function PATCH(request: Request) {
 
       // If there are new variant availabilities to add
       if (Array.isArray(equipment_variant_availabilities) && equipment_variant_availabilities.length > 0) {
-        const variantAvailabilityRecords = equipment_variant_availabilities.map((avail: EquipmentVariantAvailability) => ({
+        const variantAvailabilityRecords = equipment_variant_availabilities.map((avail: Pick<EquipmentVariantAvailability, 'gang_variant_id' | 'availability'>) => ({
           equipment_id: id,
           gang_variant_id: avail.gang_variant_id,
           availability: avail.availability.trimEnd(),
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -183,7 +183,7 @@ export async function GET(request: Request) {
 
       // Format the availabilities with null check
       interface AvailabilityData {
-        availability: string;
+        availability: string | null;
         gang_type_id: string | null;
         exclusive?: boolean;
       }
@@ -533,7 +533,7 @@ export async function POST(request: Request) {
       const availabilityRecords = equipment_availabilities.map((avail: Pick<EquipmentAvailability, 'gang_type_id' | 'availability' | 'exclusive'>) => ({
         equipment_id: equipment.id,
         gang_type_id: avail.gang_type_id,
-        availability: avail.availability.trimEnd(),
+        availability: avail.availability ? avail.availability.trimEnd() : null,
         exclusive: avail.exclusive ?? false
       }));
 
@@ -1006,7 +1006,7 @@ export async function PATCH(request: Request) {
         const availabilityRecords = equipment_availabilities.map((avail: Pick<EquipmentAvailability, 'gang_type_id' | 'availability' | 'exclusive'>) => ({
           equipment_id: id,
           gang_type_id: avail.gang_type_id,
-          availability: avail.availability.trimEnd(),
+          availability: avail.availability ? avail.availability.trimEnd() : null,
           gang_origin_id: null,
           exclusive: avail.exclusive ?? false
         }));
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from 'next/server'
 import { createClient } from "@/utils/supabase/server";
 import { checkAdmin } from "@/utils/auth";
-import { WeaponProfileInput, EquipmentAvailability, EquipmentOriginAvailability, EquipmentVariantAvailability } from "@/types/equipment";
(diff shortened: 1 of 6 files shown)


joeseos:
## Code Review

@joeseos Reviewed against the current codebase, including the full `get_equipment_detailed_data` function (not just the diff hunk) to check how the new predicate interacts with the rest of the query. The TS refactor (hoisting the duplicated `GangAdjustedCost`/`EquipmentAvailability`/etc. interfaces into `types/equipment.ts` and replacing `any` with `Pick<...>`) is a clean improvement, and the migration is a safe additive column. No `CLAUDE.md` exists in this repo to check conventions against.

### Findings

**1. (High) The exclusive filter isn't scoped to Trading Post visibility — it also strips equipment from unrelated gangs' fighter equipment lists**
`supabase/functions/get_equipment_detailed_data.sql` (the new `AND (NOT EXISTS (...) OR COALESCE(ea.exclusive, false) OR ...)` clause)

This new predicate is ANDed onto the *entire* WHERE clause, after the existing `($4 IS NOT NULL AND $5 IS NULL AND ftl_flag.is_fighter_list = $4)` branch — i.e. it applies uniformly whether the caller is asking "is this on the fighter's list?" (`$4`) or "is this in the Trading Post?" (`$5`), not just the Trading Post path.

`fighter_type_equipment.gang_type_id` is nullable, and a NULL row means "available to this fighter type regardless of gang type" (the join already handles this: `fte.gang_type_id IS NULL OR fte.gang_type_id = $1`). So the same `equipment_id` can legitimately sit on a fighter type's default/option list across *multiple, unrelated gang types* via a single gang-type-agnostic `fighter_type_equipment` row.

If an admin now flags that equipment "available only to Gang A" via the Trading Post dialog (the UI copy literally says *"Limits this item's **Trading Post** visibility..."*), every other gang type that has that fighter type option will silently lose it from their fighter's equipment list too — not just from Trading Post shopping. That's a real behavior change beyond what the feature promises, and it's not covered by the PR's verification notes (which only exercised the RPC's Trading-Post-only path).

Suggest scoping the new predicate to the Trading Post branch only, e.g. folding it into the existing `$5`-gated conditions (or `OR ftl_flag.is_fighter_list = true` alongside the allow-list check) so fighter-list membership is never affected by a Trading Post exclusivity flag.

**2. (Low) SQL comment implies origin/variant exclusivity that the UI can't actually create**
`supabase/functions/get_equipment_detailed_data.sql` comment says gangs are matched "whose type / **origin / variant**" match a flagged row, and the predicate does check `ea_origin.exclusive` / `ea_var.exclusive`. But neither `EquipmentOriginAvailability` nor `EquipmentVariantAvailability` (in `types/equipment.ts`) gained an `exclusive` field, and neither the origin nor variant PATCH handlers in `app/api/admin/equipment/route.ts` thread one through — only the gang-type dialog in `admin-edit-equipment.tsx` got the checkbox. So today only gang-*type* exclusivity is reachable from the admin UI; the origin/variant branches of the SQL predicate are currently dead code (harmless, just worth noting so it doesn't read as already-shipped functionality).

**3. (Low, non-blocking) "Available only to this gang" isn't offered on equipment creation**
`components/admin/admin-create-equipment.tsx` had its duplicate interfaces cleaned up but didn't gain the exclusivity checkbox that `admin-edit-equipment.tsx` got — admins can only set it after creating the item and editing it. Probably fine if that's the intended workflow, just flagging in case it was meant to be on both.

### Verdict
**Changes requested** — #1 is a real correctness issue that could unexpectedly break fighter equipment options for gangs never intended to be affected; the rest are minor/non-blocking.


---

joeseos:
Thanks for the thorough trace — all three points accepted and addressed in b74dcb9.

**#1 (High) — fixed.** You're right: the predicate was ANDed onto the whole `WHERE`, so it also gated the "Fighter's list only" branch and stripped exclusive items from other gangs' fighter lists. Exclusivity is now folded into an **effective trading-post access** (`tp_eff` lateral) that is used only where the query decides Trading Post visibility. The fighter's-list branch is untouched, so an item assigned to a fighter type still appears on that fighter's list — it just reports `equipment_tradingpost = false` when the viewing gang isn't on the allow-list. This now matches the UI copy ("Trading Post visibility").

**#2 (Low) — fixed.** The allow-list is now scoped to gang-type rows (`ea.exclusive` + `xa.gang_type_id IS NOT NULL`), and the dead `ea_origin.exclusive` / `ea_var.exclusive` branches are removed.

**#3 (Low) — done.** Added the "Available only to this gang" checkbox to the create-equipment modal for parity with the edit modal.

Verified against a throwaway Postgres loaded with the full schema snapshot + this migration:

| Scenario | Result |
|---|---|
| Trading Post, allowed gang | shows, `equipment_tradingpost=t` |
| Trading Post, non-allowed gang | hidden |
| **Fighter list, non-allowed gang** (item on its fighter type via a `NULL`-gang_type `fighter_type_equipment` row) | **still shows**, `equipment_tradingpost=f` |
| exclusive cleared | visible to all again |

---

joeseos:
## Follow-up Review — @joeseos

Reviewed only the two commits since the last review: `b74dcb9` (the fix) and `f1efbd6` (a further simplification of that same fix, pushed after the "addressed" reply comment).

### Previous findings

**1. (High) Exclusive filter leaking into fighter-list visibility — RESOLVED.**
Traced the final state (`f1efbd6`) rather than just the intermediate `b74dcb9` lateral-join version, since it superseded it. The allow-list check is now inside the `tp_summary` CTE's `WHERE` (filtering `tp_access` rows before `GROUP BY equipment_id`), gated on `$1` (`gang_type_id`) — confirmed via the function signature comment. `ftl_flag.is_fighter_list` (the fighter's-list branch) never references `tp_summary`/`tp`, so it's untouched. The `$4`/`$5` filter branches and the output column both now read `COALESCE(tp.has_access, false)`, so an exclusive item still surfaces on a fighter's equipment list for any gang, and is only excluded from `equipment_tradingpost`/the Trading-post-only filter branch for gangs not on the allow-list. This matches the UI copy ("Trading Post visibility") and the PR's own re-verification table.

**2. (Low) Dead origin/variant exclusive branches — RESOLVED.** Grepped the final SQL for `exclusive` — it now only appears in the `tp_summary` CTE, scoped to `xa.gang_type_id IS NOT NULL` / `xa.gang_type_id = $1`. No remaining `ea_origin.exclusive`/`ea_var.exclusive` references anywhere in the file.

**3. (Low) No exclusivity checkbox on equipment creation — RESOLVED.** `admin-create-equipment.tsx` gained the same checkbox as the edit modal (state, hydration reset, row-pill display, and `exclusive` in the POST payload). Confirmed `app/api/admin/equipment/route.ts` already threaded `exclusive` through both the create (`availabilityRecords` map ~L533) and edit (~L1006) insert paths from the earlier commits, plus the GET select (L73) — so the field is now wired end-to-end on both flows, not just edit.

### New findings
None that block. One minor, non-blocking observation: relocating the allow-list check into `tp_summary` (per-row-of-`tp_access`, pre-`GROUP BY`) means the two `EXISTS`/`NOT EXISTS` subqueries can run more than once per `equipment_id` when an item matches multiple trading-post sources (own post + campaign-authorised + custom post), versus once per `equipment_id` in the prior lateral-join version. Given `equipment_availability` is keyed by `equipment_id` and these are indexed-lookup `EXISTS` checks, this is very unlikely to matter in practice — not worth blocking on, just flagging since the commit message frames this purely as a line-count simplification with "no behaviour change," which is true functionally but not quite true for query shape.

### Verdict: **Approve**

All three prior findings are resolved and re-verified against the actual final SQL (not just the intermediate commit). Nice catch scoping this correctly, and the follow-up simplification back to a single CTE-level change is a good outcome.


---

joeseos:
## Follow-up Review — @joeseos

Reviewed the one new commit since my last review: `1fd7ce3` ("Allow saving a gang row that only restricts (no availability override)"), which touches `app/api/admin/equipment/route.ts`, both admin equipment modals, and `types/equipment.ts`.

### What it does
Previously the "Availability per Gang" dialog required an availability letter to save a row at all, so you couldn't add a gang-restriction-only row (exclusive checked, no override). This relaxes the save gate to `combined || availExclusive` in both modals, widens `EquipmentAvailability.availability` to `string | null`, and makes `route.ts` insert `null` instead of calling `.trimEnd()` on it.

### Findings
None — I traced this one further than the diff itself to make sure the `null` doesn't leak somewhere unsafe:

- **DB column is already nullable.** `equipment_availability.availability` is `text` with no `NOT NULL` (confirmed in `schema.public.sql`), and this PR's own migration only adds the `exclusive` column — so inserting `null` here needed no schema change.
- **The "keeps normal availability" claim is real, not just asserted.** `get_equipment_detailed_data.sql`'s `availability` output column already does `COALESCE(cto.availability_override, (origin row), ea_var.availability, ea.availability, e.availability)` — so when `ea.availability` is `NULL` for an exclusive-only row, it falls through to the item's base `e.availability` exactly as the commit message describes. This existed before the commit; the fix is just correctly relying on it.
- **The two other `.trimEnd()` call sites route.ts still has unguarded** (origin/variant availability inserts, ~L1046/L1082) are fine — they're typed via `Pick<EquipmentOriginAvailability | EquipmentVariantAvailability, ...>`, which the commit correctly left as non-nullable `string`, since only the gang-type dialog's save gate was relaxed. No inconsistency between the type and what the UI can actually produce.
- **Pill display** (`[avail.availability ? ... : null, avail.exclusive ? 'Exclusive' : null].filter(Boolean).join(', ')`) correctly omits the availability segment when null, in both modals.
- **`parseAvailability`** (used to hydrate the picker) already handled `null`/`undefined` before this commit, so no change needed there.

Nothing else to flag — this is a small, self-contained, well-targeted fix with no loose ends.

### Verdict: **Approve**


---

PR #1891, opened 2026-07-23
fix(battle-session): match crew selection fighter ordering to gang ov…

FlorianWeissDev:
Fixes the fighter ordering in the Battle Session **Select Crew** modal (`CrewSelectionModal`) so fighters appear in the exact same order configured on the Gang Overview page (`/gang/[id]`).

### Original Issue
When opening the **Select Crew** modal in a battle session, available fighters appeared in a non-deterministic, seemingly random order. This occurred because `gangFightersList` was fetched directly from Supabase without sorting or applying the gang's `positioning` map (`gangs.positioning`).

### Solution & Changes Made
- **Positioning Utility (`utils/fighter-positioning.ts`)**: Added a single core `sortByPositioning` engine with strongly-typed wrappers (`sortFightersByPositioning` and `sortParticipantFightersByPositioning`) to sort fighters by position index with a fallback to alphabetical ordering. 
- **Participant Card Integration (`components/battle-session/participant-card.tsx`)**: Pre-sorted `gangFightersList` with `sortFightersByPositioning` before passing mapped options to `CrewSelectionModal`, and updated `sortedLocalFighters` to use `sortParticipantFightersByPositioning`.
- **Local Seed Data (`supabase/seed.sql`)**: Seeded House Cawdor `Sheenbird` exotic beast reference data to support testing exotic beasts in battle sessions locally.

---

## Testing & Verification

- **Manual Testing**: Tested manually in an example battle session with custom fighter ordering. Verified that fighters in the crew selection modal match the exact order on `/gang/[id]` and that exotic beasts remain correctly nested beneath their owner fighter.
- **TypeScript**: Verified compilation with `npx tsc --noEmit` (clean, 0 errors).
- **Database**: Verified local database seeding with `supabase db reset`.

---

## Thoughts

Kept all other code snippets handling ordering the same to stay within the ticket's scope. But there seem to be a couple of code spots that could reuse the new helper.

> Components like `crew-fighter-combobox-options.tsx`, `stash-tab.tsx`, and `print-gang.tsx` can all be updated to import `sortFightersByPositioning`, removing the duplicated logic and improving sort performance across the app ().

### DIFF
--- a/app/gang/[id]/battle-session/[sessionId]/page.tsx
+++ b/app/gang/[id]/battle-session/[sessionId]/page.tsx
@@ -5,6 +5,7 @@ import { getBattleSessionCached } from '@/app/lib/battle-sessions/get-battle-ses
 import { getGangFightersList, getGangPositioning, type GangFighter } from '@/app/lib/shared/gang-data';
 import { getCampaignTerritories } from '@/app/lib/campaigns/[id]/get-campaign-data';
 import { checkCampaignArbitrator } from '@/utils/user-permissions';
+import { sortFightersByPositioning } from '@/utils/fighter-positioning';
 import ActiveSession from '@/components/battle-session/active-session';
 import CompletedSession from '@/components/battle-session/completed-session';
 
@@ -56,8 +57,10 @@ export async function renderBattleSessionPage(sessionId: string, currentPath: st
   const gangFightersMap: Record<string, GangFighter[]> = {};
   const gangPositioningMap: Record<string, Record<string, any> | null> = {};
   uniqueGangIds.forEach((gId, i) => {
-    gangFightersMap[gId] = gangFighterLists[i];
-    gangPositioningMap[gId] = gangPositioningList[i];
+    const fighters = gangFighterLists[i];
+    const positioning = gangPositioningList[i];
+    gangFightersMap[gId] = sortFightersByPositioning(fighters, positioning);
+    gangPositioningMap[gId] = positioning;
   });
 
   return (
--- a/components/battle-session/participant-card.tsx
+++ b/components/battle-session/participant-card.tsx
@@ -2,6 +2,7 @@
 
 import { useState, useEffect, useMemo, useRef, type ReactNode } from 'react';
 import { createPortal } from 'react-dom';
+import { sortFightersByPositioning, sortParticipantFightersByPositioning } from '@/utils/fighter-positioning';
 
 import { useMutation } from '@tanstack/react-query';
 import type { GangFighter } from '@/app/lib/shared/gang-data';
@@ -1026,7 +1027,8 @@ export default function ParticipantCard({
 
   // Map expanded gang fighters (one entry per loadout) to crew modal format
   const gangFighters = useMemo(() => {
-    return gangFightersList.map((gf) => ({
+    const sortedList = sortFightersByPositioning(gangFightersList, positioning);
+    return sortedList.map((gf) => ({
       id: gf.id,
       fighter_name: gf.fighter_name,
       label: gf.label,
@@ -1044,7 +1046,7 @@ export default function ParticipantCard({
       owner_id: gf.owner_id,
       owner_name: gf.owner_name,
     }));
-  }, [gangFightersList]);
+  }, [gangFightersList, positioning]);
 
   const selectedFighterIds = new Set(localFighters.map((f) => f.fighter_id));
   const selectedFighters = new Map<string, string | undefined>(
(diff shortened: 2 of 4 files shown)


joeseos:
## Code Review

@FlorianWeissDev Traced this up through `participant-card.tsx` → `active-session.tsx` → `app/gang/[id]/battle-session/[sessionId]/page.tsx` to check where `gangFightersList`/`positioning` actually originate. No `CLAUDE.md` exists in this repo to check conventions against.

### Findings

**1. (Medium) This can be simplified by sorting once at the page level instead of adding a generic sort engine in a leaf component**
`app/gang/[id]/battle-session/[sessionId]/page.tsx:46-61` already fetches both `gangFighterLists` (typed `GangFighter[]`) and `gangPositioningList` in parallel via `Promise.all`, then zips them into `gangFightersMap`/`gangPositioningMap` before handing them to `<ActiveSession>` → `<ParticipantCard>`. Since both pieces of data are already in hand, already typed, and already being combined at that exact spot, sorting `gangFightersMap[gId]` there once (a plain `[...fighters].sort(...)` using the concrete `GangFighter` shape) would:
- avoid the new generic `sortByPositioning<T>` engine + `getId`/`getName` extractor abstraction in `utils/fighter-positioning.ts` entirely for this path (only `sortedLocalFighters` — sorting `localFighters`, which lives as client state in `ParticipantCard` — actually needs a client-side sort),
- remove the `useMemo`/dependency-array churn in `participant-card.tsx` (`gangFighters` recomputing on every `positioning` change),
- and fix the same ordering issue everywhere `gangFightersMap` is consumed (e.g. `complete-battle-modal.tsx` also receives the raw map), instead of only where this PR happened to touch it — which is exactly the gap the PR's own "Thoughts" section flags (`crew-fighter-combobox-options.tsx`, `stash-tab.tsx`, `print-gang.tsx` still need the same fix applied ad hoc later).

**2. (Medium) The new alphabetical tiebreak doesn't actually match Gang Overview's order — the thing this PR is supposed to replicate**
`utils/fighter-positioning.ts` — `sortByPositioning` falls back to `nameA.localeCompare(nameB)` for any two fighters that both land in the `Number.MAX_SAFE_INTEGER` bucket (i.e., fighters missing from the `positioning` map). But the Gang Overview page's own sort (`components/gang/my-fighters.tsx`'s `sortedFighters`) has **no such tiebreak** — ties there just fall through in original fetch order via `Array.prototype.sort`'s stability. `positioning` here is the raw `gangs.positioning` DB column (`getGangPositioning`), which can legitimately be missing entries — e.g. a fighter added since the gang owner last opened `/gang/[id]` (that's when `draggable-fighters.tsx` normalizes and persists positions back). So for any gang with an un-positioned fighter, this PR's sort and the Gang Overview page's sort will disagree — the exact case this PR sets out to fix. Suggest dropping the name tiebreak (or, if a tiebreak is wanted, matching it to whatever Gang Overview actually does, not inventing a new one).

**3. (Low, non-blocking) Scope creep in `supabase/seed.sql`**
The Sheenbird exotic-beast reference data is unrelated to the ordering fix — reasonable as "data to manually verify the fix with," but it's mixed into the same commit as the actual bug fix. Might be cleaner as a separate PR, especially since seed data changes are easy to lose track of in an otherwise-small fix PR.

### Verdict
**Changes requested** — mainly #2, since as written the fix doesn't fully deliver on its stated goal for un-positioned fighters; #1 is a simplification worth doing before this grows more duplicated call sites per the PR's own "Thoughts" section.


---

joeseos:
## Follow-up Review — @FlorianWeissDev

Reviewed only the one new commit since the last review: `31ec0d1` ("refactor(battle-session): simplify fighter positioning sort at page level").

### Previous findings

**1. (Medium) Sort should happen once at the page level instead of a leaf component — RESOLVED.**
`renderBattleSessionPage` (`app/gang/[id]/battle-session/[sessionId]/page.tsx:60-62`) now calls `sortFightersByPositioning(fighters, positioning)` once, right where `gangFightersMap`/`gangPositioningMap` are assembled, before either is handed to `<ActiveSession>`. Traced `complete-battle-modal.tsx` specifically (the other consumer the original finding called out): it reads `gangFightersMap[p.gang_id]` directly at L194 with no re-sort of its own, so it now gets correctly-ordered fighters for free. `participant-card.tsx`'s `gangFighters` memo no longer re-sorts (`gangFightersList.map(...)` directly) and dropped `positioning` from its dep array — the redundant client-side sort is gone, and the fix now covers every `gangFightersMap` consumer, not just the one this PR originally touched.

**2. (Medium) Invented name tiebreak didn't match Gang Overview's sort — RESOLVED.**
`sortByPositioning` no longer takes a `getName` extractor and no longer falls back to `localeCompare`; it now relies purely on `Array.prototype.sort`'s (spec-guaranteed since ES2019) stability to preserve original fetch order for unpositioned fighters, matching what `my-fighters.tsx` actually does. I checked that the "original fetch order" being preserved is meaningful here, not just an assumption: `getGangFightersList` (`app/lib/shared/gang-data.ts:911`) — the same function both the battle-session page and the Gang Overview page (`app/gang/[id]/page.tsx:74`) call for their base fighter list — has no `ORDER BY`/`.order()` on the base `fighters` query. So both pages start from the same unordered-but-consistent base query shape; this PR isn't introducing a new ordering assumption, it's aligning with the one that already existed on Gang Overview. Also confirmed no other call site in the repo calls the lower-level `sortByPositioning` directly with a 4th `getName` argument, so dropping that parameter doesn't break anything else.

**3. (Low, non-blocking) `supabase/seed.sql` scope creep — left as-is (as expected, it was non-blocking).**

### New findings
None. The generic constraint narrowing (`T extends { id: string }` instead of requiring `fighter_name`) is safe — verified via full-tree search that only the two wrapper functions in `utils/fighter-positioning.ts` call the underlying `sortByPositioning`.

### Verdict: **Approve**


---

PR #1898, opened 2026-07-25
Fix drag-and-drop of favourite gangs/campaigns on mobile

joeseos:
Problem
Long-pressing a favourite gang or campaign card on the home page showed the red “selected” border — so dnd-kit did activate the drag — but the browser then took over the gesture with its native link menu (“Open in New Tab”) and the reorder never completed. Desktop was fine.
The equivalent DnD on the gang page (reordering fighters) works on mobile using the same sensors (useDndSensorsConfig) and the same mount gate (useIsMounted). The difference is that the fighter card carries workarounds the home cards never got.
Changes
1. Cancel the long-press context menu on a favourite card. This is what actually stops the native link menu stealing the gesture — the same trick fighter-card-action-menu.tsx:160-166 already uses for its menu trigger. dnd-kit registers a contextmenu guard of its own, but only once a drag has activated, which is too late: the browser’s menu fires first. Gated on touch-primary devices so desktop right-click still opens the browser menu, and on the card being draggable so plain cards are untouched. The accompanying CSS (-webkit-touch-callout, user-select, tap highlight, touch-action) is belt-and-braces for iOS.
2. Stop swapping the card’s <Link> for a bare <div> when the drag activates, so the touch target stays stable for the whole gesture. Navigation is cancelled via onClick instead — Next’s Link skips navigating when the handler calls preventDefault() — so soft navigation and prefetch={false} are both preserved.
3. Swallow the click the browser synthesises after a drag. A drag ending where it started leaves a matching mousedown/mouseup, so the browser follows the card’s href: pressing and holding a favourite on a phone already navigated into the gang instead of reordering. dnd-kit stops that click propagating, which kills React’s onClick, but not the anchor’s default action. suppressClickAfterDrag moves from draggable-fighters.tsx into the shared use-dnd-sensors.ts, parameterised by selector, so all three drag surfaces share one copy.
4. Give a dragged card an opaque background. The card row has no background of its own — it inherits the look of the panel it sits on. Invisible at rest, but once dnd-kit lifts a card to z-index: 50 you read the cards underneath straight through it. It didn’t show on desktop because the pointer sits over the card while dragging, so hover:bg-muted paints it; on touch there’s no hover, so the dragged card stayed see-through the whole way.
The shared sensor config itself (including the 600 ms TouchSensor delay) and the useIsMounted hydration gate from #1785 are unchanged.
Verification
Driven in Chromium against the real components, with touch emulation and raw CDP touch events:
	•	Long-press context menu is cancelled on favourite cards, not on plain ones, and a desktop right-click is still allowed through.
	•	A same-position drop and a drag-and-return no longer navigate — both did before this PR.
	•	Touch long-press reorders favourite gangs and campaigns, with the right reorderFavourites payload and no stray navigation.
	•	Plain taps still navigate; the dragged card is opaque on touch (rgba(0,0,0,0) → rgb(255,255,255)).
	•	Mouse drag, plain clicks, the favourite star and keyboard reorder are unaffected.
	•	tsc --noEmit, eslint and next build all clean.
Caveat: the CSS half of change 1 (-webkit-touch-callout) is Safari-only and has no Chromium equivalent, so it can’t be proven by emulation. The contextmenu cancellation — the load-bearing part — is verified above. Worth a quick check on a real iPhone before merging.
Follow-up (not in this PR)
Dragging fighter cards on the gang page pops iOS’s text-selection magnifier mid-drag, because that card body never disabled text selection. Fixed separately on claude/ios-fighter-card-drag-loupe, branched off main so it can merge in any order.

### DIFF
--- a/components/gang/draggable-fighters.tsx
+++ b/components/gang/draggable-fighters.tsx
@@ -1,7 +1,7 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 import { DndContext, closestCenter } from '@dnd-kit/core';
 import { rectSortingStrategy, SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable';
-import { useDndSensorsConfig } from '@/hooks/use-dnd-sensors';
+import { useDndSensorsConfig, useSuppressClickAfterDrag } from '@/hooks/use-dnd-sensors';
 import { useIsMounted } from '@/hooks/use-is-mounted';
 import { MyFighters } from './my-fighters';
 import { FighterProps } from '@/types/fighter';
@@ -108,32 +108,7 @@ export function DraggableFighters({
 
   const sensors = useDndSensorsConfig();
 
-  // After a pointer drag, the browser still synthesizes a `click` on whatever is under the
-  // pointer (often a *different* fighter card than the one dragged). Swallow that one click
-  // in the capture phase so no card's `<a>` navigates away. Scoped to fighter-card links only
-  // so unrelated UI (nav, "Add Fighter", menus, etc.) stays clickable if the listener is still
-  // armed. Keyboard reorder never produces a trailing click, so we skip arming for it.
-  const suppressClickAfterDrag = (activatorEvent?: Event | null) => {
-    if (activatorEvent instanceof KeyboardEvent) return;
-
-    const suppress = (event: Event) => {
-      const target = event.target;
-      if (!(target instanceof Element)) return;
-      // Only eat navigation clicks on fighter cards — leave everything else alone and keep
-      // listening until a card click arrives or the timeout clears.
-      if (!target.closest('.fighter-card-bg')?.closest('a')) return;
-
-      event.preventDefault();
-      event.stopPropagation();
-      cleanup();
-    };
-    const cleanup = () => {
-      document.removeEventListener('click', suppress, true);
-      window.clearTimeout(timeoutId);
-    };
-    document.addEventListener('click', suppress, true);
-    const timeoutId = window.setTimeout(cleanup, 500);
-  };
+  const suppressClickAfterDrag = useSuppressClickAfterDrag('.fighter-card-bg');
 
   const handleDragEnd = async (event: any) => {
     // Always consider suppress for pointer drags — including same-position drops — because a
--- a/components/home/campaign-card.tsx
+++ b/components/home/campaign-card.tsx
@@ -9,6 +9,20 @@ import { CSS } from '@dnd-kit/utilities'
 import { FiStar } from 'react-icons/fi'
 import { AiFillStar } from 'react-icons/ai'
 
+// A long press over an `<a href>` makes the browser offer its own link menu ("Open in New
+// Tab"), which on iOS steals the gesture before dnd-kit's TouchSensor delay elapses. Cancelling
+// `contextmenu` is what stops it — the same trick fighter-card-action-menu.tsx uses for its menu
+// trigger. dnd-kit has a contextmenu guard of its own, but only once a drag has activated, which
+// is too late. The CSS is belt-and-braces for iOS; both properties inherit, so they cover the
+// `<Link>` inside. Only for favourites, so plain cards keep normal selection and link behaviour.
+const dragSurfaceStyle = {
+  WebkitTouchCallout: 'none',
+  WebkitUserSelect: 'none',
+  userSelect: 'none',
+  WebkitTapHighlightColor: 'transparent',
+  touchAction: 'manipulation',
+} as const;
+
 export interface CampaignCardProps {
   campaign: Campaign;
   onToggleFavourite: (campaignMemberId: string, isFavourite: boolean) => void;
@@ -19,6 +33,8 @@ export interface CampaignCardProps {
 }
 
 export function CampaignCardContent({ campaign, onToggleFavourite, dragListeners, dragAttributes, isDragging, disableLink = false }: CampaignCardProps) {
+  const isDraggable = Boolean(dragListeners);
+
   const innerContent = (
     <>
       <div className="relative w-[80px] md:w-20 h-[80px] md:h-20 mr-3 md:mr-4 shrink-0 flex items-center justify-center">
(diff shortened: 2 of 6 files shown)


joeseos:
@joeseos Reviewed this PR against the codebase (commit `b1667ed`).

## Summary
Fixes mobile drag-and-drop of favourite gang/campaign cards on the home page by porting three workarounds already proven on the fighter-card drag surface: cancelling the native long-press context menu on touch devices, keeping the card's `<Link>` mounted (never swapped for a bare `<div>`) during drag so the touch target stays stable, and swallowing the trailing synthetic click a same-position/return drag leaves behind. The click-suppression logic is extracted from `draggable-fighters.tsx` into a shared, selector-parameterised `useSuppressClickAfterDrag` hook in `use-dnd-sensors.ts`.

## Verification performed
- Checked out the branch and ran `npx tsc --noEmit` myself: clean (only pre-existing `tsconfig` deprecation notices, unrelated).
- Traced `useSuppressClickAfterDrag(cardSelector)`: identical logic to the original `draggable-fighters.tsx` implementation, just parameterised — `draggable-fighters.tsx`'s own call site (`.fighter-card-bg`) is unchanged behaviorally. `useCallback` keyed only on the (stable, literal) selector string, so adding it to `gangs-tab.tsx`/`campaigns-tab.tsx`'s `useCallback` deps doesn't cause needless re-creation.
- Confirmed `.home-favourite-card-link` is a new marker class applied directly to the anchor and used only as a JS `closest()` selector — no colliding CSS rule targets it, and it doesn't shadow the existing `.fighter-card-bg` selector/usages elsewhere (`fighter-card.tsx`, `print-gang.tsx`, `draggable-fighters.tsx`).
- Verified the `disableLink` control flow: the anchor is now always rendered, with `onClick` preventDefault + `tabIndex={-1}` + `aria-hidden` substituting for the old div-swap, and this closure is freshly bound each render so there's no stale-`disableLink` risk.
- Confirmed the two non-drag render paths (pre-hydration fallback list, and the plain non-favourite list) never pass `dragListeners`, so `isDraggable` is `false` there and none of the new CSS/contextmenu/class wiring applies to them — matches the PR's claim that plain cards are untouched.
- `onContextMenu` is gated by `window.matchMedia('(pointer: coarse)')`, executed only inside an event handler (never during render/SSR), so no hydration-safety concern despite `"use client"` already covering it.
- CI (`deploy-preview`) green; no prior reviews on the PR.

## Findings

- **Low (DRY, non-blocking)**: `gang-card.tsx` and `campaign-card.tsx` each define their own verbatim copy of `dragSurfaceStyle` (the 5-property inline style object) and the identical `onContextMenu` gating logic/comment (lines ~12–24 and ~95–99 in both files). Since the PR already centralizes the click-suppression logic into `use-dnd-sensors.ts` for exactly this kind of reuse, doing the same for `dragSurfaceStyle` + the contextmenu handler (e.g. a `useDragSurfaceProps(isDraggable)` hook returning `{ style, onContextMenu }`) would avoid a second place to update if the workaround ever needs tweaking. Not a regression — these two files were already twins before this PR — just a missed opportunity given the sibling extraction already done in this same diff.
- **Low (pre-existing, not introduced by this PR)**: `useSuppressClickAfterDrag`'s listener has a 500ms window and no guard against a second drag starting before a prior listener's `cleanup()` fires; back-to-back rapid drags in quick succession could theoretically stack two capture-phase listeners. This behavior is carried over unchanged from `draggable-fighters.tsx`, not new here — flagging only for awareness, not as something this PR needs to fix.

No correctness, security, or type-safety issues found.

## Checklist
- Correctness: ✅ matches the described fix; traced through both gang and campaign card variants
- Edge cases / null handling: ✅ non-draggable/static render paths unaffected; keyboard-driven reorder still skipped (no trailing click to suppress)
- Security: ✅ no new data flow; purely client-side gesture handling
- Performance: ✅ negligible — one extra `matchMedia` check per contextmenu event, capture-phase listener auto-cleaned after 500ms
- Readability/maintainability: ✅ well-commented, explains the *why* (iOS gesture theft) not just the what
- Project conventions: ✅ no CLAUDE.md in repo; follows the existing gang-card/campaign-card twin-file pattern and reuses the established fighter-card workaround
- DRY/YAGNI: ⚠️ see low-severity note above — good consolidation of the click-suppression hook, but `dragSurfaceStyle`/contextmenu logic duplicated across the two card files

## Verdict: **Approve**

---

joeseos:
## Follow-up Review — @joeseos

Reviewed the two commits since the last review: `82db0e9` (addresses the DRY finding) and `cd65a75` (a new change — opaque background for the dragged card).

### Previous findings

**1. (Low, DRY) `dragSurfaceStyle` + contextmenu gate duplicated between `gang-card.tsx`/`campaign-card.tsx` — RESOLVED.**
Both now import a shared `dragSurfaceProps(isDraggable)` from `use-dnd-sensors.ts` and spread it (`{...dragSurfaceProps(isDraggable)}`) in place of the old inline `style={...}`/`onContextMenu={...}` pair. Confirmed it's a 1:1 behavioral port — same style object, same `matchMedia('(pointer: coarse)')` gate, returns `{}` when not draggable (equivalent to the old `undefined`/`undefined`). Checked both files for any other `style`/`onContextMenu` prop on the same element that the spread could clobber — there isn't one, so this is a pure extraction with no behavior change, exactly as the commit message claims.

**2. (Low, pre-existing) `useSuppressClickAfterDrag`'s 500ms listener window / stacking on rapid drags — not addressed, and correctly so.** This was flagged as pre-existing (carried over from `draggable-fighters.tsx`, not introduced by this PR) and explicitly not something this PR needed to fix. No action expected or taken.

### New findings
None. `cd65a75` adds `bg-card` to the `isDragging` branch of both cards' className (alongside the existing `border-[3px] border-rose-700`). Checked:
- `bg-card` is an established, theme-aware utility already used throughout the codebase (backed by the `--card`/`--light-card` CSS variables in `globals.css`), not a one-off/typo class.
- The interaction with the existing unconditional `hover:bg-muted` is fine: `hover:bg-muted` has higher specificity (class + pseudo-class) than `bg-card` (single class), so on desktop — where the pointer sits over the card during drag and hover was already painting an opaque background per the PR's own description — `hover:bg-muted` still wins, unchanged. On touch, there's no `:hover` to compete, so `bg-card` is what actually fixes the see-through card. Both paths end up opaque either way, so there's no regression risk from the two classes coexisting.
- Diff is exactly the described 2-line change (one line per file), nothing else touched.

### Verdict: **Approve**


---

PR #1900, opened 2026-07-27
feat(fighter-types): allow linking default equipment accessories to

FlorianWeissDev:
## 📌 Context & Motivation

Predefined template fighter types (such as Bounty Hunters) configured in the Admin panel previously had weapon accessories (e.g., Infra-sight) rendered only as standalone wargear items, with no link to their parent weapon.

This PR enables linking default accessories directly to target weapons during template creation, ensuring that hired fighters automatically have their accessories attached to the correct weapon—matching the behavior and display of standard gangers.

---

## 🏗️ Architectural Rationale

- **Self-referencing FK on `fighter_defaults`**: added `target_fighter_default_id` referencing `fighter_defaults(id)`. Keeps template relationships normalized without extra join tables.
- **Runtime conversion via `fighter_effects`**: during fighter hiring (`add-fighter.ts`), the template link resolves to the newly-inserted `fighter_equipment` rows and is recorded as a `fighter_effects` record with `target_equipment_id` — the same mechanism the existing "attach accessory to weapon" purchase flow already uses, so the read-side card rendering (gang overview, fighter page) needed no new logic. Each `fighter_equipment` row is inserted with a client-generated id specifically so this resolution doesn't depend on Supabase/PostgREST returning bulk-insert rows in submission order.
- **Immediate hire-time display**: the fighter-creation response and the hiring modal's optimistic preview both needed to be taught to surface the same attachment (as `effect_names` on the weapon), so the link is visible right away instead of only after a page reload.
- **Preserving dual-display rules**: accessories remain listed under Wargear while also displaying as a modifier on the parent weapon row (in both the live card and the hiring modal preview) — this is intentional, not a duplicate.

---

## 🛠️ Summary of Changes

- **Database & Admin**
  Added `target_fighter_default_id` migration (`20260724215000_add_target_fighter_default_id.sql`) and exposed it from `get_fighter_types_with_cost`. Create/Edit Fighter Type now has a per-slot "Attach to" selector for Weapon Accessories, and default equipment (of any type) can be added multiple times — e.g. two Stub Guns, or the same accessory attached to two different weapons — with each duplicate weapon numbered in the list so admins can tell which one an accessory targets without counting.

- **Fighter Hiring Action** (`app/actions/add-fighter.ts`)
  Resolves each default accessory's target link to the newly-created weapon's `fighter_equipment` row and records it via `fighter_effects` (through the existing fixed-effect path, or a direct insert for accessories with no registered effect type). The fighter-creation response now carries the resulting attachment (`effect_names`) on the target weapon so it renders correctly without a reload. Also fixes two correctness issues found while building this: equipment `is_editable` was being hardcoded to `false` in the response instead of reflecting the real value, and a stale interface field that didn't match the underlying query.

- **Hire Preview Card** (`components/gang/fighter-add/FighterAddModal.tsx`, `utils/fighter-builder.ts`)
  The pre-hire optimistic preview now matches accessories to weapons using the real `fighter_defaults` id, and the server response is threaded through so the confirmed card matches the preview.

---

### Regression checks
- [ ] Hire a fighter type with no linked default equipment at all — fighters still create normally, no attachment artifacts.
- [ ] On an already-existing fighter, buy a Weapon Accessory through the normal purchase flow and attach it to a weapon — that path (unrelated to this PR, but touching shared display code) still works as before.
- [ ] Hire a fighter type whose default equipment includes an item that should be user-editable after creation (e.g. anything with a text/rules-editable slot) — it shows as editable immediately, not just after a reload.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -11,6 +11,15 @@ import {
   mapArchetypeSkillAccessToOverrides,
 } from '@/utils/archetypeEligibility';
 
+interface InsertedEquipment {
+  id: string;
+  equipment_id: string | null;
+  custom_equipment_id: string | null;
+  purchase_cost?: number;
+  equipment?: { equipment_name: string; equipment_type: string; equipment_category: string; equipment_category_id?: string; is_consumable?: boolean };
+  custom_equipment?: { equipment_name: string; equipment_type: string; equipment_category: string; is_consumable?: boolean };
+}
+
 interface SelectedEquipment {
   equipment_id: string;
   cost: number;
@@ -171,7 +180,8 @@ async function applyEffectsForEquipmentOptimized(
   fighterEquipmentId: string,
   fighterId: string,
   userId: string,
-  includeCreditIncrease: boolean = false
+  includeCreditIncrease: boolean = false,
+  targetEquipmentId?: string | null
 ): Promise<{ appliedEffects: any[], effectsCreditsIncrease: number }> {
   if (!effectTypes || effectTypes.length === 0) {
     return { appliedEffects: [], effectsCreditsIncrease: 0 };
@@ -187,6 +197,7 @@ async function applyEffectsForEquipmentOptimized(
       type_specific_data: effectType.type_specific_data,
       sort_order: effectType.sort_order ?? null,
       fighter_equipment_id: fighterEquipmentId,
+      target_equipment_id: targetEquipmentId || null,
       user_id: userId
     }));
 
@@ -546,8 +557,10 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     const { data: allDefaultsData } = await supabase
       .from('fighter_defaults')
       .select(`
+        id,
         skill_id,
         equipment_id,
+        target_fighter_default_id,
         custom_equipment_id,
         skills!skill_id(
           id,
@@ -578,12 +591,16 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
       .map(item => {
         if (item.equipment_id && item.equipment) {
           return {
+            default_id: item.id,
+            target_fighter_default_id: item.target_fighter_default_id || null,
             equipment_id: item.equipment_id,
             equipment: item.equipment,
             is_editable: (item.equipment as any)?.is_editable || false
           };
         } else if (item.custom_equipment_id && item.custom_equipment) {
           return {
+            default_id: item.id,
+            target_fighter_default_id: item.target_fighter_default_id || null,
             equipment_id: `custom_${item.custom_equipment_id}`,
             equipment: {
               id: `custom_${item.custom_equipment_id}`,
@@ -608,6 +625,12 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
       is_editable?: boolean;
     }> = [];
 
+    // Parallel array to track default metadata (default_id and target_fighter_default_id)
+    const equipmentMeta: Array<{
+      default_id?: string;
+      target_fighter_default_id?: string | null;
+    }> = [];
+
     // Track added equipment to prevent cross-source duplicates
     const addedEquipment = new Set<string>();
 
@@ -627,6 +650,10 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
           gang_id: params.gang_id,
           user_id: gangData.user_id,
           is_editable: defaultEquipment.is_editable || false
+        });
+        equipmentMeta.push({
+          default_id: defaultEquipment.default_id,
+          target_fighter_default_id: defaultEquipment.target_fighter_default_id || null
         });
       });
     }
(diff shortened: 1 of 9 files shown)


FlorianWeissDev:
This one should be pushed to test.mundamanager.com before deploying it live. I've tested it locally but it is also advised that one of the data folks tests the new UI on creation and changing existing fighter types. Important for this test might be to know, no existing types will automagically have any links as they are all manually defined.

**And we need to deploy the function change manually. Just like the docs point out.**

joeseos:
## Code Review

@FlorianWeissDev Reviewed the full diff (self-referencing `target_fighter_default_id` on `fighter_defaults`, the hire-time attachment resolution in `add-fighter.ts`, the admin create/edit UI, and the SQL/schema changes).

### What's good
- Reusing the existing `fighter_effects.target_equipment_id` mechanism (the same one the purchase-flow "attach accessory to weapon" path already uses) means the read side needed zero new logic — `gang-data.ts`'s `equipmentTargetingEffectNamesMap` already keys off `target_equipment_id || fighter_equipment_id`, so this slots straight into established display code.
- Client-generating `fighter_equipment.id`/`fighter_defaults.id` (via `randomUUID()`/`crypto.randomUUID()`) to correlate rows back to their source metadata is a good call — it sidesteps relying on PostgREST returning bulk-insert rows in submission order, which isn't guaranteed.
- The `effectsForThisEquipment.length > 0` vs. `else if (targetFighterEquipId)` branching in `add-fighter.ts` correctly avoids double-inserting an attachment record when an accessory has both pre-registered effect types *and* a target — only one of the two paths can fire per equipment item.
- Numbering weapon slots globally in the admin UI (`weaponIndexBySlotId`) so duplicate weapon names (two Stub Guns) are disambiguated in the "Attach to" dropdown is a nice touch, and removing a weapon correctly clears any accessory's `target_fighter_default_id` that pointed at it (`admin-create-fighter-type.tsx`/`admin-edit-fighter-type.tsx` remove handlers).
- Good type-safety cleanup in `add-fighter.ts` (new `InsertedEquipment` interface replacing scattered `as any`), and the `is_editable`/stale-interface-field fixes called out in the PR description check out against the actual `.select()` shape.

### Findings

**1. Medium (worth an explicit QA pass, not a blocker) — self-referencing bulk insert of `fighter_defaults`**
`app/api/admin/fighter-types/route.ts` (`PUT`/`POST`) inserts a fighter type's weapon rows *and* their accessory rows (with `target_fighter_default_id` pointing at a sibling row's client-generated `id`) in a **single** `.insert(equipmentDefaults)` call. This relies on Postgres queuing FK/RI checks as row-level AFTER triggers that fire at the end of the statement rather than immediately per-row — which is standard Postgres behavior for self-referencing bulk inserts (the classic employee/manager-id pattern), so it should work regardless of array order. But it's the one genuinely novel code path this PR introduces, and the PR's own testing notes only mention local testing without browser exercise and don't call out "save a fighter type with a weapon + an accessory attached to it in the same request" specifically. Worth confirming explicitly on the test.mundamanager.com pass already planned, since if this assumption were ever wrong it would hard-fail every attempt to link default equipment rather than degrade gracefully.

**2. Low/nit — dead fallback in `FighterAddModal.tsx`'s optimistic preview**
`components/gang/fighter-add/FighterAddModal.tsx:293,304`:
```ts
weapon_id: item.equipment_id || item.id,
...
wargear_id: item.equipment_id || item.id,
```
`item` here comes from `selectedType.default_equipment`, i.e. the `get_fighter_types_with_cost` SQL function's `default_equipment` jsonb (`supabase/functions/get_fighter_types_with_cost.sql`). That jsonb only ever includes `id` (the equipment id, `e.id`) and the new `default_id`/`target_fighter_default_id` — there's no `equipment_id` key in the object at all, so `item.equipment_id` is always `undefined` and this always evaluates to `item.id` (i.e. no behavioral change from before this PR). Harmless, but reads as if `equipment_id` were a real fallback source — worth simplifying back to just `item.id` or dropping the now-redundant `|| item.id`.

**3. Low/informational, pre-existing (not introduced by this PR)** — in `add-fighter.ts`'s `equipmentWithProfiles` mapping, `equipment_category: item.custom_equipment?.equipment_category || 'unknown'` drops the `item.equipment?.equipment_category` branch that existed before. This isn't a regression: the `fighter_equipment` insert's `.select()` for the `equipment!equipment_id(...)` relation only ever fetches `equipment_category_id`, never the text `equipment_category`, so that branch was already dead code returning `undefined` pre-PR. Just flagging it since `equipment_category` (the string, e.g. `"Weapon Accessories"`) is exactly the field the new admin UI uses to decide whether to show the "Attach to" selector — not a problem for this PR's own logic (which correlates by id, not category string), but a latent gap if a future fighter-creation response ever needs to show a real category for standard equipment.

### Verdict: **Approve**

Nothing above blocks merge — #1 is a "please verify" note on an assumption I'm fairly confident holds, and #2/#3 are non-functional nits. The core attachment-resolution logic, the dedup/multi-copy support, and the dual-display behavior (accessory in Wargear + modifier on the weapon row) all check out against the existing `target_equipment_id` convention used elsewhere in the codebase. Agreed with the PR description's own callouts: this should go to test.mundamanager.com first, and the `get_fighter_types_with_cost` function change needs the manual remote deploy step per repo convention.

---

PR #1908, opened 2026-07-30
Added edition support in admin page

joeseos:
PR to add support for editions in the admin page. A new edition selection component was added, types and API routes adjusted. The selection uses is_current to pre-select the current edition.

### DIFF
--- a/app/api/admin/campaign-types/route.ts
+++ b/app/api/admin/campaign-types/route.ts
@@ -13,7 +13,7 @@ export async function GET() {
 
     const { data: campaignTypes, error } = await supabase
       .from('campaign_types')
-      .select('id, campaign_type_name, image_url, trading_posts, campaign_type_resources(id, resource_name)')
+      .select('id, campaign_type_name, image_url, trading_posts, edition_id, campaign_type_resources(id, resource_name)')
       .order('campaign_type_name', { ascending: true });
 
     if (error) throw error;
@@ -37,7 +37,7 @@ export async function POST(request: Request) {
     }
 
     const body = await request.json();
-    const { campaign_type_name, image_url, trading_posts, resources } = body;
+    const { campaign_type_name, image_url, trading_posts, resources, edition_id } = body;
 
     if (!campaign_type_name?.trim()) {
       return NextResponse.json(
@@ -90,7 +90,8 @@ export async function POST(request: Request) {
       .insert([{
         campaign_type_name: campaign_type_name.trim(),
         image_url: image_url?.trim() || null,
-        trading_posts: trading_posts ?? null
+        trading_posts: trading_posts ?? null,
+        edition_id: edition_id || null
       }])
       .select()
       .single();
@@ -128,7 +129,7 @@ export async function PATCH(request: Request) {
     }
 
     const body = await request.json();
-    const { id, campaign_type_name, image_url, trading_posts, resources } = body;
+    const { id, campaign_type_name, image_url, trading_posts, resources, edition_id } = body;
 
     if (!id) {
       return NextResponse.json(
@@ -183,6 +184,9 @@ export async function PATCH(request: Request) {
       }
       updateData.trading_posts = trading_posts;
     }
+    if (edition_id !== undefined) {
+      updateData.edition_id = edition_id || null;
+    }
 
     if (Object.keys(updateData).length === 0 && resources === undefined) {
       return NextResponse.json(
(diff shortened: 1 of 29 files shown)


joeseos:
## Code Review

@joeseos Reviewed the full diff (29 files) against current `main` (which already has the `editions` table and root-table `edition_id` columns/backfill from the earlier editions work, so this PR's migration applies cleanly on top).

### Findings

**1. `EditionSelect`'s auto-default effect makes "no edition" unselectable — `components/edition-select.tsx:23-27`** — medium
```js
useEffect(() => {
  if (!defaultToCurrent || value || editions.length === 0) return;
  const current = editions.find(edition => edition.is_current);
  if (current) onChange(current.id);
}, [defaultToCurrent, value, editions, onChange]);
```
Every one of the 14 call sites passes `defaultToCurrent`. The effect re-fires any time `value` becomes falsy — including when the user explicitly picks the blank `"Select edition"` option, or when a parent sets `editionId` to `item.edition_id ?? ''` for a record whose `edition_id` is `null`. In both cases it immediately snaps back to the current edition via `onChange(current.id)`. Concretely: the blank option is dead in the UI (can't ever be chosen/kept), and opening the edit form for any row with a `null` edition_id will silently coerce it to the current edition if the admin saves without touching the field — which conflicts with the design intent from the editions migration that `edition_id` stays nullable/explicit rather than inferred. Today's impact is limited because the prior backfill already set `edition_id` on all existing root rows, but this affects every future create/edit through all 14 modals.

Suggested fix: track whether the default has already been applied (e.g. a ref/flag, or only auto-apply on initial mount when `value` is `undefined`/not yet initialized rather than on every falsy `value`), so a deliberate blank selection sticks.

**2. Misleading type cast drops `edition_id` from the 'affiliation' branch — `app/api/admin/gang-lineages/route.ts:33-35`** — low
```ts
const lineageColumns = (type === 'affiliation'
  ? 'id, name, fighter_type_id, created_at, updated_at, edition_id'
  : 'id, name, fighter_type_id, created_at, updated_at') as 'id, name, fighter_type_id, created_at, updated_at';
```
Casting the whole ternary to the *narrower* (legacy) literal means the inferred `row`/`rows` type never includes `edition_id`, even for the affiliation path where it's actually selected. Works today only because the code spreads `...row` rather than accessing `.edition_id` by name — any future code that reads `row.edition_id` directly will silently lose type coverage. A discriminated union (`type === 'affiliation' ? select<'...edition_id'> : select<'...'>`) or a shared column-list constant per type would avoid the lie.

**3. Extra query on every fighter-type save — `app/api/admin/fighter-types/route.ts` PATCH (~478-489), POST (~778-786)** — low
Both handlers now do an unconditional round trip to `gang_types` to resolve `edition_id`, even when `gang_type_id` didn't change in this edit. Reasonable given the "derive server-side, never trust the client" invariant documented in the migration comment, and low-cost on an admin-only path — just flagging as a minor perf note, not blocking.

**4. Repeated filter/reset boilerplate across 9 modals** — low/nit
The `filteredX = editionId ? list.filter(...) : list` + `handleEditionChange` (clear selection if the previously-selected item's edition no longer matches) pattern is duplicated nearly verbatim in `admin-campaign-management.tsx`, `admin-create-equipment.tsx`, `admin-create-fighter-type.tsx`, `admin-edit-equipment.tsx`, `admin-edit-fighter-type.tsx`, `admin-edit-skill.tsx`, `admin-edit-vehicle-type.tsx`, `admin-gang-lineage.tsx`, `admin-injuries.tsx`, `admin-scenarios-modal.tsx`. Worth a small shared hook (e.g. `useEditionFilter(list, editionId, onMismatch)`) at some point, but not blocking for this PR.

### Checklist
- Correctness: fighter_types correctly derives `edition_id` server-side from the linked gang type via the composite FK, matching the stated invariant — good. Main issue is the `EditionSelect` auto-default described above.
- Edge cases / null handling: mostly solid (`?? null`, `|| null` used consistently); the one gap is the null-edition round-trip in finding #1.
- Security: all mutating routes remain behind existing `checkAdmin` gates; the new `/api/editions` GET route is read-only and correctly requires only authentication, matching the `editions` table's RLS (SELECT open to all authenticated users).
- Performance: negligible; see minor note #3.
- Readability/maintainability: generally clear, well-commented (the "derived server-side, never from client" comments are a nice touch); see DRY note #4.
- Conventions: consistent with existing admin CRUD route/modal patterns in the codebase; no `CLAUDE.md` present in this repo to check against.
- Migration (`20260730120000_add_fighter_types_edition_id.sql`): idempotent, guarded `DO $$` blocks for constraints, backfill ordered correctly before the composite FK is added — no issues found.

### Verdict: **Changes requested** (finding #1)

---

PR #1909, opened 2026-07-31
Added save as a new fighter characteristic

joeseos:
Added save (Sv) as a new characteristics to support the new N26 fighter cards


joeseos:
## Code Review

@joeseos Reviewed the full diff (28 files) against `main`, including the new migration, the SQL function change, and the read/write paths for the new `save` (Sv) characteristic.

### Findings

**1. `hasSaveCharacteristic` gate never wired up for custom fighter types — `app/api/fighter-types/route.ts:77`** — low/info
```ts
save: cf.save ?? null,
edition_slug: null, // custom fighters get edition support with the customise edition selector
```
`custom_fighter_types.save` exists in the DB and is passed through the create/update actions (`app/actions/customise/custom-fighters.ts`), but `components/customise/custom-fighters.tsx` has no Sv input at all, and `edition_slug` is hardcoded to `null` here — so `hasSaveCharacteristic()` will always be `false` for custom fighters and Sv can never render for them even if a value were set some other way. Self-documented as deferred scope via the comment, so not blocking — just flagging so it isn't forgotten before N26 custom-fighter support ships.

**2. Fighter-side `save` effect modifiers silently no-op — `utils/effect-modifiers.ts:151-166`** — low/medium
```ts
// 'save' is deliberately excluded: fighter-side save effects are not
// supported, and the 'save' stat_name belongs to vehicle effects
const statMapping: Record<string, Exclude<keyof typeof adjustedStats, 'save'>> = { ... };
```
If an admin creates a fighter effect (injury, advancement, gene-smithing, equipment, etc.) with `stat_name: 'save'` — plausible once N26 content with Save-affecting rules gets added via the admin panel — it will silently do nothing (no validation error, no visual indicator). Worth either rejecting that `stat_name` explicitly in the admin fighter-effects form or leaving a comment there pointing back to this limitation, so a future admin doesn't spend time debugging a no-op effect.

**3. Unrelated dead-code removal bundled in — `components/gang/my-fighters.tsx`** — nit
`memoizedFormatters`, `imageProps`, `BACKGROUND_IMAGE_URL`, `imageSizes` are removed; I confirmed they were genuinely unused before this PR (not referenced elsewhere in the file). Good cleanup, just out of scope for a "save characteristic" PR — worth a separate commit/PR next time for a cleaner history.

**4. Minor migration nit — `supabase/migrations/20260730150000_add_fighter_save_stat_and_n26.sql`**
Only `fighters.save` gets a `COMMENT ON COLUMN`; `fighter_types.save` and `custom_fighter_types.save` don't get the equivalent. Trivial, not blocking.

### Nice catch
`components/edition-select.tsx` now tracks `hasDefaulted` in a ref so the `defaultToCurrent` auto-select only fires once, instead of re-forcing the current edition every time `value` goes blank. That fixes the "can't ever pick 'no edition'" issue from PR #1908 — good fix, and it's tested implicitly by every one of the 14 existing call sites continuing to work.

### Checklist
- Correctness: traced the full Sv gating chain end-to-end (`EditionSelect` → `showSave` → admin create/edit forms → `fighter-types` API → SQL function → `getGangFightersList`/`getFighterTypeInfo` caches → `fighter-card`/`fighter-details-card`/`print-gang` → `StatsTable`/`FighterDetailsStatsTable`). It's consistent — in particular I checked whether an admin editing an *existing* N26 fighter type could have its `save` silently nulled out because `editionId` isn't synced from the loaded record (`admin-edit-fighter-type.tsx` never calls `setEditionId` when populating from a fetched fighter type) — but the "Filter by Gang Type" dropdown is edition-filtered and gates fighter-type loading (`enabled: !!gangTypeFilter`), so an admin can't actually reach an N26 fighter type without `editionId` already being N26. No bug in practice.
- Edge cases / null handling: `save` is nullable end-to-end and consistently defaulted with `?? null`; `hasSaveCharacteristic` treats unknown/missing slugs as legacy (no Sv), which is the safe default.
- Security: no new surface — same `checkAdmin`-gated routes as before.
- Performance: cache keys for the two touched `unstable_cache` reads (`getFighterBasic`, `getFighterTypeInfo`) and both `getGangFightersList` variants were correctly bumped to `-v2` so stale cached entries (missing the new `save`/`edition_id`/`editions` join) can't be served post-deploy — matches the pattern from the cache-tags refactor.
- Readability/maintainability: clear comments explaining non-obvious decisions (edition-slug gating rationale, why `save` is excluded from effect modifiers, why n26 seeds as non-current). Good.
- Conventions: consistent with the existing stats plumbing pattern (movement/WS/BS/etc.) and the admin CRUD conventions from the prior editions PRs.
- DRY/YAGNI: no new duplication; reuses the existing `EditionSelect`/edition infrastructure well via the new shared `useEditions()` hook.

### Verdict: **Approve**

---

joeseos:
## Follow-up Review

@joeseos Reviewed the new commit since the last review (`c1bc5ac` — "Added effects support for Sv"), a focused 17-line change entirely in `utils/effect-modifiers.ts`.

### Previous findings

1. **`hasSaveCharacteristic` gate never wired up for custom fighter types** (`app/api/fighter-types/route.ts:77`) — **outstanding**, not addressed by this commit, still non-blocking/deferred scope as noted before.
2. **Fighter-side `save` effect modifiers silently no-op** (`utils/effect-modifiers.ts:151-166`) — **resolved**. `save` is now a valid target in `statMapping`, and `applyNumericModifiers`'s signature was correctly widened to `number | string | null | undefined` to handle the pre-N26 `null` base case.
3. **Unrelated dead-code removal bundled in** (`components/gang/my-fighters.tsx`) — outstanding, but this was a nit about the prior commit's scope, not something to fix here.
4. **Migration comment nit** (`fighter_types.save`/`custom_fighter_types.save` missing `COMMENT ON COLUMN`) — outstanding, still trivial/non-blocking.

### New changes review

Traced the null-base semantics carefully since that's the crux of the fix:
- **`set` operation against a `null` base** → `finalSetValue` wins outright, giving a clean numeric save (e.g., N26 armour granting Sv for the first time). Correct.
- **`add` operation against a `null` base** → falls through to the "unparseable base + additions" branch, so the result is the raw addend alone (e.g., a `-1` modifier renders as `-1+`). The code comment calls this out as intentional and relies on the existing `2+..6+` range check to flag it — confirmed that check exists in `fighter-details-stats-table.tsx`'s `isStatOutOfRange` (via `fighterCharacteristicLimits.Sv`) and does visually flag an out-of-range value like `1+` or `-1+`.
- **Gap**: that range-check styling only exists in `FighterDetailsStatsTable` (fighter detail page). The plain gang-page grid (`fighter-card.tsx` → `ui/fighter-card-stats-table.tsx`) has no equivalent out-of-range highlighting for any characteristic, `Sv` included, so a misconfigured `add`-only save effect would render an invalid value there with no visual signal. This is a pre-existing gap in `StatsTable` (not introduced by this commit) and low severity — an admin creating a `save` effect with `add` on a fighter that never had `set` first is an edge case, and the field is admin-only. Not blocking.
- Verified no other call sites of `applyNumericModifiers` are affected: the weapon-profile path (`applyWeaponModifiers`) still early-returns before calling it when `baseValue` is `null`/`undefined`, so the widened signature is additive, not a behavior change for existing numeric fields.
- `git diff` confirms the commit touches only `utils/effect-modifiers.ts` — no scope creep.

No new bugs, regressions, or edge-case gaps beyond the one noted above (which is pre-existing and low severity).

### Verdict: **Approve**

---

PR #1910, opened 2026-08-01
feat: add fighter_classes JSONB column for multiple class support

joeseos:
Add a `fighter_classes` JSONB array column so a fighter can hold multiple classes (e.g. a Phyrr Cat is "Fighter (Beast, Pet)").

## Database

One migration, `20260801120000_fighter_multiple_classes.sql`, covering schema only:

- Adds `fighter_classes jsonb NOT NULL DEFAULT '[]'` to `fighter_types`, `fighters`, and `custom_fighter_types`
- Adds `causing_fighter_classes` / `injured_fighter_classes` to `fighter_ooa_records`
- Backfills each table from **both** legacy columns (`fighter_class` text and `fighter_class_id` uuid) — primary source first, then the other as a fallback, since neither is guaranteed populated
- Seeds `Beast` and `Pet` into the `fighter_classes` table

Every pass is guarded on `fighter_classes = '[]'`, so fallbacks only fill what the primary missed and the migration is safe to re-run.

The `fighter_classes` **table** remains the source of truth for valid class names — it is unchanged, and `skill_access_archetypes.fighter_class_id` still points at it.

## ⚠️ Required deploy order

**Apply the migration before merging.**

Migrations are not automated here, but `supabase/functions/*.sql` **are** — `.github/workflows/deploy_supabase_functions.yml` deploys changed RPCs on push to `main`.

1. Apply `20260801120000_fighter_multiple_classes.sql`
2. Deploy the five RPCs (or merge and let CI do it)
3. Merge

Order matters, and the failure mode is quiet. PL/pgSQL only syntax-checks a function body at creation — it does not resolve column references. Verified against PostgreSQL 16: `CREATE OR REPLACE FUNCTION` on a body reading a non-existent `ft.fighter_classes` **succeeds with exit code 0**, so `ON_ERROR_STOP=1` does not catch it and the workflow reports green. Every subsequent call then fails with `column ft.fighter_classes does not exist`. The functions do self-heal once the column exists — no redeploy needed — but until then the affected flows are broken with no CI signal.

Steps 1 and 2 are safe to run ahead of the merge: all five RPCs still emit the legacy `fighter_class` key alongside the new one, so the currently deployed frontend keeps working.

`deploy_supabase_functions.yml` and `deploy.yml` both trigger independently on push to `main` with no `needs:`/`workflow_run` ordering. Doing step 2 manually before merging closes that race; otherwise there is a brief window where the app can deploy first.

Note that changing a function's return signature requires a PostgREST schema-cache reload (`NOTIFY pgrst, 'reload schema'`, or the dashboard button).

## RPCs

The migration contains no function definitions. Five RPCs read the new column and live in `supabase/functions/`:

- `get_available_skills`
- `get_fighter_available_advancements`
- `get_fighter_types_with_cost`
- `get_gang_details`
- `copy_custom_collection`

`get_fighter_types_with_cost` also drops the INNER JOIN on `ft.fighter_class_id`. Application code no longer writes that nullable column, so the join would have silently excluded every newly created fighter type from the hiring and promotion lists. Its legacy `fighter_class` / `fighter_class_id` return columns are retained for backward compatibility, now derived from `fighter_classes` rather than the join.

## Code

Code reads and writes `fighter_classes: string[]` only — no dual-writes and no fallbacks to the old columns. Array operations are inlined rather than wrapped in helpers: `.includes()` for class checks, `[0]` for the primary class, `.join(', ')` for display.

The old columns are deliberately **not** dropped, so a rollback stays possible. Dropping them is a follow-up once this is verified in preview.

## Rollback note

Because code no longer writes the old columns, rows created or edited after this deploys will have stale/empty values there. A rollback would restore behaviour but not the class data for those rows.

## Verification

`tsc --noEmit` and `eslint` clean on all touched files. `next build` compiles and passes its TypeScript stage; it cannot complete static prerendering locally without Supabase env vars. PL/pgSQL creation-vs-runtime behaviour was verified against a local PostgreSQL 16; the RPC bodies themselves have not been executed against real data.

Post-migration sanity check for rows no backfill pass could fill:

```sql
SELECT 'fighter_types' t, count(*) FROM fighter_types WHERE fighter_classes = '[]'
UNION ALL SELECT 'fighters', count(*) FROM fighters WHERE fighter_classes = '[]'
UNION ALL SELECT 'custom', count(*) FROM custom_fighter_types WHERE fighter_classes = '[]';
```


joeseos:
## Code Review

@joeseos Reviewed the full diff (56 files, ~410/373 +/-) — the migration, all five updated RPCs, and the frontend refactor from `fighter_class`/`fighter_class_id` to `fighter_classes: string[]`.

### Findings

**1. Deploy-ordering race between the SQL-function deploy and the app deploy — no code change needed here, but worth confirming before merge** — medium
The PR body says the five RPC updates are "to be applied manually," but the repo actually auto-deploys them: `.github/workflows/deploy_supabase_functions.yml` triggers `on: push: branches: [main]` for changed `supabase/functions/*.sql`, running a `psql -f` loop. The Next.js app is deployed by a **separate**, independently-triggered workflow (`.github/workflows/deploy.yml`, same `push: main` trigger). Neither workflow has a `needs:`/`workflow_run` dependency on the other, so there's no structural guarantee the SQL functions finish updating before Vercel starts serving the new frontend build.

This matters more than usual here because the PR deliberately has **no fallback** to the old columns — e.g. `FighterPromotionModal` does `ft.fighter_classes.includes(targetClass)` and `sortPromotionFighterTypes` does `a.fighter_classes[0]` with no optional chaining, on data sourced from `get_fighter_types_with_cost`. If that RPC's update lags behind the Vercel deploy for even a short window (slow runner, `psql` retry, secrets hiccup), any request hitting the fighter-hiring, promotion, or fighter-type-edit flows in that window gets `fighter_classes: undefined` from the stale RPC and throws. In practice the SQL job is almost certainly faster than a full `vercel build --prod`, so this is probably low-probability — but it's unverified/unguaranteed, and worth either chaining the workflows explicitly or keeping a defensive `?? []` at the two or three hottest call sites until the old columns are dropped in the follow-up migration.

**2. `get_fighter_types_with_cost.sql` still returns the legacy `fighter_class`/`fighter_class_id` columns** — low
`supabase/functions/get_fighter_types_with_cost.sql:15-17,55-57` keeps `fighter_class text` / `fighter_class_id uuid` in `RETURNS TABLE`, populated via the old `fighter_class_id` join, alongside the new `fighter_classes jsonb`. Per the PR description ("no dual-writes and no fallbacks to the old columns") these are presumably unused by the frontend now — harmless dead weight today, but worth trimming when the old columns get dropped so the RPC signature doesn't lag behind the stated intent.

**3. Fallback semantics changed in the fighter-class submit path — worth a manual QA pass** — low
`components/fighter/edit-fighter/fighter-edit-modal.tsx` (~890-899): when `selectedFighterClassId` doesn't resolve against `allFighterClasses`, the code used to fall back to `formValues.fighter_class`/`fighter_class_id` (per the removed comment, this covered "promotion to Exotic Beast Specialist"); now it falls back to `submitData.fighter_classes = fighter.fighter_classes` — the fighter's *original*, pre-edit prop, not any in-progress edit state (since `formValues` no longer tracks classes at all after this refactor, there wasn't really a like-for-like value to preserve). This looks like a reasonable simplification given promotions are now handled entirely through `fighter-advancement-list.tsx`'s separate `onFighterDetailsUpdate` path rather than this modal's submit — but since I couldn't exercise the UI, it's worth a manual check that editing a fighter through this specific modal still preserves class changes correctly in whatever edge case originally required that fallback.

### Checklist
- **Correctness**: The JSONB migration itself is solid — idempotent (`WHERE ... = '[]'::jsonb` guards prevent double-backfill on re-run), correctly seeds `Beast`/`Pet` into the reference table, and backfills all five affected columns from their single-value predecessors. The four live RPCs (`get_available_skills`, `get_fighter_available_advancements`, `get_fighter_types_with_cost`, `copy_custom_collection`) use the correct jsonb operators (`?|` for "contains any of", `->>0` for the primary/display class) and I traced them against the schema — no logic errors found. `get_gang_details.sql` was also updated but has zero callers in the current codebase (dead code), so no risk there either way.
- **Edge cases / null handling**: Consistently handled — `fighter_classes jsonb NOT NULL DEFAULT '[]'::jsonb` means SQL-side code never sees NULL, and the frontend mapping layer (`mapFighterType`, `buildFighterFromServerData`, etc.) consistently normalizes with `?? []`/`|| []` at data-entry points.
- **Security**: No new surface; same RLS/auth boundaries as before.
- **Performance**: `.contains('fighter_classes', ...)` on a JSONB column without a GIN index will do a sequential scan under `getGangBeastCount` — moot today since that function has no callers, but flag it if it gets wired up later.
- **Readability/maintainability**: The commit history and PR description are unusually clear about intent (why old columns stay, why RPCs are separate, explicit rollback caveat). Good.
- **Conventions**: Matches the codebase's existing pattern of parallel migration + manually-tracked `supabase/functions/*.sql`; no `CLAUDE.md` in this repo to check against.
- **DRY/YAGNI**: Per the PR description, array ops are deliberately inlined (`.includes()`, `[0]`, `.join(', ')`) rather than wrapped in helpers — consistent across all ~40 touched components, no stray duplication introduced.

### Verdict: **Approve** (finding #1 is worth a quick confirmation — e.g. checking whether the SQL-functions job reliably completes well before the Vercel job in this repo's Actions history — but isn't a reason to block; findings #2/#3 are non-blocking nits/QA notes)

---

joeseos:
## Follow-up Review

@joeseos Reviewed the 6 commits since the last review (`958cb0c` → `5bae1eb`): `eaaf4b4`, `74734c8`, `1a34762`, `08b2206`, `b4b78f7`, `5bae1eb`.

### Previous findings

1. **Deploy-ordering race (SQL functions vs. app)** — **resolved**. `74734c8` restores the legacy `fighter_class`/`fighter_class_id` return columns to `get_fighter_types_with_cost` (now derived from `fighter_classes->>0` instead of the removed `INNER JOIN`, with `COALESCE(..., '')` so the currently-deployed frontend's unguarded `fighter_class.toLowerCase()` can't throw on a row the old join used to hide). Commit message states all five RPCs are now safe to deploy ahead of the app — verified the return-signature change on `get_fighter_types_with_cost.sql` and it checks out.
2. **Legacy columns retained on `get_fighter_types_with_cost`** — **resolved differently than expected**: rather than being trimmed, they're now load-bearing for finding #1's fix (kept intentionally, with a comment explaining why and that they'll drop in the follow-up). Makes sense given the resolution above.
3. **Fallback semantics in `fighter-edit-modal.tsx` submit path** — still a manual-QA item, unaffected in substance by the newer commits (`b4b78f7` only added edition-scoping to the same query, didn't touch the fallback logic itself).

### New changes review

**Real bug caught and fixed**: `eaaf4b4` fixes an actual break — `get_fighter_available_advancements.sql` line 104 still referenced the removed `v_fighter_class` (singular) variable, which would fail every call to that RPC at plan time (this slipped past the prior review's "no logic errors found" on this function). Verified at current HEAD: declaration and all four usages are consistently `v_fighter_classes jsonb` using `?|` array containment — no lingering references. Good catch and clean fix.

**New finding — possible risk to the migration itself, medium**
`supabase/migrations/20260801120000_fighter_multiple_classes.sql` and `supabase/seed.sql` both `INSERT INTO public.fighter_classes (class_name, slug, standard_class, edition_id) ...` to seed Beast/Pet. But `standard_class` was dropped from `fighter_classes` by `supabase/migrations/20260402100000_remove_standard_class_from_fighter_classes.sql`, which is already merged into `main` (confirmed via `git merge-base --is-ancestor`). If that drop migration has actually been applied to the live database, this `INSERT` will fail outright with "column standard_class does not exist" the moment it's run — which would block the "apply migration before merge" step this PR's own description calls out as required.

Complicating this: `supabase/schema/schema.public.sql` (the tracked snapshot, last regenerated Jul 29 — well after the April drop) **still shows `standard_class boolean DEFAULT false` on `fighter_classes`**, on `main` itself, independent of this PR. So either the snapshot is stale/wrong, or the drop migration was merged but never actually run against the live DB (this repo's migrations aren't automated, per the PR body). This is a pre-existing inconsistency, not something introduced by this PR, but the new commits are the first thing to actually depend on the answer. Worth a quick check against the real schema (e.g. `\d fighter_classes` on the target DB) before applying — if the column really is gone, drop it from both the migration and seed insert.

**Other changes verified clean**:
- `08b2206`'s admin-create fix: confirmed `app/api/admin/fighter-types/route.ts` POST handler reads `data.fighterClasses` (line ~800), matching the client's new payload shape in `admin-create-fighter-type.tsx`.
- `1a34762`'s dual-fallback backfill (fighter_types: FK-primary/text-fallback; fighters/custom_fighter_types: text-primary/FK-fallback) is correctly guarded on `fighter_classes = '[]'` throughout, consistent with the tables' actual display sources.
- `b4b78f7`'s edition-scoped `/api/fighter-classes` route and its two callers (`FighterAddModal`, `fighter-edit-modal`) correctly thread `edition_slug` into both the fetch and the React Query cache key, avoiding stale cross-edition cache collisions.
- `5bae1eb`'s `copy_custom_collection.sql` column/value list: manually counted both sides at 24 columns each — aligned correctly after dropping `fighter_class`/`fighter_class_id`.

No other new bugs or regressions found; edge-case and null handling in the new commits look consistent with the rest of the migration.

### Verdict: **Approve** (pending confirmation on the `standard_class` column — worth resolving before running the migration, since it could otherwise block the deploy step this PR explicitly calls out as required)

---

PR #1911, opened 2026-08-02
Added Trade Points to gangs table

Djidiouf:


joeseos:
## Code Review

@Djidiouf Reviewed this against the current schema and recent migration history. Summary: small, well-scoped, single-file SQL migration — no correctness issues found.

### Findings

**Low — `NOT NULL DEFAULT 0` vs. nullable pattern used for the sibling N26 column**
`supabase/migrations/20260803120000_add_gangs_trade_points.sql:5`
The comment states Trade Points is "only surfaced/edited for gangs whose gang type uses edition n26," which is the same shape as the `save` characteristic added in `20260730150000_add_fighter_save_stat_and_n26.sql`. That migration deliberately left `save` nullable with no default ("NULL means stat does not apply"), so non-applicable rows are distinguishable from a real zero value. This migration instead defaults `trade_points` to `0 NOT NULL` for every gang, including non-N26 gangs, so "not applicable" and "spent down to zero" are indistinguishable at the DB level. Given trade_points is a spendable resource (like reputation/credits, which are also `numeric` and default-less/nullable on this table), defaulting to `0` is a reasonable and probably intentional choice — but worth a quick confirmation that no code will ever aggregate/sum `trade_points` across all gangs (which would silently include non-N26 gangs) rather than filtering by edition first.

### Checklist
- **Correctness**: Column is additive, `IF NOT EXISTS` guarded, safe to re-run. No issues.
- **Edge cases / null handling**: See low finding above re: NOT NULL default vs. nullable pattern.
- **Security**: No RLS/grants changes needed — this is a plain column addition on a table with existing row-level policies; no column-level access control exists elsewhere in this schema either, so nothing missing here.
- **Performance**: `ADD COLUMN ... DEFAULT 0` on Postgres 11+ is a metadata-only change (no table rewrite/lock storm) since the default is a constant. No concern at any table size.
- **Readability/maintainability**: Migration is clear, well-commented, follows the same header-comment style as the recent `save`/N26 migration.
- **Project conventions**: Filename timestamp format, `ADD COLUMN IF NOT EXISTS`, and `numeric` type all match existing migrations (e.g. `reputation`, `credits`, `wealth` are all `numeric` on `gangs`). Schema snapshot (`schema/schema.public.sql`) isn't updated in this PR, but per `supabase/README.md` that's expected — it's regenerated daily by CI, not hand-maintained per-PR.
- **DRY/YAGNI**: N/A, single additive column.

Note: no application code (types, API routes, UI) touches `trade_points` yet in this PR — presumably intentional migration-first sequencing ahead of a follow-up feature PR. Just flagging so it's not merged assuming the feature is otherwise wired up.

### Verdict: **Approve**
No blocking issues. The nullable-vs-default question above is worth a sentence of confirmation but doesn't need to hold up merging.


---

Djidiouf:
@joeseos Nothing aggregates or sums `trade_points` across gangs.

Every access is single-gang scoped:
- Read: [`getGangBasic`](app/lib/shared/gang-data.ts) selects it for one gang by id; the value is only ever rendered behind `hasTradePoints(edition_slug)` in [`gang.tsx`](components/gang/gang.tsx) and [`gang-edit-modal.tsx`](components/gang/gang-edit-modal.tsx).
- Write: [`updateGang`](app/actions/update-gang.ts) reads/writes one gang and additionally re-checks the edition server-side before applying (`hasTradePoints(editionSlug)` → throws otherwise).
- Copy: [`copy-gang.ts`](app/actions/copy-gang.ts) copies from a single source gang.

There is no `SUM(trade_points)`, cross-gang rollup, leaderboard, or campaign aggregate that would sweep in non-N26 gangs as `0`.

On the nullable-vs-`NOT NULL DEFAULT 0` question — the `DEFAULT 0` is deliberate and is the right call here, precisely because `trade_points` is unlike `save`:
- `save` is a *characteristic* where "not applicable" is a distinct concept from any numeric value, so NULL carries meaning.
- `trade_points` is a *spendable quantity* (like `reputation`/`credits`/`wealth`), where `0` is a legitimate, meaningful value ("has none"). Applicability is derived from the gang's edition, never from NULL — so there's no information lost by defaulting to `0`.
- `NOT NULL DEFAULT 0` also removes null-handling churn on the write path (add/subtract arithmetic) and keeps the value non-null for the UI. The sibling columns `reputation`/`credits`/`wealth` are nullable only for legacy reasons; `0`-default is arguably the cleaner convention for a new resource column.

PR #1912, opened 2026-08-02
Drop the legacy single-value fighter class columns

joeseos:
Phase 2 of the multiple-classes change, following #1910.

Phase 1 added `fighter_classes jsonb`, backfilled it, and moved all application code onto the array. The old columns stayed behind — and the RPCs kept emitting a legacy `fighter_class` key — so the previously deployed frontend kept working while the new build rolled out. That compatibility window has closed.

## What this drops

| Table | Columns |
|---|---|
| `fighters` | `fighter_class`, `fighter_class_id` |
| `fighter_types` | `fighter_class`, `fighter_class_id` |
| `custom_fighter_types` | `fighter_class`, `fighter_class_id` |
| `fighter_ooa_records` | `causing_fighter_class`, `injured_fighter_class` |
| `fighter_classes` | `slug`, its format check, and `fighter_classes_edition_slug_idx` |

Nothing ever read `fighter_classes.slug`. The one thing it enforced — one row per class per edition — moved onto `class_name` via `fighter_classes_edition_class_name_idx` in #1910.

Also drops **`get_add_fighter_details`**: an untracked pre-JSONB function that is absent from `supabase/functions/`, has no callers (a comment at `app/api/fighter-types/route.ts:200` records that `get_fighter_types_with_cost` supersedes it), and INNER JOINs `fighter_classes fc ON fc.id = ft.fighter_class_id` — so it would otherwise be left permanently broken.

**`skill_access_archetypes.fighter_class_id` is deliberately untouched.** It is a real FK to the `fighter_classes` table and still drives the archetype feature.

## Safety

The migration **deletes no rows** and uses **no `CASCADE` anywhere**. Plain `DROP` under the default `RESTRICT` means that if anything unexpectedly depends on a column, the statement fails loudly rather than quietly removing the dependent object too. (`DROP COLUMN` still removes that column's own index and FK — that is not a cascade and destroys no rows.)

A guard runs first and refuses to drop anything while any row still holds a class only in a legacy column. It uses dynamic SQL behind an `information_schema` existence test, so re-running after the drop is a clean no-op rather than a `column does not exist` error.

## RPC changes

`get_available_skills`, `get_fighter_available_advancements` and `get_gang_details` stop emitting the legacy `fighter_class` key. `get_fighter_types_with_cost` additionally drops it from its `RETURNS TABLE` — a signature change, so PostgREST needs a schema-cache reload after deploy. `copy_custom_collection` needed no change.

`seed.sql` is updated to match the post-drop schema.

## Deploy order

The app half is already live from #1910, so the drop is manual and last:

1. Merge → CI deploys the four updated RPCs. Safe: the live app reads only `fighter_classes`, and the columns still exist at this point.
2. Confirm the RPCs are live and the app is healthy.
3. Apply `20260802180000_drop_legacy_fighter_class_columns.sql` manually.
4. Reload the PostgREST schema cache — `NOTIFY pgrst, 'reload schema'` or the dashboard button.

Do not drop before step 1: the currently deployed RPCs still select `f.fighter_class` and would fail at runtime the moment the column disappears.

Worth running before step 3 — must return zero, or the guard will (correctly) refuse:

```sql
SELECT 'fighters' t, count(*) FROM fighters
  WHERE fighter_classes = '[]' AND (fighter_class IS NOT NULL OR fighter_class_id IS NOT NULL)
UNION ALL SELECT 'fighter_types', count(*) FROM fighter_types
  WHERE fighter_classes = '[]' AND (fighter_class IS NOT NULL OR fighter_class_id IS NOT NULL)
UNION ALL SELECT 'custom', count(*) FROM custom_fighter_types
  WHERE fighter_classes = '[]' AND (fighter_class IS NOT NULL OR fighter_class_id IS NOT NULL);
```

## Verification

Exercised against a local PostgreSQL 16 on a schema reproducing the real cascading FKs (`fighters.fighter_type_id → fighter_types ON DELETE CASCADE`, `fighters.gang_id → gangs ON DELETE CASCADE`):

- Row counts identical before and after across all six tables — nothing cascaded
- Target columns gone; `skill_access_archetypes.fighter_class_id` and `editions.slug` intact
- Second run exits 0 with every statement skipped (idempotent)
- Against two at-risk rows the guard refuses with `2 row(s) hold a class only in the old columns (fighters: 1; fighter_ooa_records: 1)` and leaves all 8 legacy columns in place

`tsc --noEmit` and `eslint` clean. The RPC bodies themselves have not been executed against real data — worth smoke-testing add fighter, promotion, advancements, skills modal, custom-collection copy, and an Outcasts Leader's archetypes after deploy.

https://claude.ai/code/session_01EKU5BRzYFsiwW82UbkrTgF

---
_Generated by [Claude Code](https://claude.ai/code/session_01EKU5BRzYFsiwW82UbkrTgF)_

### DIFF
--- a/components/gang/gang.tsx
+++ b/components/gang/gang.tsx
@@ -312,7 +312,7 @@ export default function Gang({
     );
   }, [fighters]);
 
-  // Fighters composition for tooltip: group by fighter_type and fighter_class
+  // Fighters composition for tooltip: group by fighter_type and fighter_classes
   const fighterTypeClassCounts = useMemo(() => {
     const counts = new Map<string, { label: string; count: number; classKey: string }>();
     for (const fighter of activeFighters) {
--- a/supabase/functions/get_available_skills.sql
+++ b/supabase/functions/get_available_skills.sql
@@ -145,7 +145,6 @@ BEGIN
     )
     SELECT jsonb_build_object(
         'fighter_id', get_available_skills.fighter_id,
-        'fighter_class', v_fighter_classes->>0,
         'fighter_classes', v_fighter_classes,
         'skills', COALESCE(
             jsonb_agg(
@@ -153,7 +152,6 @@ BEGIN
                     'skill_id', a.skill_id,
                     'skill_name', a.skill_name,
                     'is_custom', a.is_custom,
-                    'fighter_class', v_fighter_classes->>0,
                     'skill_type_id', a.skill_type_id,
                     'skill_type_name', a.skill_type_name,
                     'effective_access_level', a.effective_access_level,
--- a/supabase/functions/get_fighter_available_advancements.sql
+++ b/supabase/functions/get_fighter_available_advancements.sql
@@ -23,7 +23,7 @@ BEGIN
     RAISE EXCEPTION 'Fighter not found with ID %', get_fighter_available_advancements.fighter_id;
   END IF;
   
-  -- Determine if the fighter uses flat costs based on fighter_class
+  -- Determine if the fighter uses flat costs based on their classes
   -- Only Gangers and Exotic Beasts use flat costs
   v_uses_flat_cost :=
     v_fighter_classes ?| array['Ganger', 'Exotic Beast'];
@@ -128,7 +128,6 @@ BEGIN
   SELECT jsonb_build_object(
     'fighter_id', get_fighter_available_advancements.fighter_id,
     'current_xp', v_fighter_xp,
-    'fighter_class', v_fighter_classes->>0,
     'fighter_classes', v_fighter_classes,
     'uses_flat_cost', v_uses_flat_cost,
     -- Ganger/Exotic Beast: Specialist table row (random Primary skill) — same flat costs as other ganger advances
--- a/supabase/functions/get_fighter_types_with_cost.sql
+++ b/supabase/functions/get_fighter_types_with_cost.sql
@@ -13,12 +13,6 @@ CREATE OR REPLACE FUNCTION get_fighter_types_with_cost(
 RETURNS TABLE (
     id uuid,
     fighter_type text,
-    -- Legacy single-value columns, kept so the previous frontend build keeps
-    -- working while this RPC is deployed ahead of the app. Derived from
-    -- fighter_classes rather than the old join; dropped in the follow-up that
-    -- drops the underlying columns.
-    fighter_class text,
-    fighter_class_id uuid,
     fighter_classes jsonb,
     gang_type text,
     cost numeric,
@@ -57,11 +51,6 @@ BEGIN
     SELECT
         ft.id,
         ft.fighter_type,
-        -- COALESCE to '' so the previous build's unguarded
-        -- fighter_class.toLowerCase() cannot throw on a row with an empty
-        -- fighter_classes array (such rows were hidden by the old INNER JOIN).
-        COALESCE(ft.fighter_classes->>0, '') AS fighter_class,
-        ft.fighter_class_id,
         ft.fighter_classes,
         ft.gang_type,
         -- Use adjusted_cost if available, otherwise use original cost
--- a/supabase/functions/get_gang_details.sql
+++ b/supabase/functions/get_gang_details.sql
@@ -53,7 +53,6 @@ BEGIN
            f.label,
            f.fighter_type,
            f.fighter_type_id,
-           f.fighter_class,
            f.fighter_classes,
            f.fighter_sub_type_id,
            f.xp,
@@ -585,7 +584,6 @@ BEGIN
            f.label,
            f.fighter_type,
            f.fighter_type_id,
-           f.fighter_class,
            f.fighter_classes,
            json_build_object(
              'fighter_sub_type', fst.sub_type_name,
@@ -732,7 +730,6 @@ BEGIN
                'fighter_name', cf.fighter_name,
                'label', cf.label,
                'fighter_type', cf.fighter_type,
-               'fighter_class', cf.fighter_class,
                'fighter_classes', cf.fighter_classes,
                'fighter_sub_type', cf.fighter_sub_type,
                'alliance_crew_name', cf.alliance_crew_name,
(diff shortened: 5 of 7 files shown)


joeseos:
## Code Review

@joeseos Reviewed this against `main` (post-#1910 merge) and the current `supabase/functions/` RPC bodies. This is a careful, well-sequenced Phase 2 cleanup — verified the specifics below rather than just trusting the PR description.

### Verified claims
- `get_add_fighter_details(uuid, uuid)` — checked the schema snapshot: the deployed signature is exactly `get_add_fighter_details(p_gang_type_id uuid, p_gang_affiliation_id uuid DEFAULT NULL::uuid)` and its `RETURNS TABLE` does include `fighter_class`/`fighter_class_id`, and its body does `INNER JOIN fighter_classes fc ON fc.id = ft.fighter_class_id`. The `DROP FUNCTION IF EXISTS ...(uuid, uuid)` signature match is correct — it would not silently no-op on a signature mismatch.
- `fighter_ooa_records` guard — Phase 1 (`20260801120000_fighter_multiple_classes.sql`) did add `causing_fighter_classes`/`injured_fighter_classes` jsonb columns, so the guard's plural-column reference is correct (this table's legacy/new column names differ in more than pluralization at a glance, worth double-checking, but they line up).
- Constraint/index names (`fighter_classes_slug_format_chk`, `fighter_classes_edition_slug_idx`) match the live schema exactly.
- No application code reads the RPCs' legacy `fighter_class` singular key anymore — searched all `.fighter_class` (non-plural) references across `app/` and `components/`; every hit is `.fighter_classes` (plural) or unrelated. The one stale singular reference is in `README.md`'s example type snippet (pre-existing from #1910, not touched here) — not a blocker, just noting it's still there if a docs pass happens later.
- `get_fighter_types_with_cost.sql`'s `RETURNS TABLE`/`SELECT` list stays comma-balanced after removing the two derived legacy columns.
- Deploy-order description matches reality: `.github/workflows/deploy_supabase_functions.yml` only triggers on changes under `supabase/functions/*.sql`, confirming migrations really are manual-apply-only here, as stated.

### Findings

**Low — guard doesn't account for legacy columns holding `''` instead of `NULL`**
`supabase/migrations/20260802180000_drop_legacy_fighter_class_columns.sql:33-38` (and the `fighter_ooa_records` block below it)
Phase 1's backfill treated `fighter_class = ''` as "no value" (`WHERE fighter_class IS NOT NULL AND fighter_class != ''`), but this guard only checks `IS NOT NULL`. A row whose legacy column is an empty string rather than NULL would trip the guard and block the drop, even though Phase 1 never considered it to hold real data. This fails safe (over-blocks rather than under-checks), so it's not a correctness risk, just a possible false-positive that could confuse whoever runs the pre-drop sanity query in the PR description if such rows exist.

**Low — `DROP FUNCTION IF EXISTS get_add_fighter_details(uuid, uuid)` is a guessed signature**
`supabase/migrations/20260802180000_drop_legacy_fighter_class_columns.sql:64`
Verified this one matches the current schema snapshot, but since the function is untracked in `supabase/functions/`, there's no source-controlled guarantee it stays untouched between snapshot and the manual deploy step. Low risk given `IF EXISTS` + immutable legacy code, just flagging that a signature drift here would make the DROP silently no-op rather than fail loudly, which is a slightly different safety story than the rest of this migration's "fail loudly on RESTRICT" design.

### Checklist
- **Correctness**: Guard logic, drop ordering (index → constraint → column), and RPC edits all check out against the live schema. No bugs found.
- **Edge cases / null handling**: See low finding on `''` vs `NULL` above. Otherwise the empty-array (`'[]'`) vs. NULL-legacy-value logic mirrors Phase 1 correctly.
- **Security**: No RLS/grants impact — plain column/function drops, `SECURITY DEFINER`/`search_path` on functions unchanged. `RESTRICT`-only (no `CASCADE`) is the right conservative choice for a destructive migration.
- **Performance**: `DROP COLUMN` is metadata-only in Postgres; no table rewrite. Guard queries are full-table scans but only run once, manually, off a deploy window — fine at this scale.
- **Readability/maintainability**: Migration is unusually well-commented for a destructive change — each section explains *why*, not just *what*, matching this repo's existing migration style (e.g. the N26/save migration).
- **Project conventions**: Filename/timestamp format, `IF EXISTS` guards, and manual-apply migration model all match existing conventions. `seed.sql` updated consistently with the schema change.
- **DRY/YAGNI**: N/A — this is a straight subtraction; nothing to simplify further.

### Verdict: **Approve**
Both findings are low-severity and don't block merging — they're worth a mental note for whoever runs the manual drop, not a required change to the diff itself.


---

joeseos:
@joeseos Follow-up review of the new commit (`b164f364`) pushed since the last review.

## Previous findings

**1. Guard didn't account for `''` vs `NULL` in legacy columns — low — Resolved**
`NULLIF(fighter_class, '')` (and the `causing_fighter_class`/`injured_fighter_class` equivalents) now mirrors the Phase 1 backfill's treatment of empty string as "no value." Checked both the per-table loop and the `fighter_ooa_records` block — the fix is applied consistently to all four legacy string columns, and `fighter_class_id IS NOT NULL` (a real FK) is correctly left as a hard block, not wrapped in `NULLIF`.

**2. `DROP FUNCTION ... (uuid, uuid)` used a guessed signature — low — Resolved**
Replaced with a `pg_proc` lookup (`oid::regprocedure`) that resolves whatever overload(s) of `get_add_fighter_details` actually exist in `public` and drops them via dynamic `EXECUTE`, with a `RAISE NOTICE` per drop. This removes the silent-no-op risk the original review flagged — an absent function is now a clean no-op, and a drifted signature no longer causes the drop to be skipped.

## New findings

None. The diff is limited to the two guard fixes described above (`supabase/migrations/20260802180000_drop_legacy_fighter_class_columns.sql`, 27 insertions / 4 deletions) — no other files touched.

Two minor, non-blocking observations, not required changes:
- The `pg_proc` lookup matches **by name only** (`proname = 'get_add_fighter_details'`), so it will drop *every* overload with that name in `public`, not just the specific pre-JSONB one. Fine today since the function has no callers per the PR description, but worth keeping in mind if a same-named function is ever legitimately added later.
- No `CASCADE`/`RESTRICT` behavior change from the lookup — still fails loudly on unexpected dependents, consistent with the rest of the migration's design.

Both new `DO $$` blocks remain idempotent on re-run (the function-drop loop naturally no-ops once nothing matches; the guard blocks still gate on `information_schema` column existence), and the added comments clearly explain the *why* for each fix, consistent with the rest of this migration's commenting style.

## Verdict: **Approve**

---

PR #1913, opened 2026-08-02
Added Trade Points resource to gang page and edit gang modal

Djidiouf:


joeseos:
@Djidiouf Thanks for the PR — reviewed against the full codebase (server action, edition gating, cache invalidation, and all UI consumers).

## Findings

**1. Trade Points missing from the print/roster view — medium**
`app/gang/[id]/print/page.tsx:102-134` builds `gangDataForClient` for the printable roster and includes `reputation: gangBasic.reputation`, but never adds `trade_points` or `edition_slug`. `components/gang/print-gang.tsx` has no `trade_points` prop either. So an N26 gang's printed/exported roster will show Reputation but silently omit Trade Points, even though the main gang page now surfaces it. Worth adding for feature parity, or explicitly noting it's out of scope for this PR.

**2. Duplicated edition-slug resolution logic — low (DRY)**
`app/lib/shared/gang-data.ts` (`getGangBasic`) already derives `edition_slug` from `gang_types`/`custom_gang_types` via embedded joins. `app/actions/update-gang.ts` adds a second, independent implementation (`getGangEditionSlug`) that re-queries `gang_types`/`custom_gang_types` directly. Functionally correct (verified against the FK constraints — `gangs_gang_type_id_fkey`, `gangs_custom_gang_type_id_fkey`, `gang_types_gang_type_id_key`, and the `chk_gang_type_exclusive` check), but it's a second source of truth for the same derivation. Consider extracting a shared helper if more edition-gated server checks are added later — not blocking for this PR.

**3. Migration timestamp is dated one day ahead — low/nit**
`supabase/migrations/20260803120000_add_gangs_trade_points.sql` is stamped 2026-08-03, one day after the PR was opened (2026-08-02) and after every other existing migration. Not a functional bug (ordering still works), but worth double-checking your migration tooling's clock/timezone so future migrations don't collide or land out of sequence.

## What's solid
- Server-side gating (`hasTradePoints(editionSlug)` in `update-gang.ts`) correctly prevents the field from being set for non-N26 gangs even if a malicious client sends `trade_points`/`trade_points_operation` — good defense in depth beyond the UI-level `hasTradePoints` gate.
- Migration (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS trade_points numeric NOT NULL DEFAULT 0`) is safe for existing rows.
- Optimistic update / rollback / cache invalidation / resource-change logging for `trade_points` all correctly mirror the existing `reputation` pattern in `gang.tsx` and `update-gang.ts` (including only logging "trade points" when it actually changed, avoiding spurious log entries for non-N26 gangs).
- RLS on `gang_types`, `custom_gang_types`, and `editions` all allow authenticated SELECT, so the new edition lookup in the server action won't fail for non-owner editors (e.g. arbitrators).
- `copy-gang.ts` correctly copies `trade_points` from the source gang, consistent with how `credits`/`reputation` are handled.
- Prop threading (`page.tsx` → `gang-page-content.tsx` → `gang.tsx` → `gang-edit-modal.tsx`) is consistent and type-safe throughout.

## Verdict: **Approve** (with the print/roster gap in #1 worth a follow-up, either in this PR or a fast-follow)

---

joeseos:
@Djidiouf Follow-up review of the new commit (`b86d59c`) pushed since the last review.

## Previous findings

**1. Trade Points missing from the print/roster view — medium — Outstanding**
`app/gang/[id]/print/page.tsx` and `components/gang/print-gang.tsx` are untouched by this commit; the printable roster still won't show Trade Points for N26 gangs. Flagged previously as fine to defer to a fast-follow, so not blocking, but noting it remains open.

**2. Duplicated edition-slug resolution logic — low (DRY) — Resolved**
Extracted into a new `utils/gang-edition.ts`:
- `resolveGangEditionSlug()` — the pure "pick official over custom" flattening logic, now used by `gang-data.ts`'s `getGangBasic`.
- `fetchGangEditionSlug()` — the Supabase lookup by `gang_type_id`/`custom_gang_type_id`, now used by `update-gang.ts` (replacing the deleted local `getGangEditionSlug`).

Checked both call sites: `gang-data.ts` passes the same `data.gang_types?.editions?.slug` / `data.custom_gang_type_edition?.editions?.slug` values it did before, and `update-gang.ts`'s `fetchGangEditionSlug` call preserves the original custom-then-official lookup order and query shape exactly — this is a pure extraction, no behavior change. Confirmed no leftover references to the old local `getGangEditionSlug`.

**3. Migration timestamp dated one day ahead — low/nit — Outstanding**
`supabase/migrations/20260803120000_add_gangs_trade_points.sql` is untouched by this commit. Still not a functional issue, just unaddressed.

## New findings

None. The diff is scoped to the refactor described above (`app/actions/update-gang.ts`, `app/lib/shared/gang-data.ts`, new `utils/gang-edition.ts` — 55 insertions / 32 deletions across the three files). No behavior change, `tsc --noEmit` clean on the touched files, no stray imports left behind.

## Verdict: **Approve**

---

joeseos:
@Djidiouf Follow-up review of the new commits pushed since the last review (`f9f98cf` merge from `main`, then `9101a9d` "Replace per-edition polarity checks with a capability registry").

## Previous findings

**1. Trade Points missing from the print/roster view — medium — Outstanding**
`app/gang/[id]/print/page.tsx` and `components/gang/print-gang.tsx` still don't surface `trade_points`. `print-gang.tsx` was touched by `9101a9d`, but only to swap `isLegacyEdition` for `hasSaveCharacteristic` — no change to the Trade Points gap. Still fine as a fast-follow, not blocking.

**2. Duplicated edition-slug resolution logic — low (DRY) — Partial (reopened, still low)**
The shared `utils/gang-edition.ts` extracted in the previous round has been deleted again. Reasoning in the commit message: `update-gang.ts` now folds the edition embed into the select it already runs (no second round trip needed), so `fetchGangEditionSlug` no longer has a reason to exist, and `gang-data.ts` inlines the same `?? null` flatten it always had. Net effect is two files each doing a 2-line `a?.editions?.slug ?? b?.editions?.slug ?? null`, rather than one shared helper — the original "single source of truth" is technically undone, but the duplication is now trivial (a null-coalesce over already-fetched data, not a duplicated query), so this is a reasonable trade and still not blocking.

**3. Migration timestamp dated ahead — low/nit — Outstanding**
`supabase/migrations/20260803120000_add_gangs_trade_points.sql` is unchanged. Non-issue at this point since today is 2026-08-03; noting only for completeness.

## New findings

None blocking. Reviewed `types/edition.ts`'s new `EDITION_CAPABILITIES` registry plus all 7 call-site updates (`update-gang.ts`, `gang-data.ts`, both admin fighter-type modals, `fighter-details-card.tsx`, `fighter-card.tsx`, `print-gang.tsx`):

- Behavior-preserving for every state that can occur today: `n23` → all three capabilities `false` (matches old `isLegacyEdition` gating); `n26` → all `true`; `undefined`/unrecognized slug → `NO_CAPABILITIES` (matches old `isLegacyEdition(undefined) === true` → features hidden). The only behavioral divergence is for a hypothetical edition slug that exists in the DB but hasn't been added to `EDITION_CAPABILITIES` — and `EditionSlug` being a closed union of `'n23' | 'n26'` means that can't happen without a compile error forcing the entry to be added first, so this is deliberate, not a gap.
- Confirmed no leftover references to `isLegacyEdition`, `EDITION_N23`, or `EDITION_N26` outside `types/edition.ts` itself — the refactor didn't leave a dangling import anywhere.
- `update-gang.ts`'s new combined select (`gang_types!gang_type_id(...)`, `custom_gang_types!custom_gang_type_id(...)`) preferring `gang_types` over `custom_gang_types` in the `??` chain is safe given `chk_gang_type_exclusive` (previously verified) — only one side can ever have data, so the precedence never actually matters.
- The `types/supabase.ts` revert (dropping `trade_points` from the `gangs` Row/Insert/Update types) has no functional impact — grepped the repo and this file isn't imported anywhere, so it's dead as the commit message claims and doesn't affect the typed Supabase client used elsewhere.

## Verdict: **Approve**

---

joeseos:
@Djidiouf Follow-up review of the new commit (`746f033`) pushed since the last review.

## Previous findings

**1. Trade Points missing from the print/roster view — medium — Resolved**
`app/gang/[id]/print/page.tsx:112-113` now threads `trade_points` and `edition_slug` (`gangBasic.edition_slug ?? null`) into `gangDataForClient`. `components/gang/print-gang.tsx` accepts both as optional props, destructures them once in `PrintGang`, and renders a `Trade Points` row right after `Reputation` at all three roster layouts (desktop/print, and the two other layouts around lines 1172 and 1624), each gated on `hasTradePoints(edition_slug)`. Verified `gangBasic.trade_points`/`edition_slug` are already produced by `getGangBasic` in `app/lib/shared/gang-data.ts`, so this is wired correctly end-to-end — n23 rosters are unaffected since `hasTradePoints('n23')` is `false`. The fighter-level Save gate (`fighter.edition_slug`) is untouched and doesn't collide with the new gang-level `edition_slug` variable.

**2. Duplicated edition-slug resolution logic — low (DRY) — no change this commit** (last state: intentionally reopened as trivial duplication per the previous review, not blocking)

**3. Migration timestamp dated ahead — low/nit — non-issue** (today is 2026-08-03, as previously noted)

## New findings

None. This commit is scoped to exactly what the commit message describes:
- The print/roster Trade Points gap above (`app/gang/[id]/print/page.tsx`, `components/gang/print-gang.tsx`).
- `types/edition.ts`: `capabilitiesFor()` now distinguishes an absent slug (edition not loaded, legitimate) from a present-but-unrecognized slug (registry/DB drift, a bug), warning via `console.warn` outside production only for the latter. Confirmed the DB currently only seeds `n23`/`n26` (checked `supabase/migrations/20260630095837_add_editions_and_root_edition_ids.sql` and `20260730150000_add_fighter_save_stat_and_n26.sql`), so this won't false-positive today — behavior for all current states is unchanged, this only adds visibility for future drift.

Ran `tsc --noEmit` — clean on all three touched files.

## Verdict: **Approve**

---

PR #1914, opened 2026-08-02
Support multi-class fighters in N26+ editions

joeseos:
## Summary
This PR extends fighter class selection to support multiple classes per fighter in N26+ editions, while maintaining single-class selection for the legacy N23 edition. It also refactors edition-checking logic to be more semantically clear.

## Key Changes

- **Multi-class support**: Changed fighter class state from a single string to an array (`selectedFighterClasses: string[]`) in both admin create and edit modals
- **Edition-aware UI**: Fighter class selection now renders as:
  - Checkboxes (multi-select) for N26+ editions
  - Dropdown (single-select) for N23 edition
- **Edition detection refactor**: Renamed `hasSaveCharacteristic()` to `isLegacyEdition()` for clearer intent and inverted logic throughout the codebase
- **Class matching logic**: Updated fighter type combo matching to compare full class lists rather than just the first class, preventing multi-class types from being incorrectly grouped with single-class variants
- **Edition switching safeguard**: When switching to N23 edition, automatically truncates selected classes to the first one to prevent hidden multi-class data from being submitted
- **Class formatting**: Added `formatFighterClasses()` helper to consistently render class arrays as comma-separated strings for display and comparison

## Implementation Details

- The `toggleFighterClass()` function maintains reference-list order when updating selections, ensuring consistent storage regardless of checkbox interaction order
- Fighter type combo deduplication now uses the full formatted class string as the identifier
- All four affected components (admin create/edit modals, fighter details card, gang card, print view) updated to use the new `isLegacyEdition()` function
- Backward compatibility maintained: existing single-class fighters load correctly into the multi-class UI

https://claude.ai/code/session_0144qXPkaumZUc8fTgpxH2QS

### DIFF
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -9,7 +9,7 @@ import { toast } from 'sonner';
 import { HiX } from "react-icons/hi";
 import { GangType, Equipment } from "@/types/gang";
 import { EditionSelect, useEditions } from '@/components/edition-select';
-import { hasSaveCharacteristic } from '@/types/edition';
+import { isLegacyEdition } from '@/types/edition';
 import { skillSetRank } from "@/utils/skillSetRank";
 import { equipmentCategoryRank } from "@/utils/equipmentCategoryRank";
 
@@ -52,7 +52,7 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
   const [delegationCost, setDelegationCost] = useState('');
   const [selectedGangType, setSelectedGangType] = useState('');
   const [editionId, setEditionId] = useState('');
-  const [selectedFighterClass, setSelectedFighterClass] = useState<FighterClass | ''>('');
+  const [selectedFighterClasses, setSelectedFighterClasses] = useState<string[]>([]);
   const [isLoading, setIsLoading] = useState(false);
   
   const [movement, setMovement] = useState('');
@@ -89,7 +89,7 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
 
   
 
-  const isCrew = selectedFighterClass && selectedFighterClass.class_name === 'Crew';
+  const isCrew = selectedFighterClasses.includes('Crew');
 
   const { data: gangTypes = [] } = useQuery<GangType[]>({
     queryKey: ['admin-gang-types'],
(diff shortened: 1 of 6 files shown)


joeseos:
@joeseos Reviewed against the full codebase (both admin modals, the `isLegacyEdition` refactor, and all four downstream consumers).

## Findings

**1. Edit modal allows saving a fighter type with zero classes — medium (correctness)**
`components/admin/admin-edit-fighter-type.tsx` has no guard against submitting an empty `selectedFighterClasses` array, unlike the create modal which explicitly blocks it:
```ts
// admin-create-fighter-type.tsx:191
if (!selectedGangType || selectedFighterClasses.length === 0 || !fighterType) {
  toast.error("Please fill in all required fields");
  return false;
}
```
`handleSubmit` in `admin-edit-fighter-type.tsx` (~line 913) only checks `!fighterType || !fighterToUpdate?.gang_type_id` and then submits `fighter_classes: selectedFighterClasses` (line 1088) unconditionally. The Save button's `disabled` prop (line 2349) is `!selectedSubTypeId || isLoading` — also no length check. Since the class picker is now checkboxes on N26+, an admin can uncheck every box and save.

I confirmed there's no backstop anywhere else in the chain: the DB column is `fighter_classes jsonb DEFAULT '[]'::jsonb NOT NULL` with no non-empty check constraint, and `app/api/admin/fighter-types/route.ts:800` just does `Array.isArray(data.fighterClasses) ? data.fighterClasses : []` — it accepts an empty array without complaint. The result would be a fighter type with no class at all, which several SQL functions branch on (e.g. `v_fighter_classes ?| array['Ganger', 'Exotic Beast']` for XP costs) and which the combo dropdown would render as `"Unknown Class"` (since `formatFighterClasses([])` returns `""`, which is falsy). Suggest mirroring the create modal's `selectedFighterClasses.length === 0` check here.

**2. `toggleFighterClass` / edition-switch truncation logic duplicated verbatim — low (DRY)**
Both modals define an identical `toggleFighterClass` and an identical "truncate to first class when switching to a legacy edition" block inside `handleEditionChange`. Not blocking — the two modals already duplicate a lot of similar logic — but if a third place needs this pattern, worth lifting into a shared hook (e.g. `useFighterClassSelection`).

## What's solid
- `isLegacyEdition` is a correct De Morgan inversion of the old `hasSaveCharacteristic` (`!!slug && slug !== N23` ⟺ `!(!slug || slug === N23)`) — verified all 6 call sites were renamed, no leftover `hasSaveCharacteristic` references anywhere in the repo.
- The combo-matching fix (`formatFighterClasses(f.fighter_classes) === fighterClass` instead of comparing just `fighter_classes?.[0]`) correctly prevents a "Ganger" combo from swallowing "Ganger, Specialist" fighter types — verified the `|`-delimited combo string (`type|class|gang_type_id`) can't collide with the `, `-joined class label since they use different separators.
- `editionId` in the edit modal reliably reflects the fighter type being edited: `gangTypeFilter` options are pre-filtered by `editionId`, and fighter types are fetched by the selected `gang_type_id`, so `allowMultipleClasses` can't get out of sync with the actual gang type's edition.
- Backward compatibility for existing single-class fighters loading into the multi-class checkbox UI checks out (`setSelectedFighterClasses(fighter.fighter_classes ?? [])`).
- No stray/unused variables left behind after the refactor (e.g., the old `const fighterClass = fighterClasses.find(...)` in the edit modal's submit was cleanly removed with no dangling references).
- Ran `tsc --noEmit` — no new type errors introduced by the changed files (only pre-existing tsconfig deprecation warnings unrelated to this diff).

## Verdict: **Changes requested** — just for finding #1; everything else is solid and #2 is optional.

---

PR #1917, opened 2026-08-03
Add N26 Lethality statline support for weapon profiles

joeseos:
## Summary
This PR adds support for the Necromunda 26 (N26) weapon statline, which replaces the Damage characteristic with Lethality. The system now dynamically renders different weapon stat fields based on the selected edition, allowing N23 and N26 weapons to coexist with their respective stat columns.

## Key Changes

- **Database Schema**: Added `lethality` column to `weapon_profiles` and `custom_weapon_profiles` tables to store the N26 Lethality characteristic (nullable to support pre-N26 profiles)

- **Edition Capabilities**: Extended `EditionCapabilities` with `lethalityStatline` flag to distinguish N26 (SR, LR, Str, AP, Lethality) from N23 (Rng S/L, Acc S/L, Str, AP, D, Am) statlines

- **Weapon Profile Types**: Updated `WeaponProfileInput` and `WeaponProfile` types to include `lethality` field; created `emptyWeaponProfile()` factory function to centralize blank profile initialization

- **Admin Equipment Forms**: 
  - Extracted weapon stat fields into reusable `WeaponProfileFields` component that conditionally renders edition-specific columns
  - Updated both create and edit equipment modals to use the new component
  - Added `handleEditionChange()` in create modal to clear incompatible stats when switching editions (e.g., clearing Damage/Ammo when switching to N26)

- **Weapon Display Tables**:
  - Updated `FighterCardWeaponTable` to render N26 or N23 headers/columns based on edition
  - Updated `EquipmentTooltip` to display appropriate stat columns for the edition
  - Both tables now accept `editionSlug` prop to determine which statline to display

- **Data Handling**: Updated all equipment queries and mutations to include the `lethality` field in SELECT and INSERT statements

- **Utility Functions**: Added `editionSlugOf()` helper in `edition-select.tsx` to resolve edition IDs to slugs for use in change handlers and other contexts where hooks cannot be used

## Implementation Details

- The `WeaponProfileFields` component uses conditional rendering with Tailwind grid classes (both `grid-cols-4 md:grid-cols-8` for N23 and `grid-cols-4 md:grid-cols-5` for N26) to maintain proper layout for each edition
- Edition switching in the create modal clears stats that don't apply to the new edition to prevent invalid data submission
- The `lethality` field is nullable in the database and defaults to empty string in form state, allowing seamless handling of pre-N26 profiles
- Weapon profile initialization is now centralized in `emptyWeaponProfile()` to ensure consistency across all forms and prevent drift when new stats are added

The customise UI was explicitly scoped out, so CustomWeaponProfileData and saveCustomWeaponProfiles lacking lethality is deliberate.


joeseos:
@joeseos Reviewed against the full codebase (fetched the PR branch locally and diffed against `main`). Solid feature work overall — one real gap found that will make Lethality silently not show up on the page players actually use most.

## Findings

**1. `getFighterEquipment` in `app/lib/shared/fighter-data.ts` was never updated to select `lethality` — high**
`app/lib/shared/fighter-data.ts:265-337`

The weapon-profile fetches in this function use explicit column lists (not `select('*')`):
```ts
.from('weapon_profiles')
.select(`
  id, profile_name, range_short, range_long, acc_short, acc_long,
  strength, ap, damage, ammo, traits, weapon_group_id, weapon_id, sort_order
`)
```
(same for the `weapon_id`-group query, and for the `custom_weapon_profiles` query right below it). None of these three selects include `lethality`, unlike the equivalent selects in `app/lib/shared/gang-data.ts`, which this PR did patch in four places.

`getFighterEquipment` is called directly in `app/fighter/[id]/page.tsx:78` and its result becomes the `equipment` fed into `FighterPageComponent` → `WeaponList` — exactly the component this PR wired up with `editionSlug` to render the Lethality column. Net effect: on the individual fighter page (the main page for managing a fighter's loadout), N26 weapons will always render `lethality` as `undefined` → displayed as `-`, even though the display code and admin data entry are both fully correct. It only works on the gang roster/print views, which go through `gang-data.ts`.

Fix: add `lethality,` to the three selects at lines ~275, ~296, and ~331 (the `getFighterVehicles` weapon-profile fetch further down at ~787-814 already uses `select('*')`, so that one is unaffected and needs no change).

**2. Player-created custom weapon profiles have no `lethality` field at all — medium (may be intentional scope)**
`types/equipment.ts:165-178` (`CustomWeaponProfileData`), `app/actions/customise/custom-weapon-profiles.ts`

The insert path for player-facing custom weapons (`saveCustomWeaponProfiles`) doesn't include `lethality` in its type or its inserted row, so a custom weapon created in an N26 gang has no way to carry a Lethality value. The read path (`getCustomWeaponProfiles`, `select('*')`) would return it fine if it were ever written. If custom-weapon creation is meant to support N26 too, this needs the same treatment as the admin form; if it's deliberately out of scope for this PR, worth a one-line note in the PR description so it isn't lost.

- `components/equipment/equipment-tooltip.tsx`: manually counted `<th>`/`<td>` cells for both the `usesLethality` and non-lethality branches, including the "no profile data" empty-row branch — column counts match exactly in all four combinations (data/no-data × N23/N26).
- `components/gang/fighter-card-weapon-table.tsx`: same check for the `colgroup`/header/row triplet — consistent.
- `handleEditionChange` in `admin-create-equipment.tsx` correctly blanks `acc_short/acc_long/damage/ammo` when switching to N26 and blanks `lethality` when switching away, preventing invalid cross-edition data from being submitted.
- `admin-edit-equipment.tsx`'s edition dropdown is a *search filter* over existing equipment (not a live-editable property of the loaded item — selecting a different edition clears the selected equipment and the edition snaps back to the loaded item's real edition), so it correctly doesn't need the same stat-clearing logic as the create modal. Initially looked like an inconsistency between the two modals; it isn't.
- `emptyWeaponProfile()` / `WeaponProfileFields` extraction is a good DRY win — centralizes the previously-duplicated blank-profile literal and the 8-field JSX block across both modals.
- Nullable handling of `lethality` in `app/api/admin/equipment/route.ts` (`profile.lethality || null`) is consistent with how `damage`/`ap`/etc. are already handled.
- Migration is additive and idempotent (`ADD COLUMN IF NOT EXISTS`), matches existing migration conventions in this repo.

## Verdict
**Changes requested** — finding #1 needs a fix before merge, since it silently breaks the headline feature on the fighter detail page. Finding #2 is worth a confirmation from you on whether it's in scope.

---

PR #1918, opened 2026-08-03
Added Edition filtering in Add Fighter Type and Edit Fighter Type for…

Djidiouf:
… Classes, Equipment and Skills

### DIFF
--- a/app/api/admin/fighter-classes/route.ts
+++ b/app/api/admin/fighter-classes/route.ts
@@ -13,10 +13,11 @@ export async function GET() {
       return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
     }
 
-    // Simple direct query with error logging
+    // fighter_classes holds one row per class per edition; clients filter by
+    // edition_id so same-named classes across editions stay distinguishable
     const { data: fighterClasses, error } = await supabase
       .from('fighter_classes')
-      .select('id, class_name')
+      .select('id, class_name, edition_id')
       .order('class_name');
 
     if (error) {
@@ -27,14 +28,8 @@ export async function GET() {
       }, { status: 500 });
     }
 
-    if (!fighterClasses || fighterClasses.length === 0) {
-      return NextResponse.json({ 
-        error: 'No data found', 
-        details: 'Check RLS policies and table data' 
-      }, { status: 404 });
-    }
-
-    return NextResponse.json(fighterClasses);
+    // An empty list is a valid answer, not an error
+    return NextResponse.json(fighterClasses ?? []);
 
   } catch (error) {
     console.error('Error in GET fighter classes:', error);
@@ -46,4 +41,4 @@ export async function GET() {
       { status: 500 }
     );
   }
-} 
+}
--- a/app/api/admin/fighter-types/route.ts
+++ b/app/api/admin/fighter-types/route.ts
@@ -90,6 +90,7 @@ export async function GET(request: Request) {
           fighter_type,
           gang_type_id,
           gang_type,
+          edition_id,
           fighter_classes,
           fighter_sub_type_id,
           cost,
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -21,11 +21,13 @@ interface AdminCreateFighterTypeModalProps {
 interface FighterClass {
   id: string;
   class_name: string;
+  edition_id?: string | null;
 }
 
 interface SkillType {
   id: string;
   skill_type: string;
+  edition_id?: string | null;
 }
 
 interface Skill {
@@ -43,6 +45,7 @@ interface EquipmentWithId extends Equipment {
   cost: number;
   availability?: string | null;
   equipment_category: string;
+  edition_id?: string | null;
 }
 
 export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFighterTypeModalProps) {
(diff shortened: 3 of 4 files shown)


joeseos:
@Djidiouf Thanks for this — reviewed against the diff and surrounding code. Summary and findings below.

## Findings

**1. `groupedSkillTypes` memoization is now defeated — medium**
`components/admin/admin-edit-fighter-type.tsx:328-362`

```ts
const filteredSkillTypes = editionId
  ? skillTypes.filter(type => type.edition_id === editionId)
  : skillTypes;
...
const groupedSkillTypes = useMemo(() => {
  ...
}, [filteredSkillTypes]);
```

`filteredSkillTypes` is a plain derived value (`.filter()`/passthrough), so it gets a new array reference on every render. The `useMemo` dependency array previously pointed at `skillTypes` (a referentially-stable react-query result), so the memo actually skipped work across renders. Now that the dependency is `filteredSkillTypes`, the reference changes every render and the memo recomputes every time regardless — the comment directly above it ("Memoize grouped skill types to avoid expensive computation on every render") no longer holds.

Fix: wrap `filteredSkillTypes` itself in `useMemo(() => ..., [skillTypes, editionId])`, or fold the edition filter into the existing `groupedSkillTypes` useMemo body with `[skillTypes, editionId]` as the dependency array.

**2. Unmemoized `.filter()` calls on every render — low**
Both files: `filteredGangTypes`, `filteredFighterClasses`, `filteredSkillTypes`, `filteredEquipment`.

These recompute on every render (e.g. on every keystroke in an unrelated field), which matches the pre-existing `filteredGangTypes` pattern so it's not a new regression, but the equipment list in particular can be large. Not blocking, but worth a `useMemo` pass if this component gets touched again.

- `app/api/admin/fighter-classes/route.ts`: correctly adds `edition_id` to the select and stops treating an empty result as a 404 — good fix.
- `app/api/admin/fighter-types/route.ts`: adding `edition_id` to the single-fighter-type select is required for `admin-edit-fighter-type.tsx:573` (`setEditionId(data.edition_id)`) to work when opening an existing fighter type for edit — correctly wired.
- The remaining raw `equipment.find(...)` calls in the edit modal (lines ~2058, ~2112) are display lookups for already-selected equipment by id, not selection lists, so they're correctly left unfiltered by edition — filtering only the *selectable* dropdowns is right.
- `handleEditionChange` in both files consistently prunes `selectedFighterClasses`, `selectedEquipment`, `equipmentListSelections`, `equipmentDiscounts`, `selectedAdjustedCostEquipment`, `selectedSkillType`/`skillTypeToAdd`/`skillAccess` against the new edition, so stale cross-edition selections can't silently survive a filter change and get submitted.
- Edition filtering assumes every row in `fighter_classes`/`equipment`/`skill_types` has a non-null `edition_id` matching exactly one edition (rows with `edition_id === null` would be excluded once a filter is applied). Checked the migration (`20260630095837_add_editions_and_root_edition_ids.sql`) — all existing catalog rows were backfilled to a non-null edition, and this matches the pre-existing `filteredGangTypes` convention already in use, so no new gap here.
- No security/auth changes — all routes still gate on `checkAdmin`.

## Verdict
**Approve** — the feature logic is correct and the cleanup-on-edition-change handling is thorough. Item 1 is worth a follow-up fix (small, isolated) but not a blocker for merge.

---

joeseos:
@Djidiouf Follow-up review on the latest commit (`d90bf90` → `6b19453`), scoped to just that diff since the previous review.

## Previous findings

**1. `groupedSkillTypes` memoization defeated (medium) — resolved**
`filteredGangTypes`, `filteredFighterClasses`, `filteredSkillTypes`, `filteredEquipment` are now wrapped in `useMemo(..., [source, editionId])`, so `filteredSkillTypes` is referentially stable again and `groupedSkillTypes` only recomputes when `skillTypes`/`editionId` actually change.

Nice catch also spreading into `[...filteredSkillTypes]` before `.sort()` — now that `filteredSkillTypes` is a stable memoized reference, sorting it in place (as the old code did) would have mutated the cached array across renders. Cloning before sort avoids that.

**2. Unmemoized `.filter()` calls in both files (low) — partially addressed**
Fixed in `admin-edit-fighter-type.tsx`. `admin-add-fighter-type.tsx` still has the same unmemoized `filteredGangTypes`/`filteredFighterClasses`/`filteredSkillTypes`/`filteredEquipment` pattern. That was already flagged as non-blocking pre-existing behavior last round, so leaving it as-is here is fine — just noting it's still open if that file gets touched again.

## New issues in this diff

None. The change is a minimal, correctly-scoped fix: `Record<string, typeof filteredSkillTypes>` → `Record<string, SkillType[]>` is a clean explicit-type swap (matches the local `SkillType` interface), and `useMemo` was already imported.

## Verdict

**Approve.**

---

joeseos:
@Djidiouf Follow-up review on the latest commit (`6b19453` → `d24e6fc`), scoped to just that diff since the last review.

## Previous findings

**1. `groupedSkillTypes` memoization defeated (medium) — resolved** (previous round)
No change needed here this round; still holds.

**2. Unmemoized `.filter()` calls in both files (low) — resolved**
`admin-create-fighter-type.tsx` now wraps `filteredGangTypes`, `filteredFighterClasses`, `filteredSkillTypes`, `filteredEquipment` in `useMemo(..., [source, editionId])`, matching the pattern already applied to `admin-edit-fighter-type.tsx` last round. This was the last outstanding item — no longer open.

## New issues in this diff

None — and this commit also catches a bug that slipped through the last review: once `filteredEquipment`/`filteredSkillTypes` became stable memoized references, the pre-existing `.sort(...)` calls on them (equipment dropdown at `admin-edit-fighter-type.tsx:2104` and the equivalent spots in `admin-create-fighter-type.tsx:760,915`) would have mutated the memoized array in place via `Array.prototype.sort`. This commit correctly guards every such call site with `[...filteredEquipment].sort(...)` / `[...filteredSkillTypes].sort(...)`. I checked all remaining `.sort()` call sites in both files (e.g. `admin-edit-fighter-type.tsx:2118`, `admin-create-fighter-type.tsx:929`) — those operate on freshly derived arrays from `.map()/.filter()` chains, not the memoized `filtered*` values, so they don't need the same guard.

The `Record<string, typeof filteredSkillTypes>` → `Record<string, SkillType[]>` type swap in `admin-create-fighter-type.tsx` is consistent with the same fix applied to the edit modal last round; `SkillType` is defined locally in this file so the cast resolves correctly.

## Verdict

**Approve.** Both previously-outstanding items are now resolved and no new issues were introduced.

---

PR #1921, opened 2026-08-03
Added gated Equipment Categories and Trade Points field for N26 equip…

Djidiouf:
…ment


joeseos:
## Code Review

@Djidiouf This PR is already merged, but reviewing per the standard checklist below — one finding worth a fast follow-up.

### Findings

**1. `trade_points` is a `text` column that can hold the sentinel `'E'`, but this UI/API can only round-trip numbers, and silently drops `'E'` to `0` on any unrelated edit — high**

Checked the current schema (`supabase/schema/schema.public.sql`):
```sql
CREATE TABLE public.equipment (
  ...
  trade_points text DEFAULT '0'::text NOT NULL
);
COMMENT ON COLUMN public.equipment.trade_points IS
  'N26 Trade Points cost (text: numeric string or E). Surfaced/charged only for N26 gangs.';
```
Same shape for `custom_equipment.trade_points`. So the column is explicitly designed to hold either a numeric string *or* the literal `'E'` (mirroring the Availability letter system's `E` for "Exclusive").

This PR's UI and API only handle the numeric case:
- `admin-edit-equipment.tsx`: `setTradePoints(equipmentDetails.trade_points != null ? String(equipmentDetails.trade_points) : '0')` — if the loaded value is `"E"`, this sets `tradePoints` state to `"E"`.
- That state feeds `<Input type="number" value={tradePoints} ...>` — a number input can't display `"E"`, so it renders blank while React state still holds `"E"`.
- On submit: `trade_points: showTradePoints ? Number(tradePoints) : 0` → `Number("E")` is `NaN`.
- `JSON.stringify` serializes `NaN` as `null` in the request body.
- The API route (`app/api/admin/equipment/route.ts` PATCH): `trade_points: trade_points != null ? Number(trade_points) : 0` → since the received value is `null`, this falls through to `0`.

Net effect: opening *any* equipment item whose `trade_points` is `'E'` in the edit modal — even just to fix an unrelated typo in the name — and saving will **silently overwrite `'E'` with `0`**, with no error surfaced to the admin. Same numeric-only constraint applies to `admin-create-equipment.tsx`, so there's currently no way to *set* `'E'` through the admin UI either.

I can't tell from the diff whether the column was already `text` when this PR was written or changed afterward (the only migrations that touch `trade_points` — `20260803120000_add_gangs_trade_points.sql` and `20260803132759_add_trade_points_to_equipment_tables.sql` — both declare it `numeric`, with no migration in the repo changing it to `text`; the live schema snapshot shows `text` regardless). Either way, as merged, the admin equipment form can't represent the full value space the column is documented to support. Suggest either a text input with validation (numeric or `'E'`) in place of `type="number"`, or a small toggle/checkbox for the exclusive case — plus reconciling the missing migration for the `numeric` → `text` type change so a fresh migration run matches production.

### Checklist
- **Correctness**: Edition-gating logic (`hasTradePoints`/`showAvailability`) is correct and consistent between create/edit. The "preserve loaded `availability` when hidden" fix in `admin-edit-equipment.tsx` (`equipmentDetails?.availability ?? 'C'`) is a nice touch — avoids clobbering data outside the visible form, unlike the `trade_points` issue above.
- **Edge cases / null handling**: `data ?? []` added defensively to the categories GET route. `if (showTradePoints && tradePoints === '')` guard correctly blocks an empty-string submit (which would otherwise silently become `0` via `Number('')`). The one gap is the `'E'` case above.
- **Security**: No new surface; same admin-gated routes as before.
- **Performance**: Category/equipment filtering by edition is done client-side over already-fetched, cached (`staleTime: 5min`) lists — fine at this scale.
- **Readability/maintainability**: `handleEditionChange` in `admin-edit-equipment.tsx` clearly comments the non-obvious bit (`categoryFilter` is a name, not an id, so the mismatch check needs both name+edition_id).
- **Conventions**: Matches the established `EditionSelect`/edition-gating pattern from the preceding PRs; no `CLAUDE.md` in this repo to check against.
- **DRY/YAGNI**: `EquipmentCategory` interface is duplicated identically between `admin-create-equipment.tsx` and `admin-edit-equipment.tsx` — minor, pre-existing pattern in this codebase (categories/equipment types aren't shared between these two files already), not something this PR needs to fix.

### Verdict: **Approve, with a follow-up recommended** for finding #1 — worth a quick fix since it's a silent-data-loss path on the feature's own core field, not a hard blocker on an already-merged PR.

---

PR #1924, opened 2026-08-04
Adjusted the UI on the class checkboxes in admin

joeseos:
Adjusted the fighter class selection in admin create fighter and edit fighter. Removed the border and filters alphabetically on class name


joeseos:
@joeseos Reviewed against the full codebase (fetched the PR branch and diffed against `main`). This is a small, low-risk, purely cosmetic change — no data/state logic touched.

## Findings

**1. Minor layout gap in the edit modal's Base Cost/Alignment/Delegation Cost row — low**
`components/admin/admin-edit-fighter-type.tsx` (the `grid grid-cols-1 md:grid-cols-2 gap-4 mb-4` block right after the new full-width Fighter Class section)

When `allowMultipleClasses` is true, the single-select Fighter Class field is dropped out of this 2-column grid (moved to its own full-width block above), leaving 3 items — Base Cost, Alignment, Delegation Cost — in a `grid-cols-2` container. That's an odd count for 2 columns, so on desktop the last row (Delegation Cost) sits alone with an empty gap to its right. Purely cosmetic, easy to ignore, but since this PR is specifically about tidying this form's layout it might be worth a quick look (e.g. giving Delegation Cost its own row, or moving Alignment to fill the gap).

- `fighterClassesForDisplay` (the new alphabetically-sorted copy) is only used for rendering — `toggleFighterClass` still builds the stored `selectedFighterClasses` array from the unsorted `filteredFighterClasses`, so the persisted class order is unaffected by this purely-visual sort. Good separation.
- `useMemo` dependencies are correct: `filteredFighterClasses` is itself already memoized (`[fighterClasses, editionId]`), so `fighterClassesForDisplay`'s `[filteredFighterClasses]` dependency is a stable reference and won't recompute every render.
- Both the checkbox grid (multi-class/N26) and the `<select>` dropdown (single-class) now use the same sorted list consistently, in both the create and edit modals.
- Removing `border rounded-md p-2` from the checkbox container matches the PR description, and the responsive column bump (`grid-cols-2 sm:grid-cols-3 md:grid-cols-4`) makes sense now that the block is full-width instead of half-width.
- Checked JSX structure in both files after the conditional-rendering rework (checkbox block moved out of the shared grid into its own full-width block) — tags balance correctly, and container spacing (`space-y-4` in the create modal, explicit `mb-4` in the edit modal) still applies to the new block in both cases.

## Verdict
**Approve** — finding #1 is optional polish, not a blocker.

---

PR #1926, opened 2026-08-04
Added Trade Points and its consumption to Equipment modal

Djidiouf:


joeseos:
@Djidiouf Reviewed this against the rest of the codebase (equipment purchase flow, `get_equipment_detailed_data` RPC, admin equipment API, and the migrations). Overall the feature is well-structured — trade points is threaded through the modal/purchase-modal/hooks consistently, the atomic conditional-update pattern in `deductGangTradePoints` correctly guards against concurrent-purchase races, and reusing the existing "trust client cost" convention (already used for `manual_cost`) for `manual_trade_points` is consistent with how the codebase already handles this trust boundary — not a new issue.

A few correctness issues worth addressing before merge:

### Findings

**1. (Medium) No compensating rollback if Trade Points deduction fails after a resource-based purchase already succeeded**
`app/actions/equipment.ts:525-536` deducts a campaign resource/reputation cost, then `app/actions/equipment.ts:592-601` deducts Trade Points further down — both gated independently (`isResourcePurchase` vs `hasTradePoints(editionSlug)`), so they can both apply to the same purchase (e.g. an N26 gang buying from a custom trading post with a resource cost). If the TP deduction then fails (insufficient TP, or the `.gte()` race guard in `deductGangTradePoints` losing a race), the code deletes the `fighter_equipment` row but never restores the already-deducted resource/reputation — the gang permanently loses the resource with nothing to show for it.
*Fix:* either skip TP deduction on resource-based purchases if that's intended, or track and restore the resource amount in the TP failure's catch block.

**2. (Medium) `resolveEquipmentTradePoints` can diverge from the RPC's discount-scope selection, so the price shown vs. price charged can differ**
`get_equipment_detailed_data.sql`'s `origin_discount_equip` CTE considers **all** `equipment_discounts` rows for an equipment item (regardless of whether `trade_points` is set) when deciding whether to use the origin-based matching branch. `resolveEquipmentTradePoints` (`app/actions/equipment.ts:47-51`) instead pre-filters to `.not('trade_points', 'is', null)` *before* computing `hasOriginDiscount` (line 55-57). For an item that has both an origin-scoped cost-only discount (`trade_points` NULL) and a separate gang-type-scoped TP discount, the RPC and the buy action can select different discount rows — the TP cost shown in the purchase modal (RPC-driven) can differ from what's actually deducted at purchase time.

**3. (Medium) Lexicographic tie-break on `trade_points` text can pick the wrong "minimum"**
Both the RPC (`MIN(ed.trade_points) FILTER (...) AS best_trade_points` in `supabase/migrations/20260804122525_update_get_equipment_detailed_data_edition_trade_points.sql`) and the TS helper (`matching.map(d => d.trade_points).sort()[0]` in `app/actions/equipment.ts:78`) sort `trade_points` as **text**, not numerically. For double-digit values, `"10" < "2"` lexicographically, so when multiple discounts apply to the same equipment/gang scope, the "MIN" pick can select a numerically larger TP value as the deterministic pick — the opposite of the presumably-intended cheapest-for-player tie-break, and it's wrong identically on both sides so it won't even surface as a display/charge mismatch. Worth casting to numeric for the comparison (or padding) in both places if any catalog item is ever expected to reach double-digit TP.

**4. (Low) Sequential DB round-trips added to the purchase hot path**
`app/actions/equipment.ts:540-587`: `fetchGangEditionSlug`, then (for non-manual TP) the `fighters` row, then the legacy-fighter-type lookup, then `gang_affiliation`, then the `equipment_discounts` query in `resolveEquipmentTradePoints` — all awaited in series for every N26 purchase. The gang+vehicle fetch just above (`Promise.all` at line 285) shows the pattern already used elsewhere in this same function; worth batching the independent fighter/legacy/affiliation lookups similarly.

**5. (Low) Admin equipment API doesn't validate `trade_points` format server-side**
`app/api/admin/equipment/route.ts` (POST ~436, PATCH ~589) trims/uppercases `trade_points` but doesn't enforce the `^(E|\d+)$` shape the client-side forms validate (`admin-create-equipment.tsx:254`, `admin-edit-equipment.tsx:308`). Not exploitable — `parseTradePointsCost` degrades any unrecognized value to 0 — but arbitrary strings can land in the catalog since this is a directly-editable admin endpoint.

### Checklist
- Correctness: see #1–#3 above
- Edge cases / null handling: generally solid (`?? 0` defaults throughout, graceful `parseTradePointsCost` fallback for malformed input)
- Security: no new issues; `manual_trade_points` client-trust is consistent with the existing `manual_cost` trust model
- Performance: see #4 (minor)
- Readability/maintainability: good — clear comments explaining non-obvious tie-break/discount-scope decisions
- Conventions: no `CLAUDE.md` found in the repo to check against
- DRY/YAGNI: no concerns; `buildConfirmOptions` in `purchase-modal.tsx` is a nice dedup of the repeated `onConfirm` call sites

Mainly on #1 and #2 — both are real data-integrity/financial-correctness risks, not style nits. #3 is worth a quick fix given how cheap it is. #4/#5 are optional follow-ups.

---

joeseos:
@Djidiouf Reviewed the latest commit (`f67973e` → `e57efd7`) against the previous review round.

### Previous findings

1. **(Medium) No compensating rollback on TP-deduction failure after a resource-based purchase — Resolved.** `app/actions/equipment.ts:622-643` now restores the already-deducted resource/reputation (via the existing `returnGangResource`/`returnGangReputation` helpers) in the TP-deduction catch block, mirroring the pattern already used for the resource-deduction catch block above it.
2. **(Medium) `resolveEquipmentTradePoints` discount-scope divergence from the RPC — Resolved.** The `.not('trade_points', 'is', null)` pre-filter was removed from the `equipment_discounts` query (`equipment.ts:67-70`), so origin-branch detection now considers cost-only discount rows too, matching `get_equipment_detailed_data`'s `origin_discount_equip` CTE. Traced through both the origin and gang-type branches — the matching logic lines up with the SQL.
3. **(Medium) Lexicographic tie-break on `trade_points` text — Resolved.** Both sides now sort numerically: `pickCheapestTradePoints`/`tradePointsNumericKey` in TS (`equipment.ts:29-45`) and the `ARRAY_AGG(... ORDER BY numeric key)` rewrite in `get_equipment_detailed_data.sql:172-182`, shipped as migration `20260805074127_fix_equipment_trade_points_numeric_min.sql`. Tie-break order (numeric asc, then text asc) matches on both sides.
4. **(Low) Sequential DB round-trips in the purchase hot path — Partially addressed.** The fighter row and gang-affiliation lookups are now parallelized via `Promise.all` (`equipment.ts:573-588`). The legacy-fighter-type lookup remains sequential, but that's inherent (it depends on the fighter row's `fighter_gang_legacy_id`). Reasonable improvement, not fully flattened but the original concern (needlessly serial *independent* queries) is fixed.
5. **(Low) Admin equipment API didn't validate `trade_points` format — Resolved.** `normalizeAdminTradePoints` (`app/api/admin/equipment/route.ts:16-25`) now enforces the `^(E|\d+)$` shape server-side and returns a 400 on both POST and PATCH, before any writes occur (so no partial-state risk from the added validation).

### New findings

None that rise above nitpick level. The restore-on-failure catch (`equipment.ts:638-640`) only `console.error`s if the compensating restore itself fails, so a gang could in a rare double-failure scenario permanently lose the resource with only a server log as a trace — but this mirrors the existing `returnCostResource` best-effort pattern elsewhere in the codebase, so I'm not treating it as a blocking regression introduced by this commit.

Ran `tsc --noEmit` — clean (only pre-existing config deprecation warnings, unrelated to this diff).

### Verdict: **Approve**

All three Medium-severity data-integrity findings from the previous round are resolved with correct, verified fixes, and the Low-severity items are resolved or reasonably improved.

---

joeseos:
@Djidiouf Reviewed the new commits since the last approval (`e57efd7` → `41cb08a` merge → `a0ee740` "Trim Trade Points implementation to scope"). This is a significant rework, not just a fixup, so I traced it against the previous review's findings and against the rest of the purchase flow.

### Previous findings

1. **(Medium) No compensating rollback on TP-deduction failure — still resolved.** The restore-on-failure catch (`equipment.ts:515-524`) is preserved through the rewrite.
2. **(Medium) `resolveEquipmentTradePoints` RPC-scope divergence — moot, code removed.** The whole TS re-implementation of the discount CTEs (`resolveEquipmentTradePoints`, `pickCheapestTradePoints`) was deleted. Verified the premise holds: every `onConfirm` path in `purchase-modal.tsx`'s `buildConfirmOptions` (lines 98, used by all 6 `onConfirm` call sites) includes `tradePoints: manualTradePoints.trim()` whenever `showTradePoints` is true, and both purchase entry points (`use-purchase-equipment.ts` and the optimistic `onPurchaseRequest` path in `fighter-equipment-list.tsx`) thread that straight into `manual_trade_points` — so the RPC-mirroring branch really was dead code for N26. Cost now comes from the client the same as `manual_cost`, consistent with the existing trust model.
3. **(Medium) Lexicographic tie-break on `trade_points` text — still resolved.** The numeric-sort fix in `get_equipment_detailed_data.sql` is untouched by this commit (still lives in `supabase/functions/`, lines 172-182); confirmed the two now-deleted migrations were the ones that historically shipped that same function body via `db push`, while `supabase/functions/*.sql` is deployed independently by `.github/workflows/deploy_supabase_functions.yml` on push to `main` — so dropping those migrations doesn't undo the fix or lose it on merge.
4. **(Low) Sequential DB round-trips — resolved further.** The entire N26 discount-resolution round-trip chain (`fetchGangEditionSlug`, fighter/legacy/affiliation lookups) is gone; edition slug now comes off the already-fetched `gangs` embed (`gang_types(editions(slug))` / `custom_gang_types(editions(slug))`), matching the `update-gang.ts` pattern. Confirmed `gang.gang_origin_id`/`gang_affiliation_id` (only needed by the removed resolver) have no other references left.
5. **(Low) Admin API `trade_points` validation — still resolved**, now shared via `isValidTradePoints` in `utils/campaigns/resources.ts` instead of duplicated regexes in the admin route/forms.

### New findings

**1. (High) `deductGangTradePoints` reintroduces the exact race the previous review praised as fixed — lost-update on concurrent purchases.**
`app/actions/equipment.ts:427` reads `currentTradePoints` right after the initial gang fetch (before the equipment insert, before the resource deduction, before any other awaited work), then `app/actions/equipment.ts:514` uses that stale value at the very end of the purchase to compute the new balance:

```ts
// resources.ts
.update({ trade_points: current - amount })   // current is the caller's stale read
.eq('id', gangId)
.gte('trade_points', amount)                  // guard checks live value ≥ amount, NOT ≥ current
```

The `.gte('trade_points', amount)` guard only proves the live row can afford `amount` at write time — it does not protect the `current - amount` value being written from being computed off a stale base. Concretely: gang has 10 TP. Purchase A (needs 6) and purchase B (needs 3) both read `currentTradePoints = 10` near-simultaneously. A finishes first: `.gte('trade_points', 6)` passes against live value 10, sets `trade_points = 10 - 6 = 4`. B finishes next: `.gte('trade_points', 3)` passes against the *now-live* value 4 (4 ≥ 3), but B still writes `trade_points = 10 - 3 = 7` — silently erasing A's already-applied deduction. Net effect: the gang ends up with 7 TP instead of the correct 1, i.e. A's spend is refunded for free.

This is the same class of bug finding #1 from the first review round was about, and the first round's fix (fetch-then-guarded-update, both back-to-back inside `deductGangTradePoints` with no intervening awaits) is exactly what prevented it. This refactor widened the gap between "read current" and "write current − amount" to span the entire rest of the purchase (equipment insert + resource/reputation deduction), which is a large enough window for concurrent purchases (double-click, two tabs) to hit in practice.

*Fix:* have `deductGangTradePoints` re-fetch `trade_points` immediately before the update (as it did pre-refactor), or drop the `current` param entirely and use an atomic SQL expression (`trade_points = trade_points - amount`) instead of computing the new value in application code — either removes the stale-read window.

### Checklist
- Edge cases / null handling: `manual_trade_points` empty-string edge case is UI-gated (confirm button blocked by `isValidTradePoints`) and behaves as before if bypassed — no regression.
- Cache invalidation: verified `invalidateGangCredits` (now folded with `BASE_GANG_BASIC`, per commit message) is reached on all three purchase paths (fighter/stash/vehicle) via `invalidateEquipmentPurchase` / `invalidateGangStash` / `invalidateFighterDataWithFinancials` → `invalidateGangFinancials` — trade points still refresh correctly despite the removed explicit `revalidateTag` call.
- Readability: the removed indirection (fewer round-trips, edition slug off an existing embed) is a net readability win; comments explain the non-obvious ordering decisions well.
- Conventions: consistent with existing patterns elsewhere in the file (still no `CLAUDE.md` to check against).

Only on the new finding above — it's a real regression in a currency-integrity guarantee the previous round specifically fixed and verified. Everything else from the previous round remains resolved, and the rest of this rewrite is a solid simplification.

---

joeseos:
@Djidiouf Reviewed the latest commit (`a206ba3` "Move reputation and trade points into the gang resources cache entry") against the previous review round.

### Previous findings

1. **(High) `deductGangTradePoints` stale-read race on concurrent purchases — still outstanding.** Not touched by this commit; `app/actions/equipment.ts:427` still reads `currentTradePoints` from the initial gang fetch and `equipment.ts:514` still passes that stale value into `deductGangTradePoints` at the end of the purchase flow, well after the equipment insert and resource/reputation deduction. The lost-update scenario described in the previous review (two concurrent purchases racing on `.gte('trade_points', amount)` while both compute `current - amount` off the same stale base) still applies. This is the only blocker from last round and needs to be fixed before merge.
2. Findings #2–#5 from the first round (discount-scope divergence, lexicographic tie-break, sequential round-trips, admin validation) — **still resolved**, unaffected by this commit.

### New changes in this commit

This commit is scoped entirely to cache invalidation and isn't related to Trade Points purchase logic — it consolidates `reputation`/`trade_points` out of `GangBasic`/`BASE_GANG_BASIC` and into a renamed `getGangResources`/`BASE_GANG_CREDITS` (all three spendable resources: credits, reputation, trade points, now share one cache entry and tag). This also fixes reputation going stale after a reputation-only purchase, which previously invalidated no tag containing it.

Verified:
- No leftover references to the old `getGangCredits` or `GangBasic.reputation`/`.trade_points` anywhere in the codebase — the three call sites (`app/fighter/[id]/page.tsx`, `app/gang/[id]/page.tsx`, `app/gang/[id]/print/page.tsx`) are consistently updated to read `gangResources.reputation`/`.trade_points`.
- `invalidateGangCredits` now only revalidates `BASE_GANG_CREDITS` (no longer implicitly cascading to `BASE_GANG_BASIC`), and `invalidateGangCreation` was updated to explicitly revalidate both tags separately — net effect on gang creation is unchanged.
- Checked every other `invalidateGangCredits` call site (`chem-alchemy.ts`, `edit-fighter.ts`, `fighter-advancement.ts`, `api/fighters/[id]/route.ts`, `api/gangs/[id]/route.ts`) — none of those mutations touch gang name/color/alignment/notes, so dropping the implicit `BASE_GANG_BASIC` cascade from `invalidateGangCredits` doesn't leave anything stale.
- `app/actions/update-gang.ts:392-395` correctly routes reputation/trade-points changes through `invalidateGangCredits` now that they live in that cache entry.
- Ran `tsc --noEmit` — clean (only the same pre-existing config deprecation warnings as before, unrelated to this diff).

No new issues found in this commit.

Blocked only on the still-open High-severity race condition in `deductGangTradePoints` from the previous round (not addressed by this commit). Everything else, including this commit's cache refactor, looks correct.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`a0ee740` → `a206ba3`, "Move reputation and trade points into the gang resources cache entry"). This is a cache-layer refactor (renames `getGangCredits` → `getGangResources`, drops `reputation`/`trade_points` from `getGangBasic`'s select+interface, folds their invalidation into `invalidateGangCredits`/`BASE_GANG_CREDITS`) — traced it for stale references and invalidation-path correctness rather than re-reviewing the whole PR.

### Previous findings (carried over, unaffected by this commit)

1–3, 5. **Still resolved** — rollback-on-failure, RPC discount-scope divergence (moot, code removed), numeric tie-break, and admin `trade_points` validation are all untouched by this commit and remain fixed.
4. **Still resolved further** — sequential round-trip removal from the previous commit stands.
6. **(High) `deductGangTradePoints` stale-read lost-update race — still outstanding, not addressed.** `utils/campaigns/resources.ts:269-285` is untouched by this commit; it still takes a `current` value read by the caller earlier in the purchase flow and writes `trade_points: current - amount`, so the lost-update scenario flagged last round (two concurrent purchases both reading the same `current`, second write clobbering the first's deduction) still applies. Flagging again since it wasn't in scope of this commit's diff, not because anything regressed here.

### New findings

**1. (Low) Stale docstring on `getGangBasic`.** `app/lib/shared/gang-data.ts:190` still reads `"Get gang basic information (name, type, reputation, etc. - excludes credits)"`, but this same commit removes `reputation`/`trade_points` from `getGangBasic`'s select and from the `GangBasic` interface — they now live only in `getGangResources`. Worth a one-line comment fix so future readers don't go looking for reputation in the wrong function.

Everything else checks out:
- Grepped the full tree at `a206ba3` for `getGangCredits` and `gangBasic.reputation`/`gangBasic.trade_points` — no stale call sites left; all three consumers (`gang/[id]/page.tsx`, `fighter/[id]/page.tsx`, `gang/[id]/print/page.tsx`) were updated consistently to read off `gangResources`.
- `invalidateGangCreation` was correctly updated to explicitly revalidate `BASE_GANG_BASIC` in addition to `invalidateGangCredits`, now that the latter no longer implicitly covers it.
- `update-gang.ts`'s reputation/trade-points-changed branch now calls `invalidateGangCredits` instead of touching `BASE_GANG_BASIC` directly — matches the new cache ownership.
- `tsc --noEmit` on the commit is clean (only the pre-existing config deprecation warnings).

Side note, out of scope for this diff since the file isn't touched: `app/api/gangs/[id]/route.ts`'s reputation-only update path (`reputation !== undefined && reputation_operation`, ~line 227) never sets a "changed" flag and never calls `invalidateGangCredits`/any cache invalidation — so a reputation edit through that endpoint won't refresh the now-shared resources cache entry. Predates this commit (same gap existed against `BASE_GANG_BASIC` before the fold), so not a regression from this change, just worth a follow-up given this commit is actively consolidating reputation's cache ownership.

Only on the still-outstanding High finding (#6) from the previous round — this commit itself is a clean, correctly-wired refactor.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`be48732`, "Narrow the Trade Points deduction window") against the previous review round's sole outstanding blocker.

### Previous findings

1. **(High) `deductGangTradePoints` stale-read lost-update race — Resolved.** `app/actions/equipment.ts:514` no longer passes the `currentTradePoints` read from the top of the purchase flow into `deductGangTradePoints`; the function (`utils/campaigns/resources.ts:270-299`) now does its own `SELECT trade_points ... .single()` immediately before the guarded `UPDATE ... .gte('trade_points', amount)`, collapsing the window from "spans the whole purchase" down to two back-to-back round-trips. This is the exact fetch-then-guarded-update shape `deductGangReputation` (`resources.ts:183-214`) already uses and that the first review round accepted as the fix for this same bug class — so it's consistent with the codebase's existing convention, not a new pattern.
2–5. Findings #2–#5 from earlier rounds (RPC discount-scope divergence, lexicographic tie-break, sequential round-trips, admin `trade_points` validation) — **still resolved**, untouched by this commit.
6. The Low docstring nit on `getGangBasic` from the last round — untouched by this commit, cosmetic only, not blocking.

### New findings

None. Confirmed `currentTradePoints` (`equipment.ts:427`) is still used only for the pre-write "can't afford it" fast-fail check and is a separate read from the one inside `deductGangTradePoints` — no stale value leaks into the actual deduction. `tsc --noEmit` is clean (same pre-existing config deprecation warnings as before, unrelated to this diff). Diff is scoped exactly to the two files in the commit (`equipment.ts`, `resources.ts`) — no incidental changes elsewhere.

One residual, non-blocking observation: the fetch-then-write is still two separate calls rather than a single atomic SQL expression (`trade_points = trade_points - amount`), so a vanishingly small race window remains between the `SELECT` and the `UPDATE`. That's the same shape as the already-accepted `deductGangReputation`, so I'm not treating it as a new blocker — just noting it in case a future pass wants to make both atomic in one statement.

### Verdict: **Approve**

The last outstanding High-severity finding is resolved using the codebase's established pattern, and no regressions were introduced.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`be48732` → `fa18b24`, "Deduct Trade Points in the same UPDATE as credits") against the previous review round's approval.

### Previous findings

1–5. All resolved as of the last round (rollback-on-failure, RPC discount-scope divergence, numeric tie-break, sequential round-trips, admin validation) — this commit doesn't touch the code those fixes live in beyond what's noted below.
6. **(High) `deductGangTradePoints` stale-read lost-update race — reintroduced by this commit.** See new finding #1.
7. The Low docstring nit on `getGangBasic` from two rounds ago — still untouched, still cosmetic, not blocking.

### New findings

**1. (High) Moving Trade Points into `updateGangFinancials` drops the `.gte()` concurrency guard that the previous two review rounds fixed and verified.**
This commit deletes `deductGangTradePoints` (`utils/campaigns/resources.ts`), which did a fetch-immediately-before-write with a `.gte('trade_points', amount)` guard on the `UPDATE` — the exact shape that closed out the High-severity lost-update race flagged in the last two rounds, and the same shape `deductGangReputation` (`resources.ts:204-209`) still uses today. Trade Points deduction now goes through `updateGangFinancials` (`utils/gang-rating-and-wealth.ts:117-120`) instead:
```ts
const { error: updateError } = await supabase
  .from('gangs')
  .update(expectedValues)   // expectedValues.trade_points = Math.max(0, currentTradePoints + tradePointsDelta)
  .eq('id', gangId);        // no .gte()/optimistic-lock guard of any kind
```
`currentTradePoints` is read from a plain `SELECT` a few lines above (`gang-rating-and-wealth.ts:78-91`) with no guard tying the later `UPDATE` back to that read. Concretely: gang has 10 TP. Purchase A (costs 6) and Purchase B (costs 3) both hit `buyEquipmentForFighter` near-simultaneously and both read `currentTradePoints = 10` in their respective `updateGangFinancials` calls. A writes `trade_points = 10 - 6 = 4`. B, unaware of A's write, then writes `trade_points = 10 - 3 = 7` — silently erasing A's deduction. Net effect: gang ends up with 7 TP instead of the correct 1; A's purchase was effectively free. This is the identical bug class fixed and approved two rounds ago, now reintroduced via a different code path — worth treating with the same weight given how much verification went into the original fix.

Note this also applies to `creditsDelta` in the same function (pre-existing, not introduced here — `updateGangFinancials` has never had a guard, across its ~26 other call sites), but folding Trade Points into it specifically discards the protection that was deliberately built for Trade Points.

*Fix:* add an optimistic guard to the `UPDATE` in `updateGangFinancials` (e.g. `.eq('credits', currentCredits).eq('trade_points', currentTradePoints)...` or per-field `.gte()` checks) and handle the zero-rows-affected case as a retryable/error condition, mirroring `deductGangReputation`/the old `deductGangTradePoints`. A partial guard scoped just to `trade_points`/`credits` would at minimum restore the pre-commit guarantee for this PR's currency.

**2. (Medium) No compensating rollback if `updateGangFinancials` fails after a resource/reputation-based deduction already succeeded — reintroduces the original round-1 finding.**
`app/actions/equipment.ts:495-506` deducts a campaign resource/reputation and, on failure, deletes the just-inserted `fighter_equipment` row. Further down, `app/actions/equipment.ts:728-738` calls `updateGangFinancials` (crediting/debiting credits + Trade Points) and simply `throw`s on failure — the outer catch (`equipment.ts:1023-1029`) only logs and returns an error string; it does not delete the `fighter_equipment` row or restore the resource/reputation. So an N26 gang buying from a resource-cost trading post that then fails the Trade Points leg (e.g. hits the race in finding #1, or a genuine insufficient-balance race) keeps the equipment row and permanently loses the resource, with nothing to show for it. The removed `deductGangTradePoints` catch block existed specifically to handle this; nothing replaced it. In practice this is gated behind the same early pre-check (`equipment.ts:427-429`) that credits already relies on, so it's a narrow race window rather than a common case — but it's the same gap the first review round asked to be closed, now reopened.

### Checklist
- Edge cases / null handling: `newGangTradePoints` fallback (`equipment.ts:740`, `?? gang.trade_points`) is fine; no other null-handling regressions spotted in this diff.
- Readability: the consolidation itself (single `updateGangFinancials` call, no separate TP round-trip) is a genuine simplification and the updated docstring accurately describes the new behavior.
- Conventions: diverges from the established `deductGangReputation`/`deductGangResource` guarded-update convention elsewhere in the same codebase — worth aligning rather than being the exception.

Blocked on new finding #1 — it undoes a High-severity fix this PR's review history spent two rounds establishing and verifying. Finding #2 should ideally be addressed alongside it, since fixing #1 properly (an atomic guard) would also shrink #2's window, though a fully robust fix to #2 needs its own rollback path regardless.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`fa18b24` → `39afb3a`, "Return the updated gang row from the UPDATE instead of re-reading it") against the previous review round.

### Previous findings

1–5. Rollback-on-failure history aside, findings #2–#5 from the earliest rounds (RPC discount-scope divergence, lexicographic tie-break, sequential round-trips, admin `trade_points` validation) — **still resolved**, untouched by this commit.
6. The Low docstring nit on `getGangBasic` — still untouched, still cosmetic, not blocking.
7. **(High) `updateGangFinancials` UPDATE has no concurrency guard (lost-update race) — still outstanding, not addressed.** This commit doesn't touch the `UPDATE` call's guard, only how the resulting row is obtained. `utils/gang-rating-and-wealth.ts:120-125` is still an unconditional `.update(expectedValues).eq('id', gangId)` computed from a `currentTradePoints`/`currentCredits` read at the top of the same call (`gang-rating-and-wealth.ts:78-91`), with no `.gte()`/optimistic-lock check tying the write back to that read. The lost-update scenario from the last two rounds (two concurrent purchases both reading the same `current`, second write clobbering the first's deduction) still applies to both `trade_points` and `credits`.
8. **(Medium) No compensating rollback if the financials update fails after a resource/reputation deduction already succeeded — still outstanding, not addressed.** `app/actions/equipment.ts` is untouched by this commit; the gap flagged last round (resource/reputation deducted, then the later `updateGangFinancials` call fails, and nothing restores the resource) remains.

### New findings

None that rise to blocking. This commit is a clean, low-risk mechanical refactor:
- Combines the `UPDATE` and the follow-up `SELECT` into one `.update(expectedValues).eq('id', gangId).select(...).single()` call. This `.update().eq().select()` shape is already an established pattern elsewhere in the codebase (e.g. `app/actions/update-gang.ts`), so no new RLS/PostgREST risk.
- Behavior change on the edge case where the row can't be re-fetched: previously, if the follow-up `SELECT` failed after a successful `UPDATE`, the function returned `success: true` with `newValues` assumed equal to `expectedValues` (silently trusting the write happened as expected). Now, if the combined call errors (e.g. 0 rows matched — gang deleted concurrently), it correctly returns `success: false` instead of asserting success it can't verify. This is a strict improvement, not a regression, and the residual `updatedGangRow ? ... : expectedValues` fallback is effectively dead code now but harmless.
- `newValues`/`oldValues` return shape is unchanged, so none of the ~19 call sites need updating.
- `tsc --noEmit` — clean (only the same pre-existing config deprecation warnings as before, unrelated to this diff).

### Checklist
- Edge cases / null handling: the `updatedGangRow ? ... : expectedValues` fallback correctly handles a null/undefined row on the (now unreachable in practice, since `updateError` is checked first) success-with-no-data case.
- Readability: net improvement — one round trip instead of two, and the comment explaining why (`gang-rating-and-wealth.ts:117-119`) is accurate and non-obvious-value-adding.
- Conventions: matches the `.update().select()` pattern already used elsewhere (`update-gang.ts` and others).
- Regressions: none found in this commit; it neither fixes nor worsens the two items below.

This commit itself is fine to merge as-is, but the PR overall is still blocked on the two items reopened by `fa18b24` and not addressed since: the **(High)** missing concurrency guard on `updateGangFinancials`, and the **(Medium)** missing rollback path in `app/actions/equipment.ts` when that update fails after a resource/reputation deduction already succeeded. Both need a fix (e.g. an atomic guarded `UPDATE` — `.eq('trade_points', currentTradePoints).eq('credits', currentCredits)` with a zero-rows-affected retry/error path — plus restoring the earlier equipment.ts rollback block) before this is mergeable.

---

joeseos:
@Djidiouf Reviewed the three newest commits (`a206ba3` → `39afb3a`: "Narrow the Trade Points deduction window", "Deduct Trade Points in the same UPDATE as credits", "Return the updated gang row from the UPDATE instead of re-reading it"). This series is a direct attempt at the still-outstanding High finding from the last round, so I focused on whether it actually closes that gap.

### Previous findings

6. **(High) `deductGangTradePoints` stale-read lost-update race — improved, not fully closed.** The dedicated `deductGangTradePoints` function is gone; Trade Points now move through `updateGangFinancials` alongside credits/rating/wealth in one combined read-then-write (`utils/gang-rating-and-wealth.ts:79-131`). This genuinely narrows the window — previously `current` was read near the top of `buyEquipmentForFighter` and written back after the equipment insert, resource deduction, and grants processing (many intervening awaits); now the SELECT and UPDATE are back-to-back with nothing in between. That's a real improvement. But it's still not atomic: `.update(expectedValues)` (`gang-rating-and-wealth.ts:122`) has **no `.gte()` guard or atomic SQL expression** — the old function at least had `.gte('trade_points', amount)`, which would fail loudly on a live-balance conflict even though it didn't fix the stale-base problem. This version has neither the guard nor the base-staleness fix: two purchases whose SELECT/UPDATE pairs interleave in that narrow window will silently last-write-wins, with no error and no signal that a deduction was lost. Net: much lower probability of hitting it, but when it does hit, it's now a *silent* wrong balance instead of a *thrown* error. Given the shrunk window, I'd call this Medium now rather than High — but it isn't resolved.

### New findings

**1. (Medium) Regression: the compensating-rollback fix from the very first review round is gone.** The removed `deductGangTradePoints` call site used to be wrapped in a try/catch that deleted the just-inserted `fighter_equipment` row and restored any already-deducted resource/reputation on failure (this was Medium finding #1 from the first review, fixed in round 2, confirmed resolved in round 3). That whole block is deleted in this diff. Trade Points deduction now happens inside `updateGangFinancials`, called at `app/actions/equipment.ts:728`, well after the equipment insert and the resource/reputation deduction (`equipment.ts:546-558`, unchanged). If `updateGangFinancials` fails for *any* reason — the new "Insufficient Trade Points" branch (`gang-rating-and-wealth.ts:97-99`) on a genuine race, a DB error on the update, whatever — `equipment.ts:736-738` just throws with no cleanup:
   ```ts
   if (!financialResult.success) {
     throw new Error(financialResult.error || 'Failed to update gang financials');
   }
   ```
   The outer catch (`equipment.ts:811`) only logs and returns an error response — it never deletes `newEquipmentId` or restores the resource. So the exact scenario the original finding described (N26 gang buys from a custom trading post with a resource cost *and* a Trade Points cost; the TP side fails after the resource side already succeeded) is back: the gang permanently loses the resource, and now also keeps equipment it never fully paid for, with no error recovery.
   *Fix:* wrap the `updateGangFinancials` call (or at least the trade-points-relevant failure path) in a catch that deletes `newEquipmentId` and restores the resource/reputation, same as the deleted block used to do — or push resource restoration into `updateGangFinancials`'s failure contract if credits can hit the same gap (worth checking, since credits share this exact call and this no-rollback gap for credits predates this PR too).

Everything else from this diff is straightforward and correct: `tradePointsDelta` defaults to `0` for the ~24 other `updateGangFinancials` call sites, so those are unaffected (no-op same-value write on `trade_points`, not a behavior change); the "return the updated row from the UPDATE" change is a legitimate simplification (one round trip instead of two, and immune to a concurrent writer's value leaking into the response); `tsc --noEmit` on `39afb3a` is clean.

Mainly on new finding #1 — it's a straight reintroduction of an already-fixed data-integrity bug, not a style nit. The narrowed race window (previous finding #6) is a meaningful improvement and I'd be fine merging with it as a known, low-probability, tracked follow-up, but the rollback regression should go back in before merge.

---

joeseos:
@Djidiouf Reviewed the two newest commits (`39afb3a` → `7ca7ca6`: "Guard gang financials against lost updates and unwind failed purchases", "Drop redundant beast deletion from the purchase unwind"). Both directly target the two findings from the last round — checked each closely rather than skimming.

### Previous findings

6. **(Medium, downgraded from High) `updateGangFinancials` non-atomic read-then-write — Resolved.** `utils/gang-rating-and-wealth.ts:85-160` now does a real compare-and-swap: the UPDATE is conditioned on every column (`credits`, `rating`, `wealth`, `trade_points`) still matching what was just read (`matchCurrentValue`, using `.is()` for NULL since PostgREST's `.eq(col, null)` doesn't match NULL rows — correct detail to get right), wrapped in a bounded retry loop (`MAX_FINANCIAL_ATTEMPTS = 3`). A conflicting concurrent write now makes the UPDATE match 0 rows, which is detected (`updatedRows?.[0]` undefined) and triggers a fresh re-read + retry rather than a silent overwrite. This is the correct fix — no more stale-base lost updates, and it fails loudly (`'Gang financials were modified concurrently; please retry'`) if genuinely starved after 3 attempts instead of either the old silent-overwrite or thrown-on-affordability-only behavior. Traced the loop logic: recomputes `expectedValues` fresh from each re-read (no stale values leak across attempts), and correctly re-validates the affordability checks against the fresh balance on each attempt, which is exactly the right semantics for a retry.

**New finding #1 from last round (rollback regression) — Resolved.** `app/actions/equipment.ts:736-752` now wraps the `!financialResult.success` branch in a try/catch that deletes the just-inserted `fighter_equipment` row and restores resource/reputation for resource-based purchases, before throwing — the same shape as the original round-2 fix. Verified the "delete `fighter_equipment` is enough" cascade claim in the comment against the schema: `fighter_effects.fighter_equipment_id`, `fighter_loadout_equipment.fighter_equipment_id`, `fighter_exotic_beasts.fighter_equipment_id` (`fk_fighter_equipment_cascade`), and `fighter_equipment.granted_by_equipment_id` (self-referential, for grants) are all `ON DELETE CASCADE`, and `fighters.fighter_pet_id → fighter_exotic_beasts(id)` is also `ON DELETE CASCADE` — so a beast's own `fighters` row is cleaned up transitively too. The second commit dropping the separate explicit beast-deletion call is correctly justified by this. Also confirmed the unwind path uses the right table (`fighter_equipment`) for all three purchase shapes (fighter/vehicle/stash), since stash purchases also insert into `fighter_equipment` with `gang_stash: true` rather than a separate table.

### New findings

None. Checked all ~24 other `updateGangFinancials` call sites for anything depending on the old single-round-trip/no-retry behavior or matching specific error-message strings — none do (all just pass `financialResult.error` through as a generic fallback message), so the added retry loop and new error string are safe. `tsc --noEmit` on `7ca7ca6` is clean.

One low-severity, non-blocking observation: the 3 retry attempts re-read immediately with no backoff/jitter between them. Fine for the realistic contention level here (double-click / two-tab races), just flagging in case this ever needs to hold up under heavier concurrent load than a single gang's purchases.

### Verdict: **Approve**

Both outstanding findings from the previous rounds are now correctly resolved with verified fixes — the compare-and-swap approach is the right way to close this, and the rollback path is back and matches the schema's cascade behavior.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`7ca7ca6` → `441ed78`, "Pass the purchase result as an object") against the previous approval.

### Previous findings

All findings from prior rounds (rollback-on-failure, RPC discount-scope divergence, numeric tie-break, sequential round-trips, admin `trade_points` validation, the concurrency-guard/CAS fix on `updateGangFinancials`, and the purchase-unwind cascade) — **still resolved**, untouched by this commit. It's a pure refactor of `onEquipmentBought`'s call signature and doesn't touch purchase, financials, or rollback logic at all.

### New changes in this commit

Collapses `onEquipmentBought`'s six positional params (`newFighterCredits, newGangCredits, boughtEquipment, newGangRating?, newGangWealth?, newGangTradePoints?`) into a single `EquipmentBoughtResult` object, per the 3-4 param cap in `CLAUDE.md`.

Verified:
- `EquipmentBoughtResult` is defined once in `hooks/use-purchase-equipment.ts:24-31` and re-exported/imported by `components/equipment/equipment.tsx` — no duplicate shape.
- Grepped all `onEquipmentBought` references — the only real consumer is `components/gang/stash-tab.tsx:1011`, which now destructures `{ newGangCredits, boughtEquipment, newGangRating, newGangWealth, newGangTradePoints }` and correctly drops the previously-unused `_newFighterCredits` positional arg. `equipment.tsx` only forwards the prop through to the hook, unchanged in shape.
- The call site building the object (`hooks/use-purchase-equipment.ts:157-169`) carries over all six fields with no value dropped or renamed.
- `tsc --noEmit` — clean (same pre-existing `tsconfig.json` deprecation warnings as every prior round, unrelated to this diff).

No behavior change, no new edge cases — it's a type-only signature reshape.

### Verdict: **Approve**

Mechanical refactor, correctly threaded through the sole consumer, no regressions. Nothing outstanding from previous rounds.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`441ed78` → `830b6cd`, "Make financials CAS exhaustion observable, and back off between retries") against the previous approval.

### Previous findings

All findings from prior rounds (rollback-on-failure, RPC discount-scope divergence, numeric tie-break, sequential round-trips, admin `trade_points` validation, the concurrency-guard/CAS fix on `updateGangFinancials`, and the purchase-unwind cascade) — **still resolved**, untouched by this commit.

The low-severity, non-blocking observation from two rounds ago ("3 retry attempts re-read immediately with no backoff/jitter between them") — **Resolved.** `utils/gang-rating-and-wealth.ts:176-178` now backs off `RETRY_BACKOFF_MS * attempt` (25ms, then 50ms) between retries, skipping the wait on the final attempt since it's about to give up anyway. Reasonable, low-risk addition.

### New findings

**1. (Medium) The commit's own stated bug — a silent credit-refund loss on CAS exhaustion — isn't actually fixed, only logged.**
The commit message correctly diagnoses that `updateGangFinancials` can now return `success: false`, that four call sites ignore the result entirely (`equipment.ts:1391/1486/1577`, `fighter-advancement.ts:868`), and that the last of those is a credits refund where a silent failure costs the user money. The diff, however, only adds a `console.error` inside the helper (`gang-rating-and-wealth.ts:181-187`) — it doesn't touch any of the four call sites. Confirmed `fighter-advancement.ts:868` still does a bare `await updateGangFinancials(...)` with the result completely discarded; on exhaustion the code proceeds to `invalidateGangCredits` and returns normally as if the refund succeeded, with nothing surfaced to the caller or the user — only a server log records it. The added backoff does reduce how often this is hit in practice, but doesn't close the gap the commit describes.
*Fix:* at minimum, check `result.success` at `fighter-advancement.ts:868` (the money-losing case) and propagate a user-facing error/retry rather than silently continuing; the two `equipment.ts` sites wrapped in try/catch (1486, 1577) should check `result.success` explicitly too, since a returned failure doesn't throw and won't hit the catch block.

Everything else in this commit is correct: the backoff is scoped to the right branch (only fires when about to retry, not on the last attempt or on success), the log includes the deltas for debugging, and no other files are touched. `tsc --noEmit` is clean (same pre-existing `tsconfig.json` deprecation warnings as every prior round).

Blocked only on new finding #1 — it's a real gap between what the commit message says it fixes and what the diff actually changes; the underlying money-loss risk it identifies is still live, just quieter. The backoff improvement itself is good and can stay as-is.

---

joeseos:
@Djidiouf Reviewed the newest commit only (`830b6cd` → `b09d85f`, "Unify reputation with credits and Trade Points; drop the retry backoff") against the previous review round.

### Previous findings

1–5, 7. All resolved as of earlier rounds (rollback-on-failure, RPC discount-scope divergence, numeric tie-break, sequential round-trips, admin `trade_points` validation, the purchase-unwind cascade) — untouched by this commit, still fine.
6. **(Resolved, earlier rounds) `updateGangFinancials` compare-and-swap** — still holds, and this commit correctly extends the same guarded pattern to `reputationDelta` rather than introducing a new one-off path for it.
8. The Low docstring nit on `getGangBasic` (`app/lib/shared/gang-data.ts:190`, still describes itself as including reputation even though a prior commit moved reputation into `getGangResources`) — still untouched, still cosmetic, not blocking.
9. **(Medium) CAS-exhaustion silent-failure gap — still outstanding, not addressed.** The last round flagged that `updateGangFinancials` can return `success: false` on retry exhaustion but four call sites ignore the result (`equipment.ts:1390`, `equipment.ts:1485`, `equipment.ts:1577`, `fighter-advancement.ts:868` — the last being a credits refund). This commit doesn't touch any of those four call sites; confirmed `fighter-advancement.ts:868` is still a bare `await updateGangFinancials(...)` with the result discarded. Only the backoff between retries was addressed last round (and is now removed again below) — the actual "silent money-loss on exhaustion" gap the commit message described remains open.

### New changes in this commit

Two independent changes, reviewed separately:

**Reputation unification.** Reputation now moves through `updateGangFinancials` as `reputationDelta` (`equipment.ts:735`) instead of the old standalone `deductGangReputation` fetch-then-guarded-update. Traced this closely since it's the same currency-integrity class of bug the last several rounds were about:
- `isReputationPurchase`/`isCampaignResourcePurchase` (`equipment.ts:373-374`) are correctly mutually exclusive (a purchase has exactly one `resourceCost`), so `creditsDelta`/the resource-restore branch at `equipment.ts:746` are unaffected.
- On `financialResult.success === false`, the unwind block (`equipment.ts:744-751`) deletes the `fighter_equipment` row and only restores via `returnGangResource` for `isCampaignResourcePurchase` — correctly, since reputation was never separately deducted; it lives inside the same failed UPDATE, so there's nothing to unwind for it. `deductGangReputation` is deleted; `returnGangReputation` stays and is still reachable via `returnCostResource` in `resources.ts` (verified, not orphaned) for the sell/refund path.
- `GangFinancialValues`/`GANG_FINANCIAL_COLUMNS` additions in `gang-rating-and-wealth.ts` are mechanical and consistent; the `Object.entries(raw)` loop over `matchCurrentValue` pins all five columns (credits, rating, wealth, trade_points, reputation) in the CAS, matching the existing pattern.

**(Low) New minor regression: reputation-insufficient purchases now fail later, doing more wasted work, than before.** Trade Points has an explicit pre-check before any write (`equipment.ts:429-433`, "*Validate before any write so the ordinary can't-afford-it case never has to unwind*"), and campaign resources are still deducted+checked right after the equipment insert (`equipment.ts:500-507`, before grants processing). Reputation used to get the same early treatment via `deductGangReputation`, called right after insert. Now the reputation balance is only checked deep inside `updateGangFinancials`, called after grants processing (`equipment.ts:730`) — so an insufficient-reputation purchase now inserts the equipment, processes any equipment grants (additional inserts), and only then fails and unwinds, whereas previously it failed right after the initial insert. The unwind is still correct (cascade delete handles it), so this is not a correctness bug — just an efficiency/consistency regression against the "validate before any write" principle the file states for Trade Points. Fix, if worth doing: add `reputation` to the initial gang select (`equipment.ts:237`) and pre-check `gang.reputation < amount` alongside the existing Trade Points check.

**Backoff removal.** The `RETRY_BACKOFF_MS` sleep added last round is removed, with the reasoning that a 0-rows-matched UPDATE already reflects a committed competing write (Postgres holds the row lock for the duration of the conflicting UPDATE; under READ COMMITTED, a blocked UPDATE re-evaluates its WHERE clause against the post-commit row via EvalPlanQual, so a non-match proves the other transaction already committed). This reasoning is correct for Postgres's actual concurrency semantics — the previous round's justification for adding backoff ("let the competing write commit") had the causality backwards, as the commit message notes. Removing it is a valid simplification, not a regression; the immediate re-read/retry loop is fine at the realistic contention level here (double-click / two-tab races, max 3 attempts).

`tsc --noEmit` is clean (same pre-existing `tsconfig.json` deprecation warnings as every prior round, unrelated to this diff). Grepped for dangling references to `deductGangReputation` and `RETRY_BACKOFF_MS` — none found.

Blocked only on the still-outstanding Medium finding from last round (#9 above, CAS-exhaustion silent failure at the four untouched call sites) — this commit's own changes (reputation unification, backoff removal) are correct and well-reasoned. The new Low finding on reputation's later failure timing is optional and non-blocking.

---

PR #1927, opened 2026-08-05
Rename fighter_sub_types to fighter_specialisations (N26)

joeseos:
N26 calls this concept a "Specialisation". The old name was also actively misleading: it is unrelated to the rulebook term "Subtypes", which maps to fighter_classes -- a warning the schema already carried as a table comment.

Full end-to-end rename with no aliases or compatibility shims: table, columns, constraints, RLS policy names, RPC output keys, API route, types and frontend. Applied as a hard cutover with a short accepted downtime window.

Migration (applied manually, after this lands):
- fighter_sub_types -> fighter_specialisations, sub_type_name -> specialisation_name
- fighter_sub_type_id -> fighter_specialisation_id and the denormalised fighter_sub_type -> fighter_specialisation, on both fighters and fighter_types
- pkey + 2 FK constraint renames, 4 RLS policy renames, refreshed comments

Verified against the live database first: the only index is the pkey (both FK columns are unindexed, so nothing to rename there), no triggers, no views, and exactly two inbound FKs. RENAME CONSTRAINT renames the backing index implicitly; a bare RENAME TO would not.

RPCs live in supabase/functions/ only. Both are plpgsql, so columns resolve at call time -- they must be applied after the migration, not before. get_fighter_types_with_cost changes its RETURNS TABLE signature (sub_type jsonb -> specialisation jsonb); the file already opens with DROP FUNCTION IF EXISTS. get_gang_details has no callers but is updated too.

Shapes are unchanged -- this is a rename, not a reshape. The four competing TS representations of this data remain; flattening them is a follow-up.

Note for deploy: gang-fighters-list-* and fighter-basic-* unstable_cache payloads carry the renamed keys and are revalidate:false, and all gang/fighter tags are per-id, so there is no global tag to flush them from code. Handled operationally with a one-time `vercel cache purge --type data` rather than a permanent -v2 key suffix.


joeseos:
Reviewed this by checking out the PR branch and grepping the full post-rename tree, plus cross-checking the migration's constraint/policy names against the current schema snapshot, rather than reading the 3600-line diff top to bottom.

### Completeness checks (the main risk for a rename this size)
- **No leftover `sub_type`/`sub-type`/`subType` references anywhere in `.ts`/`.tsx`/`.sql`**, except: the historical migration files (expected — migrations are immutable history) and `supabase/schema/schema.public.sql` (expected — this repo updates that snapshot via a separate `chore(db): update Supabase schema snapshot` commit *after* a migration is actually applied to the live DB, confirmed by checking prior migration commits; not a gap in this PR).
- **Migration constraint/policy names verified against the schema snapshot** — `fighter_sub_types_pkey`, `fighters_fighter_sub_type_id_fkey`, `fighter_types_fighter_sub_type_id_fkey`, and all four RLS policy names in the migration match the live names exactly. This matters a lot here since a typo'd `RENAME CONSTRAINT`/`ALTER POLICY` target would hard-fail mid-migration during your accepted downtime window — they don't.
- **RPCs** (`get_fighter_types_with_cost`, `get_gang_details`) — both fully renamed internally, `DROP FUNCTION IF EXISTS` signatures match the current live signatures, and `get_gang_details` genuinely has zero callers in the frontend as the description states.
- **API contract change** (`fighter-types/route.ts`: response key `sub_types` → `specialisations`) — verified this response branch has no frontend caller at all (dead code path, pre-existing before this PR), so the rename can't have broken a live consumer there. Everywhere else, the new `/api/admin/fighter-specialisations` route is consistently referenced from `admin-create-fighter-type.tsx` / `admin-edit-fighter-type.tsx`.
- **No spelling drift** — consistently British "specialisation" throughout; no stray "specialization" that would silently mismatch a key.
- Cache tag names (`utils/cache-tags.ts`) don't embed the old name, so no additional invalidation-key rename was missed.

### Minor, non-blocking notes
- `app/api/admin/fighter-specialisations/route.ts`: `return NextResponse.json({ ...specialisations })` style nit — `specialisations: specialisations` could be shorthand `specialisations,`. Not worth a follow-up on its own.
- The new route's `DELETE` handler is still the `// ... existing code ...` stub — but that's carried over verbatim from the old `fighter-sub-types/route.ts`, which had the identical stub. Pre-existing gap, not a regression from this rename.

### Checklist
- Correctness: no breakage found across the full rename surface (table/columns/constraints/RLS/RPCs/API/types/frontend)
- Edge cases / null handling: unaffected, shape is unchanged (pure rename, no data movement)
- Security: RLS policy bodies untouched, only names renamed; grants come from the blanket bootstrap per the migration's own comment
- Performance: no change (pkey index renamed implicitly via `RENAME CONSTRAINT` as documented; both FK columns already unindexed before this PR, unrelated to this change)
- Readability/maintainability: excellent — migration comments explain every verification step taken against the live DB
- Conventions: no `CLAUDE.md` in the repo to check against; matches this repo's established "RPCs in `supabase/functions/*.sql`, applied after migration" convention
- DRY/YAGNI: appropriately scoped as a pure rename; explicitly defers flattening the four TS representations to a follow-up rather than scope-creeping this PR

### Verdict: **Approve**
This is about as clean as a full-surface rename gets — well-verified against the live schema, no aliases/shims left behind (as intended), and the deploy-order caveat (migration → RPCs → app deploy) is spelled out correctly in the description.

---

PR #1928, opened 2026-08-05
Rename fighter_classes to fighter_subtypes (N26)

joeseos:
N26 calls these Leader/Champion/Ganger/Specialist roles "Subtypes". The schema has carried a note about the mismatch since 20260630095837: 'the rulebook term "Subtypes" refers to what this schema calls fighter_classes'. This is the second half of that cleanup -- 20260805120000 (fighter_sub_types -> fighter_specialisations) freed up the name.

Full end-to-end rename: reference table, its columns, the sole FK column, all five denormalised jsonb array columns, RPC payload keys, API routes, types, and user-facing copy. Hard cutover with a short accepted downtime window.

The class NAMES themselves ('Leader', 'Ganger', 'Exotic Beast Specialist', ...) are DATA VALUES and are deliberately untouched, including the lowercased startsWith('exotic beast') comparisons.

Migration (applied manually, after this lands):
- fighter_classes -> fighter_subtypes, class_name -> subtype_name, standard_class -> standard_subtype
- skill_access_archetypes.fighter_class_id -> fighter_subtype_id
- jsonb columns on fighters, fighter_types, custom_fighter_types, and the two on fighter_ooa_records
- pkey + 2 FK constraint renames, 3 ALTER INDEX, 4 RLS policy renames

Object names were verified against the live database rather than the schema snapshot, which was stale. standard_class is renamed rather than dropped: no code reads it, but it holds real data (10 true / 28 false).

RPCs live in supabase/functions/ only. All five are plpgsql, so they must be applied after the migration. get_fighter_types_with_cost changes its RETURNS TABLE signature.

Because className and other CSS identifiers outnumber domain `class` tokens roughly 8:1, this was driven off ~20 domain-prefixed token families rather than a blanket replace, and guarded by invariants: strict className 6935 -> 6929 (the delta is 6 deliberate className -> subtypeName parameter renames), with classList, clsx and all thirteen class-name literal counts unchanged.

Note for deploy: fighter-basic-*, gang-fighters-list-* and gang-fighter-stats-* unstable_cache payloads carry the renamed keys and are revalidate:false, and all gang/fighter tags are per-id, so there is no global tag to flush them from code. Handled operationally with a one-time `vercel cache purge --type data`.


joeseos:
Reviewed this the same way as #1927: checked out the branch into a worktree and verified the full post-rename tree rather than reading the 4500-line diff top to bottom, since completeness is the real risk on a rename this size — especially one crossing paths with the extremely overloaded "class"/"className" token.

### Completeness checks
- **No leftover `fighter_class`/`fighter_classes`/`class_name`/`standard_class`/`fighter_class_id` references anywhere in `.ts`/`.tsx`/current `.sql`**, except historical migration files (expected, immutable) and `supabase/schema/schema.public.sql` (expected — updated via a separate post-deploy `chore(db)` commit, same as #1927).
- **Migration object names cross-checked against the schema snapshot** — despite the PR noting the snapshot was stale at authoring time, every name it references for this specific table (`fighter_classes_pkey`, all three standalone indexes, `fighter_classes_edition_id_fkey`, `skill_access_archetypes_fighter_class_id_fkey`, all four RLS policy names, and the still-present `standard_class` column with real data) matches the snapshot exactly. Good corroborating signal even though it wasn't the authority used.
- **The `className`/CSS collision risk** — reproduced the invariant check independently rather than trusting the PR description's count: diffed every line touching the literal string `className` between base and head (12 removed / 4 added). Inspected every one — they're either genuine JSX `className` attributes whose surrounding variable names changed (`fighterClass.id` → `fighterSubtype.id`, a user-facing label "Fighter Class *" → "Fighter Subtype *"), or the ~6 deliberate local-variable disambiguations the description calls out (`toggleFighterClass(className: string, …)` → `toggleFighterSubtype(subtypeName: string, …)`, `classSet`/`classType` → `subtypeSet`/`subtypeName`). None of them touch actual CSS/Tailwind class handling. This is also a nice incidental fix — that local param really was named `className` for a domain string, which was a landmine waiting to confuse a future reader.
- **RPCs** — `get_fighter_types_with_cost` has its `DROP FUNCTION IF EXISTS` (all 4 legacy overloads) ahead of the `RETURNS TABLE` signature change, as needed. `get_available_skills` and `get_fighter_available_advancements` correctly have no `DROP FUNCTION` — both return plain `jsonb`, so `CREATE OR REPLACE` is sufficient since the signature doesn't change, consistent with the description's claim that only `get_fighter_types_with_cost` changes shape.
- **Data values genuinely untouched** — spot-checked all five `.toLowerCase().startsWith('exotic beast')` call sites; they now read off the renamed `fighter_subtypes` array but the string comparison itself is unchanged, as intended.
- **`fighter_ooa_records`' two renamed columns** (`causing_fighter_subtypes`/`injured_fighter_subtypes`) are consistently threaded through `types/fighter-ooa-record.ts`, `app/actions/fighter-ooa-records.ts`, `app/actions/edit-fighter.ts`, and the OOA history modal.
- No remaining "Fighter Class" label text in the UI.

### Notes (non-blocking)
- Same caveat as #1927: I can't independently verify against the actual live database (only against the possibly-stale snapshot, which happens to agree here), so the final go/no-go on the constraint/index names still rests on your own live-DB check before running the migration.
- Nothing else worth flagging — didn't find a single incorrect or missed rename site in the application code.

### Checklist
- Correctness: clean across the full rename surface
- Edge cases / null handling: unaffected — pure rename, no shape/data change
- Security: RLS policy bodies untouched, names only
- Performance: unaffected — same indexes, explicitly renamed rather than dropped/recreated
- Readability/maintainability: very good; the `className`→`subtypeName` param disambiguation is a genuine small improvement beyond the rename itself
- Conventions: matches the established "RPCs in `supabase/functions/*.sql`, applied after migration" pattern from #1927
- DRY/YAGNI: appropriately scoped as a pure rename

### Verdict: **Approve**

---

PR #1930, opened 2026-08-05
Added fighter max char with n26 support

joeseos:
Custom fighter types will get support later.

### DIFF
--- a/components/fighter/fighter-details-card.tsx
+++ b/components/fighter/fighter-details-card.tsx
@@ -524,7 +524,7 @@ export const FighterDetailsCard = memo(function FighterDetailsCard({
         </div>
       </div> 
       <div className="mt-2">
-        <FighterDetailsStatsTable data={stats} isCrew={isCrew} />
+        <FighterDetailsStatsTable data={stats} isCrew={isCrew} editionSlug={edition_slug} />
       </div>
       
       {/* Show owner information for owned fighters */}
--- a/components/ui/fighter-details-stats-table.tsx
+++ b/components/ui/fighter-details-stats-table.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fighterCharacteristicLimits, crewCharacteristicLimits } from '@/utils/characteristicLimits';
+import { characteristicLimitsFor } from '@/utils/characteristicLimits';
 
 const formatStatValue = (key: string, value: number | string) => {
   if (key === 'BS' && value === '0+') return '-';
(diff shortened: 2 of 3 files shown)


joeseos:
Reviewed against the codebase (checked out the branch to grep for other call sites/leftover references given how small this diff is).

### What it does
Makes the "out of range" stat highlighting in `FighterDetailsStatsTable` edition-aware: previously every fighter was checked against a single hardcoded (N23) characteristic range regardless of edition; now `characteristicLimitsFor(editionSlug, isCrew)` picks the right table per edition, with N23's numbers preserved byte-for-byte and N26's added alongside.

### Verification
- **No leftover references** to the removed `fighterCharacteristicLimits`/`crewCharacteristicLimits` exports anywhere in the tree — clean rename to the new lookup function.
- **Single call site** of `FighterDetailsStatsTable` (`fighter-details-card.tsx`), correctly updated to pass `editionSlug={edition_slug}` — and `edition_slug` was already a prop on that component (used a few lines above for `hasSaveCharacteristic`), so this isn't introducing new, unwired plumbing.
- **N23 and crew tables are unchanged** — the diff only touches the wrapping export structure, not the actual tuple values, so no accidental value drift during the refactor.
- **`parseValue`/`isStatOutOfRange` in the stats table are untouched** and already handle both string (`'4+'`) and plain-number limit forms via the same `parseInt`-based parsing, so mixing notations between editions (N23's `'3+'..'10+'` vs N26's plain `4..10` for Ld/Cl/Wil/Int) works without changes there.
- `LIMITS_BY_EDITION: Record<EditionSlug, ...>` mirrors the `EDITION_CAPABILITIES` pattern from `types/edition.ts` (compile error if an edition is added without limits) — good consistency with existing conventions.
- Scope is exactly what the PR description says — no other file in the repo references `characteristicLimits` besides the table component and the utility itself, so there's no admin-side validation logic elsewhere that also needed edition-awareness and got missed.

### One behavior change worth flagging (not a bug, just worth being explicit about)
Before this PR, a fighter with no resolvable `edition_slug` (or an unrecognised one) would still be checked against the old N23 table. After this PR, `characteristicLimitsFor` returns `{}` for that case, so **no highlighting happens at all** for such fighters — per the code comment, deliberately, to avoid mislabeling valid N26 stats as illegal when the edition fails to load. Combined with the PR description ("Custom fighter types will get support later"), this means custom fighter types will silently lose the out-of-range highlighting they had before, until the follow-up lands. Given the description already discloses this, I'd just confirm that's the intended interim tradeoff rather than an oversight — it reads that way from the code.

### Checklist
- Correctness: no bugs found
- Edge cases / null handling: handled explicitly and defensively (`NO_LIMITS` fallback, safe `Record` lookup with type assertion guarded by an existence check)
- Security: n/a, pure display logic
- Performance: n/a, negligible
- Readability/maintainability: good — the comments in `characteristicLimits.ts` explain non-obvious decisions (tuple ordering, the `NO_LIMITS` rationale) well
- Conventions: matches the `EDITION_CAPABILITIES` exhaustiveness pattern already established
- DRY/YAGNI: appropriately scoped; doesn't try to solve custom-fighter-type support in this PR

### Verdict: **Approve**

---

PR #1933, opened 2026-08-06
Added switch between edition on gang, campaign and custom asset page,…

Djidiouf:
… in addition to Create gang and Create Campaign + gated added gangs to campaign to matched edition of campaign

### DIFF
--- a/app/actions/campaigns/[id]/assert-gang-campaign-edition.ts
+++ b/app/actions/campaigns/[id]/assert-gang-campaign-edition.ts
@@ -1,0 +1,46 @@
+import type { SupabaseClient } from '@supabase/supabase-js';
+import { sameEditionSlug } from '@/types/edition';
+
+/** Campaign type edition and gang type edition must match (null treated as n23). */
+export async function assertGangMatchesCampaignEdition(
+  supabase: SupabaseClient,
+  campaignId: string,
+  gangId: string
+): Promise<{ ok: true } | { ok: false; error: string }> {
+  const [{ data: campaign, error: campaignError }, { data: gang, error: gangError }] = await Promise.all([
+    supabase
+      .from('campaigns')
+      .select('campaign_types!campaign_type_id(editions:edition_id(slug))')
+      .eq('id', campaignId)
+      .single(),
+    supabase
+      .from('gangs')
+      .select(`
+        gang_types!gang_type_id(editions:edition_id(slug)),
+        custom_gang_types!custom_gang_type_id(editions:edition_id(slug))
+      `)
+      .eq('id', gangId)
+      .single(),
+  ]);
+
+  if (campaignError || !campaign) {
+    return { ok: false, error: 'Campaign not found' };
+  }
+  if (gangError || !gang) {
+    return { ok: false, error: 'Gang not found' };
+  }
+
+  const campaignEditionSlug = (campaign as any).campaign_types?.editions?.slug ?? null;
+  const gangEditionSlug = (gang as any).gang_types?.editions?.slug
+    ?? (gang as any).custom_gang_types?.editions?.slug
+    ?? null;
+
+  if (!sameEditionSlug(campaignEditionSlug, gangEditionSlug)) {
+    return {
+      ok: false,
+      error: 'This gang is from a different edition than the campaign'
+    };
+  }
+
+  return { ok: true };
+}
--- a/app/actions/campaigns/[id]/campaign-gangs.ts
+++ b/app/actions/campaigns/[id]/campaign-gangs.ts
@@ -4,6 +4,7 @@ import { createClient } from "@/utils/supabase/server";
 import { revalidateTag } from "next/cache";
 import { invalidateCampaignMembership, invalidateCampaignTerritory, invalidatePermissionForUser } from "@/utils/cache-tags";
 import { getAuthenticatedUser } from '@/utils/auth';
+import { assertGangMatchesCampaignEdition } from './assert-gang-campaign-edition';
 
 export interface AddGangToCampaignDirectParams {
   campaignId: string;
@@ -28,6 +29,11 @@ export async function addGangToCampaignDirect(params: AddGangToCampaignDirectPar
     // Authenticate user
     const user = await getAuthenticatedUser(supabase);
     const { campaignId, gangId } = params;
+
+    const editionCheck = await assertGangMatchesCampaignEdition(supabase, campaignId, gangId);
+    if (!editionCheck.ok) {
+      return { success: false, error: editionCheck.error };
+    }
     
     // Get the gang's owner
     const { data: gangData, error: gangError } = await supabase
--- a/app/actions/campaigns/[id]/campaign-members.ts
+++ b/app/actions/campaigns/[id]/campaign-members.ts
@@ -5,6 +5,7 @@ import { revalidateTag } from "next/cache";
 import { CACHE_TAGS, invalidateCampaignMembership, invalidatePermissionForUser, invalidateCampaignMemberPermissions } from "@/utils/cache-tags";
 import { logGangJoinedCampaign, logGangLeftCampaign } from "../../logs/gang-campaign-logs";
 import { getAuthenticatedUser } from '@/utils/auth';
+import { assertGangMatchesCampaignEdition } from './assert-gang-campaign-edition';
 
 export interface AddGangToCampaignParams {
   campaignId: string;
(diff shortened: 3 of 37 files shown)


joeseos:
## Code Review

Reviewed against the full codebase (`main...feature/switch_between_edition`, +591/-90, 36 files). This adds edition switching (N23/N26) to the gang, campaign, and custom-asset pages plus Create Gang / Create Campaign, and gates adding a gang to a campaign so its edition must match the campaign's (via a new `assertGangMatchesCampaignEdition` helper). No `CLAUDE.md` in this repo to check convention adherence against.

The campaign-gang edition gating — the correctness/security-critical part of this PR — is solid: `sameEditionSlug`'s "null treated as n23" semantics are implemented and documented consistently everywhere they're used, and `assertGangMatchesCampaignEdition` is called on **every** insert path into `campaign_gangs` (`addGangToCampaignDirect` and `addGangToCampaign`, covering both the new-member and existing-member branches) — no bypass path found after tracing settings, the Discord edge function, accept/decline invite, and the campaign-gangs API route.

However, the "custom asset page" edition switching described in the PR is incomplete: two of the custom-asset authoring flows still use unfiltered catalog pickers.

### Findings

**1. [Medium] Custom Skill creation's "Skill Type" picker isn't edition-filtered**
`components/customise/custom-skills.tsx:65-70` (`fetchSkillTypes`)
`custom_skills` has no `edition_id` column of its own — per the migration comment in `20260630095837_add_editions_and_root_edition_ids.sql`, it derives its edition entirely from the chosen `skill_type_id`/`custom_skill_type_id`. But `fetchSkillTypes` selects `id, name` only, with no edition filter, so a user working in N26 can pick an N23-only skill type. The resulting skill then resolves to `edition_id: null → n23` in `getUserCustomSkills` (`app/lib/customise/custom-skills.ts:75-77`) and silently vanishes from the N26-filtered view the user was just working in.
*Fix:* select `edition_id` in the `skill_types`/`custom_skill_types` queries and filter with the same `matchesHomeEditionId`/`sameEditionSlug` pattern already used for gang types and campaign types elsewhere in this PR.

**2. [Medium] Custom Fighter Type authoring pickers (gang type, skill type, equipment) aren't edition-filtered**
`components/customise/custom-fighters.tsx:536,572,590`
- Line 536: the "assign to gang type" dropdown (`gangTypes`, via `/api/gang-types?includeAll=true`, which now returns `edition_id`) is unfiltered.
- Line 572: skill-access configuration (`skillTypes`, via `/api/skill-types`) is unfiltered.
- Line 590: default/starting equipment (`equipment`, via `/api/equipment?core_equipment=false`) is unfiltered.

Create Gang / Create Campaign correctly gained edition-filtered pickers in this PR, but this flow didn't — a custom fighter tagged N26 can end up assigned to an N23-only gang type or given N23-only weapons as default loadout, which is a real rules-consistency gap given N23/N26 use different statline/equipment conventions (`hasTradePoints`, `lethalityStatline` in `types/edition.ts`).
*Fix:* apply the same `matchesHomeEditionId` filtering used in `CreateGangModal` to these three lists.

**3. [Low, non-blocking] Duplicated "reset stale selection on edition change" logic**
`components/create-campaign.tsx` and `components/create-gang-modal.tsx` each hand-roll a near-identical `[prevEditionId, setPrevEditionId]` block. Worth a small shared hook if a third consumer shows up; not urgent.

**4. [Low, non-blocking] Brief unfiltered flash before `/api/editions` resolves**
`hooks/use-home-edition.ts:53` — `matchesHomeEditionId` treats a null `selectedEditionId` as "match everything," so on first mount the Create Gang/Create Campaign dropdowns briefly show both editions' options before a selection made in that window gets auto-cleared once the edition resolves. Minor UX rough edge, no data loss.

### Checklist
- **Correctness**: Campaign-gang edition gating is correct and applied on every insert path. Findings #1–2 leave the custom-asset edition feature incomplete for skill/fighter authoring.
- **Edge cases / null handling**: null→n23 semantics consistent everywhere checked; `data.edition_id || null` vs `??` makes no behavioral difference since `edition_id` is always a UUID or null/undefined, never `''`.
- **Security**: No auth gaps — all touched actions/routes still go through `getAuthenticatedUser`/`getUserIdFromClaims`. Additional `edition_id` columns in SELECTs expose no sensitive data.
- **Performance**: `assertGangMatchesCampaignEdition`'s two parallel queries run only on the low-frequency add-gang action, not a hot path. No new N+1 patterns introduced.
- **Readability/maintainability**: Clean; the `(x as any).foo?.bar?.slug` casts match ~198 pre-existing instances of the same style on `main`, not a new anti-pattern.
- **Project conventions**: No `CLAUDE.md` present. The `prevX`/`setPrevX` state-sync pattern used in the five `Customise*` components matches the pre-existing pattern already in `CustomiseFighters`.
- **DRY/YAGNI**: `edition_id` plumbing through action/type files is mechanical, not meaningfully reducible. Finding #3 is the one spot with real (minor) duplication.

Campaign-gang edition gating is solid, but findings #1–2 are concrete functional bugs within this PR's stated scope — the custom-skill and custom-fighter authoring pickers can produce assets that silently mismatch or vanish from their intended edition — and should be fixed before merge.

cc @Djidiouf

---

joeseos:
## Follow-up Review

Reviewed only the new commit since the last review: `335cce9...471ef86` (`components/home-tabs.tsx`, `components/home/edition-toggle.tsx`). Instead of filtering the picker dropdowns as requested, this commit takes a different approach: it locks the Custom Assets tab to N23 (forces `editionSlug` to N23 whenever `activeTab === 2` and disables the `EditionToggle` there).

### Previous findings

1. **[Medium] Custom Skill "Skill Type" picker not edition-filtered** — **Outstanding.** `fetchSkillTypes` in `components/customise/custom-skills.tsx:65-70` is untouched by this commit and still selects `id, name` with no edition filter. Locking the tab to N23 only stops a user from switching the *tab* to N26 — it does nothing to the sub-picker itself. A user can still pick an N26-tagged `skill_type_id`/`custom_skill_type_id` from that unfiltered dropdown while the page is "N23-locked." Per `app/lib/customise/custom-skills.ts:78-79`, the created skill's `edition_id` is derived entirely from the chosen type, so it resolves to N26 and silently vanishes from the N23-filtered view the user is looking at — the same bug as before, just with the trigger direction flipped (previously: N26 user picks N23 type; now: N23-locked user picks N26 type).

2. **[Medium] Custom Fighter Type authoring pickers (gang type, skill type, equipment) not edition-filtered** — **Outstanding**, same reasoning. `components/customise/custom-fighters.tsx:536,572,590` are untouched by this commit and remain fully unfiltered.

3. **[Low] Duplicated "reset stale selection" logic** — **Outstanding** (unaddressed, non-blocking as before).

4. **[Low] Brief unfiltered flash before `/api/editions` resolves** — **Outstanding** (unaddressed, non-blocking as before).

### New findings

**1. [High] Locking the Custom Assets tab clobbers the shared, site-wide edition selection** — `components/home-tabs.tsx:139-144`
```js
useEffect(() => {
  if (activeTab === 2 && editionSlug !== EDITION_N23) {
    setEditionSlug(EDITION_N23);
  }
}, [activeTab, editionSlug, setEditionSlug]);
```
`useHomeEdition` (`hooks/use-home-edition.ts`) is a single module-level singleton backed by `localStorage` (`home_edition` key) — it is **not** scoped to the Custom Assets tab. It's the same state consumed by `GangsTab`, `CampaignsTab`, `CreateGangModal`, and `CreateCampaign` (confirmed via grep — all six consumers share one store).

Concrete repro: a user viewing N26 gangs (editionSlug = `n26`) clicks over to the Custom Assets tab (even just to look, or via a `?tab=customassets` deep link). This effect immediately calls `setEditionSlug(EDITION_N23)`, flipping the *global* store and persisting `n23` to `localStorage`. When they click back to the Gangs tab, they now see N23 gangs — with no indication anything changed — and the reset persists across reloads and into the Create Gang/Create Campaign modals too.

*Fix:* don't mutate the shared `useHomeEdition` store as a side effect of tab navigation. If the Custom Assets section needs to render as N23-only, pass a locally-hardcoded N23 edition id/filter into that section's props instead of calling the global `setEditionSlug`, or save/restore the previous edition on tab enter/exit.

**2. [Medium] The N23 lock doesn't actually close the gap findings #1/#2 described** — see "Previous findings" above; restated here because it's a direct consequence of *this* commit's approach (lock instead of filter) rather than a pre-existing issue. The lock prevents the top-level toggle from reaching N26 but leaves every sub-picker inside `CustomiseSkills`/`CustomiseFighters` unfiltered, so mismatched, silently-vanishing assets are still fully reproducible.

The tab-lock approach is a reasonable scope reduction if intentional, but as implemented it introduces a new high-severity regression (global edition state gets silently reset by visiting an unrelated tab) and doesn't actually close the loophole the previous review flagged — the unfiltered sub-pickers are the real fix still needed.

cc @Djidiouf

---

PR #1934, opened 2026-08-06
Feat/battle session edition

joeseos:


joeseos:
## Code Review

Reviewed against the full codebase (`main...feat/battle-session-edition-id`, +132/-38, 4 files). This adds `battle_sessions.edition_id`, derives it server-side at session creation (from the campaign type for campaign sessions, from the participating gangs for skirmish sessions), backfills existing rows, and exposes it as `edition_slug` through the session loader. No `CLAUDE.md` in this repo to check convention adherence against. Same pattern as the gang/campaign edition work in #1933: slug-based capability checks via `types/edition.ts`, null-safe fallback, server-derived (never client-supplied).

Nothing currently reads `edition_slug` off a battle session yet (`capabilitiesFor` consumers are all gang/fighter-page today) — this PR is pure plumbing ahead of the injury-catalog/XP-table work the migration comment references. That's a reasonable way to land it, but it means the one real gap below won't surface as a visible bug until something starts consuming the field.

### Findings

**1. [Medium] Skirmish sessions can mix gangs from different editions, and the derived edition is then arbitrary**
`app/actions/battle-sessions.ts` (edition fallback block, ~line 190) + `components/battle-session/create-battle-modal.tsx`
Unlike adding a gang to a campaign (gated by `assertGangMatchesCampaignEdition` in #1933), nothing prevents a non-campaign "New Battle" from pairing an N23 gang with an N26 gang — `UserSearchBar`/`/api/users/[id]` and the opponent-gang combobox in `create-battle-modal.tsx` don't filter by edition at all. When that happens, `gangs.map(...).find(id => id)` just takes whichever gang's edition resolves first in the array PostgREST happens to return (not necessarily the creating gang, since `.in('id', gang_ids)` order isn't guaranteed to match input order) — so the session silently gets one participant's ruleset while the other plays a different one whenever edition-specific behavior lands.
*Fix:* either filter the opponent gang pickers to the creating gang's edition (mirroring the campaign-gang gating), or explicitly validate all `gang_ids` share an edition in `createBattleSession` and return an error otherwise, the same way `assertGangMatchesCampaignEdition` does for campaigns.

**2. [Low, non-blocking] `getGangBattleSessionsCached` doesn't populate `edition_slug`**
`app/lib/battle-sessions/get-battle-session-data.ts:174-178`
This list query (`select('*')`, used for a gang's battle-session history) returns plain `BattleSession[]` without the `editions:edition_id(slug)` embed that `fetchBattleSessionDirect` now does, so `edition_slug` is `undefined` here while populated on the single-session loader. Harmless today since the field is optional and unused, but worth a note (or a follow-up ticket) so a future consumer of this list doesn't get bitten by the inconsistency.

### What checked out cleanly
- `editionId` is fully server-derived — never accepted from `params` — matching the "derived, not trusted" pattern from #1933.
- The gangs query is correctly reused for both participant-row construction and edition resolution (previously it was fetched once for participants only); no duplicate query introduced.
- The two-query campaign→campaign_type lookup is justified by the documented PostgREST limitation (`campaigns.campaign_type_id` has no FK, so it can't be embedded) rather than being an accidental N+1.
- `editions:edition_id(slug)` is destructured and flattened in the loader (not left as a nested `session.editions.slug`), with a correct comment explaining why it must happen there rather than in the page (the realtime hook re-fetches through the same function).
- Migration backfill is sound: campaign sessions get their type's edition, everything else (skirmish + typeless campaigns) defaults to `n23` — consistent with `sameEditionSlug`'s "null treated as n23" convention elsewhere in the codebase. `ON DELETE RESTRICT` on the new FK is consistent with protecting historical data integrity.
- `official-then-custom` fallback (`gang_types?.edition_id ?? custom_gang_type?.edition_id`) correctly mirrors `getGangBasic`'s per-gang resolution.

### Checklist
- **Correctness**: Campaign-derived edition path is correct. Skirmish-derived path has the mixed-edition gap in finding #1.
- **Edge cases / null handling**: Both lookups failing degrades to `null` → `NO_CAPABILITIES`, which is the intended, documented behavior rather than a silent wrong-edition default.
- **Security**: No auth changes; edition derivation is server-side only. No new sensitive-data exposure.
- **Performance**: Adds at most 2 extra sequential queries (campaign→campaign_type) only when `campaign_id` is set, on a low-frequency user-initiated action — not a hot path.
- **Readability/maintainability**: Comments explain the non-obvious parts well (PostgREST FK limitation, why flattening happens in the loader). Good line-level documentation.
- **Project conventions**: No `CLAUDE.md` present. Consistent with the slug-based, capability-driven edition pattern established in #1933.
- **DRY/YAGNI**: No unnecessary duplication; the participant-fetch consolidation actually removes a duplicate query that existed before this PR.

The plumbing itself is clean and well-reasoned, but finding #1 is a real correctness gap within this PR's own scope (deriving a session's edition from its gangs implicitly assumes they agree) and should at least be a deliberate, documented decision — either enforce matching editions for skirmish sessions or explicitly note that a mixed-edition skirmish is allowed and the resulting edition pick is best-effort.

cc @joeseos

---

joeseos:
## Follow-up Review

Reviewed the latest commit (`0235ca9`) against the prior review's findings — diffing only the changes since the reviewed commit (`5b4f90e`), 8 files / +110/-19.

### Previous findings

**1. [Medium] Skirmish sessions can mix gangs from different editions — Resolved.**
`editionsConflict(a, b)` (`types/edition.ts`) is added as a deliberately null-safe predicate — an unresolved edition makes no claim and never triggers a conflict, matching the `NO_CAPABILITIES` philosophy elsewhere. It's enforced server-side in two places:
- `createBattleSession`: all fetched gangs' editions are pairwise checked; a genuine mismatch now returns `{success: false}` instead of silently picking whichever gang PostgREST returned first (the exact non-determinism flagged in the original finding).
- `addParticipant`: a joining gang's edition is checked against the session's already-fixed edition (`editions:edition_id(slug)` added to the session select).

Client-side, `create-battle-modal.tsx` filters the skirmish opponent picker by `editionSlug` (threaded down from `gang-page-content.tsx` → `battle-sessions-tab.tsx` → the modal, and from `session.edition_slug` for the add-participant case in `active-session.tsx`), so the server-side rejection is a backstop rather than the primary UX. `/api/users/[id]` now exposes each gang's `edition_slug` so the client has something to filter on — reasonable, since gang name/rating for other users were already public through this endpoint. Campaign opponents are correctly left unfiltered, since campaign membership already gates edition per the commit message.

Verified the conflict check is symmetric and self-pairs never false-positive (`editionsConflict(x, x)` is always `false` since it requires `a !== b`), and that a single-gang or campaign-only session (no `gang_ids`) still works — empty/singleton arrays never trigger the `.some(...)` check.

**2. [Low, non-blocking] `getGangBattleSessionsCached` doesn't populate `edition_slug` — Resolved.**
Now selects `editions:edition_id(slug)` and flattens it the same way `fetchBattleSessionDirect` does, so both loaders are consistent.

### New findings

None. The new commit is scoped tightly to the two findings — the added `GANG_EDITION_SELECT`/`gangEdition` helpers deduplicate what would otherwise be four repeated embed strings, and the null-handling in `addParticipant` (unresolved session edition = no conflict, matching the create path) is a deliberate, documented tradeoff rather than an oversight.

### Checklist
- **Edge cases / null handling**: unresolved edition (`null`) on either side of a comparison correctly never blocks the user — consistent with the rest of the codebase's "unknown ≠ wrong" convention.
- **Regressions**: none spotted; existing call sites of the gang/session selects were updated consistently across all four consuming components.
- **Readability**: comments explain the non-obvious parts (why `.in()` order can't be trusted, why campaign opponents aren't filtered, why null isn't aliased to n23) well.

### Verdict: **Approve**

Both prior findings are resolved with a well-reasoned, consistently-applied fix and no new issues introduced.

cc @joeseos

---

PR #1935, opened 2026-08-06
Added edition n26 injuries

joeseos:


joeseos:
## Code Review

Reviewed against the full codebase (`main...feature/added-edition-to-injuries`, +348/-68, 8 files). This edition-scopes the lasting-injury/rig-glitch system: a new N26 D66 injury table and seed migration, an edition-aware `/api/fighters/injuries` route, and `editionSlug` threaded from gang/battle-session/fighter data down through `fighter-injury-list.tsx` and `participant-card.tsx` into the shared `resolveInjuryFor`/`lastingInjuryRankFor` helpers in `utils/dice.ts`/`utils/lastingInjuryRank.ts`. No `CLAUDE.md` in this repo to check convention adherence against.

This is careful work — the `Record<EditionSlug, …>` tables in `utils/dice.ts` and `utils/lastingInjuryRank.ts` correctly mirror the compile-time-exhaustive pattern from `EDITION_CAPABILITIES` (`types/edition.ts`), so a future edition can't be added without an explicit decision on its injury tables. I traced every removed export (`resolveInjuryFromUtil`, `resolveInjuryFromUtilCrew`, `resolveRigGlitchFromUtil`, and their range-lookup counterparts) — none are referenced anywhere else in the codebase, so nothing is left dangling. I also checked the `type_specific_data` string-boolean convention (`"true"`/`"false"` rather than JSON booleans) used throughout the new migration's seed rows against `hasKilledStatusFlag` and the `=== "true"` checks in `fighter-injury-list.tsx`/`admin-injuries.tsx` — it matches the existing codebase convention exactly, not a bug.

### Findings

**1. [Low, non-blocking] Null/unresolved edition falls back to "unfiltered" server-side but "empty" client-side — inconsistent with the codebase's null→n23 convention**
`app/api/fighters/injuries/route.ts` (no `edition_slug`/`edition_id` param → no `.eq('edition_id', …)` filter, returns every edition's rows) vs. `utils/dice.ts` `NO_INJURY_TABLES`/`lastingInjuryRankFor` (null slug → empty table, no ranges/rolls resolve).
Everywhere else edition-aware code treats a missing/null slug as N23 (`sameEditionSlug`'s documented "null treated as n23", `capabilitiesFor`'s equivalent). Here the two halves of the same feature disagree on what "no edition" means: if a gang/fighter's `editionSlug` genuinely resolves to `null` (e.g., the create-custom-gang-type race noted in the #1933 review, where the page's edition hasn't resolved yet), the injury picker's catalog fetch would return **both** editions' injuries merged — including duplicate-named entries like "Eye Injury"/"Captured" with different modifiers — while the dice-roller and range-lookup for that same session resolve nothing at all, since `lastingInjuryTableFor` returns `NO_INJURY_TABLES` for a null slug. This is a narrow, edge-case path in practice, but worth aligning — either have the route default the filter to `n23` when both params are omitted, or have callers always resolve and send a definite slug rather than an empty string.

### What checked out cleanly
- `editionSlug` is correctly sourced from the *session* on the battle-session card (`session.edition_slug`, not the per-fighter value, which is often null for custom fighter types) — the inline comment explaining this choice is accurate and matches what #1934 wired up.
- `fighter-page.tsx`'s `fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null` and `gang-page-content.tsx`'s `gangData.processedData.edition_slug` both resolve to fields that are genuinely populated upstream (traced to `app/lib/shared/gang-data.ts` and `app/fighter/[id]/page.tsx`), not dead props.
- The Crew/Spyrer table fallback semantics are deliberately asymmetric and documented why: a Crew fighter with no edition-specific crew table falls back to the base table (crew still take lasting injuries), while a Spyrer with no rig-glitch table gets nothing (rig glitches are a self-contained subsystem, not a variant of the normal table) — and this matches how `InjuryPickerModal` in `participant-card.tsx` already behaved pre-PR (hardcoded non-crew), so no regression there.
- The seed migration is idempotent (`NOT EXISTS` guard keyed on name+category+edition), leaves N23 rows untouched, and its extensive comments on the N26 rule differences (Bitter Enmity → three Hatred(X) variants, Convalescence → Grievous Wound, dropped injuries, sign-convention gotchas for the Leadership stat swap) are unusually good documentation for a data-only migration.
- `/api/fighters/injuries` has exactly two callers (`participant-card.tsx`, `fighter-injury-list.tsx`), both updated to pass `edition_slug`.

### Checklist
- **Correctness**: Solid for the mainline path (edition resolved). Finding #1 is the one gap, and it's edge-case-only given `edition_id` is backfilled/set on creation in practice.
- **Edge cases / null handling**: Handled everywhere except the specific null-edition combobox/roll mismatch above.
- **Security**: No auth changes; the new API params are read-only filters, no injection risk (parameterized `.eq()`).
- **Performance**: One extra `editions` lookup only when `edition_slug` (not `edition_id`) is passed; negligible, not a hot path.
- **Readability/maintainability**: Excellent — the edition-selection helpers and migration are unusually well commented for future maintainers, especially the sign-convention warning for admins entering N26 stat modifiers by hand.
- **Project conventions**: No `CLAUDE.md` present. Follows the `EDITION_CAPABILITIES`-style exhaustive-record pattern established elsewhere.
- **DRY/YAGNI**: `injuryTableOpts` is computed once and reused across range display, roll resolution, and grouping in `fighter-injury-list.tsx` — good, avoids the three call sites disagreeing.

### Verdict: **Approve**

The one finding is a narrow edge case (an unresolved/null edition slug) that's already a known, documented rough edge from the sibling edition-switching work rather than something this PR introduces outright — worth a follow-up to make the "no edition" default consistent between the API and the client tables, but not a blocker.

cc @joeseos

---

PR #1939, opened 2026-08-06
Added support to select the hatred target when applying injuries with…

joeseos:
… the hatred condition


joeseos:
## Code Review

Reviewed against the full codebase (`main...feature/support-for-hatred-injuries`, +881/-252, 19 files). This adds Hatred (X) target selection for N26's three Enmity injuries (Eternal/Bitter/Personal — against a gang type, gang, or fighter respectively), generalizing the old hardcoded "Bitter Enmity → enemy gang" support into a data-driven system: the effect type declares `type_specific_data.hatred_target`, and both the picker UI and the `add_fighter_injury` RPC read that instead of branching on the injury's name. No `CLAUDE.md` in this repo to check convention adherence against.

This is a clean generalization — `utils/injuryTarget.ts` replaces `utils/bitterEnmityDisplay.ts` cleanly (I checked every removed export is either re-pointed at the new module or genuinely gone, no dangling imports), the RPC's `campaign_gangs` migration comment explains *why* three separate injuries now exist instead of one, and the admin-injuries.tsx fix (spreading `type_specific_data` before the form's own five keys) fixes a real pre-existing bug where saving any injury via the modal silently wiped `skill_id`/`hatred_target`/`special_rules_to_add`/`d66` — worth calling out as a good catch bundled into this PR.

### Findings

**1. [Medium] `add_fighter_injury.sql`'s "shares a campaign" check for `gang`/`fighter` Hatred targets doesn't filter by `status = 'ACCEPTED'`, unlike the function's own permission check just above it**
`supabase/functions/add_fighter_injury.sql` (the `v_shares_campaign` `EXISTS` queries for `hatred_target = 'gang'` and `hatred_target = 'fighter'`)
Both queries join `campaign_gangs cg1`/`cg2` on `campaign_id` with no status predicate:
```sql
SELECT EXISTS (
  SELECT 1 FROM campaign_gangs cg1
  INNER JOIN campaign_gangs cg2 ON cg1.campaign_id = cg2.campaign_id
  WHERE cg1.gang_id = v_gang_id AND cg2.gang_id = in_hatred_target_id
) INTO v_shares_campaign;
```
Contrast with the arbitrator-permission check ~40 lines earlier in the same function, which correctly requires `cg.status = 'ACCEPTED'`. `campaign_gangs.status` is `'PENDING'` until an owner/arbitrator accepts it (`app/actions/campaigns/[id]/campaign-gangs.ts`), so as written, a gang that merely has a *pending* invite to a campaign the target gang is also (pending or accepted) in would satisfy this check — the RPC is `SECURITY DEFINER` and directly callable, so this is the actual enforcement boundary, not just a UI nicety. In practice the battle-session picker sidesteps this entirely (it scopes `candidateGangs` to the session's actual participants, which is correct regardless), but the fighter page's standalone "Add Injury" flow builds its candidate list from the fighter's campaign memberships and would inherit this if the underlying `/api/campaigns/campaign-gangs` route doesn't itself filter to accepted members.
*Fix:* add `AND cg1.status = 'ACCEPTED' AND cg2.status = 'ACCEPTED'` to both `EXISTS` checks, mirroring the permission check above.

**2. [Low, non-blocking] Stale `utils/hatredTarget.ts` references — the file is `utils/injuryTarget.ts`**
Five comments across this PR point at a filename that was never created (the module is named `injuryTarget.ts` throughout, per its own `export type HatredTargetKind` and `export function readHatredTarget`):
- `supabase/migrations/20260806170000_add_hatred_target_to_enmity_injuries.sql`: "Values must match HatredTargetKind in utils/hatredTarget.ts."
- `supabase/functions/add_fighter_injury.sql`: "see utils/hatredTarget.ts" (denormalised-snapshot comment)
- `types/fighter-effect.ts`: "Read both via utils/hatredTarget.ts rather than touching these directly."
- `types/fighter.ts`: "See utils/hatredTarget.ts."
Harmless (comments only, doesn't affect behavior), but worth a find-and-replace before merge so a future `grep` for the filename doesn't come up empty.

### What checked out cleanly
- `add_fighter_injury.sql`'s `gang_type` target validation correctly scopes to the fighter's own gang's edition (`COALESCE(gt.edition_id, cgt.edition_id)`, mirroring the official-then-custom fallback used elsewhere) and uses `IS NOT DISTINCT FROM` so NULL editions compare correctly rather than silently failing every match.
- `fighter-skills-list.tsx`'s badge rendering was correctly generalized from `item.injury_name === BITTER_ENMITY_EFFECT_NAME` to "does this skill have a recorded hatred target at all" — necessary now that three injuries (not one) can grant a target, and it still degrades gracefully to an unlinked span for `gang_type` targets (no page to link to).
- `getFighterSkills`/`getGangFightersList` (`app/lib/shared/fighter-data.ts`, `app/lib/shared/gang-data.ts`) were updated symmetrically via the same `readHatredTarget` helper — no risk of the two loaders drifting from each other.
- `readHatredTarget`'s fallback to the legacy `bitter_enmity_target_gang_*` keys for the ~760 historical un-backfilled instances is a reasonable, clearly-documented permanent shim rather than a forgotten migration step.
- The RPC's `DROP FUNCTION` list and the note explaining *why* (Postgres can't rename a parameter via `CREATE OR REPLACE`) is accurate and the 5-UUID signature drop is present.

### Checklist
- **Correctness**: Solid outside finding #1. Name/colour denormalization, edition-scoped gang-type validation, and the skirmish-mode "no target required" path all check out.
- **Edge cases / null handling**: `IS NOT DISTINCT FROM` for edition comparison, optional-target handling (`IN in_hatred_target_id IS NOT NULL`), and legacy-data fallback are all handled deliberately.
- **Security**: Finding #1 is the one real gap — everything else (admin/arbitrator permission check, ownership checks) is unchanged and correct.
- **Performance**: No new hot-path queries; the RPC's extra lookups only run when a target is actually supplied.
- **Readability/maintainability**: Excellent — the "why three injuries, why data-driven, why this replaces the old hardcoded check" reasoning is documented at both the migration and the RPC.
- **Project conventions**: No `CLAUDE.md` present. Consistent with the edition-scoped, data-driven pattern established in the prior PRs in this series.
- **DRY/YAGNI**: `readHatredTarget`/`injuryAggregationLabel`/`hatredTargetHref` are shared and reused correctly rather than reimplemented per call site.

Finding #1 is a real gap in the RPC's own validation logic, inconsistent with the pattern the function already establishes for itself one block earlier — worth fixing before merge since this function is the actual security boundary. Finding #2 is a trivial cleanup that can ride along with it.

cc @joeseos

---

PR #1942, opened 2026-08-06
Resolve editions server-side instead of fetching /api/editions per page

joeseos:
PR #1933 introduced two ways to ask what edition a row belongs to: a slug
joined server-side (gangs, campaigns, members table) and a uuid compared
client-side (custom assets, create modals, campaign edit modal). The uuid
path needed the editions table in the browser purely to translate the
hardcoded EDITION_N23 constant into a uuid, so /api/editions was fetched on
every home page and campaign page load.

Rows now carry their edition as a slug, resolved by the server fetch that
already loads them — the pattern app/lib/shared/gang-data.ts and
get-user-gangs.ts were already using:

- New app/lib/editions.ts caches the editions table with unstable_cache
  (revalidate: false, GLOBAL_EDITIONS tag) and exposes getEditionIdBySlug
  for the write path.
- The remaining fetchers join editions:edition_id (slug) and flatten it via
  withEditionSlug: the six app/lib/customise/* loaders, app/page.tsx,
  app/campaigns/[id]/page.tsx, and the gang-types / gang-variant-types routes.
- matchesHomeEditionId is gone; sameEditionSlug is the only predicate.
  useHomeEdition is now a pure localStorage store with no network dependency,
  so edition-scoped lists filter correctly on first paint instead of
  rendering empty until a fetch resolves.
- Customise server actions take an edition_slug and resolve the uuid
  themselves, rather than trusting one supplied by the client.
- /api/editions stays for the admin EditionSelect dropdown and is served from
  the cached lib, so even those calls cost no Supabase query.

Also drops the save-time trading-post filtering in campaign-edit-modal. It
re-filtered form state against a list that was empty whenever the editions
fetch had not resolved or had failed, so saving the modal in that state wrote
trading_posts: [] and wiped the campaign's authorised trading posts. The
checkbox list is already edition-filtered, so form state cannot hold an
out-of-edition id.


joeseos:
Reviewed against the codebase (checked out the branch to trace every consumer of the endpoints this PR changes, since the risk here is an incomplete migration rather than a logic error).

### What it does
Moves edition resolution server-side: rows now arrive from their loader already carrying `edition_slug` (via a new `withEditionSlug`/`editionSlugFromJoin` pair in `types/edition.ts` and a cached `app/lib/editions.ts`), instead of the browser fetching `/api/editions` on every page just to translate a hardcoded `EDITION_N23` constant into a uuid for client-side comparison. Also removes a save-time re-filter in `campaign-edit-modal.tsx` that could wipe a campaign's trading posts if the (now-deleted) client-side editions fetch hadn't resolved yet.

### Finding: real regression, not just a missed cleanup

**`components/fighter/injury-hatred-target-picker.tsx` still reads a field this PR removed from the API it calls.**

`app/api/gang-types/route.ts` used to `select(... , edition_id)` and return `edition_id` on every row. This PR changes that select to `editions:edition_id (slug)` and runs the result through `withEditionSlug`/`editionSlugFromJoin`, so the response no longer has an `edition_id` field at all — only `edition_slug`.

`injury-hatred-target-picker.tsx` (not touched by this PR) still declares:
```ts
interface GangTypeRow { gang_type_id: string; gang_type: string; edition_id: string | null; is_custom?: boolean; }
...
.filter(row => (editionId ? row.edition_id === editionId : false))
```
Since `/api/gang-types` no longer sends `edition_id`, `row.edition_id` is now always `undefined`, so this filter is always false — **the "Against [gang type]" Hatred-injury target picker will show zero gang types**, for every edition, after this ships. This is the exact same "fetch `/api/editions` client-side just to turn a slug into a uuid for comparison" pattern the rest of this PR eliminates (see the `useEditions({ enabled: needsGangTypes })` call a few lines above it) — it just wasn't caught because it lives outside the six loaders/components enumerated in the description.

Fix is small and actually simplifies this file further than the PR already does elsewhere: change `GangTypeRow.edition_id` → `edition_slug`, compare `row.edition_slug === editionSlug` directly, and drop the now-unnecessary `useEditions()` call and `editions.find(...).id` lookup in `gangTypeOptions` entirely — that removes another live `/api/editions` fetch this PR's description doesn't mention.

I checked the other three consumers of `/api/gang-types` (`create-gang-modal.tsx`, `gang-edit-modal.tsx`, `custom-fighters.tsx`) — none of them read `.edition_id`, so this is the only broken call site.

### Checked and looks correct
- **`campaign-edit-modal.tsx` trading-post filter removal**: verified the claim — `formValues.tradingPosts`/`customTradingPosts` can only be populated via the checkbox handlers, which only ever add ids sourced from the already edition-filtered `editionTradingPostTypes`/`editionCustomTradingPostTypes` lists. The only residual risk is if `campaignData.trading_posts` itself (the initial/DB value) ever contains a stale cross-edition id from before edition-scoping existed — in that narrow case the old buggy re-filter would have silently cleaned it up on next save, and now it won't. Given the far worse failure mode it replaces (wiping *all* trading posts whenever the client fetch hadn't resolved), this is a reasonable trade.
- **Custom-content create forms now hardcode `edition_slug: EDITION_N23`** instead of the previous `edition_id: null` default (`custom-equipment.tsx`, `custom-fighters.tsx`, `custom-gang-types.tsx`, `custom-skills.tsx`, `custom-trading-posts.tsx`, `custom-collections.tsx`). At first glance this looks like an undisclosed behavior change, but `home-tabs.tsx` already carries a comment stating the whole Custom Assets tab is intentionally N23-only until "skill-type and fighter authoring pickers are edition-filtered" — so this hardcoding is consistent with an existing, disclosed limitation, not a new one. Flagging only so it's confirmed as intentional rather than assumed, since the PR description itself doesn't mention it (it's framed purely as a caching refactor).
- **Completeness of the `/api/editions` / `matchesHomeEditionId` cleanup**: grepped for both — zero remaining `matchesHomeEditionId` references, and `/api/editions` itself is only still fetched from `edition-select.tsx` (the admin picker), exactly as the description says it should be.
- **`withEditionSlug`/`editionSlugFromJoin`**: correctly handles PostgREST's to-one embed returning either an object or (per the comment) a widened array, and `sameEditionSlug`'s `null`/`undefined` → `EDITION_N23` fallback is consistent with `getEditionIdBySlug`'s default.
- `useSyncExternalStore`-based `useHomeEdition` (now pure localStorage, no network) is a nice fix for the "edition-scoped lists render empty until fetch resolves" flash the description calls out.

### Checklist
- Correctness: one confirmed regression (above), otherwise sound
- Edge cases / null handling: good — explicit fallbacks throughout, `NO_LIMITS`-style defensive defaults
- Security: server actions resolving the edition uuid themselves rather than trusting a client-supplied one is a good tightening, not just a refactor
- Performance: does what it says — eliminates the per-page-load `/api/editions` round trip
- Readability/maintainability: comments explain non-obvious tradeoffs well (the campaign-edit-modal comment in particular)
- Conventions: matches the `withEditionSlug` pattern consistently across all touched loaders
- DRY/YAGNI: appropriately scoped, though the missed consumer shows the value of a repo-wide grep before calling a migration complete

Just for the `injury-hatred-target-picker.tsx` fix — everything else here is solid.

---

PR #1944, opened 2026-08-07
Add `is_vehicle` support and adapt fighter/vehicle UI and data flow

joeseos:
### Motivation
- Introduce explicit vehicle handling for fighters so N26 vehicle-type fighters are treated differently from N23 crew-with-vehicle records and to correctly resolve edition slugs for custom types. 
- Ensure UI and data queries correctly surface vehicle-specific equipment, stats and effects while hiding irrelevant weapon/injury sections for vehicles.

### Description
- Add an `is_vehicle` flag through the stack: include `is_vehicle` and `editions` on `custom_fighter_type` and return `is_vehicle`/`editions` from `getFighterTypeInfo` and `getFighterBasic` queries, and add `is_vehicle` to types in `types/fighter.ts` and `app/lib/shared/fighter-data.ts`.
- Compute `isVehicle` and `fighterEditionSlug` in the server page (`app/fighter/[id]/page.tsx`) so `edition_slug` prefers custom type edition or falls back to gang edition for vehicles.
- Propagate `is_vehicle` into components and use it to change UI/behavior: update `FighterDetailsCard`, `fighter-page.tsx`, and related components to hide or show vehicle-only sections, to prevent showing weapons and injuries lists for vehicle fighters, and to treat crew-hosted vehicles (N23) differently from vehicle fighters (N26).
- Fix vehicle stat and slot calculations in `FighterDetailsCard`: improve application of effect modifiers, explicitly skip slot modifiers when computing numeric stats, and count occupied slots based on explicit slot modifiers; minor whitespace/formatting cleanups.
- Adjust log modal and vehicle equipment/vehicle damages modals so vehicle IDs are only included when appropriate and pass the resolved `editionSlug` consistently.

### Testing
- Ran TypeScript typecheck and updated types to reflect new `is_vehicle` and `editions` fields, and the typecheck completed without errors. 
- Ran the existing automated test suite and UI component tests (unit/integration) locally, and they succeeded. 
- Ran linters/formatters to ensure no syntactic issues; linting passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a749b31c31c83328b80d760d09a5165)

### DIFF
--- a/app/fighter/[id]/page.tsx
+++ b/app/fighter/[id]/page.tsx
@@ -159,6 +159,13 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
 
     // Custom fighter type gang info is embedded in fighterBasic via the FK join
     const customFighterTypeInfo = fighterBasic.custom_fighter_type ?? null;
+    const isCustomFighterType = Boolean(fighterBasic.custom_fighter_type_id);
+    const isVehicle = isCustomFighterType
+      ? Boolean(customFighterTypeInfo?.is_vehicle)
+      : Boolean(fighterTypeData?.is_vehicle);
+    const fighterEditionSlug = isCustomFighterType
+      ? customFighterTypeInfo?.editions?.slug ?? (isVehicle ? gangBasic.edition_slug : null)
+      : fighterTypeData?.editions?.slug ?? null;
 
     // Campaign data already includes trading_post_names from getGangCampaigns, no processing needed
 
@@ -278,7 +285,8 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
         credits: totalCost,
         alliance_crew_name: fighterTypeData?.alliance_crew_name,
         is_spyrer: fighterTypeData?.is_spyrer || false,
-        edition_slug: fighterTypeData?.editions?.slug ?? null,
+        is_vehicle: isVehicle,
+        edition_slug: fighterEditionSlug,
         fighter_type: {
           fighter_type_id: fighterTypeData?.id || fighterBasic.custom_fighter_type_id || '',
           fighter_type: fighterBasic.fighter_type || fighterTypeData?.fighter_type || 'Unknown',
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -39,6 +39,8 @@ export interface FighterBasic {
   custom_fighter_type?: {
     gang_type_id: string | null;
     custom_gang_type_id: string | null;
+    is_vehicle?: boolean;
+    editions?: { slug: string } | null;
   } | null;
   fighter_gang_legacy_id?: string | null;
   fighter_gang_legacy?: {
@@ -159,7 +161,9 @@ export const getFighterBasic = async (fighterId: string, supabase: any): Promise
           custom_fighter_type_id,
           custom_fighter_type:custom_fighter_type_id (
             gang_type_id,
-            custom_gang_type_id
+            custom_gang_type_id,
+            is_vehicle,
+            editions:edition_id (slug)
           ),
           fighter_gang_legacy_id,
           fighter_gang_legacy:fighter_gang_legacy_id (
@@ -1183,8 +1187,8 @@ export const getFighterTypeInfo = async (fighterTypeId: string | null, supabase:
   fighter_type: string;
   alliance_crew_name?: string;
   is_spyrer?: boolean;
+  is_vehicle?: boolean;
   gang_type_id?: string | null;
-  edition_id?: string | null;
   editions?: { slug: string } | null;
 } | null> => {
   if (!fighterTypeId) return null;
@@ -1193,7 +1197,7 @@ export const getFighterTypeInfo = async (fighterTypeId: string | null, supabase:
     async () => {
       const { data, error } = await supabase
         .from('fighter_types')
-        .select('id, fighter_type, alliance_crew_name, is_spyrer, gang_type_id, edition_id, editions:edition_id (slug)')
+        .select('id, fighter_type, alliance_crew_name, is_spyrer, is_vehicle, gang_type_id, editions:edition_id (slug)')
         .eq('id', fighterTypeId)
         .single();
 
--- a/components/fighter/fighter-details-card.tsx
+++ b/components/fighter/fighter-details-card.tsx
@@ -354,6 +354,10 @@ export const FighterDetailsCard = memo(function FighterDetailsCard({
   const canShowEditButtons = userPermissions.canEdit;
   const isCrew = fighter_subtypes.includes('Crew');
   const isVehicle = is_vehicle ?? false;
+  // Only an N23 crew reads its profile off an attached vehicle record. An N26 vehicle is the
+  // fighter, so it keeps the ordinary statline. N26 has no Crew subtype today, so the guard is
+  // consistency rather than a live case — but the statline and its limits must agree either way.
+  const showsVehicleProfile = isCrew && !isVehicle;
 
   const handleImageClick = () => {
     if (canShowEditButtons) {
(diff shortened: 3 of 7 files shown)


joeseos:
Reviewed with the scope in mind — this is the first PR toward vehicle support, focused on correctly showing an N26 vehicle-as-fighter on the fighter page, with hardpoint/damage editing for N26 explicitly and visibly deferred. I focused on breakage risk and cleanliness rather than feature completeness.

### Finding: crash risk when `isCrew` and `is_vehicle` are both true

`components/fighter/fighter-details-card.tsx:355-377`:
```ts
const isCrew = fighter_subtypes.includes('Crew');
const isVehicle = is_vehicle ?? false;
...
const vehicleStats = useMemo(() =>
  isCrew ? calculateVehicleStats(vehicles?.[0]) : null,
  [isCrew, vehicles]
);
```
`calculateVehicleStats` (line ~123) dereferences its argument immediately (`baseStats.movement`, `baseStats.front`, …) with no null guard. Every other place in this same file that branches on `isCrew` for vehicle-record display was updated to `isCrew && !isVehicle` (lines 580, 591, 640) — specifically because an N26 vehicle-as-fighter has no `vehicles[0]` row to read. This one call site is the exception: it's still gated on `isCrew` alone. If a fighter ever has `fighter_subtypes` containing `'Crew'` *and* `is_vehicle: true` at the same time, `vehicles?.[0]` is `undefined`, and `calculateVehicleStats(undefined)` throws inside a `useMemo`, taking down the whole card render. The `stats` object built right below it (lines 381-388) also assumes `vehicleStats` is populated whenever `isCrew` is true, so it inherits the same gap.

Given the rest of the file treats `isCrew && isVehicle` as a real, guarded-against case, I'd add the same `!isVehicle` guard here — either `isCrew && !isVehicle ? calculateVehicleStats(...) : null`, or make `calculateVehicleStats` itself null-safe as defense in depth. If it's actually impossible for a current N26 vehicle fighter type to carry the `'Crew'` subtype, worth a one-line comment saying so, since nothing else in the file assumes that.

### Everything else checked out

- **`vehicle-lasting-damages.tsx`** is properly defensive throughout: every action that needs a real `vehicleId` (add, repair, damage-roll logging) is now gated with `!vehicleId` checks, and the one unconditional read of the `vehicle` prop (repair-cost calculation) already has its own `if (!vehicle) { setRepairCost(0) }` guard. Good, consistent handling.
- **`vehicle-equipment-list.tsx`**: the purchase path never touches `vehicleId` directly (it flows through `params.vehicle_id`, which is correctly `undefined` for N26 via `vehicleId={vehicle?.id}` in the parent), so buying vehicle equipment for an N26 vehicle-fighter routes through the ordinary fighter-equipment purchase path as intended. Only the hardpoint "Fit Weapon" modal is (correctly) disabled via `fitWeaponData && vehicleId`.
- **`fighter-page.tsx`**'s `onEquipmentUpdate`/`onDamageUpdate` handlers for the N26 branch correctly keep `fighterData.equipment` and `fighterData.vehicleEquipment` in sync (explicit comment: "On N26 this list *is* the fighter's equipment, so keep both in step"), and the damage-update reducer symmetrically writes to `fighter.effects` instead of a `vehicles[]` entry when there's no `vehicleId`. Good attention to the two data shapes.
- Server-side (`app/fighter/[id]/page.tsx`, `app/lib/shared/fighter-data.ts`): `is_vehicle`/`editions` added consistently to both the official and custom fighter-type queries, mirroring the existing `editions:edition_id (slug)` join pattern already used elsewhere in this file — no shape mismatch.

### Minor cleanliness notes (non-blocking)

- **Two separate "is this a vehicle" computations in `fighter-page.tsx`** that could in principle diverge: `transformFighterData`'s local `isVehicleFighter = Boolean(fighterData.fighter?.is_vehicle) && !fighterData.fighter?.vehicles?.[0]` (used only to pick how to build `vehicleEquipment`) vs. the component-level `isVehicle = fighterData.fighter.is_vehicle ?? false` (used to hide `WeaponList`/`InjuriesList` and gate `vehicleView`). If a fighter ever had both `is_vehicle: true` and a populated `vehicles[0]` (a transitional/bad-data state), the two would disagree — `vehicleEquipment` would build off the N23 `vehicles[0].equipment` shape while `WeaponList`/`InjuriesList` would still be hidden. Probably impossible given the data model, but since both booleans exist in the same file for the same purpose, collapsing them into one derived value (or naming them so the difference is obvious) would remove the question entirely.
- The `withoutDeleted` generic helper extracted in the vehicle-equipment `onEquipmentUpdate` handler is a nice small dedup over the old copy-pasted category-filter loop — good cleanup.
- Hoisting `editionSlug`/`vehicleView` to single `const`s at the top of the component instead of repeating `fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null` at every call site is a solid readability win.
- The diff carries a fair amount of pure whitespace/trailing-space cleanup mixed into the functional changes (e.g. throughout `fighter-details-card.tsx` and `fighter-page.tsx`) — harmless, but it does inflate the diff and makes the functional changes slightly harder to scan. Not asking for a change, just noting it for next time if you want tighter diffs.

Just for the `vehicleStats`/`calculateVehicleStats` null-safety gap — that's the one path here that can actually crash the page rather than just under-render. Everything else is solid, well-scoped, and consistent with the "N26 vehicle is the fighter" model the PR sets out to establish.

---

joeseos:
Follow-up review of the 4 commits since the last review (`01eaf10`…`d2c799c`, current HEAD). Skipped the lasting-damage button changes (`a486d3a`) per your note — that's for the later PR.

## Previous finding

**`calculateVehicleStats` crash risk when `isCrew` and `is_vehicle` are both true — Resolved ✅**
`01eaf10` collapses all five call sites onto one `showsVehicleProfile = isCrew && !isVehicle` (`fighter-details-card.tsx:357`), including the `vehicleStats` `useMemo` that previously branched on `isCrew` alone. This is the right fix — one derived value instead of five independently-maintained conditions, and it's paired with `FighterDetailsStatsTable`'s `isCrew` prop so the stats table's limit-checking (`CREW_LIMITS` vs fighter limits) can't disagree with which statline is actually rendered. Good catch pairing those two.

## N23/N26 vehicle-separation design — the part you asked me to focus on

The core abstraction is `vehicleView` in `fighter-page.tsx:603`:
```ts
const vehicleView = vehicle
  ? { vehicleId: vehicle.id, effects: vehicle.effects ?? {} }
  : isVehicle
    ? { vehicleId: null, effects: fighterData.fighter.effects }
    : null;
```
This is a clean unification — N23's attached `vehicles[0]` row and N26's vehicle-as-fighter both resolve to the same `{ vehicleId, effects }` shape, and every downstream consumer (`VehicleEquipmentList`, `VehicleDamagesList`, the log modal) reads through it rather than re-deriving N23-vs-N26 itself. `vehicleId: string | null` propagating through `VehicleEquipmentList`'s props (with the `fitWeaponData && vehicleId` guard disabling hardpoint-fitting on N26) is exactly the right way to keep hardpoint editing deferred without special-casing it at every call site.

**One thing still outstanding from the previous review, and worth resolving explicitly given it's the crux of this design:** `vehicleView`'s branch order picks the N23 path whenever `vehicle` (`vehicles?.[0]`) is truthy, *regardless of* the `isVehicle` flag — but the raw `isVehicle` boolean (not `vehicleView`) is what gates `WeaponList`/`InjuriesList`/`AddWeapon` visibility (`fighter-page.tsx:785,971,1372`) and, separately, `showsVehicleProfile` in `fighter-details-card.tsx`. Today these agree, because `isVehicleFighter` (`fighter-page.tsx:241`, used for `transformedVehicleEquipment`) and `vehicleView`'s N26 branch both require `!vehicles?.[0]`, and the file comments assert N26 fighters never carry `vehicles[0]` or the `Crew` subtype. But if that invariant is ever violated (bad data, or a future migration path that attaches a legacy `vehicles` row to an `is_vehicle: true` fighter), the UI splits: weapons/injuries would hide as if it's an N26 vehicle, while the vehicle-equipment section would read off the N23 `vehicles[0]` row instead of the fighter's own equipment — a mixed state that's hard to debug because no single flag owns the decision.

Given there are now three independent places encoding "is this N23 or N26" (`isVehicle`, `isVehicleFighter`, `vehicleView`'s `vehicle`-truthiness), and you're about to build the N26 hardpoint/lasting-damage follow-up on top of this, I'd suggest picking one owning check now rather than after more code depends on the current shape — e.g. make `vehicleView` branch on `isVehicle` first (`isVehicle ? {vehicleId: null, ...} : vehicle ? {...} : null`), matching `showsVehicleProfile`'s `isCrew && !isVehicle` pattern, so a stray `vehicles` row can never silently steer the N23 path while `isVehicle` says otherwise. Not blocking — the invariant holds today — but it's the one piece of this design with room to drift.

## Everything else in this batch

- `vehicle-lasting-damages.tsx`'s doc-comment tweak (`a486d3a`) is just wording, consistent with the button change you asked me to skip.
- No new null-handling, security, or performance issues introduced by these 4 commits.

## Verdict: **Approve**

The crash-risk fix is correct and complete. The one open item is a design-consistency question, not a live bug — worth a decision (either collapse to one flag or leave a comment recording the invariant) before the hardpoint/lasting-damage follow-up builds more on top of `vehicleView`.

---

PR #1945, opened 2026-08-07
Make the Add XP modal edition-aware for N26

FlorianWeissDev:
## Why

The XP award list in the shared Add XP modal was hard-coded to N23, so N26 gangs were offered awards their edition does not have and were missing two it does.

N26 changes the list in three ways:

| Award | N23 | N26 |
| --- | --- | --- |
| Successful Rally | ✅ | ❌ removed |
| Leader/Champion (+1) | ✅ | widened to **Leader/Champion/Brute/Vehicle** |
| Controlled an Objective (+1, once per battle) | ❌ | ✅ added |

## What changed

The award definitions move out of the modal into a new **`utils/xpCases.ts`**, following the `utils/characteristicLimits.ts` pattern for edition-keyed data: a `Record<EditionSlug, readonly XpCaseId[]>` gets the same compile-error-on-a-new-edition guarantee that `EDITION_CAPABILITIES` gives the capability flags, while `EditionCapabilities` stays boolean-only. This is *data* rather than a behaviour switch, so it does not belong in the capability record.

- **Unknown edition degrades safely.** An unset or unrecognised slug offers no presets and leaves the manual XP amount field, mirroring `NO_CAPABILITIES` rather than quietly aliasing to a real edition. That makes an empty award list reachable for the first time, so the separator between the counters and the checkboxes now only renders when both sides have entries.
- **N26's wider elite bonus is its own `eliteTargetBonus` id**, not a relabelled `leaderChampionBonus`. Breakdown ids are persisted in fighter logs and `formatXpBreakdown` is server-side with no edition context, so a single id cannot carry two meanings.
- **Shared cases keep their existing N23 labels** for that same reason, which leaves the N23 modal unchanged.
- **`app/actions/logs/log-helpers.ts` derives its two lookup maps from the registry** instead of duplicating them, but stays typed `Record<string, …>`: it indexes with ids read back from stored payloads, which may predate the union or outlive an award being retired from every edition's list. Historic `rally` entries still format correctly.

The three entry points that render the modal — the gang overview, the fighter page, and the battle session participant card — each pass their edition down. The edition is taken from the **gang**, not the fighter, since a fighter-derived slug resolves through `fighter_types` and is `null` for custom fighter types. The battle session already had `editionSlug` in scope from the battle-session edition work, so that one is a single prop.

No database or server action change: `updateFighterXpWithOoa` already accepts a free-form `xp_breakdown` map and was edition-agnostic.

## Out of scope

- Advancements and ranks.
- Persisting the objective award across modal openings — it is once per *modal opening*, not enforced once per battle.
- Gating Assistance and objectives for Vehicles and Pets. N26 does forbid those, but the modal has no fighter-type props today and the existing N23 modal does not enforce the comparable per-activation cap either, so this does not make N26 stricter than N23.

## Testing

Verified with `npm run lint` (only pre-existing errors in unrelated files remain) and `npm run build`, which type-checks the whole project.

> [!NOTE]
> Not yet exercised against a running app — the local Supabase database is on a stale schema and cannot serve sign-in, so the modal has not been clicked through in either edition.

## Worth checking on staging

1. **N23 gang, no regression.** The modal still lists Cause Serious Injury, Cause OOA, Leader/Champion, Wreck Vehicle, Successful Rally, Provide Assistance and Misc. as counters, with Battle Participation as the only checkbox. Should be identical to production.
2. **N26 gang.** Successful Rally is gone, Leader/Champion reads Leader/Champion/Brute/Vehicle, and Controlled an Objective appears as a second checkbox below Battle Participation.
3. **N26 totals.** Cause OOA 1 + Leader/Champion/Brute/Vehicle 1 totals 3 XP, and the fighter's XP and OOA count both increase correctly after saving.
4. **N26 fighter log.** The saved entry names the two new awards in full rather than printing raw ids like `eliteTargetBonus` or `objectiveControl`.
5. **An older N23 log entry containing Successful Rally still renders its label**, not a raw `rally` id. This is the main risk in deriving the log lookup maps from the new registry, and needs existing staging data rather than a freshly created entry.
6. **All three entry points** show the right list for the same fighter: gang overview card menu, fighter detail page, battle session participant card. The battle session is the one to watch, since its edition comes from the session rather than the gang.
7. **A custom fighter type in an N26 gang** still gets the N26 list, confirming the edition is read from the gang and not the fighter type.
8. **The optional OOA / wrecked-crew target picker** still opens, records targets and increments kills in both editions.

### DIFF
--- a/app/actions/logs/log-helpers.ts
+++ b/app/actions/logs/log-helpers.ts
@@ -1,26 +1,21 @@
-/** XP case id → display label (must match fighter-xp-modal xpCountCases) */
-export const XP_CASE_LABELS: Record<string, string> = {
-  seriousInjury: 'Cause Serious Injury',
-  outOfAction: 'Cause OOA',
-  leaderChampionBonus: 'Leader/Champion',
-  vehicleWrecked: 'Wreck Vehicle',
-  rally: 'Successful Rally',
-  assistance: 'Provide Assistance',
-  misc: 'Misc.',
-  battleParticipation: 'Battle Participation',
-};
+import { XP_CASES } from '@/utils/xpCases';
 
-/** XP per case (id → XP each) - must match fighter-xp-modal */
-export const XP_CASE_VALUES: Record<string, number> = {
-  seriousInjury: 1,
-  outOfAction: 2,
-  leaderChampionBonus: 1,
-  vehicleWrecked: 2,
-  rally: 1,
-  assistance: 1,
-  misc: 1,
-  battleParticipation: 1,
-};
+/**
+ * XP case id → display label / XP each, derived from the single award registry
+ * in utils/xpCases.ts.
+ *
+ * Deliberately typed Record<string, …> rather than Record<XpCaseId, …>: these
+ * are indexed with ids read back from persisted log payloads, which may predate
+ * the union or outlive an award being retired from every edition's list. Both
+ * lookups below fall back rather than assuming a hit.
+ */
+export const XP_CASE_LABELS: Record<string, string> = Object.fromEntries(
+  Object.entries(XP_CASES).map(([id, def]) => [id, def.label])
+);
+
+export const XP_CASE_VALUES: Record<string, number> = Object.fromEntries(
+  Object.entries(XP_CASES).map(([id, def]) => [id, def.xp])
+);
 
 export function formatXpBreakdown(breakdown: Record<string, number>, miscNote?: string): string {
   return Object.entries(breakdown)
--- a/components/battle-session/participant-card.tsx
+++ b/components/battle-session/participant-card.tsx
@@ -339,6 +339,7 @@ function FighterActionModal({
           currentKillCount={fighterData.kill_count}
           gangId={gangId}
           campaignId={campaignId ?? undefined}
+          editionSlug={editionSlug}
           onClose={() => setShowXpModal(false)}
           onXpUpdated={(newXp) => {
             const delta = newXp - (fighterData?.xp ?? 0);
--- a/components/fighter/fighter-page.tsx
+++ b/components/fighter/fighter-page.tsx
@@ -1271,6 +1271,7 @@ export default function FighterPage({
               is_spyrer={fighterData.fighter.is_spyrer}
               gangId={fighterData.gang?.id}
               campaignId={fighterData.fighter?.campaigns?.[0]?.campaign_id}
+              editionSlug={fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null}
               onClose={() => handleModalToggle('addXp', false)}
               onXpUpdated={handleXpUpdated}
             />
--- a/components/fighter/fighter-xp-modal.tsx
+++ b/components/fighter/fighter-xp-modal.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { useState } from 'react';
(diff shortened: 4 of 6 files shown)


joeseos:
## Code review

Overall, I think the architectural direction is good and fits the existing N23/N26 design. I would keep the edition-aware XP registry, but streamline the implementation before/after merge.

### What's good

- **Correct edition architecture.** XP awards are edition-specific data, so keeping them in a dedicated module matches the existing `characteristicLimits.ts` / `lastingInjuryRank.ts` pattern better than adding more booleans to `EditionCapabilities`.
- **Uses the central edition type.** `xpCases.ts` correctly depends on `EditionSlug` from `types/edition.ts`. This should preferably be a type-only import: `import type { EditionSlug } from '@/types/edition';`.
- **Single source of truth for XP metadata.** Deriving log labels/XP values from `XP_CASES` removes the previous duplication between the modal and `log-helpers.ts`.
- **Historic log compatibility is considered.** Keeping log lookup tolerant of string IDs makes sense because persisted breakdown IDs may outlive the current edition lists.
- **Edition plumbing does not introduce new fetching.** The gang page, fighter page and battle-session path already have edition information available and pass it into the shared modal.
- **Unknown edition behaviour is consistent with the newer edition design.** Falling back to no presets/manual XP rather than silently treating an unknown edition as N23 matches the `NO_CAPABILITIES` style.

### What I'd fix / simplify

1. **`xpCases.ts` is larger than necessary.** The file is ~113 lines for ten definitions, two edition lists and one lookup. The structure is good, but a lot of the comments repeat architecture already established elsewhere. I would keep the file but trim it substantially.

2. **Use a type-only import for `EditionSlug`.**

   ```ts
   import type { EditionSlug } from '@/types/edition';
   ```

3. **Avoid repeatedly reaching through `gangData.processedData.edition_slug`.** `GangPageContent` already uses the same slug for other edition-aware components. Define it once:

   ```ts
   const editionSlug = gangData.processedData.edition_slug ?? null;
   ```

   and reuse:

   ```tsx
   editionSlug={editionSlug}
   ```

   for both `FighterXpModal`, `InjuriesList`, and any other edition-aware child in this component.

4. **The modal derivation can be simpler.** It currently creates `awards`, then separately memoises `xpCountCases` and `xpCheckboxCases`. These lists are tiny. Three `useMemo`s feel heavier than the work being avoided. Either let `xpAwardsFor()` return materialised definitions or derive the two arrays more directly.

5. **Tighten component-local XP state typing.** The modal now has `XpCaseId`, but still uses `Record<string, number>` / `Record<string, boolean>` for newly-created state. Broad `string` typing is justified when reading historic persisted logs, but not necessarily inside the modal. Something like `Partial<Record<XpCaseId, number>>` / `Partial<Record<XpCaseId, boolean>>` would preserve more type safety.

6. **Keep `eliteTargetBonus` only if the persisted semantic distinction is intentional.** A separate ID is defensible because logs have no edition context and the N26 meaning is wider than `leaderChampionBonus`. I would keep it if those IDs are considered permanent log contracts; otherwise this is an area worth confirming because it adds another persisted case ID.

### Verdict

The PR is **not fundamentally bloated because it introduces `xpCases.ts`** — that abstraction is appropriate. The bloat is mostly in the verbosity of that file and a little unnecessary derivation/plumbing around it. I would keep the architecture and do a cleanup pass to make the implementation smaller and easier to follow.

PR #1946, opened 2026-08-07
Updated website logo and added versioning of images in PWA to increas…

Djidiouf:
…e chances of them being updated on people's mobile devices as they are cached by default with the manifest


joeseos:
## Code Review

@Djidiouf Reviewed this PR (logo/favicon update + PWA icon cache-busting) against the codebase.

### Findings

**1. Low — orphaned unused asset**
`public/images/favicon-36x36.png` is still present and its binary content was updated, but nothing in the codebase references it anymore — `app/footer.tsx` and `app/layout.tsx` were switched to `favicon-36x36-black.png` / `favicon-36x36-white.png`. Verified with a repo-wide grep: no remaining references to `favicon-36x36.png`.
*Fix:* delete the now-dead file, or confirm it's still needed for something outside this diff (e.g. og:image, external link) before keeping it.

**2. Low — both light/dark icon variants are always fetched**
`app/footer.tsx:16-29` and `app/layout.tsx:147-160` render two `<Image>` elements (black + white) and hide one via `dark:hidden` / `hidden dark:block`. CSS `display:none` doesn't stop the browser from requesting an ``'s `src`, so both ~1.2KB PNGs load on every page view instead of one. Impact is negligible at this file size, but worth knowing it's not a "free" theme switch — a `prefers-color-scheme` media query on a single inline asset (or resolving the theme server-side before render) would avoid the extra request if this pattern gets reused for larger images.

**3. Medium — cache-busting scope doesn't cover the icon most likely to be stale**
The PR's stated goal is defeating aggressive icon caching on mobile devices, and it does this by renaming the manifest's 192×192/512×512 icons to `-pwa-v2` (confirmed: `favicon-192x192-pwa-v2.png` and `favicon-512x512-pwa-v2.png` are byte-identical to the updated `favicon-192x192.png`/`favicon-512x512.png`, so this is purely a URL-busting trick, not a content change — good approach). However, `apple-touch-icon.png` (`app/layout.tsx:52`) — the icon iOS actually pins to the home screen and caches most aggressively — keeps its original filename and wasn't versioned. iOS users are the ones most likely to still see the stale icon after this ships.
*Fix:* consider the same versioned-filename treatment for `apple-touch-icon.png` if the goal is genuinely "increase chances of update on mobile devices."

### Not an issue
- `public/images/fallback-image.png` deletion: confirmed unreferenced anywhere in the codebase — safe to remove.
- Dark/light icon swap logic (`dark:hidden` on black icon, `hidden dark:block` on white icon) is correct given `darkMode: "class"` in `tailwind.config.ts`.

### Verdict: **Approve**

No correctness bugs or regressions found. The findings above are minor follow-ups, not blockers.


---

PR #1947, opened 2026-08-07
Earn N26 Advancements by rank instead of spending XP

FlorianWeissDev:
## Why

N26 replaces N23's "spend XP to buy an Advancement" economy with a rank ladder: XP only ever accumulates, and a model earns an Advancement each time its total moves it onto a new tier. That single rule change touches the fighter row, the advancement actions, the catalog and the roster UI, so this branch lays the foundation for it rather than shipping the roll UI itself.

Rebased onto `main`; #1945 has since merged, so the diff is this branch's work only.

## What changed

**`fighters.starting_xp`** — a fighter now records the XP it was recruited with. `fighter_types.starting_xp` is editable, so reading recruitment XP from the type would mean that correcting a fighter type retroactively changes how much XP every existing fighter of that type counts as having earned, and therefore how many Advancements it is owed. Set on every recruitment path — `add-fighter`, both `copy-fighter` branches, and `createExoticBeastsForEquipment`, whose `fighter_types` embed did not select it either. `copy-gang` needs nothing; it spreads `select('*')`.

**Starting XP is nullable, and `NULL` means N/A** — on all three tables, with no column default. Neither `0` nor `1` is right for the whole catalog: N23 has no Starting XP concept at all, and N26 has models whose entry reads N/A because they can never gain XP. Any default is one edition's rule written into the schema, where it silently applies to every other edition and to editions that do not exist yet, so the value is left to the catalog entries the data admins write.

N/A affects **display only**. It renders in the XP column and gives way to the real number the moment the model holds any XP, so a group house-ruling XP onto such a model still sees it, Level Up badge and all. Nothing is blocked.

Two paths leant on the old default and would have silently started producing N/A rows: custom fighter type creation (no UI for the field) and every `fighter_types` insert in `seed.sql`. Both now pin `0`. `fighters.starting_xp` is added carrying a default and then drops it, so fighters already on a roster backfill to `0` rather than being reinterpreted as N/A.

**`utils/advancementRanks.ts`** — the rank table, keyed by `EditionSlug` the way `utils/characteristicLimits.ts` and `utils/xpCases.ts` are, so adding an edition is a compile error until it states its own ranks.

An Advancement is earned by *moving onto* a tier, so the table stores each tier's lower bound and a count is the number of those bounds above the model's starting XP and at or below its current XP. It is deliberately **not** `currentXp - startingXp`: tiers widen as XP rises, and a fighter type's starting XP exists to place the model on the track that suits it. Rebasing to zero would put a Ganger recruited on 13 XP back on the 3-wide Rookie track and advance them at twice the intended rate.

| Starting XP | Next Advancement at |
|---|---|
| 1 | 4 |
| 3 | 4 |
| 4 | 7 |
| 13 (Ganger) | 19 |

The rulebook's own example Ganger profile reads `13/19` for Current/Target XP, which matches.

N23 gets an **empty** rank list rather than a capability flag — it has no XP-based ranks at all, so every count falls out as zero on its own.

**`cumulativeXp` capability** — new row in `EDITION_CAPABILITIES`, framed as the one decision the editions answer differently: whether XP is a currency or a running total. `fighter-advancement.ts` had no edition awareness at all, so the three actions that move a fighter's XP now load the edition from the fighter's gang and gate all five sites — both affordability guards, both deductions, and the refund on delete. An unresolved slug keeps the N23 behaviour, the same way `tradePoints` leaves an unknown edition on Availability.

**The cap** — gating on XP left N26 ungated, since it spends none. The cap is now the open-advancement count, the same figure the badge shows, so the two cannot disagree. Taking an Advancement consumes it implicitly by raising the count taken, so nothing needs marking as used. Starting and free skills stay exempt.

**Level Up badge** — shows `3 Level Ups` beside the killed/retired/captured status icons on both the fighter page and the gang roster. Derived on read, so nothing can drift and entering XP needs no extra write.

**N26 Advancement catalog** — thirteen `fighter_effect_types` rows in the `advancements` category with a modifier each, edition-scoped alongside their N23 namesakes, following the N26 Lasting Injury seed. No XP cost; credits straight off the N26 table (+5 psychology, +10 Initiative/Movement, +15 WS/BS, +20 S/T/W/A/Sv), flat with no Ganger special case. Save is N26-only.

Modifier signs follow the direction each characteristic improves in — target numbers down, raw values up. **Initiative and the four psychology characteristics are raw values in N26 but target numbers in N23**, so their signs are the opposite of the N23 rows of the same name, the same inversion the N26 injury seed documents.

**`get_fighter_available_advancements`** had no edition awareness and returned every row in the category, so seeding the N26 rows would have offered them to N23 fighters. It now resolves the fighter's edition from its gang the way `add_fighter_injury` does and filters on it.

**Roll logging no longer applies N23 routing to N26.** The two roll loggers are halves of one N23 rule: Gangers and Exotic Beasts roll on their own flat-cost table and log through the advancement logger, everyone else rolls for a skill and logs through the skill logger, and each guard rejects the other's fighters. Both ran unconditionally. Since the N26 roll UI renders for *every* fighter, the advancement logger threw for every N26 subtype outside the N23 list — Leaders, Champions, Prospects, Juves, Brutes, Specialists, Beasts and Pets — and, in mirror, an N26 Ganger could not log a skill roll. Both guards now apply only where the edition routes rolls by subtype. N23 is unchanged.

## Out of scope

The roll/selection UI (`fighter-advancement-list.tsx` is still N23-shaped, and `gangerCharMap` is still Ganger-gated so a rolled result does not auto-fill the purchase), N26 promotions (`PROMOTION_MAP` is pure N23 — N26 promotes Prospect→Ganger at 13 XP and Ganger→Champion at 37 XP), batch selection of several Advancements at once, maximum-characteristic enforcement, N26 skill advancements and their costs (blocked upstream — there are no N26 `skill_types` yet), vehicles, rank titles, and the Level Up badge's appearance in printed rosters.

## Testing

> [!NOTE]
> Not exercised in a browser. Local sign-in is unavailable, so everything below is SQL-level or code-level.

- Migrations applied in order to a freshly reset local DB. Round-tripped `fighter_types.starting_xp`: omitted→`NULL`, explicit null→`NULL`, `0`→`0`, `13`→`13`. Confirmed the add-then-drop-default pattern backfills existing rows to `0` while leaving later omitted inserts as `NULL`.
- `seed.sql` runs clean with all 27 fighter type rows at `0`.
- `openAdvancementsFor` exercised over the worked examples above plus N/A, house-ruled XP on an N/A model, and zeros for n23/null/unknown slugs.
- `get_fighter_available_advancements` exercised against throwaway N26 and N23 fighters inside a rolled-back transaction. The N26 fighter sees 13 characteristics, every `xp_cost` 0, credits matching the table, `uses_flat_cost` false and no Ganger-to-Specialist row. **The N23 fighter sees only N23 rows, with flat cost and the Specialist row unchanged.**
- Roll-logger fix: the gang→gang type→edition join resolves `n23`/`n26` correctly for throwaway N23 and N26 fighters, and the new select parses against PostgREST.
- `npm run build` and `npx tsc --noEmit` clean; `eslint` clean on every touched file. The pre-existing `react-hooks` errors elsewhere in the repo are untouched.

## Worth checking on staging

- **Whether `fighter_types.starting_xp` is actually populated for the N26 catalog.** Types left at `0` put every N26 fighter on the 3-wide Rookie track; types left blank read as N/A. Both are legitimate values, so nothing will error — the numbers will just look wrong.
- `supabase/functions/*.sql` is not auto-deployed. `get_fighter_available_advancements.sql` must be applied to the remote separately, or N23 fighters will be offered the seeded N26 rows.


joeseos:
@FlorianWeissDev I went through the N26 advancement changes separately from #1945. The overall direction is good, but I found a few correctness issues I think should be fixed before merge, plus some opportunities to reduce the amount of new plumbing.

## What's good

- **`cumulativeXp` belongs in `EditionCapabilities`.** This is a real behavioural decision between editions: N23 spends/refunds XP, N26 accumulates XP and earns advancements from rank thresholds. This is exactly the kind of decision the capability registry should own.
- **Persisting `fighters.starting_xp` is the right model.** Recruitment XP is historical state and should not change if the fighter type is edited later.
- **Deriving open advancements rather than storing a separate counter is good.** `earned - taken` avoids another value that can drift.
- **Server-side gating is important.** The action does not rely only on the UI badge to decide whether an N26 fighter may take another advancement.
- **Edition-scoping the advancement catalog/RPC is correct.** N23 and N26 rows can coexist without leaking into each other.
- **No N26 XP deduction/refund is handled in the right place**, in the advancement actions rather than scattered through the UI.
- The N26 advancement seed is structured and idempotent, and keeping rank-threshold data separate from boolean edition capabilities is consistent with the existing edition-specific data modules.

## What I'd fix before merge

### 1. `starting_xp` is selected for the gang roster but never returned

`getGangFightersList()` now selects `starting_xp` and adds it to `GangFighter`, but the final `result` object does not contain:

```ts
starting_xp: fighter.starting_xp,
```

`FighterCard` therefore receives `undefined`, defaults it to `0`, and calculates N26 level-ups from the wrong starting point. A Ganger recruited at 13 XP can show multiple false Level Ups.

This looks like the most immediate bug in the current roster implementation.

### 2. Fighter detail page does not count existing skill advancements correctly

`FighterDetailsCard` calculates:

```ts
countAdvancementsTaken(effects, advancements?.skills)
```

but `transformFighterData()` puts the actual DB skills in `fighter.skills`. The server payload also does not appear to populate `fighter.advancements.skills` on initial load.

So characteristic advancements are counted from `effects.advancements`, but existing skill advancements can be missed, causing the Level Up count to be too high.

I would pass/use the actual `skills` record here rather than the legacy/parallel `advancements?.skills` shape.

### 3. Use the gang edition for the Level Up logic, especially for custom fighter types

We just handled this in #1945: a fighter-derived edition slug can be null for custom fighter types, while the gang already has the authoritative edition.

On the fighter page the new badge currently gets:

```tsx
edition_slug={fighterData.fighter?.edition_slug ?? null}
```

while the XP modal correctly uses the gang fallback. I would use the same resolved gang edition for both.

The gang roster has the same underlying issue because `getGangFightersList()` currently derives `edition_slug` from `fighter_types`, which does not cover custom fighter types. Since every fighter in that view belongs to the same gang, passing/reusing the gang edition is cleaner than deriving the same edition independently for each fighter.

### 4. Non-experienced copies are put on the wrong N26 rank track

`copy-fighter.ts` currently does:

```ts
xp: copy_as_experienced ? sourceFighter.xp : 0,
starting_xp: copy_as_experienced ? sourceFighter.starting_xp : 0,
```

For N26, a "base" copy of a fighter recruited at 13 XP should still start on the 13-XP track; otherwise it advances at 4/7/10/etc. instead of having its next advancement at 19.

I think the non-experienced path should preserve recruitment XP and reset current XP to it, roughly:

```ts
starting_xp: sourceFighter.starting_xp,
xp: copy_as_experienced ? sourceFighter.xp : sourceFighter.starting_xp,
```

Same for copied exotic beasts.

### 5. The new DB `DEFAULT 1` is currently bypassed by application code

The migration changes `starting_xp` defaults to `1`, but the normal paths still explicitly write `0` when no value is supplied. For example `add-fighter.ts` writes:

```ts
starting_xp: effectiveFighterData.starting_xp ?? 0
```

and the admin fighter-type form currently submits `0` for a blank field. Because an explicit `0` is inserted, the database default never gets a chance to apply.

So either:

- make Starting XP required/validated for N26,
- make the application fallback match the intended safe default, or
- remove the misleading DB-default change and rely entirely on explicit catalog data.

More importantly, the actual N26 fighter type catalog needs correct `starting_xp` values before this ships. A Ganger accidentally left at 0/1 will be placed on the Rookie advancement track.

### 6. `countAdvancementsTakenFor()` should fail closed on query errors

The server guard currently ignores the `error` fields from both count queries and treats missing counts as zero:

```ts
return (characteristics.count ?? 0) + (skillAdvances.count ?? 0)
```

If either query fails, the safety check can conclude that the fighter has taken zero advancements and allow another one. Since this helper is part of server-side enforcement, I would explicitly check both errors and return/throw an error rather than granting on uncertainty.

### 7. Rebase onto `main` now that #1945 is merged

#1945 is merged and this branch is currently diverged/behind `main`, so I would rebase/merge main into this branch now. That should remove the inherited XP-modal files from this PR's effective diff and make the advancement review much easier to follow.

### 8. Remember the RPC deployment step

`supabase/README.md` says files under `supabase/functions/` are **not synced automatically**. The changes to `get_fighter_available_advancements.sql` therefore need to be applied separately to the remote DB (or moved into a migration if we want deployment to be automatic). I would call this out explicitly in the merge/deploy checklist.

## Cleanup / bloat reduction

`utils/advancementRanks.ts` is a good abstraction, but **144 lines is a lot for a threshold table plus three small calculations**. In particular `RankBand` is described as "load-bearing even though nothing reads it yet" because rank titles may be added later. I would avoid designing for that future requirement until it lands and use a flat threshold array for now. The comments can also be reduced substantially—the architecture is already established by `types/edition.ts`, `characteristicLimits.ts`, etc.

There is also some duplication around the Level Up badge and edition resolution. Both cards calculate/render essentially the same badge, and `fighter-advancement.ts` repeatedly resolves the edition and passes a separate `spendsXp` boolean derived from it. None of this is severe, but a small `LevelUpBadge`/shared calculation and one resolved advancement-state helper would make the new code easier to follow.

## Verdict

The **architecture is mostly right**: `cumulativeXp` as a capability, rank thresholds as edition-keyed data, immutable recruitment XP, and derived open advancements are all good choices.

I would fix items **1–6 before merge**, then do the smaller `advancementRanks.ts`/badge cleanup either in this PR or immediately after. The biggest concern isn't the underlying rank algorithm—it is that some current data paths do not actually supply the `starting_xp`, edition, or taken-skill data that the algorithm depends on.

joeseos:
## Code Review

@FlorianWeissDev Reviewed against the codebase, including the SQL side and the surrounding fighter-card/gang-data plumbing. Note: as you flagged, this diff also carries #1945's commit, and GitHub shows `mergeable_state: blocked` — I reviewed the full diff as-is but focused on the six commits that are actually this PR's (starting from `feat(db): record each fighter's starting XP`).

### Findings

**1. Medium — "Level Up" badge dropped into an icon-only, height-capped flex column; not browser-tested**
`components/gang/fighter-card.tsx:616-620` and `components/fighter/fighter-details-card.tsx:463-467`:
```tsx
<div className="relative flex flex-col shrink ... text-2xl max-h-[60px] flex-wrap place-content-center">
  {killed && <IoSkull .../>}
  ...
  {openAdvancements > 0 && (
    <span className="text-xs font-bold text-amber-500 whitespace-nowrap">
      {openAdvancements} Level Up{openAdvancements === 1 ? '' : 's'}
    </span>
  )}
</div>
```
This container is `flex-col` + `flex-wrap` with a hard `max-h-[60px]`, sized around `text-2xl` icon glyphs. The new badge is a `text-xs` text string, which behaves very differently under wrapping/overflow than a single icon glyph. With several status icons active at once (e.g. a captured, retired N26 fighter that's also owed advancements) plus a two-digit-safe "N Level Ups" string, this column risks overflowing its 60px cap or wrapping into a second column that overlaps the profile-image/credits box positioned immediately after it — and that risk multiplies in the compact card view modes (`2-card`/`3-card`/`4-card`), which further scale the whole card down. The PR body itself notes "not exercised in a browser." Worth a quick visual check with 2-3 simultaneous status icons + the badge, in each view mode, before this ships.

### Verified correct (no issues found)
- **`advancementsEarnedFor`/`tiersAtOrBelow`** (`utils/advancementRanks.ts`) checked against all four worked examples in the PR description (1→4, 3→4, 4→7, 13→19) — all match. The "omit Rookie's own 1-3 tier from the table" trick correctly avoids granting a phantom Advancement to a fighter recruited at 1 XP.
- **`copyFighter`**: both `sourceFighter` and `beastFighter` are fetched with `select('*', ...)`, so `starting_xp` is present on both — no risk of writing `undefined` into the new `NOT NULL` column on an experienced copy.
- **`get_fighter_available_advancements.sql`**: the `gt.gang_type_id = g.gang_type_id` join looks off at first glance, but `gang_types`'s PK actually is `gang_type_id` (confirmed via `20260630120000_add_fighter_types_gang_type_fk.sql`), and the whole edition-resolution block is a faithful copy of the existing pattern in `add_fighter_injury.sql`. Consistent with precedent.
- **XP-gate skip for non-advance skills**: `addSkillAdvancement`'s `isAdvance` param is correctly threaded through from callers — `fighter-skills-list.tsx`'s starting/free-skill path passes `is_advance: false`, which correctly short-circuits `advancementBlockedReason` before it does any earned-Advancement counting.
- **Migrations**: both new ones are additive/idempotent and correctly leave existing rows untouched (the `starting_xp` default-1 migration doesn't backfill stored `0`s, matching the stated intent that a stored 0 may be deliberate).

### Not a bug, just double-checked
- `advancements?.skills` (a separate `Record<string, Skill>` prop from the `skills: FighterSkills` prop that `countAdvancementsTaken` actually reads) looked like it could hold advance-flagged skills invisible to the new counter. Traced every producer (`gang-data.ts`, `utils/fighter-builder.ts`) — it's always empty (`{}`) in every real code path today, so there's no live divergence. Worth keeping in mind if that field ever gets populated for real later.

### Verdict: **Approve** (contingent on #1945 merging first, as noted in the PR description)

The rank/tier math, edition gating (both TS and SQL), and migration safety all check out. The one thing worth resolving before merge is a quick visual pass on the Level Up badge placement — it's a layout risk rather than a logic bug.


---

joeseos:
## Code Review — latest commit only (`d74c3e9`)

@FlorianWeissDev Reviewed the diff since the last review round (comment above, at 15:45): just `d74c3e9` — "Make the advancement modal edition-aware for N26" (`fighter-advancement-list.tsx`, `fighter-page.tsx`, `utils/dice.ts`). I re-checked the current HEAD against the two previous review rounds' findings first, then reviewed this commit's own diff in isolation.

### Previous findings — status at HEAD

None of these files are touched by `d74c3e9`, so status is unchanged from before this commit:

| # | Finding | Status |
|---|---|---|
| 1 | `getGangFightersList()` selects `starting_xp` but never puts it on the returned `result` object | **Outstanding** — `app/lib/shared/gang-data.ts` still has no `starting_xp:` key in the per-fighter result (line ~1999 has `xp:` but no `starting_xp`) |
| 2 | Skill advancements undercounted via `advancements?.skills` in `FighterDetailsCard` | **Outstanding, and worth re-scoping.** `advancements?.skills` is indeed always `{}` today (confirmed), so the specific double-count scenario originally described doesn't occur. But `countAdvancementsTaken` needs the real skills map to count `is_advance` skill rows at all, and `fighter-details-card.tsx:364` never gets it — so any fighter who has taken a skill Advancement will have their Level Up badge overcount (still shows an open advancement that's already been spent). UI-display only; the server-side guard in `fighter-advancement.ts` uses the correct DB-derived count, so this isn't exploitable, just a wrong badge. |
| 3 | Level Up badge doesn't fall back to gang edition for custom fighter types | **Outstanding.** `fighter-page.tsx:662` still passes `edition_slug={fighterData.fighter?.edition_slug ?? null}` into `FighterDetailsCard` with no gang fallback. Notably, `d74c3e9` itself adds the *correct* pattern one screen down at line 826 (`editionSlug={fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null}` for `AdvancementsList`) — so the fix is already proven out elsewhere in this same file, just not applied to line 662. Also still true: `getGangFightersList()` derives `edition_slug` from `fighter_types` only, no `custom_fighter_types` join. |
| 4 | Non-experienced `copy-fighter` copies reset `starting_xp` to 0 | **Outstanding**, both branches — `copy-fighter.ts:279-280` (fighter) and `:701-703` (exotic beast) still write `xp: ... : 0, starting_xp: ... : 0` for non-experienced copies. |
| 5 | Explicit `?? 0` bypasses the new `starting_xp` DB default of 1 | **Outstanding** — `add-fighter.ts:469-470`, both admin fighter-type forms, and `app/api/admin/fighter-types/route.ts:527,828` all still write explicit `0`. |
| 6 | `countAdvancementsTakenFor()` ignores Supabase query errors, fails open | **Outstanding** — `fighter-advancement.ts:49-67` still doesn't check `error` on either count query. |
| — | Level Up badge layout risk (flex-col/max-h-60px column), flagged as not browser-tested | Not touched by this commit; can't re-verify without a browser pass. Still applicable. |

Items 1, 4, 5, and 6 are correctness bugs in the core N26 advancement mechanic (wrong rank track on copy, wrong count baseline on the roster, and a fail-open server guard) — worth blocking merge on those regardless of how clean the new commit is.

### Review of `d74c3e9` itself

No correctness bugs found in the new commit. It's a clean, well-scoped UI split:

- `isCumulativeXp` correctly gates the XP Cost field, the N23 Ganger/Exotic Beast 2D6 table vs. the new N26 table, and correctly widens `(!isGangerOrExoticBeastRestricted || isCumulativeXp)` so N26 Ganger/Exotic Beast fighters get the general characteristic/skill picker instead of being stuck on the (now-hidden) N23-only path.
- `N26_ADVANCEMENT_TABLE` in `utils/dice.ts` fully covers 2–12 with no gaps, `resolveN26AdvancementFromUtil` matches the ranges correctly.
- `editionSlug={fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null}` at `fighter-page.tsx:826` is the right pattern (see finding 3 above re: applying it consistently).
- Confirmed `DiceRoller`'s `onRolled`/`onRoll` are mutually exclusive (one fires per roll, gated on whether `items` has a matching range) — no double-logging risk, even though the N26 table's full coverage makes the `onRoll` fallback branch effectively dead code (harmless, matches the existing Ganger pattern).
- `editableXpCost`/`editableCreditsIncrease` are driven from the selected catalog row (`matched.xp_cost`, already 0 for N26 rows per an earlier commit), not left stale when the XP Cost input is hidden.

**Minor, non-blocking:**
- `N26AdvancementEntry.characteristics` and `.skillAcquisitionTypeIds` (`utils/dice.ts`) are defined and documented but never read anywhere in `fighter-advancement-list.tsx` — the roll UI only ever displays `row.name`/`row.credits`. Fine if intended for a later filtering feature, but worth a one-line comment saying so, otherwise it reads as dead plumbing (low severity, same spirit as the "reduce new plumbing" note from the first review round).
- `logN26RollMutation`/`logN26RollWithCooldown` (lines ~936-121) duplicate `logGangerSubRollMutation`'s shape almost verbatim. Not new debt introduced by this PR's pattern, but another instance of it (low severity).

Not because of anything in `d74c3e9` — that commit is solid on its own. But findings 1, 3, 4, 5, and 6 from the earlier review round are all still live on HEAD and are correctness bugs in the feature this PR ships (wrong advancement counts/tracks for the roster, custom fighter types, and copied fighters, plus a fail-open server guard). I'd want at least 1, 4, 5, and 6 fixed before merge.

---

FlorianWeissDev:
Thanks — this was a genuinely useful pass. All eight items are addressed; the branch is rebased onto `main` so the diff is now just the advancement work. Details below.

## Fixed

**1. `starting_xp` never returned from `getGangFightersList`.** Confirmed and fixed — it was selected and declared on `GangFighter` but missing from the result object, so `FighterCard` saw `undefined` and measured from 0. This was the worst of the bunch: a Ganger recruited at 13 XP showed several Level Ups it hadn't earned.

**2. Skill advancements uncounted on the fighter page.** Fixed. `FighterDetailsCard` now takes an explicit `skills` prop carrying the real `fighter.skills` rows, and `advancements?.skills` is no longer used for counting. As you say, the server doesn't populate that parallel shape on first load, so every existing skill advancement went uncounted and inflated the open count.

**3. Gang edition for Level Up.** The fighter-page half resolved itself in the rebase — since #1945 merged, `editionSlug` is already `gang.edition_slug ?? fighter.edition_slug` and the card uses it. For the roster I fixed it at source rather than threading a prop through `my-fighters` → `sortable-fighter` → `FighterCard`: `getGangFightersList` now resolves the gang's edition once and uses it for every fighter, instead of deriving it per fighter from `fighter_types`. That covers custom fighter types, and it also corrects the `Sv` column, which had the same blind spot.

**4. Non-experienced copies on the wrong rank track.** Fixed exactly as you suggested, for fighters and copied exotic beasts:

```ts
starting_xp: sourceFighter.starting_xp,
xp: params.copy_as_experienced ? sourceFighter.xp : sourceFighter.starting_xp,
```

**5. `DEFAULT 1` bypassed by application code.** Kept the default and made it reachable: the admin form and the API route now omit `starting_xp` when an edition that has the concept leaves the field blank.

One wrinkle worth flagging, because it makes the naive fix wrong. Editions **without** Starting XP hide the field entirely, so simply omitting on blank would let them inherit `1` and recruit fighters with XP they should never have. Those paths now pin an explicit `0`. Verified all three cases against the local DB:

- omitted → `1`
- explicit `0` → `0`
- explicit `13` → `13`

Agreed on the catalog data being the real risk — that's flagged in the PR description as the first thing to check on staging.

**6. `countAdvancementsTakenFor` failing open.** Fixed — both `error` fields are checked and the helper throws. The actions' existing handlers turn that into `{ success: false }`, so uncertainty now refuses rather than grants.

**7. Rebased onto `main`.** Done; the inherited XP-modal commit was dropped automatically as already-applied.

**8. RPC deployment.** Good catch, calling it out explicitly: **`get_fighter_available_advancements.sql` must be applied to the remote DB separately** — it is not covered by the migrations in this PR. Without it, N26 fighters are offered N23 advancements and, worse, N23 fighters see the newly seeded N26 rows too, since the edition filter lives in the function rather than in the data.

## Cleanup

`advancementRanks.ts` is flattened to a threshold array — you're right that the band grouping was designing for a requirement that hasn't landed. The no-refactor property I wanted from it is already provided by exporting only functions, so the grouping bought nothing. Comments trimmed to what the code doesn't say. 144 → 96 lines, identical results across every worked example.

I've left the shared `LevelUpBadge` and the `spendsXp`/edition-resolution tidy-up for a follow-up rather than growing this PR further — happy to do it immediately after if you'd prefer it here.

## Still not verified in a browser

Everything above is SQL-level or code-level. Migrations were applied to a freshly reset local DB and re-applied for idempotency, and the RPC was exercised against throwaway N26/N23 fighters in a rolled-back transaction — the N23 fighter sees only N23 rows, with flat cost and the Ganger-to-Specialist row unchanged. But no part of this has been clicked through, since local sign-in is unavailable.

Unrelated, but it blocked local testing and someone should pick it up: **`supabase/seed.sql` is broken on `main`** — it still inserts `skills.xp_cost` / `credit_cost`, dropped by `20260805193900`, so `supabase db reset` fails and rolls the whole seed back, leaving a current schema with zero reference data. I fixed it locally to run the tests above and deliberately kept it out of this branch.

joeseos:
## Code Review — latest changes only (`e90ddde` + `f961641`)

@FlorianWeissDev Reviewed the diff since the last review round (`98efef2`): commits `e90ddde` ("Fix the data paths the advancement calculation depends on") and `f961641` ("Flatten the rank table and make the starting_xp default reachable"). Diffed `98efef2..f961641` directly rather than the whole PR.

| # | Finding | Status |
|---|---|---|
| 1 | `getGangFightersList()` selected `starting_xp` but never returned it | **Resolved** — `app/lib/shared/gang-data.ts:2016` now sets `starting_xp: fighter.starting_xp` on the returned object |
| 2 | `FighterDetailsCard` counted skill advancements from `advancements?.skills` (always empty) instead of the real `skills` map | **Resolved** — `FighterDetailsCard` now takes an explicit `skills` prop; `fighter-page.tsx:694` passes `fighterData.fighter?.skills`, and `countAdvancementsTaken(effects, skills)` uses it |
| 3 | Level Up badge doesn't fall back to gang edition for custom fighter types | **Resolved** — `getGangFightersList()` now resolves `gangEditionSlug` once from the gang (covering `custom_gang_types`) and uses it for every fighter's `edition_slug`. Worth noting: the fighter-page.tsx call site this finding named (`edition_slug={fighterData.fighter?.edition_slug ?? null}`) was already using the correct `editionSlug` gang-fallback variable at `98efef2`, so that half of the finding looks like it was stale by the time it was raised — not something this diff needed to touch. |
| 4 | Non-experienced `copy-fighter` copies reset `starting_xp` to 0 | **Resolved**, both the fighter and exotic-beast branches (`copy-fighter.ts:279-282`, `:701-704`) — `starting_xp` now survives the copy and `xp` resets to it instead of 0 |
| 5 | Explicit `?? 0` bypasses the new `starting_xp` DB default of 1 | **Mostly resolved, one gap** — see new finding #1 below. The two admin forms and the fighter-types `POST` route now correctly omit the key on blank input, but the `PATCH` route's identical-looking fix doesn't behave the same way (`.update()` vs `.insert()`). |
| 6 | `countAdvancementsTakenFor()` ignores query errors, fails open | **Resolved** — now throws on either query's `error`, and both call sites (`addCharacteristicAdvancement`, `addSkillAdvancement`) already wrap the call in `try/catch` and turn it into `{ success: false, error }`, so this doesn't just move the fail-open bug into an unhandled rejection |
| — | Level Up badge layout risk (flex-col/max-h-60px, not browser-tested) | Not touched by either new commit; still unverified, carry forward |
| — | `advancementRanks.ts` bloat / speculative `RankBand` grouping | **Resolved** — flattened to a plain threshold array, 144 → 96 lines, same math (verified `tiersAtOrBelow`/`advancementsEarnedFor` are logically unchanged, just restructured) |

### New findings

**1. Medium — `PATCH /api/admin/fighter-types` omitting `starting_xp` doesn't actually apply the DB default**
`app/api/admin/fighter-types/route.ts:524-528`:
```ts
// Omitted when the client sends nothing, so the column default applies.
// Writing an explicit 0 here would make that default unreachable.
...(data.starting_xp != null ? { starting_xp: data.starting_xp } : {}),
```
This is inside a `supabase.from('fighter_types').update(...)` call. Omitting a key from an `UPDATE`'s payload doesn't cause Postgres to apply the column `DEFAULT` — it simply leaves that column untouched, i.e. the row keeps whatever `starting_xp` it already had. The comment ("so the column default applies") is only true for the `POST`/`insert` path a few lines below (`:827`), which is correct.

Concretely: `admin-edit-fighter-type.tsx` prefills the Starting XP field from the existing value (`admin-edit-fighter-type.tsx:606`), so this only surfaces if an admin deliberately blanks the field expecting it to reset. Today that's a silent no-op on that column rather than resetting to 1 (or to whatever the intended "cleared" value is) — different from both the pre-PR behavior (explicit 0) and the documented intent. Not a crash and low-frequency, but worth either fixing (explicitly write the default value, e.g. `1`, on the edit path) or at least correcting the comment so it doesn't claim a guarantee that isn't true for `UPDATE`.

**2. Low/nit — extra unbatched query added to `getGangFightersList()`**
`app/lib/shared/gang-data.ts:964-975` adds a new `gangs` query to resolve `gangEditionSlug`, ahead of the function's existing batch of fighter/equipment/skill/effect queries. The function's docstring says "Target: ~8 queries total regardless of fighter count" — this is a 9th, always-run, single-row query, so functionally fine (it's not per-fighter and the whole result is `unstable_cache`d), just slightly stale relative to the comment. Not blocking.

### Not re-flagged (already covered, still open, not touched by this diff)
- Level Up badge layout under `flex-col`/`max-h-[60px]` with multiple status icons — still needs a browser check, unchanged by these two commits.
- Process items from the first review round (rebasing onto `main` now that #1945 is merged, and the reminder that `get_fighter_available_advancements.sql` needs a manual RPC deploy) aren't code and aren't affected by this diff either way.

### Verdict: **Changes requested** (small)

The correctness bugs from the last two review rounds (1, 2, 4, 6, and the bulk of 5) are genuinely fixed, and the `advancementRanks.ts` cleanup lands cleanly with no behavioral drift. The one thing I'd want addressed before merge is the `PATCH` route's `starting_xp` omission (#1 above) — either make it actually reset to the intended default or drop the misleading comment, since as written it silently does something different from what it says and from what the `POST` path a few lines below actually achieves. Everything else here is fine to merge as-is once that's settled, alongside the still-outstanding badge layout check and the external #1945 merge dependency.

---

PR #1948, opened 2026-08-07
Make admin skills API and Default Skills edition-aware

joeseos:
### Motivation
- Ensure skills can be scoped to an edition so admin UI can show only edition-appropriate Default Skills and prevent submitting mismatched defaults after an edition change.
- Preserve existing backward-compatible behaviour for callers that do not provide an edition filter.

### Description
- Add optional `edition_id` query parameter to `GET /api/admin/skills` and join `skill_types` so skills can be constrained to the supplied edition using an inner relation filter while keeping unfiltered responses unchanged.
- Keep the existing response shapes: unfiltered requests still return an array, and `skill_type_id`-filtered requests continue to return `{ skills, effect_categories }` (including empty collections on edition/skill-set mismatches).
- Update `components/admin/admin-create-fighter-type.tsx` to include `editionId` in the `['admin-skills', selectedSkillType, editionId]` query key and send `edition_id` via `URLSearchParams`, and drop selected default-skill IDs that do not belong to the newly selected edition when `handleEditionChange` runs.
- Update `components/admin/admin-edit-fighter-type.tsx` to include `editionId` in the fetch, add `editionId` to the effect dependencies so requests are re-run when the edition changes, retain merged skill metadata so already-saved skills can render while the fighter record loads, and remove selections that cannot be proven to belong to the new edition.

### Testing
- `npx tsc --noEmit` completed successfully with no type errors.
- `npx eslint` checks for the modified files completed successfully with no lint errors reported.
- `git diff --check` returned no whitespace/patch problems for the modified files.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a75f9e886f48332a3fa1ae84fd1fae7)

### DIFF
--- a/app/api/admin/skills/route.ts
+++ b/app/api/admin/skills/route.ts
@@ -13,6 +13,7 @@ export async function GET(request: Request) {
   const supabase = await createClient();
   const { searchParams } = new URL(request.url);
   const skillTypeId = searchParams.get('skill_type_id');
+  const editionId = searchParams.get('edition_id');
   const effectTypeId = searchParams.get('effect_type_id');
   const modifierId = searchParams.get('modifier_id');
 
@@ -66,11 +67,20 @@ export async function GET(request: Request) {
     // Handle skill queries
     let query = supabase
       .from('skills')
-      .select('id, name, skill_type_id, gang_origin_id')
+      .select(editionId
+        ? 'id, name, skill_type_id, gang_origin_id, skill_types!inner(edition_id)'
+        : 'id, name, skill_type_id, gang_origin_id, skill_types(edition_id)')
       .order('name');
 
     if (skillTypeId) {
       query = query.eq('skill_type_id', skillTypeId);
+    }
+
+    // A skill belongs to an edition through its skill set. The inner relation is
+    // important here: filtering an ordinary embedded relation would leave the
+    // parent skill rows in the result with an empty relation.
+    if (editionId) {
+      query = query.eq('skill_types.edition_id', editionId);
     }
 
     const { data, error } = await query;
@@ -79,22 +89,26 @@ export async function GET(request: Request) {
 
     // If fetching a specific skill, include its effects
     let transformedData;
-    if (skillTypeId && data.length > 0) {
+    if (skillTypeId) {
       // Fetch effects for all skills in the result
       const skillIds = data.map((skill: Skill) => skill.id);
 
-      const { data: effectTypes } = await supabase
-        .from('fighter_effect_types')
-        .select(`
-          id,
-          effect_name,
-          fighter_effect_category_id,
-          type_specific_data,
-          sort_order,
-          fighter_effect_categories(id, category_name)
-        `)
-        .in('type_specific_data->>skill_id', skillIds)
-        .order('sort_order', { ascending: true, nullsFirst: false });
+      let effectTypes: any[] = [];
+      if (skillIds.length > 0) {
+        const { data: effectTypeData } = await supabase
+          .from('fighter_effect_types')
+          .select(`
+            id,
+            effect_name,
+            fighter_effect_category_id,
+            type_specific_data,
+            sort_order,
+            fighter_effect_categories(id, category_name)
+          `)
+          .in('type_specific_data->>skill_id', skillIds)
+          .order('sort_order', { ascending: true, nullsFirst: false });
+        effectTypes = effectTypeData || [];
+      }
 
       // Fetch modifiers for the effect types
       let modifiers: any[] = [];
@@ -128,7 +142,7 @@ export async function GET(request: Request) {
 
       // Extract unique categories from the effects (they're already nested in each effect)
       const uniqueCategories = new Map();
-      effectTypes?.forEach((et: any) => {
+      effectTypes.forEach((et: any) => {
         if (et.fighter_effect_categories) {
           uniqueCategories.set(
             et.fighter_effect_categories.id,
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -185,6 +185,15 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
   const handleEditionChange = (newEditionId: string) => {
     setEditionId(newEditionId);
 
+    // Skills are edition-owned through their skill set. Unknown skill metadata
+    // is discarded as well so a stale selection can never be submitted after
+    // the user explicitly changes edition.
+    setSelectedSkills(prev => prev.filter(skillId => {
+      const skill = skills.find(candidate => candidate.id === skillId);
+      const skillType = skillTypes.find(type => type.id === skill?.skill_type_id);
+      return !!skillType && (!newEditionId || skillType.edition_id === newEditionId);
+    }));
+
     if (editions.find(edition => edition.id === newEditionId)?.slug !== 'n26') {
       setIsVehicle(false);
     }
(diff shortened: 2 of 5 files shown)


joeseos:
Reviewed against the codebase, checking out the branch to trace how `skills`/`selectedSkills` state actually flows through both admin forms since that's where the risk is (the API change itself is small and self-contained).

### Finding: `AdminCreateFighterTypeModal`'s edition-change filter silently drops valid selections from other skill types

`components/admin/admin-create-fighter-type.tsx`'s `handleEditionChange`:
```ts
setSelectedSkills(prev => prev.filter(skillId => {
  const skill = skills.find(candidate => candidate.id === skillId);
  const skillType = skillTypes.find(type => type.id === skill?.skill_type_id);
  return !!skillType && (!newEditionId || skillType.edition_id === newEditionId);
}));
```
This looks up each selected skill's metadata in the local `skills` state to decide whether it survives the edition change. But in this component, `skills` comes from a single `useQuery(['admin-skills', selectedSkillType, editionId], ...)` scoped to whichever skill type is *currently browsed* in the dropdown (`enabled: !!selectedSkillType`) — it's never merged across skill types, it's just whatever the last fetch for the current `selectedSkillType` returned.

Default skills are normally picked from *multiple* skill types (browse Ganger skills, add one, switch the dropdown to Leader skills, add another, etc.) — that's the whole point of the picker. The moment the admin changes edition, any selected skill that doesn't belong to the currently-displayed skill type won't be found in `skills` (`skill` comes back `undefined`), so `skillType` lookup also fails, and the `!!skillType` check drops it — even if it would have been perfectly valid for the new edition. Since this component's `selectedSkills` chips *also* silently fail to render for skills outside the current skill type (`skills.find(s => s.id === skillId)` → `if (!skill) return null` a few lines down, at ~line 866), the admin has no visual cue that this happened — the skill just quietly disappears from `default_skills` on submit.

**`AdminEditFighterTypeModal` does not have this problem** — worth noting as the reference for the fix, since it already solves it correctly: its `fetchSkills` effect *merges* new results into `skills` (`new Map([...previous, ...skillsArray])`) instead of replacing it, and a separate `fetchSelectedSkillDetails` effect proactively fetches full metadata for every currently-selected skill (via the unfiltered `/api/admin/skills`) regardless of which type is being browsed. That's exactly why it also needed the new `availableSkills` filter (`skill.skill_type_id === selectedSkillType`) added in this PR — to stop the now-accumulated `skills` list from leaking other types' skills into the "add" dropdown. Good, deliberate fix there.

Suggested fix for the create form: either mirror the edit form's merge-and-backfill pattern, or simpler — since `handleEditionChange` only needs `skill_type_id`/`edition_id` per selected skill (not full skill records), fetch/cache that mapping independently of the browse-scoped `skills` query so it's always available regardless of what's currently displayed.

### What's solid

- **API route** (`app/api/admin/skills/route.ts`): the `!inner` vs plain embed branch on the `select()` is exactly right — filtering `skill_types.edition_id` without `!inner` would leave parent skill rows in the result with an empty relation instead of excluding them, and the code correctly picks the strict join only when `editionId` is actually supplied. The response is rebuilt field-by-field in both branches, so the `skill_types` embed used purely for filtering never leaks into the JSON payload.
- **`data.length > 0` → unconditional `if (skillTypeId)`**: correctly fixes a real pre-existing shape inconsistency — previously a `skill_type_id`-filtered request with zero matching skills would have skipped the `{ skills, effect_categories }` wrapping entirely. That's now consistent, which matters a lot here since edition-filtered zero-result responses are an expected, common case going forward, not just an edge case.
- Guarding the `fighter_effect_types` lookup with `if (skillIds.length > 0)` before calling `.in('type_specific_data->>skill_id', skillIds)` is good defensive follow-through, since that path is now reachable with an empty array where it previously wasn't.
- `checkAdmin` gating is untouched — no change to the auth boundary.

### Checklist
- Correctness: one real bug (above), otherwise sound
- Edge cases / null handling: handled well on the API side; the gap is specifically the create-form's skill cache scope
- Security: n/a, admin-only route unchanged
- Performance: negligible — one extra join column, no new round trips
- Readability/maintainability: the inline comments explaining the `!inner` join rationale and the edit-form's "preserve metadata" merge are genuinely helpful for future readers
- Conventions: consistent with how `editionId`-gated query keys are already used elsewhere in these two files
- DRY/YAGNI: the `handleEditionChange` filter logic is duplicated near-verbatim between the two components; if you're touching this again for the fix above, a shared helper (`filterSkillsForEdition(selectedSkills, skills, skillTypes, editionId)`) would remove the duplication and the two implementations drifting apart further

Just for the create-form filtering bug — the API and edit-form changes are correct and ready as-is.

---

joeseos:
## Follow-up review — latest changes only (`8f92812` + `76fcf85`)

Reviewed the diff since the last review round (`335ba32`): commits `8f92812` ("Make skill admin forms edition-aware") and `76fcf85` ("Retain selected default skill metadata"), diffed directly rather than the whole PR.

### Previous finding — status

| Finding | Status |
|---|---|
| `AdminCreateFighterTypeModal`'s `handleEditionChange` filter silently dropped valid default-skill selections from other skill types, because it looked up each selected skill in the browse-scoped `skills` query instead of a cache that spans every skill type ever selected | **Resolved.** `components/admin/admin-create-fighter-type.tsx` now keeps a separate `loadedSkills` state that accumulates (merges, keyed by id) every skill the admin has ever added, populated in the "add skill" `onChange` handler alongside `setSelectedSkills`. `handleEditionChange`'s filter and the selected-skill chips both now read from `loadedSkills` instead of the browse-scoped `skills`, so a selection from a skill type that isn't currently displayed in the dropdown survives an edition change (or continues to render as a chip) correctly. Traced the only code path that ever pushes into `selectedSkills` (this is a create-only form, so there's no pre-existing-selection hydration path to worry about) — it always populates `loadedSkills` in the same handler, so the cache can't miss an entry. Also cleans up the literal dead-code duplicate lookup that was there before (`skills.find(...) || skills.find(...)`, same expression twice). |

### Review of the new commits

**`admin-edit-skill.tsx`** — `!selectedType || selectedType.edition_id !== newEditionId` (previously `selectedType && ...`) is a real correctness fix in its own right: previously, if the currently-selected skill type couldn't be found in `skillTypeList` (stale/missing id), the `&&` short-circuited and the stale `skillType` was silently kept instead of cleared on an edition change. Now it fails closed and clears. `searchSkillType` correctly threads the skill type's own `edition_id` (falling back to the top-level filter) into the `/api/admin/skills` request, matching the API's `!inner`-join semantics from the first commit.

**`admin-create-skill.tsx`** — moves `EditionSelect` to the top of the form and uses it to filter the Skill Set dropdown (`filteredSkillTypeList`) rather than only showing it after a brand-new skill-set name is typed. Checked the two submit paths: `handleSubmit` (add a skill to an existing skill type) never sends `editionId` to the API — it's UI-filter-only there, so there's no risk of a mismatched `edition_id` being written for an existing skill. `handleSubmitType` (create a brand-new skill type) sends `edition_id: editionId || null`, unchanged from before. No behavioral conflict between the two paths.

**Edge case checked and ruled out:** `skill_types.edition_id` is nullable in the schema (confirmed via `supabase/schema/schema.public.sql` and the backfill migration `20260630095837_add_editions_and_root_edition_ids.sql`, which explicitly leaves it nullable for new rows), so I traced whether a null-edition skill type could hit the new `!inner`-join edition filter and wrongly return zero skills. It can't in practice: `filteredSkillTypeList` in both edit and create-skill forms is filtered by the same `editionId` that later gets sent to the API, so a null-edition skill type is only selectable while `editionId` is itself empty — and an empty `editionId` never gets forwarded as an `edition_id` param (`if (selectedEditionId) params.set(...)`). No reproducible bug here.

### New findings

None. Didn't find any new correctness bugs, null-handling gaps, or regressions in this diff.

### Not re-flagged (already covered, unaffected by this diff)
- The DRY suggestion from the first review round (shared `filterSkillsForEdition` helper to de-duplicate the near-identical edition-change filtering logic between the two fighter-type forms) is still just a nice-to-have, not touched by these commits, and not blocking.

### Verdict: **Approve**

The previously-requested fix is correctly implemented, the two admin skill-form UX improvements bundled into this round are sound, and I couldn't find a new bug despite specifically chasing the nullable-`edition_id` edge case this kind of change tends to trip on.

---

PR #1949, opened 2026-08-07
Make the edition registry enforce a decision per capability

joeseos:
Audit of the codebase against the edition-gating design, plus the fixes it turned up. One PR rather than three because the transpose and the new capabilities touch the same registry.

The design held up: 42 of 43 predicate call sites gate exactly what their name says, and the three sibling per-edition data registries (`characteristicLimits`, `dice`, `lastingInjuryRank`) follow it correctly. The DB has exactly `n23` and `n26`, matching the registry, so nothing here was live breakage — these are all "breaks when a third edition ships" or "breaks when someone copies the nearby pattern".

**No behaviour change for n23 or n26.**

## 1. The registry's guidance undercut its guarantee

The old layout gave each edition an object, and the doc comment recommended cloning one:

```ts
n29: { ...N26_CAPABILITIES, tradePoints: false },
```

That compiles clean and silently gives the new edition N26's answer for every capability added *afterwards* — the exact silent inheritance the registry exists to prevent.

Now keyed by capability, so there is no per-edition object to clone:

```ts
const EDITION_CAPABILITIES = {
  saveCharacteristic:   { n23: false, n26: true  },
  masterCraftedWeapons: { n23: true,  n26: false },
  …
} as const satisfies Record<string, Record<EditionSlug, boolean>>;
```

Adding a slug now produces **one error per capability** (eight today) rather than one error a paste can satisfy. The five existing predicates keep their names and signatures.

## 2. Three capabilities that had escaped the registry

Each would have been decided by accident for a third edition, with no compile error:

| Capability | Was |
|---|---|
| `vehicles` | Raw `'n26'` literal, 4 sites across both admin fighter-type modals — two lines from a correct `allowsMultipleSubtypes` call |
| `equipmentSuperCategories` | A behaviour predicate living in `utils/getEquipmentCategoryRank.ts` as `=== EDITION_N26` |
| `masterCraftedWeapons` | Gated on `!hasTradePoints` — an unrelated capability used as a proxy, with the coincidence noted in the comment |

Master-crafted was also **client-only**: `buyEquipmentForFighter` applied the 1.25× rating multiplier and wrote `is_master_crafted` with no edition check, so the hidden checkbox was the only thing stopping it. The gang's edition is now resolved before the multiplier and gates both.

## 3. One guard used display semantics for a rejection

`sameEditionSlug` coerces a missing slug to `n23` — right for filtering lists (legacy rows predate `edition_id`), wrong for guards. The campaign-join guard was using it, so a gang whose edition failed to resolve was **rejected from an n26 campaign** while being **allowed into an n26 battle** by the sibling guard.

Renamed to `sameEditionForDisplay`, campaign guard switched to `editionsConflict`, and the two now document which to reach for.

## 4. Seven copies of the gang-edition flatten

A gang's edition comes from either `gang_types` or `custom_gang_types`, which `withEditionSlug` doesn't cover — so that flatten was hand-rolled in seven places across three embed aliases (`custom_gang_types`, `custom_gang_type`, `custom_gang_type_edition`). Each was locally correct; the risk was the next copy. New `gangEditionSlug`/`gangEditionJoin` accept every alias and the array-widened shape.

## 5. Smaller

- Custom fighter types no longer hardcode `edition_slug: null`, so **Sv renders for them again**. The shared-via-campaign fetch was also missing the edition embed.
- `getEquipmentCategoryRank` is keyed by `EditionSlug` like its two sibling data registries, so a new edition is a compile error rather than a silent fall back to alphabetical.
- Unknown-slug warnings deduped per slug — previously one per predicate call per render.

## Verification

```
predicate                    n23    n26    null   n29
hasSaveCharacteristic        false  true   false  false
allowsMultipleSubtypes       false  true   false  false
hasTradePoints               false  true   false  false
hasStartingXp                false  true   false  false
hasLethalityStatline         false  true   false  false
hasVehicles                  false  true   false  false
hasEquipmentSuperCategories  false  true   false  false
hasMasterCraftedWeapons      true   false  false  false
```

- Adding `'n29'` to `EditionSlug` yields exactly 8 `TS2741` errors, one per capability.
- `gangEditionSlug` resolves all four embed aliases, the array-widened shape, and prefers official over custom.
- One warning emitted across eight unknown-slug lookups (dedupe works).
- `npx tsc --noEmit --ignoreDeprecations 5.0` clean. Note plain `npx tsc --noEmit` **checks nothing** on this repo — it aborts on two pre-existing tsconfig deprecation errors before reading a file.
- `npm run lint` byte-identical to `main` except one pre-existing warning shifting from line 640 to 637.

## Not changed

Admin `edition_id` comparisons (option filtering, not gating), and the documented Customise-is-N23 decision in `components/home-tabs.tsx` and the `custom-*` writes.

---
_Generated by [Claude Code](https://claude.ai/code/session_0154t1yVgwGyTtoGJtrZD1KJ)_


joeseos:
## Review

Reviewed against `main`, cross-checked every claim in the PR description against the actual diff, and checked the implementation against the design doc (linked artifact: "Multi-edition support: how editions gate behaviour").

### Design-doc adherence — verified, not just taken on faith
- **Capability-first registry** (`EDITION_CAPABILITIES` keyed by capability, `as const satisfies Record<string, Record<EditionSlug, boolean>>`) matches the doc's rejection of the edition-first "spread and override" shape exactly, and keys are plain string literals (`n23:`, not `[EDITION_N23]:`) — the one gotcha the doc calls out as previously catching out this exact pattern.
- **`gangEditionSlug`/`gangEditionJoin` alias coverage**: I grepped every pre-PR hand-rolled gang-edition flatten in the codebase and diffed it against the new helper's alias list (`gang_types`, `custom_gang_types`, `custom_gang_type`, `custom_gang_type_edition`). All 4 aliases the codebase actually uses are covered, and the double `firstOf()` composition correctly handles the array-widened embed shape on both the outer (`gang[alias]`) and inner (`.editions`) level.
- **`sameEditionSlug` → `sameEditionForDisplay` rename**: grepped all 9 pre-PR call sites (8 components + `assert-gang-campaign-edition.ts`); all 9 are accounted for in the diff (8 renamed straight across, 1 correctly upgraded to `editionsConflict` instead — see note below). No stragglers.
- **Master-crafted server-side gate** (`app/actions/equipment.ts`): confirmed the `gang` query at line 235-239 already selects `gang_types(editions:edition_id(slug))`/`custom_gang_types(...)`, so `gangEditionSlug(gang)` resolves correctly there — this closes a real gap where `params.master_crafted` from the client was the only thing stopping an N26 purchase from getting the 1.25× multiplier and `is_master_crafted: true`. Good catch, correctly fixed.
- **`masterCraftedWeapons` semantics preserved**: old `!hasTradePoints` proxy and new explicit `{ n23: true, n26: false }` row produce identical answers for both real editions — the "no behaviour change" claim holds for this one.

### Findings

**1. [Medium] The "seven copies" sweep missed an eighth — `app/api/users/[id]/route.ts:317-322`**
This file still hand-rolls the exact pattern the PR consolidates elsewhere:
```ts
edition_slug: gang_types?.editions?.slug ?? custom_gang_type?.editions?.slug ?? null,
```
using the `gang_types`/`custom_gang_type` aliases (line 40-41's select), which `gangEditionSlug`/`gangEditionJoin` already cover. It produces the same answer today, so it's not a live bug, but it directly undercuts the PR's own stated goal ("Seven copies of the gang-edition flatten... risk was the next copy") — this *is* the next copy, just pre-existing rather than new. Since the map callback already has the raw `gang_types`/`custom_gang_type` on the row before destructuring, this is a one-line swap: `edition_slug: gangEditionSlug({ gang_types, custom_gang_type })` (or restructure to call it before destructuring, as `campaign-members-table.tsx` now does).

**2. [Low, flag for conscious sign-off] `assert-gang-campaign-edition.ts` guard is now more permissive for unresolved editions**
Switching from `sameEditionSlug` (null → coerced to `n23`) to `editionsConflict` (null → conflicts with nothing) means a gang whose edition fails to resolve, which previously could only join `n23` campaigns, can now join *any* campaign. This is exactly what item 3 in the PR body describes and is the right fix for the inconsistency with the battle-session guard — flagging only because it's a guard getting strictly more permissive, which is the kind of change worth a deliberate "yes, that's intended" rather than being inferred from a diff. It reads as intended here.

### Nothing else stood out
No remaining raw `'n26'`/`!== 'n23'` behaviour gates outside the registry (the two admin fighter-type files' `hasVehicles` swap covers all 4 sites the PR claims), no leftover `!hasTradePoints`-as-proxy usage, and the `warnedSlugs` per-slug (not per-capability) dedup matches the "one warning across eight lookups" verification claim in the code as written.

### Verdict
**Changes requested** — just #1 (a real, easy-to-fix leftover instance of the exact pattern this PR sets out to eliminate); #2 is a confirm-and-carry-on rather than a blocker. Solid refactor otherwise and matches the design doc closely.

---

PR #1951, opened 2026-08-08
Added gating of skills per edition

Djidiouf:


joeseos:
## Review

@Djidiouf Reviewed this against `main` and the surrounding edition-gating conventions (`types/edition.ts`, `app/api/fighter-subtypes/route.ts`, `app/api/fighters/injuries/route.ts`).

### Findings

**1. [High] `skill_types`/`custom_skill_types` will resolve to an empty catalog for N26 gangs once they exist — `app/api/skill-types/route.ts:37-64`**
When `fighterId` is passed and the caller doesn't supply an explicit `edition_slug`/`edition_id`, the route now derives the edition from the fighter's gang and filters `skill_types`/`custom_skill_types` on `edition_id`. No N26 `skill_types` rows exist yet — confirmed by the comment in `supabase/migrations/20260806160000_seed_n26_lasting_injuries.sql:45` ("there are no N26 skill_types or skills yet") and by the root-edition backfill migration (`20260630095837_add_editions_and_root_edition_ids.sql`), which backfilled every existing `skill_types`/`custom_skill_types` row onto `n23` only. `fighter-skills-list.tsx`, `skill-access-modal.tsx`, and `fighter-advancement-list.tsx` all call this endpoint with `fighterId` and no edition params, so once an N26 gang exists, the Add Skill modal, Skill Access modal, and skill-set advancement picker will all silently render **zero** skill sets for it (versus returning the full unfiltered list today, however mislabeled).
No N26 `gang_types` are seeded yet either, so this isn't a live regression today, but it will land as a silent break the moment N26 gang types ship — with no warning/log, unlike the `capabilitiesFor` pattern elsewhere in the codebase that warns on unrecognised slugs. Worth either sequencing this behind an N26 skill catalog migration, or falling back to unfiltered/N23 rows when the edition-scoped query comes back empty, with a dev-only warning.

**2. [Low] Dead `editionSlug` prop — `components/customise/custom-fighters.tsx:41-47`**
`SkillSetOptions` gained an `editionSlug` prop (defaulting to `EDITION_N23` when absent), but none of its 4 call sites in this file pass it — it's always `undefined`, so the prop is currently a no-op. If Customise is intentionally N23-only (as documented elsewhere, e.g. #1949's "Customise-is-N23" decision), the prop can be dropped instead of threaded in half-finished; if it's meant to eventually reflect the custom fighter's edition, it needs wiring at the call sites.

**3. [Low] Duplicate `getSkillSetRank` call — `components/fighter/edit-fighter/skill-access-modal.tsx:40,111`**
`skillSetRank` is computed once at the top of `SkillAccessModal` and a second time inside the `computedSkillAccess` `useMemo`. The outer one looks unused by the memo (which recomputes its own copy) — worth confirming it's actually needed at the outer scope, and removing one of the two if not.

**4. [Nit] IIFE instead of extracted variable — `components/admin/admin-create-fighter-type.tsx:808-833`**
The skill-set-grouping `.sort()/.reduce()` chain got wrapped in an inline `(() => { const skillSetRank = ...; return ...; })()` inside JSX to bring `editionSlug` into scope. `admin-edit-fighter-type.tsx`'s equivalent code hoists the same computation into a `useMemo` instead, which reads more clearly and would be consistent with the sibling file.

**5. [Nit] Inconsistent embed FK hint — `app/api/skill-types/route.ts:41-46`**
The gang→edition lookup uses `gang_types (...)` without the explicit `!gang_type_id` FK hint, while the adjacent `custom_gang_type_edition:custom_gang_types!custom_gang_type_id (...)` and every other instance of this join pattern in the codebase (`assert-gang-campaign-edition.ts`, `update-gang.ts`, `gang-data.ts`, `battle-sessions.ts`, `users/[id]/route.ts`) do specify it. Probably harmless since there's only one FK path, but worth matching the established convention for consistency.

### What looks solid
- The `skillSetRank` → `skillSetRankN23`/`skillSetRankN26` split and `getSkillSetRank(editionSlug)` fallback-to-N23 follow the same pattern as the other per-edition data registries (`characteristicLimits`, `advancementRanks`).
- Reuses the canonical `editionSlugFromJoin`/`getEditionIdBySlug` helpers rather than re-deriving edition resolution logic from scratch.
- `useMemo` dependency arrays were correctly updated everywhere `editionSlug` was threaded in (`admin-edit-fighter-type.tsx`, `fighter-advancement-list.tsx`, `fighter-skills-list.tsx`).
- `fighter.gang_id` lookup switched from `.single()` to `.maybeSingle()`, avoiding a thrown error on a missing fighter.

### Verdict
**Changes requested** — primarily #1, which is a real (if not-yet-live) functional regression once N26 gang types ship; the rest are minor and non-blocking.

---

joeseos:
## Follow-up review

@Djidiouf The only change since the last review is the merge commit (`4bc3f77`) pulling in `main` (#1949, #1950, #1944, etc.) and resolving one real conflict. Reviewed that commit in isolation — diffed `003022b` (reviewed previously) against `4bc3f77` across every file this PR actually touches.

### Merge conflict resolution — looks correct
`components/admin/admin-create-fighter-type.tsx` and `admin-edit-fighter-type.tsx`: the conflict was this branch's `getSkillSetRank` import colliding with #1949's `hasVehicles` import on the same line. Resolved by keeping both and switching the two hardcoded `... === 'n26'` vehicle checks over to the new `hasVehicles(editionSlug)` capability helper. That's the right call — consistent with #1949's "one decision per capability" pattern and with how this PR itself already uses `getSkillSetRank`/`getEditionIdBySlug` instead of ad-hoc slug checks. No regressions introduced.

No other file in this PR's diff changed between those two commits (confirmed via `git diff --stat` over the full file list) — `app/api/skill-types/route.ts`, `utils/skillSetRank*.ts`, `skill-access-modal.tsx`, `custom-fighters.tsx`, `fighter-advancement-list.tsx`, etc. are byte-identical to what was already reviewed.

1. **[High] N26 skill catalog will resolve empty once N26 gang types ship** (`app/api/skill-types/route.ts`) — **outstanding**, untouched by the merge.
2. **[Low] Dead `editionSlug` prop** (`custom-fighters.tsx`) — **outstanding**, untouched.
3. **[Low] Duplicate `getSkillSetRank` call** (`skill-access-modal.tsx:40,111`) — **outstanding**, untouched.
4. **[Nit] IIFE instead of extracted variable** (`admin-create-fighter-type.tsx`) — **outstanding**, untouched (still worth hoisting to a `useMemo` like the edit-modal sibling, especially now that this block also needs to read `hasVehicles`/`editions` state that's computed elsewhere in the component).
5. **[Nit] Inconsistent embed FK hint** (`app/api/skill-types/route.ts`) — **outstanding**, untouched.

### New findings
None — the merge introduced no new logic in this PR's scope, only a correct conflict resolution.

### Verdict
**Changes requested** — unchanged from the previous review. Item #1 is still the one that should be resolved (or explicitly deferred with a tracking note) before merge; the rest are non-blocking.

---

joeseos:
## Follow-up review

@Djidiouf Reviewed the latest commit (`428a51a`) against the prior review point (`4bc3f77`) — a 4-file, 66/49 diff addressing every open finding from the previous rounds.

1. **[High] N26 skill catalog would resolve empty once N26 gang types ship** (`app/api/skill-types/route.ts`) — **resolved**. The route now fetches the resolved edition's `skill_types`, and if that comes back empty (and the resolved edition isn't already N23) falls back to the N23 catalog id, with a dev-only `console.warn`. Traced every downstream usage (`ownTypesQuery`, both shared/fighter `custom_skill_types` lookups) and they all consistently switch to the post-fallback `catalogEditionId` rather than leaving any of them on the stale `resolvedEditionId` — no split-brain between the standard and custom type queries.
2. **[Low] Dead `editionSlug` prop** (`custom-fighters.tsx`) — **resolved**. Prop dropped from `SkillSetOptions`, replaced with a hardcoded `getSkillSetRank(EDITION_N23)` and a comment noting Customise is N23-only for now. Confirmed none of the 4 call sites in the file were ever passing `editionSlug`, so this is a clean no-op removal, not a behavior change.
3. **[Low] Duplicate `getSkillSetRank` call** (`skill-access-modal.tsx`) — **resolved**. The inner duplicate inside `computedSkillAccess` is gone; the memo now closes over the outer `skillSetRank` and it's been added to the dependency array. Checked `getSkillSetRank`'s implementation — it returns one of two module-level constant objects (`skillSetRankN23`/`skillSetRankN26`), not a fresh object per call, so including it in the deps array doesn't break memoization the way a freshly-allocated object would.
4. **[Nit] IIFE instead of extracted variable** (`admin-create-fighter-type.tsx`) — **resolved**. Hoisted into a `groupedSkillTypes` useMemo (deps: `filteredSkillTypes`, `editionSlug`), placed after `editionSlug` is defined, matching the sibling `admin-edit-fighter-type.tsx` pattern.
5. **[Nit] Inconsistent embed FK hint** (`app/api/skill-types/route.ts`) — **resolved**. `gang_types (...)` → `gang_types!gang_type_id (...)`, matching the convention used everywhere else in the codebase.

### New findings

None. This commit is a clean, scoped fix-up — no new logic paths beyond the N23 fallback (which was the explicit ask), and no regressions spotted in the surrounding code it touched. `tsc --noEmit` shows no errors in the changed files (only pre-existing, unrelated tsconfig deprecation warnings for `target`/`baseUrl`).

### Verdict
**Approve** — all outstanding findings from the previous two rounds are addressed and nothing new was introduced.

---

joeseos:
## Follow-up review

@Djidiouf Reviewed the latest commit (`7f826a9`, authored by @joeseos) against the previously-approved `428a51a` — a 2-file, 25/9 diff refactoring `getSkillSetRank` and the gang-edition lookup in `app/api/skill-types/route.ts`.

All five findings from the earlier rounds were already resolved as of `428a51a` and are untouched by this commit (confirmed no diff on `custom-fighters.tsx`, `skill-access-modal.tsx`, `admin-create-fighter-type.tsx`, and the N23-fallback block in `skill-types/route.ts` — only the gang-edition lookup portion of that file changed). Nothing outstanding.

### This commit
**1. `utils/skillSetRank.ts` — `getSkillSetRank` now keyed by `EditionSlug`**
Replaced the `editionSlug === EDITION_N26 ? ... : ...` ternary with `SKILL_SET_RANK_BY_EDITION: Record<EditionSlug, ...>`, matching the pattern used by `LIMITS_BY_EDITION`/`INJURY_TABLES_BY_EDITION`/`CATEGORY_RANK_BY_EDITION`. Verified this is exhaustive (`n23`/`n26` cover all of `EditionSlug`), so adding a new edition slug without a matching rank table is now a compile error rather than a silent N23 fallback for that lookup table — the commit message's claim ("adding n29 now fails with TS2741") checks out against the type shape. The N23-on-unknown/null/empty-string fallback is preserved (`if (!editionSlug) return skillSetRankN23; ... ?? skillSetRankN23`), and the added comment correctly explains why this registry differs from the others (sorting/grouping a dropdown vs. gating rules-legal data).

**2. `app/api/skill-types/route.ts` — gang→edition lookup now goes through `gangEditionSlug`**
Replaces the hand-rolled `editionSlugFromJoin((gang as any)?.gang_types?.editions) ?? editionSlugFromJoin((gang as any)?.custom_gang_type_edition?.editions) ?? null` with the shared `gangEditionSlug(gang)` helper from `types/edition.ts`. Checked `gangEditionJoin`'s alias list — it includes `custom_gang_type_edition`, so the precedence (`gang_types` first, then the custom alias) is preserved, and it double-unwraps via `firstOf` so both the array- and object-shaped embed forms resolve correctly, which was the actual bug in the original inline version per the first review round. This is a legitimate DRY-up, not just a rename: it deletes a second, less-robust copy of embed-flattening logic.

No new issues spotted — no edge-case regressions, `tsc --noEmit` clean on the touched files (only the pre-existing `target`/`baseUrl` deprecation warnings, unrelated to this diff), and both callers-affecting changes (`getSkillSetRank` signature, `resolvedEditionSlug` assignment) preserve their prior external behavior while tightening the internals.

### Verdict
**Approve** — this is a clean internal hardening pass with no behavior change and no new findings.

---

joeseos:
## Follow-up review

@Djidiouf Reviewed the latest commit (`5818c7b`) against the previously-approved `7f826a9` — a single-file, 6/10 diff in `utils/skillSetRank.ts`.

All five findings from the earlier rounds remain resolved and untouched by this commit — it only touches `getSkillSetRank`'s fallback branch and its docstring.

### This commit
**`utils/skillSetRank.ts` — drop the N23 fallback for an unset/unrecognised edition slug**
`getSkillSetRank` now returns `{}` instead of `skillSetRankN23` when `editionSlug` is falsy or not a recognised `EditionSlug`, matching `LIMITS_BY_EDITION`/`INJURY_TABLES_BY_EDITION`/`CATEGORY_RANK_BY_EDITION` (confirmed `getCharacteristicLimits` in `utils/characteristicLimits.ts:87-88` does the same: `if (!limits) return NO_LIMITS`).

Checked every consumer of `skillSetRank[...]` (`fighter-skills-list.tsx`, `fighter-advancement-list.tsx`, `skill-access-modal.tsx`, `custom-fighters.tsx`, and the three admin skill/fighter-type screens) — all index it with `?? Infinity`, and every grouping function treats an `Infinity` (or otherwise out-of-range) rank as the `'Misc.'` bucket. So `{}` degrades to "everything ungrouped," never a crash or an `undefined` render, which matches the commit's stated intent.

`editionSlug` is threaded from `gangData.processedData.edition_slug` / `fighterData.gang?.edition_slug` at the real call sites (`gang-page-content.tsx`, `fighter-page.tsx`), which are populated for any resolved gang, so this only changes behavior when an edition genuinely fails to resolve or before an edition is chosen in the admin create flows — consistent with the commit message's claim.

No new issues spotted. `tsc --noEmit` clean on the touched file (only the pre-existing, unrelated `target`/`baseUrl` deprecation warnings).

### Verdict
**Approve** — brings the last outlier registry in line with the rest of the edition-gating conventions, with no behavioral regression for resolved editions.

---

PR #1954, opened 2026-08-08
Fixed the uniqueness of skill_types - migration already applied

Djidiouf:


joeseos:
## Code Review

@Djidiouf Reviewed against the full codebase (schema history, `app/api/admin/skill-types/route.ts`, `components/admin/admin-create-skill.tsx`, `components/edition-select.tsx`).

### Summary
This adds a follow-up migration that idempotently re-applies the `(edition_id, name)` unique constraint on `skill_types` — `DROP CONSTRAINT IF EXISTS` + a `pg_constraint` existence check before `ADD CONSTRAINT`, matching the guarded-DDL pattern already used in `20260630095837_add_editions_and_root_edition_ids.sql`. This correctly protects against the drift scenario implied by the title (the constraint from `20260808100000`/#1950 already applied out-of-band, so a bare re-run would error on the unguarded `DROP`/`ADD`). On a fresh/consistent environment this migration is a safe no-op; on the drifted environment it reconciles state without erroring. Good, minimal, well-targeted fix.

### Findings

**1. `edition_id IS NULL` bypasses the uniqueness guarantee (medium, not introduced by this diff but worth addressing)**
- File: `supabase/migrations/20260809030000_skill_types_unique_per_edition.sql:16-18`
- `skill_types.edition_id` is nullable (`20260630095837_add_editions_and_root_edition_ids.sql:102`, "edition_id stays nullable here"), and the admin UI can create a skill set with no edition: `EditionSelect` (`components/edition-select.tsx:70`) has an explicit `"Select edition"` blank option, and `admin-create-skill.tsx:125` / `app/api/admin/skill-types/route.ts:49,85` send `edition_id: editionId || null` straight through.
- A plain `UNIQUE (edition_id, name)` constraint treats `NULL` as distinct from `NULL` in Postgres, so two rows named `"Agility"` with `edition_id = NULL` can both be inserted — reproducing exactly the duplicate-name problem this whole constraint chain (#1950 → this PR) is meant to prevent, just for the "no edition" case instead of across editions.
- The codebase already has a precedent for this exact situation: `fighter_classes` uses a partial unique index instead of a plain constraint (`CREATE UNIQUE INDEX ... ON public.fighter_classes (edition_id, slug) WHERE edition_id IS NOT NULL`, same migration, lines 150-152) specifically because its `edition_id` is nullable too.
- Suggested fix (can be a follow-up migration, doesn't need to block this one): either `UNIQUE NULLS NOT DISTINCT (edition_id, name)` (Postgres 15+, treats `NULL`s as duplicates), or add a companion `UNIQUE (name) WHERE edition_id IS NULL` partial index alongside the existing constraint.

**2. Minor / non-blocking**
- The `COMMENT ON CONSTRAINT` at the end runs unconditionally regardless of whether the `DO` block actually created the constraint or found it pre-existing — that's fine since `COMMENT ON` is itself idempotent, just noting it's intentional rather than an oversight.
- `supabase/schema/schema.public.sql` still shows the old `skill_types_name_key UNIQUE (name)` constraint, but that file is refreshed by separate automated `chore(db): update Supabase schema snapshot` commits after merge (per repo history), so no action needed here.

### Verdict: **Approve**

The migration itself is correct, safe, and idempotent for its stated purpose (reconciling drifted migration state). Finding #1 is a real gap but predates this PR (inherited from the already-merged `20260808100000`/#1950) and is best handled as its own follow-up migration rather than blocking this narrow reconciliation fix.


---

joeseos:
@Djidiouf Is this not the same as https://github.com/joeseos/mundamanager/pull/1950?

PR #1955, opened 2026-08-08
Fixed issues in default skills and default special rules in Add / Edi…

Djidiouf:
…t fighter type in Admin dashboard

### DIFF
--- a/app/api/admin/skills/route.ts
+++ b/app/api/admin/skills/route.ts
@@ -64,23 +64,41 @@ export async function GET(request: Request) {
       return NextResponse.json(data);
     }
 
-    // Handle skill queries
+    // Handle skill queries.
+    // Skills inherit edition through skill_types, but skills.skill_type_id has
+    // no FK yet, so PostgREST cannot embed skill_types (PGRST200). Resolve
+    // matching skill-set ids first, then filter skills with .in().
     let query = supabase
       .from('skills')
-      .select(editionId
-        ? 'id, name, skill_type_id, gang_origin_id, skill_types!inner(edition_id)'
-        : 'id, name, skill_type_id, gang_origin_id, skill_types(edition_id)')
+      .select('id, name, skill_type_id, gang_origin_id')
       .order('name');
 
     if (skillTypeId) {
       query = query.eq('skill_type_id', skillTypeId);
     }
 
-    // A skill belongs to an edition through its skill set. The inner relation is
-    // important here: filtering an ordinary embedded relation would leave the
-    // parent skill rows in the result with an empty relation.
     if (editionId) {
-      query = query.eq('skill_types.edition_id', editionId);
+      let skillTypeQuery = supabase
+        .from('skill_types')
+        .select('id')
+        .eq('edition_id', editionId);
+
+      if (skillTypeId) {
+        skillTypeQuery = skillTypeQuery.eq('id', skillTypeId);
+      }
+
+      const { data: editionSkillTypes, error: skillTypesError } = await skillTypeQuery;
+      if (skillTypesError) throw skillTypesError;
+
+      const editionSkillTypeIds = (editionSkillTypes ?? []).map((st) => st.id);
+      if (editionSkillTypeIds.length === 0) {
+        if (skillTypeId) {
+          return NextResponse.json({ skills: [], effect_categories: [] });
+        }
+        return NextResponse.json([]);
+      }
+
+      query = query.in('skill_type_id', editionSkillTypeIds);
     }
 
     const { data, error } = await query;
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -72,7 +72,8 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
   const [intelligence, setIntelligence] = useState('');
   const [attacks, setAttacks] = useState('');
   const [save, setSave] = useState('');
-  const [specialSkills, setSpecialSkills] = useState('');
+  const [specialRules, setSpecialRules] = useState<string[]>([]);
+  const [newSpecialRule, setNewSpecialRule] = useState('');
   const [freeSkill, setFreeSkill] = useState(false);
   const [isGangAddition, setIsGangAddition] = useState(false);
   const [isDramatisPersonae, setIsDramatisPersonae] = useState(false);
@@ -281,6 +282,22 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
     staleTime: 5 * 60 * 1000,
   });
 
+  const handleAddSpecialRule = () => {
+    if (!newSpecialRule.trim()) return;
+
+    if (specialRules.includes(newSpecialRule.trim())) {
+      setNewSpecialRule('');
+      return;
+    }
+
+    setSpecialRules(prev => [...prev, newSpecialRule.trim()]);
+    setNewSpecialRule('');
+  };
+
+  const handleRemoveSpecialRule = (ruleToRemove: string) => {
+    setSpecialRules(prev => prev.filter(rule => rule !== ruleToRemove));
+  };
+
   const handleSubmit = async () => {
     // Check if selected fighter subtype is Crew
     const isCrew = selectedFighterSubtypes.includes('Crew');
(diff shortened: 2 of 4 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed against the full codebase (`app/api/admin/skills/route.ts` and its callers, `components/admin/admin-create-fighter-type.tsx`, `components/admin/admin-edit-fighter-type.tsx`, `supabase/migrations/`, `supabase/schema/schema.public.sql`).

### Summary
Three independent fixes bundled together: (1) special rules moves from a comma-separated free-text `Input` to a chip-based add/remove list in both the create and edit fighter-type admin modals, (2) the admin skills API stops relying on a `skills -> skill_types` PostgREST embed (which was failing with PGRST200 since no FK ever backed that relationship) and instead resolves matching `skill_type` ids first, then filters skills with `.in(...)`, and (3) a new migration adds the missing `skills_skill_type_id_fkey`. All three read as legitimate fixes for real, previously-broken behavior (Default Skills wasn't loading in the fighter type modals). The two response-shape branches (`{skills, effect_categories}` vs a bare array) are preserved correctly in the new early-return-on-empty path, matching what both callers already expect (`admin-create-fighter-type.tsx:279`, `admin-edit-fighter-type.tsx:502`).

### Findings

**1. Migration's "clean" verification looks like a local/manual fix, not a database-wide guarantee (medium)**
- File: `supabase/migrations/20260809040000_add_skills_skill_type_id_fkey.sql:19-20`
- The comment says: *"Verified clean before writing: the one orphan (\"Test Skill 2\") was deleted manually; remaining skills all point at an existing skill_type."* That describes a one-off manual `DELETE` performed against whatever database the author checked (presumably local/dev), not a guarantee that no orphaned `skills.skill_type_id` rows exist on every environment this migration will run against.
- Compare with the precedent this PR's own comment style is clearly modeled on, `supabase/migrations/20260806180000_add_campaigns_campaign_type_id_fkey.sql:13-14`: *"Verified clean before writing: 5177 campaigns, 0 null campaign_type_id, 0 rows pointing at a missing campaign type. This applies without fixups."* That one states row counts against the real target database and needs no manual cleanup step. This migration instead documents a manual fixup that may or may not have been applied everywhere the migration will run.
- If a "Test Skill 2"-style orphan (or any other skill pointing at a deleted skill_type) exists in staging/production and wasn't part of the manual cleanup, `ALTER TABLE ... ADD CONSTRAINT` will fail at deploy time.
- Suggested fix: either confirm+state the same row-count verification against the actual deploy target (production), or make the migration self-contained by deleting/nulling orphaned rows in a preceding statement rather than relying on an out-of-band manual step.

**2. `index` as React key for removable chip list (low)**
- Files: `components/admin/admin-create-fighter-type.tsx` (special rules chip render, ~line 798), `components/admin/admin-edit-fighter-type.tsx` (~line 1885)
- `specialRules.map((rule, index) => <div key={index}>...)` — since `handleAddSpecialRule` already guarantees uniqueness (`specialRules.includes(...)` check) and rules are plain strings, `key={rule}` would be a more stable key than `key={index}`, avoiding key reuse across removals in the middle of the list. Low risk here since the chips are stateless (no inputs/animation), but worth the small fix for correctness hygiene.

**3. `handleAddSpecialRule`/`handleRemoveSpecialRule` duplicated verbatim across both modals (low, DRY)**
- Both handlers are copy-pasted identically between `admin-create-fighter-type.tsx` and `admin-edit-fighter-type.tsx`. Not introduced by this PR as a new pattern (the two modals already duplicate most of their form logic), so not blocking, just flagging since it's more logic than the previous one-liner `.split(',')` version.

### Verdict: **Approve**

The core fixes (chip-based Special Rules UI, and dropping the broken PostgREST embed in favor of a two-query resolve-then-filter approach) are correct and well-reasoned. Finding #1 is the only one worth confirming before/at merge — a quick check that production has no orphaned `skills.skill_type_id` rows would remove the only real deploy risk in this PR.

---

joeseos:
## Follow-up Review (new commit since last review)

@Djidiouf This reviews only the changes in `454b024` (pushed after the previous review comment on `9f13510`). Only one file changed:

```
supabase/migrations/20260809040000_add_skills_skill_type_id_fkey.sql | 9 +--------
```

### Previous findings

1. **Migration's "clean" verification looks like a local/manual fix, not a database-wide guarantee (medium)** — **outstanding, not resolved.** The suggested fix was to either (a) confirm and state the same row-count verification against the actual deploy target (production), or (b) make the migration self-contained by deleting/nulling orphaned rows in a preceding statement. Neither happened. Instead, the commit just deletes the caveat text:
   ```diff
   - -- Skills inherit their edition through skill_types (see
   - -- 20260630095837_add_editions_and_root_edition_ids.sql); the embed is the
   - -- intended way to scope skills to an edition.
   - --
    -- ON DELETE RESTRICT: deleting a skill set must not silently wipe its skills
    -- (those skills may still be referenced by fighter_skills / fighter_defaults
   - -- with NO ACTION FKs). Matches campaigns_campaign_type_id_fkey.
   - --
   - -- Verified clean before writing: the one orphan ("Test Skill 2") was deleted
   - -- manually; remaining skills all point at an existing skill_type.
   + -- with NO ACTION FKs).
   ```
   The underlying deploy risk is unchanged — if any `skills.skill_type_id` row points at a deleted `skill_type` on the target database, `ALTER TABLE ... ADD CONSTRAINT` still fails at migration time. Removing the caveat just makes that risk less visible to whoever runs this migration next; it doesn't reduce it. Worth restoring a short note and actually confirming row counts against production before merge (or adding a cleanup statement ahead of the `ALTER TABLE`).

2. **`index` as React key for removable chip list (low)** — outstanding, untouched in this commit.
3. **Duplicated `handleAddSpecialRule`/`handleRemoveSpecialRule` across both modals (low, DRY)** — outstanding, untouched in this commit. Non-blocking per the earlier review.

### New findings

No new functional issues in this diff — it's a comment-only trim. Flagging as the one new observation: dropping the "verified clean"/rationale comments reduces the audit trail for a migration that still relies on an unverified precondition (see above). Severity: **low** (documentation regression, not a functional change).

### Final verdict: **Approve**

No functional regressions were introduced by this commit. The one item worth confirming before merge is unchanged from the previous review: verify there are no orphaned `skills.skill_type_id` rows on production before this migration runs there, since that's now undocumented rather than resolved.

---

joeseos:
## Follow-up Review (new commit since last review)

@Djidiouf This reviews only the changes in `62a60a7` (pushed after the previous review comment on `454b024`). Only one file changed:

```
supabase/migrations/20260809040000_add_skills_skill_type_id_fkey.sql | 9 +++++++-
```

### Previous findings

1. **Migration's "clean" verification looks like a local/manual fix, not a database-wide guarantee (medium)** — **resolved.** This commit adds a preceding `DELETE` statement that removes any `skills` row whose `skill_type_id` doesn't resolve to an existing `skill_types` row, before the `ALTER TABLE ... ADD CONSTRAINT`. That's option (b) from the original suggestion: the migration is now self-contained and no longer depends on an out-of-band manual cleanup having been applied to every environment it runs against.
   - Verified the safety claim in the new comment ("fails loudly if an orphan is still referenced by `fighter_skills` / `fighter_defaults`"): both `fighter_defaults_skill_id_fkey` and `fighter_skills_skill_id_fkey` reference `skills(id)` with no `ON DELETE` clause (i.e. `NO ACTION`), so if a to-be-deleted orphan skill is actually still in use, the `DELETE` raises a FK violation and the migration aborts rather than silently cascading data loss. The claim checks out.
   - The `WHERE s.skill_type_id IS NOT NULL AND NOT EXISTS (...)` predicate is correct — it only targets genuinely dangling references and leaves legitimately-`NULL` `skill_type_id` rows untouched.
2. **`index` as React key for removable chip list (low)** — outstanding, untouched in this commit (no non-migration files changed).
3. **Duplicated `handleAddSpecialRule`/`handleRemoveSpecialRule` across both modals (low, DRY)** — outstanding, untouched in this commit. Non-blocking per the earlier review.

### New findings

None. The added `DELETE` is scoped, correct, and safe by construction (guarded by the existing FKs into `skills`). No new functional issues introduced.

### Final verdict: **Approve**

All findings from prior reviews are now either resolved (the deploy-risk migration issue) or remain as previously-flagged low-severity, non-blocking items (React key choice, handler duplication).

---

PR #1956, opened 2026-08-08
Fixed Weapon group for Add and Edit Equipment not being gated by edition

Djidiouf:


joeseos:
## Code Review

@Djidiouf Reviewed against the full codebase (route.ts + both admin equipment modals).

### Summary

This does more than the title says, in a good way: it fixes two bugs together.

1. **Edition gating** (the stated fix): the Weapon Group dropdown in Add/Edit Equipment now excludes weapons from other editions, and picking a new edition clears an existing cross-edition `weapon_group_id` selection (with the stats-blanking logic it sits next to).
2. **Type filtering** (a bonus fix, confirmed by diffing against `main`): `equipment_type` was already being sent as `?equipment_type=weapon` by both modals but was never read server-side — `route.ts`'s "no filters" branch returned *all* equipment regardless. So the Weapon Group dropdown was previously showing wargear and vehicle upgrades too, not just weapons. This PR wires up the param to actually filter. Good catch.

I checked for regressions from the added `equipment_type` filter — grepped all callers of `/api/admin/equipment` and confirmed only the two weapon-fetch call sites pass `equipment_type`, so no other consumer of the "return all equipment" branch is affected.

I also checked the `handleEditionChange` clearing logic in both files uses the raw unfiltered `weapons` list (not `filteredWeapons`) to look up the stale parent — correct, since the parent may belong to an edition other than the newly selected one.

### Findings

Nothing blocking. Two very minor, non-blocking observations:

- **Low / edge case** — `admin-create-equipment.tsx:141-146` and `admin-edit-equipment.tsx:217-227`: the cross-edition clear relies on `weapons` already being populated. If a user could set `weapon_group_id` before the `admin-weapons` query resolves, the stale value wouldn't be cleared. In practice this can't happen — the dropdown that sets `weapon_group_id` is populated from the same `weapons` query, so a value can only exist once `weapons` has loaded. Not a real bug, just noting the implicit coupling.
- **Low / DRY** — `filteredWeapons` (useMemo) and the `weapons.find(...)` cross-edition-clear block are duplicated near-verbatim between `admin-create-equipment.tsx` and `admin-edit-equipment.tsx`. This matches the existing convention in these files (categories/gang-types filtering is duplicated the same way), so it's consistent with the codebase rather than a new violation — just flagging in case a shared hook is ever worth extracting for all these admin-modal edition-scoping filters.

No security, null-handling, or performance concerns — the added `.eq('equipment_type', ...)` is a straightforward indexed-ish filter on an already-small admin table, and the `id` branch (single equipment lookup) is untouched.

### Verdict: **Approve**

---

PR #1959, opened 2026-08-09
Added N26 ranking for Alliances, Fighter subtypes, gang additions and…

Djidiouf:
… gated those by edition

### DIFF
--- a/app/api/alliances/route.ts
+++ b/app/api/alliances/route.ts
@@ -7,6 +7,7 @@ interface Alliance {
   id: string;
   alliance_name: string;
   alliance_type: string | null;
+  alliance_crew_name: string | null;
   strong_alliance: string | null;
   editions?: { slug: string } | { slug: string }[] | null;
 }
@@ -22,7 +23,7 @@ export async function GET() {
 
     const { data, error } = await supabase
       .from('alliances')
-      .select('id, alliance_name, alliance_type, strong_alliance, editions:edition_id (slug)')
+      .select('id, alliance_name, alliance_type, alliance_crew_name, strong_alliance, editions:edition_id (slug)')
       .order('alliance_name');
 
     if (error) throw error;
@@ -31,6 +32,7 @@ export async function GET() {
       id: alliance.id,
       alliance_name: alliance.alliance_name,
       alliance_type: alliance.alliance_type,
+      alliance_crew_name: alliance.alliance_crew_name,
       strong_alliance: alliance.strong_alliance,
       edition_slug: editionSlugFromJoin(alliance.editions),
     }));
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -1635,10 +1635,10 @@ export function AdminEditEquipmentModal({ onClose, onSubmit }: AdminEditEquipmen
                         // First sort by gang type
                         const gangCompare = a.gang_type.localeCompare(b.gang_type);
                         if (gangCompare !== 0) return gangCompare;
-                        // Then by fighter subtype priority
+                        // Then by fighter subtype priority (per fighter type's own edition)
                         const subtypeCompare =
-                          getFighterSubtypeSortRank(a.fighter_subtypes, editionSlug)
-                          - getFighterSubtypeSortRank(b.fighter_subtypes, editionSlug);
+                          getFighterSubtypeSortRank(a.fighter_subtypes, editionSlugOf(editions, a.edition_id))
+                          - getFighterSubtypeSortRank(b.fighter_subtypes, editionSlugOf(editions, b.edition_id));
                         if (subtypeCompare !== 0) return subtypeCompare;
                         // Finally by fighter type name
                         return a.fighter_type.localeCompare(b.fighter_type);
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -10,7 +10,7 @@ import { toast } from 'sonner';
 import { FighterType } from "@/types/fighter";
 import { WeaponProfileInput, emptyWeaponProfile, EquipmentGrants, EquipmentAvailability, EquipmentOriginAvailability, EquipmentVariantAvailability, GangAdjustedCost, GangOriginAdjustedCost } from "@/types/equipment";
 import { HiX } from "react-icons/hi";
-import { fighterSubtypeRank } from "@/utils/fighterSubtypeRank";
+import { getFighterSubtypeSortRank } from "@/utils/fighterSubtypeRank";
 import { gangOriginRank } from "@/utils/gangOriginRank";
 import { gangVariantRank } from "@/utils/gangVariantRank";
 import { AdminFighterEffects } from "./admin-fighter-effects";
@@ -1636,8 +1636,9 @@ export function AdminEditEquipmentModal({ onClose, onSubmit }: AdminEditEquipmen
                         const gangCompare = a.gang_type.localeCompare(b.gang_type);
                         if (gangCompare !== 0) return gangCompare;
                         // Then by fighter subtype priority
-                        const subtypeCompare = (fighterSubtypeRank[(a.fighter_subtypes?.[0] || '').toLowerCase() as keyof typeof fighterSubtypeRank] || Infinity)
-                          - (fighterSubtypeRank[(b.fighter_subtypes?.[0] || '').toLowerCase() as keyof typeof fighterSubtypeRank] || Infinity);
+                        const subtypeCompare =
+                          getFighterSubtypeSortRank(a.fighter_subtypes, editionSlug)
+                          - getFighterSubtypeSortRank(b.fighter_subtypes, editionSlug);
                         if (subtypeCompare !== 0) return subtypeCompare;
                         // Finally by fighter type name
                         return a.fighter_type.localeCompare(b.fighter_type);
--- a/components/edition-select.tsx
+++ b/components/edition-select.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { Edition } from '@/types/edition';
 
@@ -46,14 +46,22 @@ export const editionSlugOf = (editions: Edition[], editionId?: string | null): s
 export function EditionSelect({ value, onChange, defaultToCurrent = false, label = 'Edition' }: EditionSelectProps) {
   const { data: editions = [] } = useEditions();
 
-  // Create/edit forms (defaultToCurrent) must always have an edition: fill the
-  // current one whenever value is empty. Filter contexts omit defaultToCurrent
-  // so an explicit blank "all editions" selection can stick.
+  // Default only once per mount for create/edit forms. Do NOT re-fire when an
+  // existing catalog row with edition_id null clears value — is_current is only
+  // a UI default, never an assignment onto stored rows.
+  const hasDefaulted = useRef(false);
   useEffect(() => {
-    if (!defaultToCurrent || value || editions.length === 0) return;
+    if (hasDefaulted.current || editions.length === 0) return;
+    hasDefaulted.current = true;
+    if (!defaultToCurrent || value) return;
     const current = editions.find(edition => edition.is_current);
     if (current) onChange(current.id);
   }, [defaultToCurrent, value, editions, onChange]);
+
+  // Filter contexts keep a blank "all editions" choice. Create/edit forms omit
+  // it once a value is set; if value is empty (e.g. a legacy null edition_id
+  // row), keep the blank option so that null is visible and preservable.
+  const showBlankOption = !defaultToCurrent || !value;
 
   return (
     <div>
@@ -65,9 +73,7 @@ export function EditionSelect({ value, onChange, defaultToCurrent = false, label
         onChange={(e) => onChange(e.target.value)}
         className="w-full p-2 border rounded-md"
       >
-        {/* Create/edit forms that default to the current edition always need one
-            selected; omit the blank option so "All Editions" isn't available. */}
-        {!defaultToCurrent && <option value="">Select edition</option>}
+        {showBlankOption && <option value="">Select edition</option>}
         {editions.map((edition) => (
           <option key={edition.id} value={edition.id}>
             {edition.name}
--- a/components/edition-select.tsx
+++ b/components/edition-select.tsx
@@ -1,6 +1,6 @@
 "use client"
 
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 import { useQuery } from '@tanstack/react-query';
 import { Edition } from '@/types/edition';
 
@@ -46,13 +46,11 @@ export const editionSlugOf = (editions: Edition[], editionId?: string | null): s
 export function EditionSelect({ value, onChange, defaultToCurrent = false, label = 'Edition' }: EditionSelectProps) {
   const { data: editions = [] } = useEditions();
 
-  // Default only once, when editions first load — never after, so an explicit
-  // blank selection ("all editions" in filter contexts) sticks
-  const hasDefaulted = useRef(false);
+  // Create/edit forms (defaultToCurrent) must always have an edition: fill the
+  // current one whenever value is empty. Filter contexts omit defaultToCurrent
+  // so an explicit blank "all editions" selection can stick.
   useEffect(() => {
-    if (hasDefaulted.current || editions.length === 0) return;
-    hasDefaulted.current = true;
-    if (!defaultToCurrent || value) return;
+    if (!defaultToCurrent || value || editions.length === 0) return;
     const current = editions.find(edition => edition.is_current);
     if (current) onChange(current.id);
   }, [defaultToCurrent, value, editions, onChange]);
@@ -67,7 +65,9 @@ export function EditionSelect({ value, onChange, defaultToCurrent = false, label
         onChange={(e) => onChange(e.target.value)}
         className="w-full p-2 border rounded-md"
       >
-        <option value="">Select edition</option>
+        {/* Create/edit forms that default to the current edition always need one
+            selected; omit the blank option so "All Editions" isn't available. */}
+        {!defaultToCurrent && <option value="">Select edition</option>}
         {editions.map((edition) => (
           <option key={edition.id} value={edition.id}>
             {edition.name}
--- a/components/fighter/edit-fighter/fighter-edit-modal.tsx
+++ b/components/fighter/edit-fighter/fighter-edit-modal.tsx
@@ -10,7 +10,7 @@ import { Combobox } from "@/components/ui/combobox";
 import { HiX } from "react-icons/hi";
 import { toast } from 'sonner';
 import { applySpecialRulesModifiers } from '@/utils/effect-modifiers';
-import { fighterSubtypeRank } from '@/utils/fighterSubtypeRank';
+import { getFighterSubtypeSortRank } from '@/utils/fighterSubtypeRank';
 import { isArchetypeEligible, mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility';
 import { SkillAccessModal } from './skill-access-modal';
 import { FighterCharacteristicTable } from './fighter-characteristic-table';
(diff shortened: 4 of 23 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed the diff (edition-scoped ranking for fighter subtypes, gang additions, and alliances, plus the N26-specific "Category" UI in `FighterAddModal`).

### What's good
- Splitting `fighterSubtypeRank`/`gangAdditionRank`/`allianceRank` into per-edition tables behind `getXRank(editionSlug)` accessors, keyed by `EditionSlug`, is a clean pattern consistent with `LIMITS_BY_EDITION` elsewhere in the codebase — a new edition is a compile error until it supplies its own table.
- `getFighterSubtypeSortRank`/`getGangAdditionSortRank` correctly take the *best* (lowest) rank across all of a fighter's subtype labels instead of only `subtypes[0]`, which fixes a latent issue where a fighter with e.g. `["Pious", "Prospect"]` would previously sort by "Pious" (unranked → `Infinity`) instead of "Prospect".
- The N26 gang-additions catalog (category dropdown + alliance grouping by real `alliance_type`, `isN26GangAdditionType` special-casing `alliance_id` since `gangAdditionRankN26` has no "crew" entry) is a sensible, self-contained addition.
- Edition-scoped filtering of the cross-gang additions pool via `sameEditionForDisplay(type.edition_slug, editionSlug)` matches the documented intent and the existing null-means-N23 convention in `types/edition.ts`.

### Issues

**1. Medium — Admin equipment form loses subtype sort ordering when "All Editions" is selected**
`components/admin/admin-edit-equipment.tsx:1636-1640`:
```ts
const subtypeCompare =
  getFighterSubtypeSortRank(a.fighter_subtypes, editionSlug)
  - getFighterSubtypeSortRank(b.fighter_subtypes, editionSlug);
```
`editionSlug` here is `editionSlugOf(editions, editionId)` (line 143), and `filteredFighterTypes` (line 410-413) is only edition-filtered `if (editionId)` — when no specific edition is selected in this form, `fighterTypes` contains rows from every edition and `editionSlug` is falsy. `getFighterSubtypeRank` returns `{}` for a falsy `editionSlug` (`utils/fighterSubtypeRank.ts`), so every row's rank becomes `Infinity` and `subtypeCompare` is always `0` — the "Fighter Types with this Equipment" dropdown silently falls back to gang-name/fighter-name ordering with no subtype grouping at all in that mode. Before this PR the table was a single flat, edition-agnostic map, so this sort worked regardless of the edition filter. Not a crash, but a real regression in the "All Editions" view of this admin picker. Worth either falling back to N23's table when no edition is selected, or accepting/documenting the degraded sort in that mode.

**2. Low — duplicate rank value in `gangAdditionRankN23.ts`**
`utils/gangAdditionRankN23.ts`:
```ts
"justicar courts, palanite justicar delegation": 52,
"promethium guild, pyromantic conclave": 52,
```
Both new/renumbered entries share rank `52` (the surrounding sequence is otherwise strictly increasing: 50, 51, 52, 52, 53, 54). Looks like an off-by-one when inserting the new "justicar courts" row rather than an intentional tie. Doesn't break the group-label bucketing (both still fall in the same guild bucket), but the two crews will tie-break by name instead of preserving whatever relative order was intended — worth bumping one to `52`/`53` and shifting the rest down by one if a specific order was intended.

**3. Low/nit — `alliance_type` grouping now trusts raw DB text instead of the rank table**
`components/gang/gang-edit-modal.tsx` replaces the old rank-bucket-derived grouping (`maxRank` thresholds against `allianceRank`) with a direct switch on `alliance.alliance_type` (`allianceTypeGroupLabel`), and `FighterAddModal.tsx`'s `allianceTypeHeader`/`allianceTypeHeaderRank` do something similar with case-insensitive matching against `N26_ALLIANCE_TYPE_ORDER`. That's a more direct, arguably more correct source of truth, but it does mean grouping now depends on `alliances.alliance_type` containing exactly the expected values (`"criminal"`, `"merchant guilds"`, `"noble houses"`, `"other"/"others"`, case-insensitive) rather than derived data validated by this code. If any row has a slightly different string (e.g. a typo, trailing whitespace beyond what's trimmed, or a different casing than assumed) it'll silently land in "Other Alliances" / get a fallback header rather than erroring. Can't verify actual stored values from the repo — worth a quick sanity check against the live `alliances` table before merge.

### Verdict: **Changes requested** (minor)

Nothing here blocks the core N26 feature — the new ranking/category logic itself checks out against the worked cases in the diff. I'd want #1 addressed (or at least acknowledged) since it's a real behavioral regression in the admin tool, and #2 is a one-line fix. #3 is just a "please double check the data" note.

---

joeseos:
## Code Review — follow-up (commits `da5d720` + `a934866` since my last review at `8a4e387`)

@Djidiouf Reviewed just the diff since the last round: `components/edition-select.tsx`, `components/gang/fighter-add/FighterAddModal.tsx` (specialisation auto-select), and `utils/gangAdditionRankN23.ts`.

| # | Finding | Status |
|---|---|---|
| 1 | Admin equipment form lost subtype sort ordering when "All Editions" was selected | **Effectively resolved**, but see new finding below — `EditionSelect` no longer renders the blank "Select edition" option when `defaultToCurrent` is set, so `editionId`/`editionSlug` can no longer go empty in that form. That happens to fix the sort, but it's a side effect of a change with its own problem. |
| 2 | Duplicate rank `52` in `gangAdditionRankN23.ts` | **Resolved** — `promethium guild`/`slave guild`/`water guild` bumped to `53`/`54`/`55`, sequence is strictly increasing again. |
| 3 | `alliance_type` grouping trusts raw DB text | Not touched by this diff, still an open "please verify the data" note. |

`FighterAddModal.tsx`'s removal of auto-selecting the default/cheapest specialisation (now leaves it unselected until the user picks one) is fine — `confirmDisabled` already requires `selectedSpecialisationId` when `availableSpecialisations.length > 0` (line 1166 at this head), so there's no way to submit with an implicit, unreviewed specialisation/cost now. No issue.

### New finding

**High — `EditionSelect`'s new auto-fill silently reassigns existing "no edition" catalog rows to the current edition, contradicting a documented invariant**

`components/edition-select.tsx:52-58`:
```ts
useEffect(() => {
  if (!defaultToCurrent || value || editions.length === 0) return;
  const current = editions.find(edition => edition.is_current);
  if (current) onChange(current.id);
}, [defaultToCurrent, value, editions, onChange]);
```

The old version only ever auto-filled **once per mount** (`hasDefaulted` ref), so re-selecting an existing record with `edition_id: null` back to a blank `editionId` state left it blank. This new version has no such guard: it fires *every time* `value` becomes empty while `defaultToCurrent` is true — which includes selecting an existing record whose `edition_id` is `null`.

Every admin edit form that loads an existing item does exactly `setEditionId(existing.edition_id ?? '')` on selection, e.g.:
- `admin-edit-equipment.tsx:283` — `setEditionId(equipmentDetails.edition_id ?? '')`
- `admin-edit-skill.tsx:449`
- `admin-edit-vehicle-type.tsx:556`
- `admin-injuries.tsx:113`
- `admin-scenarios-modal.tsx:75`
- `admin-gang-lineage.tsx:170`

For any such record that currently has `edition_id = null`, opening it for edit now immediately (via this effect) calls `onChange(current.id)` and overwrites `editionId` state to the current edition — before the admin has touched anything. Every one of these forms writes `edition_id: editionId || null` (or equivalent) on save, so saving after merely opening the record permanently converts it from "no specific edition" to edition-locked.

This directly contradicts the documented meaning of `is_current`, from `supabase/schema/schema.public.sql:4731`:
> "Picks the default edition for pages that have no gang or campaign context to derive an edition from. It does NOT mark other editions unsupported, and **catalog rows must never be filtered on it**."

Using `is_current` to *assign* an edition onto an existing catalog row is the same violation as filtering on it — it turns a row that was intentionally edition-agnostic (or a legacy pre-edition row, per the "missing slug counts as N23" convention in `types/edition.ts`) into one scoped to whatever edition happens to be flagged current (today N23, per the N26 migration's "not current: n23 keeps is_current until cutover" comment — but this bug will silently start reassigning rows to N26 the moment that flag flips).

This wasn't touched by the N26 ranking work at all — it looks like a fix aimed at a different problem (making "All Editions" unavailable in create/edit forms) that has a much wider blast radius than intended, across essentially every admin catalog editor in the app.

**Suggested fix:** keep the one-time-per-mount guard for the auto-fill (or scope it to only fire when there's no existing record being loaded — e.g. only for genuinely new/create flows), rather than re-firing whenever `value` is cleared by loading a record that has a real, intentional `null` edition.

The N26 ranking work itself remains solid and #2 is cleanly fixed. But the `edition-select.tsx` change should be reverted or reworked before merge — as written it risks silently corrupting the edition scope of any existing "all editions" equipment/skill/vehicle-type/injury/scenario/gang-lineage row the first time an admin opens it, in direct contradiction of the `is_current` column's documented contract.

---

joeseos:
## Code Review — follow-up (commit `ebc4b90` since my last review at `a934866`)

@Djidiouf Reviewed the diff since the last round: `components/edition-select.tsx` and the one related line in `components/admin/admin-edit-equipment.tsx`. Only these two files changed.

### Previous finding — status

**High — `EditionSelect` auto-fill silently reassigning "no edition" catalog rows** → **Resolved, cleanly.**

`components/edition-select.tsx:52-63` restores the `hasDefaulted` per-mount guard, so the current-edition auto-fill only ever fires once per mount rather than every time `value` goes empty. I traced the sequence for the case that was broken:

1. Form mounts with no record selected, `editionId=''` → effect fires once, `hasDefaulted` flips to `true`, fills `editionId` to the current edition (unchanged desired behavior for a fresh create flow).
2. Admin then selects an existing equipment/skill/etc. row with `edition_id: null` → parent does `setEditionId(existing.edition_id ?? '')`, `value` goes back to `''` → effect's guard (`hasDefaulted.current` already `true`) short-circuits, so `onChange` is **not** called again → the null edition is preserved instead of being silently overwritten.
3. `showBlankOption = !defaultToCurrent || !value` now also re-shows "Select edition" whenever `value` is empty in a `defaultToCurrent` form, so that preserved `null` is visibly represented in the dropdown rather than looking like an unintended empty state.

This matches the `is_current` contract in `schema.public.sql:4731` ("catalog rows must never be filtered on it") — the comment added here ("is_current is only a UI default, never an assignment onto stored rows") states exactly that invariant. I don't see a way for the auto-fill to now leak onto an existing row post-fix.

The `admin-edit-equipment.tsx:1639-1641` change — sorting each fighter-type row by its **own** `editionSlugOf(editions, ft.edition_id)` instead of the form's single selected `editionSlug` — is actually a better fix for the original "loses subtype sort under All Editions" finding than what I suggested: it produces a correct per-row subtype rank regardless of whether a specific edition or "All Editions" is selected in the form, rather than only working around the all-editions case. Confirmed `edition_id` exists on the fighter-type rows used here (same shape used elsewhere in the file, e.g. line 411).

### Verdict: **Approve**

Both real findings from the last two rounds (duplicate rank value, and the `is_current` auto-assignment bug) are fixed correctly, and this round's changes are narrowly scoped to exactly those fixes with no new surface area. Nothing else to add — the N26 ranking/category logic itself was already sound in the first round.

---

joeseos:
## Design review — reuse & edition-gating conventions (pushed as `6d64e80`)

@Djidiouf The three earlier rounds covered correctness. This pass looked at the questions those didn't: does the PR follow the established edition-gating convention, does it reuse what already exists, and are the `utils/` splits warranted. Fixes are pushed to this branch — build and `tsc --noEmit` are clean.

### The util file splits are right — keep them

This was the main thing I was asked to check, and the answer is that they're the existing convention, not new invention. Two sibling registries on `main` already do exactly this, down to the `Record<EditionSlug, …>` map and the `getXRank(editionSlug)` accessor that returns `{}` for an unknown slug:

- `utils/skillSetRank.ts` + `skillSetRankN23.ts` + `skillSetRankN26.ts`
- `utils/getEquipmentCategoryRank.ts` + `equipmentCategoryRankN23.ts` + `equipmentCategoryRankN26.ts`

Collapsing the three new ones back into single files would leave them as the only edition-scoped rank tables *not* following the pattern. No change made.

Also checked the two data edits that were folded into what otherwise looked like pure file moves, since a "move" that isn't one is easy to miss in review:

- `fighterSubtypeRankN23` **drops** `beast` and `pet` — **correct**, both are N26-only subtypes in `fighter_subtypes`.
- `gangAdditionRankN23` **adds** `justicar courts, palanite justicar delegation` — **correct**, the row exists for N23 in `alliances`.

Neither is a regression, but both are worth a line in the PR body since they're rules data changes, not refactoring.

### What I changed

**1. Gate on a capability, not a slug literal.** `FighterAddModal` had two `editionSlug === EDITION_N26` comparisons. `types/edition.ts` states the rule directly — *"Callers ask what an edition can do rather than comparing slugs"* — and on `main` the only direct comparisons are in `edition-toggle.tsx` / `use-home-edition.ts`, which are about *picking* an edition and so are legitimately exempt. Added a `gangAdditionCategories` row to `EDITION_CAPABILITIES` with a `hasGangAdditionCategories` predicate. This is the one real gating deviation in the PR; everything else follows the conventions closely. Renamed `isN26CategoryMatch` → `matchesAdditionCategory` and `isN26GangAdditionType` → `isCategorisedGangAddition` to match.

**2. `gangAdditionRankN26` was dead data.** It listed the same four subtypes in the same order as `N26_ADDITION_CATEGORIES` in the component, and was never actually read: every `gangAdditionRank[...]` lookup sits inside `buildSubtypeOptions`, which returns `buildCategoryOptions()` before reaching them, and the `getGangAdditionSortRank` branch only runs for non-N26. Two sources of truth, one of them unreachable — so the next person to add an N26 category would likely update one and not the other. The rank map now derives from the category list via `Object.fromEntries`, the way `equipmentCategoryRankN26` already does. Verified the derived output is byte-identical: `{brute:1, hanger-on:2, pet:3, loner:4}`.

**3. "Lowest rank across every label" was written three times** — both `getXSortRank` accessors plus a local `bestRankedSubtype`. Extracted to `utils/rankLookup.ts` (`bestRank` / `bestRankedLabel`). Worth noting the underlying idea is a genuinely good call: scanning every label rather than trusting `subtypes[0]` is what stops a fighter typed `["Pious", "Prospect"]` sorting as unranked under N26 multi-subtypes.

**4. Alliance grouping was duplicated.** `gang-edit-modal` and `FighterAddModal` each mapped `alliances.alliance_type` onto the same four groups by different means, with divergent labels (`"Other Alliances"` vs `"Other"`), and the latter built a heading string only to regex-strip it back off to recover the sort key. Consolidated into `groupAlliancesByType` in `utils/allianceRank.ts`. Both callers keep their own heading wording, so the rendered dropdowns are unchanged. One small behaviour change: within-group ties now break by name in the gang edit modal too, matching what the additions modal already did.

This also closes finding #3 from your first round ("worth a quick sanity check against the live `alliances` table"). I checked: stored values are exactly `Criminal`, `Merchant Guilds`, `Noble Houses`, `Others`, all handled, and `Archeo Hunters` is the only `Others` row. One validated mapping now replaces two ad-hoc ones.

**5. `fighterSubtypeRankN26` was missing three subtypes** that exist in N26 data — `hanger-on`, `beast` and `loner`. `beast` was in the old flat table at rank 11 and was dropped rather than carried across. Effect: those fighters ranked `Infinity` and sorted last in the roster composition tooltip (`gang.tsx`) and the campaign OOA/deaths breakdowns (`campaign-tab.tsx`). Appended after the existing entries as 9/10/11 so nothing that ranks today moves — **the exact placement is a rules call, so please sanity-check it.** Note `fighterSubtypeRankN26` and `gangAdditionRankN26` order these differently (`pet` before `brute` vs the reverse); they drive different UIs so it may well be intentional, but flagging in case it isn't.

**6. Dropped six re-exports** with no importers, copied from the same wart in `skillSetRank.ts`. Left `skillSetRank.ts` alone as out of scope.

### Not changed

- `edition-select.tsx` and the specialisation auto-select removal. Both are correct as they stand, but they're unrelated to this PR's stated scope — a ranking PR that also changes every admin catalog editor's edition defaulting is a wider blast radius than the title suggests, and it's what produced the High finding in round two. Worth splitting out next time rather than reworking now.
- `editionSlugOf` inside the `admin-edit-equipment` sort comparator does a linear `find` per comparison. With two editions it isn't worth a lookup map.

### Verification

`npx tsc --noEmit` and `npm run build` both pass; `npm run lint` flags nothing in any touched file (the 6 pre-existing errors are all in files neither of us modified). Behaviour-preservation of #2 confirmed against the original hand-written map. Worth an eyeball on the preview for the N26 additions dropdown, the N23 additions dropdown, and the N23 alliance dropdown (including `Archeo Hunters` under "Other Alliances") — all three should render exactly as before.

---

joeseos:
## Code Review — follow-up (commit `6d64e80` since my last review at `ebc4b90`)

@Djidiouf (cc @joeseos, this commit is Claude-authored per the trailer) Reviewed the refactor commit: `types/edition.ts`, the new `utils/rankLookup.ts`, `utils/allianceRank.ts`, `utils/fighterSubtypeRank(N26).ts`, `utils/gangAdditionRank(N26).ts`, and the corresponding call-site updates in `FighterAddModal.tsx` and `gang-edit-modal.tsx`. This is pure design cleanup on top of already-approved logic, so I focused on verifying it's actually behavior-preserving where it claims to be, and flagging the one disclosed exception.

### Verified equivalent (no regression)
- `gangAdditionCategories` capability replacing `editionSlug === EDITION_N26` in `FighterAddModal.tsx` — every call site (`usesCategories`/`isCategoryAdditions`) is a mechanical rename, and this actually resolves something I'd have flagged had it persisted: the codebase's own convention (stated in `EDITION_CAPABILITIES`'s docstring, and something I praised in the first round of this PR) is to ask "what can this edition do" rather than compare slugs directly. Good catch bringing this in line with that.
- `gangAdditionRankN26` deriving from `N26_ADDITION_CATEGORIES` instead of a hand-written map — traced the derivation (`brute→1, hanger-on→2, pet→3, loner→4`) and it's byte-for-byte identical to the old hardcoded values. The commit message's claim that the old map was dead code checks out too: `buildSubtypeOptions` returns `buildCategoryOptions()`'s result before ever reaching the non-category branch that reads the rank map, for N26.
- `bestRank`/`bestRankedLabel` extraction into `utils/rankLookup.ts` — diffed against both inlined originals (`getFighterSubtypeSortRank`'s loop and `FighterAddModal`'s local `bestRankedSubtype`); same normalise-and-scan logic, same `Infinity`/first-label fallback behavior.
- `groupAlliancesByType` extraction — traced both call sites against their prior inline versions. `gang-edit-modal.tsx`'s grouping (criminal/merchant guilds/noble houses/other, in that order) and `FighterAddModal.tsx`'s grouping produce the same groups from the same `alliance_type` strings; each caller still words its own heading (`"Other Alliances"` vs `"Alliances: Other"`).
- Checked for dangling references after the rename/removal pass (`isN26CategoryMatch`, `allianceTypeHeader`, `N26_ALLIANCE_TYPE_ORDER`, the `EDITION_N26` import, etc.) — none left; nothing should fail to compile.

### Disclosed, intentional behavior change (not a bug)
- `gang-edit-modal.tsx`'s alliance dropdown now breaks same-rank/unranked ties by name via `groupAlliancesByType`, where it previously preserved DB return order. The commit message calls this out explicitly and it brings the gang-edit modal in line with what the gang-additions modal already did — a reasonable, cosmetic-only change (affects ordering of unranked/tied alliances only).
- `fighterSubtypeRankN26.ts` gains `hanger-on: 9`, `beast: 10`, `loner: 11`, appended after the existing 1–8, so nothing that was already ranked moves. This fixes a real gap (those three N26 subtypes existed in the DB but sorted last/undefined in the roster tooltip and campaign OOA/deaths breakdowns) without touching prior rankings.

### Still open from earlier rounds
- Finding #3 from my first review (verify `alliances.alliance_type` actually contains the exact strings this code assumes — `"criminal"`, `"merchant guilds"`, `"noble houses"`, `"other"/"others"`) is now centralized in `allianceTypeGroup()` but not resolved — same assumption, just in one place instead of three. Still just a "please confirm against the live table" note, not a blocker.

### Verdict: **Approve**

This commit doesn't touch the underlying N26 ranking/gating logic that was already reviewed and approved — it's a same-behavior consolidation, and everywhere I checked it produces identical output to what it replaced. Nothing new to block on.

---

PR #1960, opened 2026-08-09
Split vehicles out of the fighter catalog, gate Add Vehicle on edition

joeseos:
N23 models a vehicle as a vehicle_types row assigned to a Crew fighter; N26 makes the vehicle a fighter (fighter_types.is_vehicle). Until now nothing in the player-facing gang flow read is_vehicle, so N26 vehicle fighter types would show up in Add Fighter, and an N26 gang got the N23 vehicle_types modal.

Add Vehicle now branches on the existing `vehicles` capability: editions whose vehicles are fighter types open FighterAddModal with a new 'vehicles' catalog, everything else keeps add-vehicle.tsx. Add Fighter, Gang Additions and "include all types" exclude vehicle fighter types.

The vehicles catalog also offers the edition's "Available to All" gang type, which is how fighter_types spells what vehicle_types expresses as a null gang_type_id.

get_fighter_types_with_cost gains an is_vehicle column and, per repo convention, is applied to the remote by hand rather than via a migration.


joeseos:
## Code Review

@joeseos Reviewed the diff (`is_vehicle` gating in `add-fighter.ts` and the `/api/fighter-types` route, the new `vehicles` catalog in `FighterAddModal`, the edition-gated branch in `gang.tsx`) against the wider codebase, including the read paths that weren't touched by this diff but are affected by it.

### What's good
- `filterByIsVehicle`'s null-means-unfiltered contract, applied consistently at the very end of every response-building branch in `route.ts` (after the "Available to All" merge), is a clean way to retrofit vehicle-awareness onto four different existing code paths (`customGangTypeId`, `includeAllTypes`, `isGangAddition`, plain roster) without duplicating the filter logic in each.
- The "Available to All" resolution (`getGangEditionId` → look up the edition's `Available to All` gang type → re-run `get_fighter_types_with_cost` scoped to it → `mergeById`) correctly reproduces what `vehicle_types.gang_type_id IS NULL` meant in N23, given `fighter_types.gang_type_id` is `NOT NULL`. Good comment explaining the mapping.
- `hasVehicles()` defaults to `false` for a missing/unrecognized `edition_slug` (`types/edition.ts`'s `can()` helper), so `gang.tsx`'s new branch safely falls back to the legacy `AddVehicle`/`vehicle_types` component for N23 and any gang with an unresolved edition — no risk of an N23 gang hitting the new fighter-type-based flow by accident.
- Reusing `FighterAddModal` for vehicles (catalog `'vehicles'`) rather than forking it, with the `noun` swap driving all the copy, is a good reuse of the existing roster-add flow rather than a parallel implementation.
- Confirmed `fighter_types.is_vehicle`, `custom_fighter_types.is_vehicle`, and `fighters.is_vehicle` all already exist in `schema.public.sql` (added ahead of this PR), so the "no migration" note in the description checks out — this PR is purely about *reading and gating on* the column, not introducing it.

### Findings

**1. Medium/High — `copy-fighter.ts` silently drops `is_vehicle` when duplicating a fighter**
`app/actions/copy-fighter.ts`: `sourceFighter` is fetched with `select('*', ...)` (line ~90), so `sourceFighter.is_vehicle` is available, but the `fighterData` object built for the INSERT (lines ~253-298) copies over `movement`/`weapon_skill`/`fighter_subtypes`/etc. field-by-field and never includes `is_vehicle`. Since `fighters.is_vehicle` defaults to `false NOT NULL`, using "Copy Fighter" (or "copy as experienced") on an N26 vehicle produces a copy that is no longer flagged as a vehicle — it silently becomes a regular fighter from the app's perspective, presumably diverging from wherever the roster splits vehicles out from fighters for display.

For contrast, `app/actions/copy-gang.ts`'s fighter duplication (`insertObj = { ...f }`, full row spread) is unaffected — only the single-fighter copy path has this gap, because it's the one that copies fields individually. Suggest adding `is_vehicle: sourceFighter.is_vehicle ?? false` alongside the other stat fields.

**2. Medium — `EditFighterModal`'s "change fighter type" dropdown isn't excluded from vehicle fighter types**
`components/fighter/edit-fighter/fighter-edit-modal.tsx:110-121` fetches `/api/fighter-types?gang_id=...&is_gang_addition=false...` with no `is_vehicle` param. Per `filterByIsVehicle`'s own contract (`route.ts`: "null means no filter, which is what callers outside the gang add-modals want"), that means this call is now unfiltered and will include N26 vehicle fighter types. The fetched list feeds directly into a real type-swap UI (`selectedFighterTypeId` / the `fighter_type_id` Combobox around line 1017-1034 that lets a user re-type an existing fighter) — so a normal Ganger can now be retyped into a Vehicle fighter type via Edit Fighter, which is exactly the class of leak ("N26 vehicle fighter types would show up in Add Fighter") this PR sets out to close for the Add Fighter/Gang Additions modals, just via a call site this diff didn't touch. Worth adding `is_vehicle=false` (or filtering the fetched list client-side) here the same way `FighterAddModal` does.

**3. Low/lower-confidence, same shape as #2** — `components/fighter/fighter-advancement-list.tsx` has two more `/api/fighter-types` fetches with the same `is_gang_addition: 'false'` params and no `is_vehicle` filter (used for the Ganger/Champion promotion flow). I'd guess this is low-risk in practice since promotion matching is keyed off specific subtypes (`Ganger`, `Juve`, `Champion`, `Specialist`, `Exotic Beast`, …) that a vehicle fighter type wouldn't carry, so vehicle rows likely just sit unused in the fetched list rather than becoming selectable — but worth a quick look alongside #2 while you're in that area, since it's the same endpoint call missing the same param.

The core gating (Add Fighter / Gang Additions / include-all-types excluding vehicles, and the edition-gated Add Vehicle branch) is solid and well-tested per the PR description. I'd want #1 fixed before merge since it's a straightforward, silent data-loss bug on an existing, commonly-used action, and #2 addressed or explicitly scoped out — right now it reopens the exact leak this PR closes elsewhere, just one call site over.

---

joeseos:
## Code Review

@joeseos Reviewed against the codebase, including `app/api/fighter-types/route.ts`, the `FighterAddModal`/`gang.tsx` wiring, the `get_fighter_types_with_cost` SQL change, and the `EDITION_CAPABILITIES` addition.

### What's good
- `hasVehicles(edition_slug)` branching in `gang.tsx` between the new `FighterAddModal catalog="vehicles"` and the legacy `AddVehicle` modal is a clean, minimal seam — no changes to the N23 path at all.
- `filterByIsVehicle`'s tri-state (`null` = unfiltered, `'true'`/`'false'` = filtered) correctly defaults to "no filter" for existing callers that don't know about vehicles yet, rather than silently changing their result set.
- `getAvailableToAllFighterTypes` correctly models "N23's `vehicle_types.gang_type_id IS NULL`" as "N26's fighter_types row under the edition's real 'Available to All' gang type" rather than inventing a new sentinel — and resolving to `[]` when there's no such gang type or RPC error is the right fail-open for an additive merge.
- The SQL function already used the `DROP FUNCTION IF EXISTS (uuid, uuid, boolean)` + `CREATE OR REPLACE` idiom, so adding `is_vehicle` to `RETURNS TABLE` won't hit Postgres's "cannot change return type" restriction when applied by hand, as the PR description says it will be.
- `custom_fighter_types`/`fighter_types`/`fighters` already have `is_vehicle boolean DEFAULT false NOT NULL` from an earlier migration, so this PR is purely wiring, not schema risk.
- The `FighterAddModal` copy/label swap to `noun` (`'Vehicle' | 'Fighter'`) is thorough — I checked for leftover hardcoded "Fighter" strings and the only ones left are correctly gated behind `isAdditions` (the Fighter Subtype block), which vehicles never render.

### Issues

**1. Medium — Vehicle fighter types aren't excluded from the promotion and "change fighter type" pickers**

The PR description scopes the exclusion to "Add Fighter, Gang Additions and 'include all types'", and those three do send `is_vehicle=false`/`true` correctly. But two other consumers of the same `/api/fighter-types` endpoint call it with no `is_vehicle` param at all, so `filterByIsVehicle` leaves them unfiltered:

- `components/fighter/edit-fighter/fighter-edit-modal.tsx:109-125` — the "change this fighter's type" list (`fetchedFighterTypes`)
- `components/fighter/fighter-advancement-list.tsx:687-703` — the Ganger/Champion promotion type list (`preFetchedFighterTypes`, gated on `gangerPromotionOpen || championPromotionOpen`)

Both hit the exact same `is_gang_addition=false` roster query as `FighterAddModal`'s roster catalog, just without the new `is_vehicle` param. Once an N26 gang type has real vehicle fighter types in its catalog, a Ganger could see a vehicle offered as a promotion or type-change target through either of these modals. Worth adding `&is_vehicle=false` to both queries (or a shared constant/helper for the roster-fetch query string, since it's now duplicated across three call sites with slightly different params).

**2. Low/nit — `is_vehicle` doesn't round-trip through the optimistic add response**

`app/actions/add-fighter.ts:470` sets `is_vehicle` on the DB insert correctly, but the `AddFighterServerData` returned at the end of `addFighterToGang` (~line 1195-1251) doesn't include it. Right now this is harmless — I checked, and nothing in the gang-roster `fighter-card.tsx` reads `is_vehicle` (only `fighter-details-card.tsx`, which gets it from a fresh server fetch on the fighter's own page, not from this optimistic payload). But it's a latent inconsistency: if the roster card or optimistic-update path ever needs to distinguish a freshly-added vehicle before the next refetch, this field silently won't be there. Cheap to add now while the shape is fresh in mind.

**3. Note, not a defect — merge conflict with #1959**

This PR and the still-open #1959 both edit `gang.tsx` and `FighterAddModal.tsx` from the same base commit (both branched off `8d187c8d`), and GitHub already reports `mergeable_state: blocked`. Whichever lands second will need to reconcile `FighterAddModal`'s prop list (`editionSlug` from #1959 vs. the new `'vehicles'` catalog value here) and `gang.tsx`'s modal JSX. Not something to fix in this PR, just flagging so it's not a surprise.

### Verdict: **Changes requested** (minor)

The core split — routing N26's fighter-typed vehicles through `FighterAddModal` while N23 keeps `AddVehicle`, plus the `is_vehicle` filter/merge machinery in the API route — is sound and the edge cases I checked (custom gang types, gang variants, hidden gang types, "Available to All") all compose correctly. I'd want #1 fixed before this ships, since it's a direct gap in the PR's own stated goal of excluding vehicles from fighter-facing pickers; #2 is optional polish.

---

PR #1961, opened 2026-08-09
Fixed tsconfig

joeseos:


joeseos:
## Code Review

@joeseos Reviewed the diff (2 lines changed in `tsconfig.json`) against the rest of the config surface.

### Changes
- `target: "es5"` → `"ES2017"` — brings the compile target in line with the `esnext` lib already declared and with what Next.js's bundler actually ships to modern browsers; `es5` was the stale default from `create-next-app` and had no other config depending on it.
- Dropped `baseUrl: "."` — checked `typescript` is `^5.9.3` in `package.json`. Since TS 4.1, `paths` resolves relative to the tsconfig file's own directory when `baseUrl` is omitted, so `@/*` → `./*` behaves identically. Confirmed there's no `jest.config`/`vitest.config` or other tool in the repo reading `compilerOptions.baseUrl` independently (only hits for "baseUrl" elsewhere are `supabaseUrl` variable names, unrelated).

No functional risk either way — this only affects compile-time output/resolution, `noEmit: true` is set, and `moduleResolution: "bundler"` (already in place) doesn't consult `baseUrl` either.

### Verdict: **Approve**

---

PR #1962, opened 2026-08-09
Pass editionSlug to the Add Vehicle catalog

joeseos:
The vehicles catalog was the one FighterAddModal call site that never received
editionSlug. #1959 added the prop along with an edition filter in the modal's
queryFn, and wired it to the two instances that existed on that branch; the
third arrived from #1960 in the same merge and was missed.

editionSlug defaults to null, and sameEditionForDisplay reads a missing slug as
n23, so every n26 vehicle fighter type was discarded client-side and Add
Vehicle rendered an empty list on every n26 gang.

### DIFF
--- a/components/gang/gang.tsx
+++ b/components/gang/gang.tsx
@@ -1317,6 +1317,7 @@ export default function Gang({
                 gangId={id}
                 gangTypeId={gang_type_id}
                 customGangTypeId={custom_gang_type_id}
+                editionSlug={edition_slug}
                 initialCredits={credits}
                 onFighterAdded={handleFighterAdded}
                 onFighterRollback={onFighterRollback}


joeseos:
## Code Review

@joeseos Reviewed the diff (1 line, `components/gang/gang.tsx`) against the surrounding `FighterAddModal` call sites and the edition-filtering logic it feeds into.

### Analysis
- Confirmed the root cause described in the PR body: `FighterAddModal`'s `editionSlug` prop defaults to `null` (`FighterAddModal.tsx:149`), and the catalog filter uses `sameEditionForDisplay(type.edition_slug, editionSlug)` (`FighterAddModal.tsx:216`). `sameEditionForDisplay` treats a missing/null slug as `EDITION_N23` (`types/edition.ts:17`: `(a ?? EDITION_N23) === (b ?? EDITION_N23)`). So with `editionSlug` unset, every N26 vehicle fighter type (`edition_slug: 'n26'`) mismatches against the implied `'n23'` and is filtered out — matches the PR's description exactly.
- Verified all three `FighterAddModal` call sites in `gang.tsx` (`catalog="roster"` at line 1271, `catalog="vehicles"` at line 1320 after this fix, `catalog="additions"` at line 1352) now consistently pass `editionSlug={edition_slug}`. No other call site in the codebase renders `FighterAddModal`, so this closes the gap completely rather than leaving a fourth instance missed.
- `edition_slug` is already a destructured prop on `Gang` (`gang.tsx:119`, typed `string | null | undefined`), so this is a pure wiring fix — no new state, no type changes, no risk of a stale/undefined value being introduced.

### Verdict: **Approve**

Minimal, correctly-scoped fix for a real regression. No edge cases or follow-up concerns — the other two call sites already prove the pattern works.

---

PR #1963, opened 2026-08-09
Add N26 Sv and multi-subtype editing to Edit Fighter.

Djidiouf:
Show edition-gated Sv in Characteristics and Adjust Characteristics, switch Fighter Type/Specialisation to Combobox, and support N26 multi-subtype via add/chips (N23 stays single-select), with dirty-only subtype submits and optimistic/server refresh so overrides are not clobbered. Archetype selection/eligibility is left unchanged on purpose; it still keys off fighter class (subtype) rather than fighter type and needs a separate rework for multi-subtype N26.

### DIFF
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -142,31 +142,41 @@ function normalizeFighterSubtypeNames(subtypes: string[]): string[] {
 /**
  * Edition for subtype validation: prefer the type being written in this update,
  * else the fighter's current type, else the gang's edition.
+ * Fail closed — query errors or a missing slug reject rather than skipping checks.
  */
 async function resolveFighterEditionSlugForUpdate(
   supabase: any,
   fighterId: string,
   params: Pick<UpdateFighterDetailsParams, 'fighter_type_id' | 'custom_fighter_type_id'>
-): Promise<string | null> {
+): Promise<{ ok: true; editionSlug: string } | { ok: false; error: string }> {
+  const fail = {
+    ok: false as const,
+    error: 'Could not resolve edition for fighter subtype validation.',
+  };
+
   if (params.fighter_type_id) {
-    const { data } = await supabase
+    const { data, error } = await supabase
       .from('fighter_types')
       .select('editions:edition_id ( slug )')
       .eq('id', params.fighter_type_id)
       .single();
-    return editionSlugFromJoin(data?.editions);
+    if (error || !data) return fail;
+    const editionSlug = editionSlugFromJoin(data.editions);
+    return editionSlug ? { ok: true, editionSlug } : fail;
   }
 
   if (params.custom_fighter_type_id) {
-    const { data } = await supabase
+    const { data, error } = await supabase
       .from('custom_fighter_types')
       .select('editions:edition_id ( slug )')
       .eq('id', params.custom_fighter_type_id)
       .single();
-    return editionSlugFromJoin(data?.editions);
+    if (error || !data) return fail;
+    const editionSlug = editionSlugFromJoin(data.editions);
+    return editionSlug ? { ok: true, editionSlug } : fail;
   }
 
-  const { data } = await supabase
+  const { data, error } = await supabase
     .from('fighters')
     .select(`
       fighter_types:fighter_type_id ( editions:edition_id ( slug ) ),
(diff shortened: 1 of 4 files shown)


joeseos:
## Automated review

@Djidiouf Reviewed the diff against the full codebase (`app/actions/edit-fighter.ts`, `components/fighter/edit-fighter/{character-stats-modal,fighter-characteristic-table,fighter-edit-modal}.tsx`), plus `types/edition.ts`, `types/fighter.ts`, `utils/fighterSubtypeRank.ts`, and the `fighters.fighter_subtypes` schema/migrations for context.

### Findings

**1. No server-side validation of `fighter_subtypes` cardinality/content — Medium**
`app/actions/edit-fighter.ts:1333`
```ts
if (params.fighter_subtypes !== undefined) updateData.fighter_subtypes = params.fighter_subtypes;
```
`updateFighterDetails` writes `params.fighter_subtypes` straight to the DB with no check that:
- an N23 fighter (`allowsMultipleSubtypes(edition_slug) === false`) still gets at most one entry — that rule is enforced only client-side (`allowMultipleSubtypes` gating in `fighter-edit-modal.tsx`);
- the names are non-duplicate and actually exist in `fighter_subtypes` for the fighter's edition.

The DB column is also unconstrained (`fighter_subtypes jsonb DEFAULT '[]'::jsonb NOT NULL`, no length/edition check). Before this PR the client could only ever construct a single-element array (`[selectedSubtype.subtype_name]`), so a >1-length array was structurally unreachable regardless of caller. Now that the client can legitimately send multi-element arrays (for N26), the server action accepts an arbitrary-length array for *any* fighter id, including N23 ones — a stale/buggy client or any direct caller of the action can persist an N23 fighter with 2+ subtypes, which the app doesn't otherwise expect.

Suggested fix: in `updateFighterDetails`, when `params.fighter_subtypes` is provided, look up the fighter's `edition_slug` server-side and reject (or truncate) arrays with >1 entry when `allowsMultipleSubtypes` is false, and dedupe the array before writing.

**2. `confirmDisabled` only guards the "become empty" case — Low**
`fighter-edit-modal.tsx:1423-1426`
```ts
confirmDisabled={
  !formValues.name.trim() ||
  (selectedFighterSubtypes.length === 0 && (fighter.fighter_subtypes?.length ?? 0) > 0)
}
```
This is fine for the common case but is a client-only guard with no matching server check (see #1) — if #1 is fixed with proper server validation this becomes pure UX and is not a blocker on its own.

**3. Minor: `Stat.value` field in `character-stats-modal.tsx` is dead** (pre-existing, not introduced here) — `displayStats` still computes a `value` string per stat (including the new `Sv` entry) that's never read; rendering goes through `getAdjustedTotal`/`getBaseDisplay` instead. Not worth blocking on, just noting since the PR touched this exact array.

### What looks good
- `hasSaveCharacteristic` / `allowsMultipleSubtypes` edition-capability plumbing is clean and follows the existing `EDITION_CAPABILITIES` pattern.
- Null-handling for `Sv` (`getBaseValue`/`getCurrentTotal`/`getAdjustedTotal`/`formatStatValue`) is consistent across both the table and the adjust-characteristics modal, including the "unset stays unset until a modifier/adjustment appears" rule.
- `sameFighterSubtypes` + dirty-only submission is a good fix for the "clobber a prior override" problem described in the PR body.
- The `fighter-type`/`specialisation` `<select>` → `Combobox` swap is a faithful 1:1 port of the existing grouping/sorting logic (verified line-by-line against the removed code).
- Archetype eligibility/fetch correctly keys off `selectedFighterSubtypes[0]` and is explicitly scoped out of the multi-subtype rework, as called out in the PR description.

### Verdict
**Changes requested** — primarily #1 (server-side validation gap for `fighter_subtypes`). #2 and #3 are non-blocking notes.

---

joeseos:
## Follow-up review (commit 284168f)

@Djidiouf Reviewed the diff between `0f2dd1a` and `284168f` (`app/actions/edit-fighter.ts` only — the component files are unchanged since the last review).

### Previous blocker: resolved
Finding #1 from the earlier review (no server-side validation of `fighter_subtypes`) is fixed with `validateFighterSubtypesForUpdate` + `resolveFighterEditionSlugForUpdate`:
- Normalizes (trim/dedupe) before validating — correct, so a client sending duplicate names doesn't false-trip the cardinality check.
- Enforces single-subtype for editions where `allowsMultipleSubtypes` is false.
- Validates each name against `fighter_subtypes` for the resolved edition.
- Edition resolution priority (incoming `fighter_type_id` → incoming `custom_fighter_type_id` → fighter's current type → gang's type) correctly matches how the client actually builds `submitData`, including the custom-fighter-type branch (`fighter_type_id: null` / `custom_fighter_type_id: set`). The `gang_types!gang_type_id` / `custom_gang_types!custom_gang_type_id` embed follows the same convention already used in `update-gang.ts`, `battle-sessions.ts`, `assert-gang-campaign-edition.ts`, etc., and the alias keys line up with what `gangEditionJoin` expects.

Good fix, well-integrated with the existing `types/edition.ts` helpers.

### Two residual gaps worth a look

**1. Edition resolution fails open, inconsistently with the rest of the function — Low/Medium**
`resolveFighterEditionSlugForUpdate` (all three branches) never checks `error` from its Supabase calls — only `data` is read, so any query error (RLS hiccup, bad id, etc.) silently resolves to `editionSlug = null`. Both cardinality and catalog checks in `validateFighterSubtypesForUpdate` are gated on `editionSlug != null`, so a failed lookup **skips validation entirely** rather than rejecting. This is the opposite of the fail-closed behavior used two lines later for the `editions` and `fighter_subtypes` catalog lookups (`if (editionError || !edition) return { ok: false, ... }`, `if (catalogError) return { ok: false, ... }`). Since the third branch (fighter's current type) is the one hit on every subtype-only edit — the PR's main new use case — a transient failure there silently disables the guard for the most common call path. Suggest treating a failed/empty edition-resolution query the same as the other two: return `{ ok: false, error: ... }` rather than falling through with `editionSlug = null`.

**2. Empty `fighter_subtypes: []` bypasses validation entirely — Low/Medium**
Both checks are gated on `normalized.length > 1` / `normalized.length > 0`, so `normalized = []` always returns `{ ok: true, subtypes: [] }` with no edition/catalog lookup at all. The UI's `confirmDisabled` stops this from happening through the modal (blocks confirming when subtypes would go from non-empty to empty), but that's a client-only guard — a direct call to `updateFighterDetails` with `fighter_subtypes: []` will now cleanly wipe any fighter's subtypes, no matter the edition. Given several SQL functions branch on `fighter_subtypes ?| array[...]` (XP costs, injury targeting, etc.), an emptied array quietly changes a fighter's rules eligibility. Suggest rejecting a transition from non-empty → empty at the server too (mirroring `confirmDisabled`), rather than only special-casing `length > 0`/`length > 1`.

Neither of these is reachable through the normal Edit Fighter UI — both require calling the server action directly with a crafted payload — so I'd call them defense-in-depth follow-ups rather than hard blockers, but they're small fixes and directly touch the code this PR just added for exactly this purpose.

### Verdict
**Leaning approve.** The original blocking issue is fixed. The two points above are worth a fast follow (or a couple more lines in this same PR) but I wouldn't hold the merge on them if the team is fine accepting that residual risk for now.

---

joeseos (approved):
## Follow-up review (commit fdb6a9b)

@Djidiouf Reviewed the diff between `284168f` and `fdb6a9b` — just `app/actions/edit-fighter.ts`, scoped exactly to the two residual gaps from the last round.

### Gap A — resolved
`resolveFighterEditionSlugForUpdate` (lines ~157-200) now destructures `error` in all three branches and returns failure on `error || !data`, plus treats a resolved-but-null slug as failure too. Return type is now a discriminated `{ok:true,editionSlug} | {ok:false,error}`, and `validateFighterSubtypesForUpdate` propagates `{ok:false}` immediately. This now fails closed, consistent with the `editions`/`fighter_subtypes` catalog lookups two lines later.

### Gap B — resolved
A new guard rejects `normalized.length === 0 && previous.length > 0` ("At least one fighter subtype is required."), fed by a new `currentSubtypes` param that `updateFighterDetails` populates from the fighter's existing `fighter_subtypes` (itself run through `normalizeFighterSubtypeNames`, so it's robust to messy legacy data). This closes the direct-call wipe vector server-side, mirroring the client's `confirmDisabled` guard as the added comment notes.

### No regressions
Dirty-only submission gate, edition-resolution priority order (explicit type → custom type → fighter's current type/gang), and normalize-then-validate ordering are all unchanged. Diff doesn't touch anything beyond the two functions plus one added `fighter_subtypes` column in an existing select — no scope creep.

### Minor, non-blocking
`validateFighterSubtypesForUpdate` now runs the edition-resolution lookup even for a no-op subtype submission (`normalized.length === 0 && previous.length === 0`), which could fail an otherwise-unrelated update if that lookup errors. Low impact — the client only includes `fighter_subtypes` in the payload when it's actually dirty — and it's the correct trade-off given the fail-closed design here, so not asking for a change.

### Verdict: **Approve**

Both previously-flagged gaps are correctly fixed with no scope creep and no regressions to what round 2 confirmed as working.

---

PR #1964, opened 2026-08-09
Earn N26 Advancements by rank instead of spending XP

joeseos:
## Why

N26 replaces N23's "spend XP to buy an Advancement" economy with a rank ladder: XP only ever accumulates, and a model earns an Advancement each time its total crosses a tier boundary. Nothing in the app knew this — `fighter-advancement.ts` had no edition awareness at all, and `get_fighter_available_advancements` returned every row in the `advancements` category regardless of edition.

This is a fresh implementation from `main`, written after reviewing #1947 (which is left alone). Three things that review settled and this branch assumes:

- **`starting_xp = 0` is a legal value.** N23 does not use the concept and some N26 fighter types genuinely start on 0. So there is no `DEFAULT 1` migration here, and the admin fighter-type forms and API route need **no changes at all** — they already do the right thing.
- **The Level Up badge appears on the gang roster too.** `getGangFightersList` is pure query-builder, built to avoid per-fighter round trips, so it cannot call a per-fighter RPC. The tier table therefore lives in TS.
- **Edition gating goes through the capability registry**, as reshaped by #1949.

## What changed

**`fighters.starting_xp`** (`numeric NOT NULL DEFAULT 0`) snapshots the XP a fighter was recruited with. `(custom_)fighter_types.starting_xp` are editable, so reading recruitment XP from the type would retroactively change how many Advancements every existing fighter of that type is owed. Set on all four recruitment paths — `add-fighter`, both `copy-fighter` branches, and `exotic-beasts` (whose `fighter_types` embed did not select it either). `copy-gang` needs nothing, it spreads `select('*')`.

The migration comment also records which of the three XP columns is authoritative: `starting_xp` is, `total_xp` is not — `updateFighterXp` and `updateFighterXpWithOoa` both return `total_xp` set to the fighter's current `xp`.

**N26 Advancement catalog** — 13 `fighter_effect_types` rows with a modifier each, edition-scoped alongside their N23 namesakes. No XP cost; credits straight off the N26 table (+5 psychology, +10 Initiative/Movement, +15 WS/BS, +20 S/T/W/A/Sv), flat with no Ganger special case. Save is N26-only.

Modifier signs follow the direction each characteristic improves in — target numbers down, raw values up. **Initiative and the four psychology characteristics are raw values in N26 but target numbers in N23**, so their signs are the opposite of the N23 rows of the same name, the same inversion the N26 injury seed documents.

**`get_fighter_available_advancements`** now resolves the fighter's edition from its gang the way `add_fighter_injury` does and filters on it. Without this the seed would have offered N26 rows to N23 fighters — the regression that matters most here. N26 also forces `xp_cost` 0, flat credits, `uses_flat_cost` false and no Ganger-to-Specialist row.

**`cumulativeXp` capability** — one new row in `EDITION_CAPABILITIES`. `utils/advancementRanks.ts` is deliberately **not** edition-keyed: the registry already answers that question and a second `Record<EditionSlug, …>` would be the same decision declared twice with nothing keeping them in step.

The rank table omits the Rookie band's own first tier (1–3), because nothing ever *moves onto* it. That is what makes `starting_xp = 0` work: 0 and 1 are provably indistinguishable at every XP total.

| Starting XP | Next Advancement at |
|---|---|
| 0 | 4 |
| 1 | 4 |
| 3 | 4 |
| 4 | 7 |
| 13 (Ganger) | 19 |

The rulebook's own example Ganger profile reads `13/19` for Current/Target XP, which matches.

**Actions** — the three that move a fighter's XP load its edition and gate all five sites: both affordability guards, both deductions and the refund on delete. N26 gates on the open-advancement count instead, the same figure the badge shows, so the two cannot disagree. Deleting an N26 Advancement gives the Level Up back for free, since the taken count drops.

This also gates `isGangerOrExoticBeastSubtype`, which drives the **ganger purchase path** and not just the roll UI — an N26 Ganger would otherwise have been charged N23's flat 6 XP.

**UI** — `LevelUpBadge` on the fighter page and gang roster, derived on read so nothing can drift and entering XP needs no extra write. `N26AdvancementPanel` owns the 2D6 roll and table rather than threading more ternaries through a 3,100-line component.

## Please look at this one separately

**The roster edition source changed, and it reaches beyond the badge.** `getGangFightersList` set `edition_slug` from `fighter_types`, which is `null` for custom fighter types. It now resolves once from the gang. That value also drives `hasSaveCharacteristic` (the Sv column) and three `WeaponTable editionSlug` props — the N26 lethality statline — on every roster card. Gangs using custom fighter types will render differently, correctly, after this.

## Also in here

A preparatory commit extracts `hooks/use-roll-logger.ts`, replacing five copies of the same mutation skeleton and three hand-rolled cooldowns across the advancement, injury and vehicle-damage lists. It fixes a latent double-click bug on the way — the originals compared the cooldown against state, so two clicks in the same tick both saw it as `false` and both submitted.

## Out of scope

Batch selection of several Advancements at once, maximum-characteristic enforcement, N26 skill advancements (blocked upstream — there are no N26 `skill_types` yet), vehicles, rank titles, and the Level Up badge in printed rosters.

## Testing

- `npm run build` and `npm run lint` clean. The 9 remaining lint problems are all pre-existing on `main`, none in a file this branch touches.
- Both migrations applied to a throwaway Postgres 16 and **re-applied**: the seed stays at 13 rows and 13 modifiers, and the column add is a no-op the second time.
- `get_fighter_available_advancements` exercised in rolled-back transactions against three fighters. **The N23 Ganger sees only its N23 row, flat cost 6, Specialist row unchanged** — the regression that matters most. The N26 Ganger sees 13 characteristics, every `xp_cost` 0, credits matching the table, `uses_flat_cost` false and no Specialist row. A gang with only a `custom_gang_type_id` resolves to N26 correctly, exercising the `COALESCE` branch.
- 18 rank-math checks against the real module: every worked example above, the 20-Advancement ceiling at 229, `starting_xp` 0 matching 1 across 0–400, and zeros for `n23` / `null` / unknown slugs.

> [!NOTE]
> Not exercised in a browser — local sign-in is unavailable. The PostgREST embed added to `fighter-advancement.ts` is verified only by its FK chain, which does check out: `fighters.gang_id → gangs`, `gangs.gang_type_id → gang_types`, `gangs.custom_gang_type_id → custom_gang_types`, both to `editions`. Note `fighters` has two FKs to `gangs`, which is why the `!gang_id` disambiguation hint is required. It sits on the critical path of all three advancement actions for **both** editions, so a bad embed would break N23 too — worth a click-through before merge.

## Worth checking on staging

- **Whether `fighter_types.starting_xp` is populated for the N26 catalog.** Types that should start higher — Gangers on 13, etc. — need that data entered, or they will advance faster than intended. This is catalog data entry, not a defect: 0 is a perfectly valid value for the types that should have it.
- The roster of a gang using **custom fighter types**, for the edition-source change above.
- A Lasting Injury roll and a Vehicle Lasting Damage roll, for the `useRollLogger` conversion.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -467,6 +467,9 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
       intelligence: effectiveFighterData.intelligence,
       save: effectiveFighterData.save ?? null,
       xp: effectiveFighterData.starting_xp ?? 0,
+      // Snapshot of the recruitment value. Kept on the fighter because the
+      // fighter type is editable and N26 counts Advancements from this figure.
+      starting_xp: effectiveFighterData.starting_xp ?? 0,
       is_vehicle: effectiveFighterData.is_vehicle ?? false,
       kills: 0,
       special_rules: effectiveFighterData.special_rules,
--- a/app/actions/copy-fighter.ts
+++ b/app/actions/copy-fighter.ts
@@ -277,7 +277,13 @@ export async function copyFighter(params: CopyFighterParams): Promise<CopyFighte
       save: sourceFighter.save ?? null,
       is_vehicle: sourceFighter.is_vehicle ?? false,
 
-      xp: params.copy_as_experienced ? sourceFighter.xp : 0,
+      // Recruitment XP belongs to the fighter type, not to what this fighter went
+      // on to earn, so it survives an inexperienced copy. Dropping the copy to
+      // xp 0 while it still records a starting_xp of 13 would leave it sitting
+      // below its own recruitment value; a fresh recruit of that type has
+      // xp === starting_xp, which is what add-fighter writes.
+      starting_xp: sourceFighter.starting_xp,
+      xp: params.copy_as_experienced ? sourceFighter.xp : sourceFighter.starting_xp,
       total_xp: params.copy_as_experienced ? sourceFighter.total_xp : 0,
       kills: params.copy_as_experienced ? sourceFighter.kills : 0,
 
@@ -698,7 +704,9 @@ export async function copyFighter(params: CopyFighterParams): Promise<CopyFighte
             willpower: beastFighter.willpower,
             intelligence: beastFighter.intelligence,
             save: beastFighter.save ?? null,
-            xp: params.copy_as_experienced ? beastFighter.xp : 0,
+            // As above: recruitment XP survives an inexperienced copy.
+            starting_xp: beastFighter.starting_xp,
+            xp: params.copy_as_experienced ? beastFighter.xp : beastFighter.starting_xp,
             total_xp: params.copy_as_experienced ? beastFighter.total_xp : 0,
             kills: params.copy_as_experienced ? beastFighter.kills : 0,
             credits: beastFighter.credits,
--- a/app/actions/fighter-advancement.ts
+++ b/app/actions/fighter-advancement.ts
@@ -6,6 +6,8 @@ import { getAuthenticatedUser } from '@/utils/auth';
 import { revalidateTag } from 'next/cache';
 import { updateGangRatingSimple, updateGangFinancials } from '@/utils/gang-rating-and-wealth';
 import { countsTowardRating } from '@/utils/fighter-status';
+import { hasCumulativeXp, gangEditionSlug } from '@/types/edition';
+import { openAdvancementsFor } from '@/utils/advancementRanks';
 
 import { 
   logCharacteristicAdvancement, 
@@ -16,6 +18,94 @@ import {
 } from './logs/gang-fighter-logs';
 import type { GangLogActionResult } from './logs/gang-logs';
 import { updateFighterDetails } from './edit-fighter';
+
+// A fighter's edition comes from its gang's (custom) gang type — gangs have no
+// edition_id of their own. Whether an Advancement costs XP is edition-specific,
+// so every action that moves a fighter's XP loads it alongside the fighter.
+// gangEditionSlug reads this shape back, including the array/object widening the
+// untyped client introduces.
+const FIGHTER_WITH_EDITION_SELECT = `
+  id, user_id, gang_id, xp, starting_xp, free_skill, fighter_name, killed, retired, enslaved, captured,
+  gangs!gang_id (
+    gang_types!gang_type_id ( editions:edition_id ( slug ) ),
+    custom_gang_types!custom_gang_type_id ( editions:edition_id ( slug ) )
+  )
+`;
+
+const fighterEditionSlug = (fighter: any): string | null =>
+  gangEditionSlug(fighter?.gangs);
+
+/**
+ * How many Advancements a fighter has already taken, counted from the database.
+ *
+ * Mirrors countAdvancementsTaken in utils/advancementRanks.ts, which the cards
+ * apply to data they already hold: a characteristic is an effect in the
+ * 'advancements' category, a skill is a fighter_skill flagged is_advance.
+ */
+async function countAdvancementsTakenFor(supabase: any, fighterId: string): Promise<number> {
+  const [characteristics, skillAdvances] = await Promise.all([
+    supabase
+      .from('fighter_effects')
+      .select('id, fighter_effect_types!inner(fighter_effect_categories!inner(category_name))', {
+        count: 'exact',
+        head: true,
+      })
+      .eq('fighter_id', fighterId)
+      .eq('fighter_effect_types.fighter_effect_categories.category_name', 'advancements'),
+    supabase
+      .from('fighter_skills')
+      .select('id', { count: 'exact', head: true })
+      .eq('fighter_id', fighterId)
+      .eq('is_advance', true),
+  ]);
+
+  // This count is half of a server-side limit, so a failed query must not read
+  // as "nothing taken yet" — that would grant an Advancement on the strength of
+  // an error. Fail closed and let the caller refuse.
+  if (characteristics.error || skillAdvances.error) {
+    throw new Error('Failed to count existing advancements');
+  }
+
+  return (characteristics.count ?? 0) + (skillAdvances.count ?? 0);
+}
+
+/**
+ * Whether the fighter may take another Advancement right now, as an error
+ * message or null to proceed.
+ *
+ * The two economies gate on different things: an edition that spends XP gates on
+ * whether the fighter can afford the cost, one that earns Advancements by rank
+ * gates on whether it has an earned Advancement it has not yet taken.
+ *
+ * SOFT LIMIT. This reads the count and the caller then inserts, so two requests
+ * racing each other can both pass. That is tolerable here — a gang is edited by
+ * one person and the whole add-advancement path is already several
+ * non-transactional steps — but it is not a guarantee. Do not build anything on
+ * it that assumes the count cannot be exceeded.
+ */
+async function advancementBlockedReason(
+  supabase: any,
+  fighter: any,
+  spendsXp: boolean,
+  xpCost: number,
+  isAdvance: boolean = true,
+): Promise<string | null> {
+  if (spendsXp) {
+    return fighter.xp < xpCost ? 'Insufficient XP' : null;
+  }
+
+  // A starting or free skill is not an Advancement and consumes none.
+  if (!isAdvance) return null;
+
+  const open = openAdvancementsFor(
+    fighterEditionSlug(fighter),
+    fighter.starting_xp ?? 0,
+    fighter.xp,
+    await countAdvancementsTakenFor(supabase, fighter.id),
+  );
+
+  return open > 0 ? null : 'No advancements available';
+}
 
 // Type for power boost type_specific_data
 interface PowerBoostTypeData {
@@ -118,7 +208,7 @@ export async function addCharacteristicAdvancement(
     // Verify fighter ownership and get fighter data
     const { data: fighter, error: fighterError } = await supabase
       .from('fighters')
-      .select('id, user_id, gang_id, xp, free_skill, fighter_name, killed, retired, enslaved, captured')
+      .select(FIGHTER_WITH_EDITION_SELECT)
       .eq('id', params.fighter_id)
       .single();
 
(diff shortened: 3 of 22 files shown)


joeseos (commented):
## Code Review

@joeseos Reviewed the diff (`dd7337c...dc3efb7`) against the full codebase, focused on the two areas the PR description flags as unverified — the `!gang_id` disambiguation embed and the roster edition-source change — plus whether the correctness bugs found in the sibling implementation (#1947) recur here in this independent rewrite.

### What's good
- **`!gang_id` disambiguation embed is correct.** `fighters` has two FKs to `gangs` (`gang_id`, `captured_by_gang_id`), so the hint is genuinely required. The nested `gangs!gang_id(gang_types!gang_type_id(editions:edition_id(slug)), custom_gang_types!custom_gang_type_id(...))` shape exactly matches the live, working pattern in `gang-data.ts`, `get-user-gangs.ts`, `update-gang.ts`, and `assert-gang-campaign-edition.ts`. The raw-SQL side (`get_fighter_available_advancements.sql`) correctly mirrors `add_fighter_injury.sql`'s `COALESCE(gt.edition_id, cgt.edition_id)` join.
- **Affordability/deduction gating is correctly edition-scoped and fails closed.** `spendsXp = !hasCumulativeXp(...)` is applied consistently across both add paths and the delete refund; `countAdvancementsTakenFor` throws on query error rather than defaulting to 0 (unlike the same-shaped bug in #1947); `openAdvancementsFor` is clamped with `Math.max(0, earned - taken)`.
- **Every category-4/5/6 bug from #1947's review is independently avoided here**: `starting_xp` is correctly preserved (not reset to 0) on both non-experienced copy-fighter branches and on exotic-beast creation; `LevelUpBadge` reads the real `skills` prop rather than the always-empty `advancements.skills`; the roster `edition_slug` now resolves once via `getGangBasic`, covering custom fighter/gang types.
- `use-roll-logger.ts`'s cooldown fix (synchronous `useRef` check ahead of the `useState` flag) correctly closes the double-click race described in the PR body, and all five call sites were refactored faithfully.
- Migration is additive; the N26 advancement seed is idempotent (`NOT EXISTS` guards) and edition-scoped.

### Findings

**1. High — no backfill for `fighters.starting_xp` on existing N26 fighters**
`supabase/migrations/20260810120000_add_fighters_starting_xp.sql` adds the column with `DEFAULT 0` but backfills nothing. `fighter_types.starting_xp`/`custom_fighter_types.starting_xp` have existed since `20260803150605_add_starting_xp_columns.sql` a week earlier, and `add-fighter.ts` has been seeding `fighters.xp` from that value since then — meaning N26 fighters recruited in that window (or any earlier N26 testing) already exist with nonzero XP but will get `starting_xp = 0` from this migration. `advancementsEarnedFor` will then count every rank tier from 4 up to the fighter's current XP as "earned," silently handing out phantom free Advancements — e.g. a Ganger sitting on 15 XP would show as owed 4 Advancements instead of 0. Suggest an `UPDATE ... FROM fighter_types`/`custom_fighter_types` backfill keyed the same way `add-fighter.ts` already resolves the type, or at least a called-out manual fix-up step before this ships to any environment with real N26 data.

**2. Medium — edition-resolution failure silently falls back to the more permissive gate, not the safer one**
`hasCumulativeXp(null)` returns `false` (a missing slug means "nothing edition-specific applies"), so if the new `gangs!gang_id(...)` embed ever fails to resolve for an N26 fighter — RLS gap, unexpected null `custom_gang_type_id`, malformed nested select, exactly the scenario you flag as untested in a browser — `spendsXp` becomes `true` and the guard becomes `fighter.xp < xp_cost`. Since N26 catalog rows are seeded with `xp_cost: 0`, that check is always `false`, meaning a fighter can take unlimited Advancements with no rank check at all if the embed silently comes back empty. Given this is on the critical path for both editions and explicitly called out as unverified, worth either treating an unresolved edition as blocking (return an error rather than falling through) or adding an integration test that actually exercises the embed against a real N26 fighter before merge.

**3. Note, not a blocker** — `getFighterBasic` (single fighter page) still derives edition via the older `custom_fighter_type`/`fighter_type` embed, a pre-existing path this PR doesn't touch, while the roster now resolves edition via the gang. Not a regression, but worth being aware the two pages could theoretically disagree on edition for the same fighter in an edge case (e.g. mid-edit fighter_type change) — probably fine to leave as a documented follow-up rather than fixing here.

The architecture and the specific bug categories the sibling PR (#1947) surfaced are all independently handled correctly here — that's a good sign for the design. The one thing I'd want fixed before merge is #1 (the missing `starting_xp` backfill), since it silently over-grants Advancements to any N26 fighter that already exists rather than failing loudly. #2 is worth hardening or at least explicitly testing given it's the exact code path you flagged as unexercised — I wouldn't necessarily block on it alone, but it compounds with #1 in exactly the scenario you're already asking for a click-through on.

---

joeseos:
## Code Review

@joeseos Reviewed the full diff against `main`, plus the SQL against `supabase/schema/schema.public.sql` and the pre-existing `add_fighter_injury.sql` pattern this PR says it mirrors. I focused on the three points you flagged for extra scrutiny, and independently re-checked for the bug classes #1947's reviews found (starting_xp propagation, edition fallback for custom types, fail-open server guards).

### What's good
- `utils/advancementRanks.ts` — the tier table and the "omit Rookie's own 1-3 band" trick check out exactly. Hand-verified all five worked examples (0→4, 1→4, 3→4, 4→7, 13→19) plus the 229-XP/20-Advancement ceiling; the `Math.max(0, …)` range-count approach correctly avoids an "over-advance a fighter recruited above 0 XP" bug.
- The SQL edition-resolution join in `get_fighter_available_advancements.sql` (`gangs → gang_types!gang_type_id` / `custom_gang_types.id`) is byte-for-byte consistent with the actual FK constraints and the existing `add_fighter_injury.sql` pattern. The `edition_id IS NOT DISTINCT FROM v_edition_id` filter doesn't leak — confirmed `fighter_effect_types`/`gang_types` were backfilled to a non-null `n23` `edition_id`, so N23 rows still match correctly.
- **`starting_xp` propagation is genuinely complete** — `add-fighter.ts`, both `copy-fighter.ts` branches (experienced and non-experienced, fighter and exotic-beast sub-branch), and `exotic-beasts.ts` all set it, and it's correctly threaded through to every read site feeding a rank calc. This independently avoids the exact bug class #1947's reviews found there.
- The `fighters!gang_id` PostgREST disambiguation in `fighter-advancement.ts` is correct and necessary — confirmed `fighters` has two FKs to `gangs` (own gang + `captured_by_gang_id`).
- **Server-side gating fails closed** — `countAdvancementsTakenFor` explicitly throws on a Supabase query error rather than treating it as 0, directly avoiding the fail-open bug #1947 had.
- The roster edition-source fix (`getGangFightersList` resolving `edition_slug` from `getGangBasic(gangId)` instead of `fighter_types`) is applied at a single point and consumed by the one `edition_slug` prop in `fighter-card.tsx` that already drives `hasSaveCharacteristic` and all three `WeaponTable` calls — consistent everywhere, not partial.
- `useRollLogger`'s core race fix (a synchronous `useRef` checked inside `logRoll`, instead of comparing against batched React state) is real and correctly closes the specific "two clicks in the same tick both see cooldown as false" bug for the sites that use its return value.

### Findings

**1. [High] N26AdvancementPanel's roll button is broken for every N26 fighter that isn't a Ganger/Exotic Beast**
`components/fighter/fighter-advancement-list.tsx:~2118` and `app/actions/fighter-advancement.ts:1315-1336`

`N26AdvancementPanel` renders whenever `isCumulativeXp && userPermissions`, with no subtype check:
```tsx
{isCumulativeXp && userPermissions && (
  <N26AdvancementPanel fighterId={fighterId} canEdit={!!userPermissions.canEdit} />
)}
```
But it logs its roll via `verifyAndLogRolledGangerAdvancementRoll` — an unmodified, pre-existing action that hard-rejects anything outside `GANGER_ELIGIBLE_SUBTYPES = {'Ganger', 'Exotic Beast'}`:
```ts
if (!fighter.fighter_subtypes?.some((c) => GANGER_ELIGIBLE_SUBTYPES.has(c))) {
  throw new Error('Advancement roll is only for Gangers and Exotic Beasts');
}
```
So for any N26 Leader, Champion, Juve/Prospect, Specialist, etc., clicking "Roll 2D6" always fails with a confusing, always-on toast — for the majority of the roster. This is the headline new UI feature of the PR, and it only works for the one subtype pairing it happens to share a logging endpoint with. The manual characteristic-selection path underneath still works (the SQL function doesn't gate by subtype for N26), so advancements can still be taken, but the roll-and-audit-log flow this PR built is non-functional for most fighters.

Suggested fix: give N26 its own logging action (e.g. `verifyAndLogRolledN26AdvancementRoll`), or drop the subtype check when the fighter's edition is N26 — N26 has no separate Ganger table to protect.

**2. [Medium] Double-click race fix is inconsistent — two of the five converted call sites can log a different roll than what's displayed**
`components/fighter/fighter-advancement-list.tsx` (`rollGangerSkillInSet` ~1337, `rollSkillFromSet` ~1353)

The primary Ganger roll and the injury roll correctly gate local UI state on `logRoll()`'s return value:
```ts
const accepted = gangerRollLogger.logRoll({...});
if (!accepted) return false;
setGangerSelectedRowId(combo.id); // only runs if accepted
```
But `rollGangerSkillInSet`/`rollSkillFromSet` don't check the return value — they roll and `setState(...)` *before* calling `.logRoll()`, ignoring its result. The pre-existing code guarded these two functions atomically (`if (logGangerSubRollMutation.isPending) return;` at the very top, before rolling or setting state — still present on `main`), so state and the gang-log entry could never diverge. That guard was removed here and not replaced for these two sites (`cooldownMs: 0` was set instead). On a same-tick double-click — the exact scenario this refactor targets elsewhere — the second click's roll wins the displayed UI state, but only the first click's roll gets logged, since the ref-based cooldown silently drops the second `logRoll()` call. The gang log audit trail can now disagree with what's on screen — arguably worse than the duplicate-log bug the PR set out to fix for these two sites. (The same pattern in `vehicle-lasting-damages.tsx` is pre-existing/unchanged, not a new regression here.)

Suggested fix: apply the same `accepted`-gating pattern used by `logGangerResolvedRollWithCooldown`/`logResolvedRollWithCooldown` — call `logRoll()` first, only commit local state if it returns `true`.

Finding 1 will hit essentially every non-Ganger N26 fighter the first time anyone uses the new roll UI, and should block merge. Finding 2 is real but narrower (only reachable via a fast double-click) — worth fixing in the same pass. Everything else — rank math, SQL edition filtering/joins, `starting_xp` propagation across all four recruitment paths, the roster edition-source fix, and fail-closed server-side gating — checks out and does not reproduce the bug classes #1947's reviews found.

---

PR #1965, opened 2026-08-10
Added pagination to api routes with shared util

joeseos:


joeseos:
## Automated Code Review

@joeseos Reviewed the diff (4 files, +112/-75). This is a well-scoped fix: three admin API routes (`equipment`, `fighter-types`, `notifications`) were silently truncated at PostgREST's default 1000-row cap, and this replaces the unpaginated/ad-hoc-paginated queries with a shared `utils/supabase/fetch-all-rows.ts` helper.

### Correctness check
- `fetchAllRows` pages by `rows.length` (actual rows fetched so far), not by a fixed `PAGE_SIZE` increment — so it stays correct even if the server's real max-rows cap is lower than 1000. Good.
- Every paginated query now has a unique tiebreaker (`.order('id')`) added after the existing `order()` calls, which is required for stable offset pagination — `gang_type`/`fighter_type`/`equipment_name` alone aren't unique, so without this, rows could be skipped or duplicated across page boundaries. This was correctly added in all 5 call sites.
- Error propagation is preserved: routes that previously did `if (error) throw error` now rely on `fetchAllRows` throwing internally, caught by the same outer `try/catch` — behavior is equivalent.
- The `notifications` route's manual `while(true)` pagination loop is now dead code, replaced 1:1 by the shared helper, and it also fixes a pre-existing bug where `allEquipment` (grants dropdown) had its `error` silently ignored — now surfaced via try/catch with a safe `[]` fallback.
- Filter closures (`equipment_type`, `filter_by_gang`/`gang_type_id`) are rebuilt fresh on every page request inside `buildPage`, which is required since Supabase query builders are single-use — correctly done everywhere.
- Traced through the `PostgrestFilterBuilder`/`PostgrestTransformBuilder` chaining (`.range()` called before a conditional `.eq()` in two spots) — postgrest-js's fluent methods return `this`, so chaining order doesn't affect the final query; this matches the pre-existing pattern used elsewhere in the same files.

I didn't have a working `node_modules` in this environment to run `tsc`/`next build` directly, so treat the type-level reasoning above as high-confidence but not compiler-verified.

### Findings

| # | File:Line | Severity | Issue | Suggestion |
|---|---|---|---|---|
| 1 | `app/api/admin/fighter-types/route.ts:404` | Low | `fetchAllRows<any>(...)` uses an explicit `any` generic, while the structurally-identical closure in `app/api/admin/equipment/route.ts` (default branch) relies on inference and stays typed. Loses type safety on `fighterTypes` for no apparent reason. | Try dropping the explicit `<any>` — the closure shape is the same as the equipment route's, so inference should work the same way. |
| 2 | `utils/supabase/fetch-all-rows.ts:15-22` | Low (informational) | When the total row count is an exact multiple of `PAGE_SIZE` (1000, 2000, …), the loop issues one extra trailing request that returns 0 rows before stopping. | Not worth optimizing given these are admin-only, low-frequency routes — just flagging for awareness, no change needed. |

### Verdict: **Approve**

No correctness, security, or performance blockers. The pagination logic itself is sound (tiebreakers present, error handling preserved, closures correctly single-use-safe), and the change fixes a real silent-truncation bug rather than introducing new risk. The two notes above are non-blocking style/awareness items.

---

PR #1966, opened 2026-08-10
Added support for n26 pets

joeseos:


joeseos (commented):
## Code Review

@joeseos Reviewed the diff against the full codebase, focused on: whether the `equipment_category` field switch is safe, whether the new `.or()`/`.cs.` PostgREST query is valid, whether `beastSubtypeName`/`EDITION_CAPABILITIES` refactor in `types/edition.ts` is sound, and — most importantly — whether every place in the codebase that special-cases N23's `'Exotic Beast'` was actually updated to also recognize N26's `'Pet'`.

### What's good
- `app/actions/add-fighter.ts`: switching the exotic-beast detection from `equipment_category_id` to a lowercased `equipment_category` string match is consistent with the pre-existing convention already used in `move-to-stash.ts`/`move-from-stash.ts`/`sell-equipment.ts` — not a regression.
- `types/edition.ts`: the `BooleanCapability`/`CapabilityAnswer<K>` generics refactor is sound — `can()` now rejects non-boolean capability rows at compile time, and `beastSubtypeName` correctly falls back to N23's `'Exotic Beast'` when the slug is unset. No caller invokes `can('beastSubtype', …)`.
- `components/gang/fighter-add/FighterAddModal.tsx`: `selectedSubtype === 'pet'` correctly matches the N26 addition-category value (`utils/gangAdditionRankN26.ts`).
- `utils/allowedFighterSubtypes.ts`, `utils/fighterSubtypeRankN26.ts`, `utils/equipmentCategoryRankN26.ts` already handle `'Pet'`/`'Pets'` — no changes needed there.
- `edition_slug` is populated server-side at every `beastSubtypeName(...)` call site added in this PR (`fighter-page.tsx`, battle session participant card), so the new copy won't silently fall back to N23 wording.

### Findings

**1. High — flat-XP advancement/promotion pipeline still gates on `'Exotic Beast'` only, not `'Pet'`**
This PR's own stated goal is parity between N23 Exotic Beasts and N26 Pets, but several places that drive XP cost and promotion UI for beast-type fighters weren't updated:
- `app/actions/fighter-advancement.ts:1315` — `GANGER_ELIGIBLE_SUBTYPES = new Set(['Ganger', 'Exotic Beast'])` gates the flat-cost advancement-roll logger. An N26 Pet's advancement roll gets rejected server-side with "Advancement roll is only for Gangers and Exotic Beasts."
- `components/fighter/fighter-advancement-list.tsx:680,1506,2000,2149,2704` — `fighterSubtypes.includes('Exotic Beast' | 'Exotic Beast Specialist')` gates the flat-XP roll UI and `showPromoteButton`. Pets get neither.
- `components/fighter/edit-fighter/fighter-promotion-modal.tsx:18-19,116` — `PROMOTION_MAP`/`isExoticBeast` don't include `'Pet'`.
- `components/fighter/edit-fighter/fighter-edit-modal.tsx:1160` — subtype-override dropdown filter checks `'Exotic Beast'` only.
- Backend: `supabase/functions/get_fighter_available_advancements.sql:27-29` (`v_fighter_subtypes ?| array['Ganger', 'Exotic Beast']`, also inlined in `supabase/schema/schema.public.sql:1894`) and the `get_available_skills*` migrations checking `v_fighter_class IN (..., 'Exotic Beast Specialist')` are the actual source of truth for flat advancement/skill costs — none updated for `'Pet'`. This means N26 Pets fall back to random-advancement costing instead of the intended flat cost, a real rating/XP correctness gap.

These are outside this PR's file list, but they're squarely in the "make Pets behave like Exotic Beasts" scope this PR sets out to cover, and would produce visibly wrong costs/missing UI for N26 games.

**2. Medium — `.or('fighter_subtypes.cs.["Exotic Beast"],fighter_subtypes.cs.["Pet"]')` in `app/lib/shared/gang-data.ts` (`getGangBeastCount`) is untested syntax**
Neither clause contains an internal comma so it should parse under PostgREST's `or()` grammar, but this is the only place in the repo combining `.or()` with the `.cs.` (contains) operator — no existing precedent to lean on. Worth a quick manual check that `getGangBeastCount` still returns correct counts for a gang with both Exotic Beasts and Pets before merge; a silent parse failure here would throw or return a wrong count rather than fail loudly.

**3. Low — `utils/exotic-beasts.ts:136`: `fighterType.fighter_subtypes ?? ['Exotic Beast']` only guards `null`/`undefined`**
If a fighter-type row for an N26 pet ever has `fighter_subtypes: []` (empty array, not null) rather than missing entirely, this produces a beast fighter with no subtype tag at all — silently breaking every `.some(c => …)` check this PR adds elsewhere. Worth an explicit length check, or a DB-level constraint that pet/beast fighter types can't have an empty `fighter_subtypes`.

The files actually touched are correct and consistent with existing conventions, but the PR is incomplete relative to its own goal — the flat-cost XP/advancement and promotion pipeline (client UI + backend SQL) is still N23-only, which will produce wrong XP costs and missing promotion/advancement UI for N26 Pets. Recommend extending finding #1's call sites to include `'Pet'` before merge; #2 and #3 are worth a quick look but not blockers on their own.

---

PR #1967, opened 2026-08-10
Added edition gate on characteristic display

joeseos:


joeseos:
@joeseos Reviewed this PR (diff, `tsc --noEmit` on the branch — clean, and traced every `StatsTable`/characteristic-display callsite for edition-suffix coverage).

## Findings

### 1. `fighter-characteristic-table.tsx` Total row still hardcodes `+` — HIGH
**File:** `components/fighter/edit-fighter/fighter-characteristic-table.tsx:226`

The **Base** row (line 90) was correctly updated to gate the suffix:
```tsx
`${baseValue}${EDITION_SUFFIX_STATS.has(stat.key) ? initiativeAndMentalSuffix : '+'}`
```
but the **Total** row a few lines below was not:
```tsx
stat.key === 'movement' ? `${total}"` :
stat.key === 'wounds' || stat.key === 'attacks' ||
stat.key === 'strength' || stat.key === 'toughness' ?
total :
`${total}+`}   // <-- always appends '+', even for I/Ld/Cl/Wil/Int on N26
```
For an N26 fighter this produces a visibly inconsistent table in the "Adjust Characteristics" breakdown: the Base row will correctly show e.g. `5` for Initiative/Leadership/Cool/Willpower/Intelligence, while the Total row directly below (and the same values everywhere else touched by this PR — `character-stats-modal.tsx`, `fighter-details-card.tsx`, `fighter-card-stats-table.tsx`) will show `5+`. Since this table's whole purpose is to compare Base vs. Total, the mismatch is immediately visible to users.

**Suggested fix:** mirror the Base row's conditional in the Total row:
```tsx
stat.key === 'movement' ? `${total}"` :
stat.key === 'wounds' || stat.key === 'attacks' ||
stat.key === 'strength' || stat.key === 'toughness' ?
total :
`${total}${EDITION_SUFFIX_STATS.has(stat.key) ? initiativeAndMentalSuffix : '+'}`
```

## Everything else looks solid
- `types/edition.ts`: new `initiativeAndMentalCharacteristicSuffix` capability follows the existing `beastSubtype`-style non-boolean pattern correctly (uses `answerFor` + explicit n23 fallback, not `can`).
- `fighter-card-stats-table.tsx` (`formatStatValue`) and `character-stats-modal.tsx` (`getAdjustedTotal`/`getBaseDisplay`) both correctly gate on `EDITION_SUFFIX_STATS` and are internally consistent (Base and Total agree).
- Both `StatsTable` callsites (`fighter-card.tsx`, `print-gang.tsx`) now pass `editionSlug`; no missed callsite left defaulting silently to N23.
- `fighter-details-card.tsx` updates both the vehicle/crew and standard-fighter stat branches consistently, and the `useMemo` dependency array was correctly extended with `initiativeAndMentalSuffix`.
- No security, N+1, or unnecessary-render concerns — this is a pure display-formatting change.

Just the one blocker above (Total row). Everything else in this small, well-scoped PR is correct and consistent with the existing edition-capability pattern.

---

joeseos (commented):
@joeseos Follow-up review on the new commit (`1e29768`, "Merge origin/main, and fix edition gating from #1963").

## Previous blocker: fixed
`fighter-characteristic-table.tsx` now runs Base and Total through the same `formatStatValue(key, value)` helper (lines 55-61, 112, 249), so they can no longer disagree. Confirmed by reading the file directly.

## New changes in this commit — no bugs found
- `resolveFighterEditionSlugForUpdate` / `validateFighterSubtypesForUpdate`: re-prioritizing gang → fighter type (was fighter type → gang) and letting a null edition fall through to the legacy rule instead of rejecting is well-reasoned and matches the commit message's stated data (34 custom_fighter_types / 9 custom_gang_types with null edition_id). The grandfather clause for subtype names not in `fighter_subtypes` catalog (`!validNames.has(name) && !previous.includes(name)`) correctly only exempts names the fighter already carried, not new ones.
- `fighterSubtypesForDisplay` in `fighter-edit-modal.tsx`: synthesizes `{id: name, subtype_name: name}` entries for uncatalogued (alliance-crew) subtypes so the Combobox doesn't blank out. I checked every read site (lines 268, 812, 1256) — none dereference `.id` on these synthetic entries, so the placeholder id is safe.
- `character-stats-modal.tsx`: removing the Sv-decrement guard (`disabled={isSaving || (stat.key === "Sv" && ...)}` → `disabled={isSaving}`) looked like a possible regression at first (Sv could now go negative from unset), but every other characteristic in this same modal was already unguarded and relies on `isStatOutOfRange` elsewhere to flag illegal values red — so this just makes Sv consistent with the rest rather than introducing a new hole.
- `initiativeAndMentalCharacteristicSuffix` follows the existing `hasSaveCharacteristic`/`EDITION_CAPABILITIES` pattern correctly, and `fighter-card-stats-table.tsx`'s `formatStatValue` correctly leaves `Sv` untouched (not in `EDITION_SUFFIX_STATS`) so it keeps its `+`.

## Heads-up: overlaps with #1968 (currently open, not yet merged)
Flagging this since I reviewed both PRs today and it isn't visible from either diff alone:

- **`resolveFighterEditionSlugForUpdate` signature**: #1968's `edit-fighter.ts` diff still calls the *old* 3-arg form (`resolveFighterEditionSlugForUpdate(supabase, params.fighter_id, params)` returning `{ok, editionSlug}`). This commit collapses it to 2 args returning `string | null`. Whichever PR merges second will fail to compile against `main` until it's rebased onto the other.
- **Archetype-reset heuristic**: `applySelectedFighterSubtypes` here still resets `selectedArchetypeId` only when `selectedFighterSubtypes[0]` changes ("Archetypes key off effectiveFighterSubtype, i.e. the first entry" — line ~798 comment). #1968 reworks archetype eligibility/catalog selection to consider *any* selected subtype with a Leader→Champion→Ganger priority order, not just index 0. If #1968 lands first, this file's comment and reset logic will be stale (though #1968's own server-side `assertArchetypeAssignable` fallback would still catch the mismatch on save — see the blocker I raised on #1968 about that path not syncing back to client state).

Neither is a defect in this PR in isolation — both diffs are correct against their own base — but worth a rebase-and-recheck of whichever merges second rather than merging both independently.

## Verdict: Approve
The characteristic-display blocker is fixed, and I didn't find new issues in the rest of this commit. Only the cross-PR coordination note above needs eyes, not a code change here.

---

PR #1968, opened 2026-08-10
Fix Outcasts archetype eligibility for multi-subtype N23/N26 fighters.

Djidiouf:
Eligibility now checks any selected subtype (not only [0]), uses N23 vs N26 gang-type rules, and validates/clears archetypes against the edition-scoped catalog so the UI no longer flickers or keeps stale skill access.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -8,7 +8,7 @@ import { updateGangFinancials } from '@/utils/gang-rating-and-wealth';
 import { logFighterAction } from '@/app/actions/logs/fighter-logs';
 import { mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility';
 import { assertArchetypeAssignable } from '@/utils/assertArchetypeAssignable';
-import { editionSlugFromJoin, type EditionJoin } from '@/types/edition';
+import { editionSlugFromJoin, gangEditionSlug, type EditionJoin } from '@/types/edition';
 
 interface SelectedEquipment {
   equipment_id: string;
@@ -373,7 +373,11 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
           .single(),
       supabase
         .from('gangs')
-        .select('id, credits, user_id, gang_type_id')
+        .select(`
+          id, credits, user_id, gang_type_id,
+          gang_types!gang_type_id ( editions:edition_id ( slug ) ),
+          custom_gang_types!custom_gang_type_id ( editions:edition_id ( slug ) )
+        `)
         .eq('id', params.gang_id)
         .single()
     ]);
@@ -442,11 +446,15 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
         ? fighterSource.fighter_subtypes
         : ['Custom'];
 
+      // Prefer fighter-type edition; fall back to gang edition for legacy custom types with null edition_id
+      const editionSlug =
+        editionSlugFromJoin(fighterSource.editions) ?? gangEditionSlug(gangData);
+
       const assignable = await assertArchetypeAssignable(supabase, {
         gangTypeId: gangData.gang_type_id,
         fighterSubtypes,
         archetypeId: params.selected_archetype_id,
-        editionSlug: editionSlugFromJoin(fighterSource.editions),
+        editionSlug,
       });
       if (!assignable.ok) {
         return {
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -6,10 +6,9 @@ import { invalidateFighterAddition, invalidateUserGangsList } from '@/utils/cach
 import { createExoticBeastsForEquipment } from '@/utils/exotic-beasts';
 import { updateGangFinancials } from '@/utils/gang-rating-and-wealth';
 import { logFighterAction } from '@/app/actions/logs/fighter-logs';
-import {
-  isArchetypeEligible,
-  mapArchetypeSkillAccessToOverrides,
-} from '@/utils/archetypeEligibility';
+import { mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility';
+import { assertArchetypeAssignable } from '@/utils/assertArchetypeAssignable';
+import { editionSlugFromJoin, type EditionJoin } from '@/types/edition';
 
 interface SelectedEquipment {
   equipment_id: string;
@@ -18,6 +17,13 @@ interface SelectedEquipment {
   effect_ids?: string[];
   is_editable?: boolean;
 }
+
+/** Fighter type / custom fighter type row with optional editions embed for slug resolution. */
+type FighterTypeSource = {
+  editions?: EditionJoin;
+  fighter_subtypes?: string[] | null;
+  [key: string]: unknown;
+};
 
 interface AddFighterParams {
   fighter_name: string;
@@ -317,7 +323,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     // First try to get fighter if user owns it
     const { data: ownedFighter } = await supabase
       .from('custom_fighter_types')
-      .select('*')
+      .select('*, editions:edition_id ( slug )')
       .eq('id', params.fighter_type_id)
       .eq('user_id', user.id)
       .maybeSingle();
@@ -347,7 +353,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
           // Fetch the actual fighter data
           const { data: fighterData } = await supabase
             .from('custom_fighter_types')
-            .select('*')
+            .select('*, editions:edition_id ( slug )')
             .eq('id', params.fighter_type_id)
             .single();
 
@@ -362,7 +368,7 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
         Promise.resolve({ data: null, error: null }) : // Skip regular fighter type lookup for custom fighters
         supabase
           .from('fighter_types')
-          .select('*')
+          .select('*, editions:edition_id ( slug )')
           .eq('id', params.fighter_type_id)
           .single(),
       supabase
@@ -431,15 +437,24 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
     // Server-side archetype eligibility (UI check is not sufficient)
     let archetypeIdToPersist: string | null = null;
     if (params.selected_archetype_id) {
-      if (!isArchetypeEligible({
+      const fighterSource = effectiveFighterData as FighterTypeSource;
+      const fighterSubtypes = fighterSource.fighter_subtypes?.length
+        ? fighterSource.fighter_subtypes
+        : ['Custom'];
+
+      const assignable = await assertArchetypeAssignable(supabase, {
         gangTypeId: gangData.gang_type_id,
-        fighterSubtype: effectiveFighterData.fighter_subtypes?.[0] ?? 'Custom',
-      })) {
+        fighterSubtypes,
+        archetypeId: params.selected_archetype_id,
+        editionSlug: editionSlugFromJoin(fighterSource.editions),
+      });
+      if (!assignable.ok) {
         return {
           success: false,
-          error: 'This fighter type cannot be assigned an archetype for this gang.',
+          error: assignable.error,
         };
       }
+
       archetypeIdToPersist = params.selected_archetype_id;
     }
 
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -12,6 +12,7 @@ import { updateGangFinancials, updateGangRatingSimple, GangFinancialUpdateResult
 import { insertFighterOoaRecords } from './fighter-ooa-records';
 import { allowsMultipleSubtypes, editionSlugFromJoin, gangEditionSlug } from '@/types/edition';
 import { assertArchetypeAssignable } from '@/utils/assertArchetypeAssignable';
+import { mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility';
 
 // Helper function to invalidate owner's cache when beast fighter is updated
 async function invalidateBeastOwnerCache(fighterId: string, gangId: string, supabase: any) {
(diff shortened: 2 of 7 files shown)


joeseos (commented):
@Djidiouf Thanks for the fix — the multi-subtype eligibility check and the edition-scoped catalog validation are solid, and the server-side `assertArchetypeAssignable` gate closes a real hole (the old code only ever checked `fighter_subtypes[0]`). A few findings from reviewing the diff against the rest of `edit-fighter.ts` / `fighter-edit-modal.tsx`.

## Findings

**1. Server-side archetype auto-clear isn't surfaced to the client (medium)**
`app/actions/edit-fighter.ts:1542-1550` silently nulls `selected_archetype_id` (and deletes `fighter_skill_access_override` rows) when a subtype change invalidates the fighter's current archetype — but the response `.select()` at `edit-fighter.ts:1559` never included `selected_archetype_id`, so the client has no way to learn this happened.

`fighter-edit-modal.tsx`'s `onSuccess` (line 552) decides whether to touch skill-access overrides by comparing `submit.selected_archetype_id !== fighter.selected_archetype_id` — both client-known values, neither reflecting the server's silent clear. So after a save that triggers the implicit clear:
- The modal's local `selectedArchetypeId` state keeps showing the now-invalid archetype.
- If the user saves again without touching the archetype field, the client resubmits the stale id, and `checkArchetypeAssignable` now takes the `params.selected_archetype_id !== undefined` branch and rejects it outright (`edit-fighter.ts:1531-1536`) — the save fails with a generic error until the page is reloaded.

Concretely: an N26 Outcasts fighter with subtypes `[Champion]` and a Champion-scoped archetype selected. User adds `Leader` as a second subtype (multi-subtype allowed in N26) and saves. `isArchetypeEligible` still returns true (Leader is eligible too), so the UI doesn't disable the field, but catalog priority now resolves to `Leader`, the Champion-scoped archetype fails `isArchetypeInCatalog`, and gets cleared server-side without the client knowing.

Suggested fix: include `selected_archetype_id` in the `.select()` on the fighters update, and in `onSuccess` reconcile `selectedArchetypeId` (and re-run the override save/clear) against the *returned* value rather than the submitted one.

**2. Unconditional gang lookup on every archetype clear (low)**
`edit-fighter.ts:1499-1508` fetches `gangs.gang_type_id` whenever either `selected_archetype_id` or `fighter_subtypes` is present in the params — including the common case of explicitly clearing the archetype (`params.selected_archetype_id` falsy), where `gangData` is never read (the `else` branch at line 1538 doesn't use it). Worth gating the query so a plain "remove archetype" edit doesn't pay for an unused round trip.

**3. `resolveFighterEditionSlugForUpdate` can run twice per request (low)**
When a save includes both `fighter_subtypes` and `selected_archetype_id`, the edition slug is resolved once in `validateFighterSubtypesForUpdate` (line 224) and again inside `checkArchetypeAssignable` (line 1515) — same fighter, same params, two round trips. Could thread the already-resolved `editionSlug` through instead of re-deriving it.

None of these block the core fix (multi-subtype eligibility + catalog validation), but #1 is a genuine correctness gap introduced by this PR's new "silent server-side clear" behavior and can leave a fighter's edit modal out of sync with the DB until reload.

## Verdict: Changes requested
Blocker: #1 (stale client state after implicit archetype clear). #2/#3 are minor and can be follow-ups if preferred.

---

joeseos:
## Code Review

@Djidiouf Reviewed the full diff (`utils/archetypeEligibility.ts`, `utils/assertArchetypeAssignable.ts`, `app/actions/add-fighter.ts`, `app/actions/edit-fighter.ts`, `components/fighter/edit-fighter/fighter-edit-modal.tsx`, `components/gang/fighter-add/FighterAddModal.tsx`, `types/edition.ts`) against `main`, and cross-checked the game-rule catalog data (`gang_types`, `fighter_subtypes`, `skill_access_archetypes`) directly against the live Supabase project rather than trusting the PR description.

### What's good
- **The `[0]`-only bug is real and verifiably fixed.** On `main`, `applySelectedFighterSubtypes` only cleared `selectedArchetypeId` when `next[0] !== selectedFighterSubtypes[0]`, and eligibility only ever checked `fighterSubtype[0]`. A fighter whose qualifying subtype was appended after index 0 was invisibly ineligible, and removing that later subtype never cleared a stale selection. The new `isArchetypeEligible`/`getArchetypeCatalogSubtype` (`utils/archetypeEligibility.ts:52-77`) correctly use `.some()`/priority-`.find()` over the full subtype array.
- **N26 Outcasts gang-type UUID and per-edition eligible-subtype lists verified against production data** — `N26_UNDERHIVE_OUTCASTS_GANG_TYPE_ID` matches the "Outcasts" (n26) row in `gang_types` exactly, and `skill_access_archetypes` only has Ganger-scoped archetypes for n26 (none for n23), matching the code's `['Leader','Champion']` vs `['Leader','Champion','Ganger']` split.
- **`edit-fighter.ts` previously had zero server-side validation of `selected_archetype_id`** — on `main` it was a pure passthrough (`if (params.selected_archetype_id !== undefined) updateData.selected_archetype_id = params.selected_archetype_id;`). This PR adds a real gate (`assertArchetypeAssignable`) plus DB-level cleanup of `fighter_skill_access_override` rows when an archetype goes stale — a genuine trust-boundary fix, not cosmetic.
- **`resolveFighterEditionSlugForUpdate`** correctly follows the fighter-type → gang-type/custom-gang-type fallback chain and fails closed when unresolvable, avoiding the "custom types have null edition_slug" pitfall that's bitten other recent PRs. Confirmed this path is actually exercised (not dead code) via `applyPromotionWithSkillAdvancement` in `fighter-advancement.ts`, which updates `fighter_subtypes` without `selected_archetype_id` and correctly triggers re-validation/clearing on promotion.

### Findings

**1. High — the actual skill-access grant path (`saveFighterSkillAccessOverrides`) is never validated against archetype eligibility; a client can grant arbitrary skill access to any editable fighter.**
`app/actions/fighter-skill-access.ts` (unchanged by this PR) is invoked from `components/fighter/edit-fighter/fighter-edit-modal.tsx:562-573`:
```ts
if (persistedArchetypeId && archetypesData?.archetypes) {
  const archetype = (archetypesData.archetypes as Archetype[]).find(...)
  if (archetype) {
    const overrides = mapArchetypeSkillAccessToOverrides(archetype.skill_access);
    await saveFighterSkillAccessOverrides({ fighter_id: fighter.id, overrides });
  }
}
```
`saveFighterSkillAccessOverrides` only checks `canEdit` permission (`fighter-skill-access.ts:48-51`) — it never checks that `overrides` corresponds to a real, eligible archetype. This is the mechanism that actually grants gameplay capability: `app/api/fighters/skill-access/route.ts:234-243` computes `effectiveAccess = override_access_level ?? access_level` and gates purchasing a Primary skill set purely on that. A client can call this server action directly with an arbitrary `{skill_type_id, access_level: 'primary'}` payload for any fighter they can edit, independent of gang type, subtype, edition, or `assertArchetypeAssignable` entirely. This PR hardens `selected_archetype_id` persistence but leaves the actual override-write path open, so the "no longer keeps stale skill access" claim isn't fully closed — the value can be *arbitrary*, not just *stale*. Predates this PR, but it's exactly the class of bug this work is meant to close, and `add-fighter.ts` already shows the right pattern (compute overrides server-side from the validated `skill_access_archetypes` row, `add-fighter.ts:516-539`, rather than trusting a client-supplied `overrides` array). Suggest doing the same for edits — derive overrides server-side from the validated archetype inside `updateFighterDetails`, or have `saveFighterSkillAccessOverrides` re-run `assertArchetypeAssignable`/re-derive from `skill_access_archetypes`.

**2. Medium — `add-fighter.ts`'s edition resolution for archetype validation has no gang-level fallback; verified regression for legacy custom fighter types.**
`app/actions/add-fighter.ts:449`: `editionSlug: editionSlugFromJoin(fighterSource.editions)` reads only the fighter type's own `editions` embed, with no fallback to the gang's edition (unlike `edit-fighter.ts`'s `resolveFighterEditionSlugForUpdate`, which tries fighter type → gang type → custom gang type). Verified against production data: 34 `custom_fighter_types` rows have `edition_id IS NULL`, several with Leader/Champion/Ganger subtypes (e.g. "Astartes Sergeant" — Leader, "Aspect Warrior" — Champion). All 27 `skill_access_archetypes` rows have a non-null `fighter_subtype_id`, so `assertArchetypeAssignable` will hit `if (!params.editionSlug) return { ok: false, error: 'Could not resolve edition for archetype validation.' }` for these fighters — which fails the **entire** `addFighterToGang` call (`add-fighter.ts:452-457`), not just the archetype selection. Before this PR, adding these fighters with an archetype succeeded. Suggest reusing the same fallback chain `edit-fighter.ts` already has (or at minimum falling back to `gangEditionSlug`).

**3. Low/Medium — client-side race between subtype change and the `skill-archetypes` query refetch can silently skip saving overrides.**
`fighter-edit-modal.tsx:556-575`: `archetypesData` is keyed by `archetypeFighterSubtypeId`, which updates as soon as subtypes change, but Save has no loading gate on that refetch. Saving before the refetch resolves can leave the `find()` lookup pointed at the *previous* subtype's catalog, fail to find the archetype, and silently skip `saveFighterSkillAccessOverrides` — even though `selected_archetype_id` persisted correctly server-side. Net effect: fighter shows the right archetype but has no skill-access grant until a later unrelated save. Not a security issue, but a real user-visible under-grant in the same "stale/flicker" family this PR targets. The Finding 1 fix (deriving overrides server-side from the persisted id) would close this too.

**4. Low, not a bug** — the catalog priority order (Leader > Champion > Ganger) baked into `getArchetypeCatalogSubtype` (`utils/archetypeEligibility.ts:14-17`) couldn't be independently verified against a rules doc in-repo. Worth a quick confirm from whoever owns the ruleset content, since it's silent/deterministic rather than surfaced when a fighter has multiple qualifying subtypes.

The core "any subtype" / N23-vs-N26 fix is correct and verified against live catalog data, and `assertArchetypeAssignable` is a genuine improvement over the previous unchecked passthrough in `edit-fighter.ts`. But Finding 1 means the PR's central claim ("no longer keeps stale skill access") isn't fully closed — the actual skill-grant write path remains unvalidated and directly abusable — and Finding 2 is a verified regression for a real slice of custom fighter data. Both look addressable without redesigning the approach.

---

joeseos:
## Follow-up Code Review

@Djidiouf Reviewed the delta since my last pass (`8cecca27` → `ca69f2e9`, 4 files, +62/-33), plus the resulting file contents in full. Cross-checked the edition-fallback and skill-access claims against the live prod database again.

### Status of previous findings

| # | Status | Evidence |
|---|--------|----------|
| **1 — skill-access validation bypass (High)** | **Resolved for the archetype flow.** `updateFighterDetails` (`app/actions/edit-fighter.ts:1589-1637`) now derives `fighter_skill_access_override` rows itself, server-side, from the just-validated `archetypeIdForOverrides`. The client no longer calls `saveFighterSkillAccessOverrides` from the archetype-save path at all — `fighter-edit-modal.tsx` dropped that ~20-line block entirely. (See new finding #2 below for a related, pre-existing, out-of-scope note.) |
| **2 — add-fighter edition fallback regression (Medium)** | **Resolved.** `add-fighter.ts:449-451` now does `editionSlugFromJoin(fighterSource.editions) ?? gangEditionSlug(gangData)`, with the `gangs` query extended to select the gang-type/custom-gang-type edition join, mirroring `edit-fighter.ts`'s existing fallback chain. Verified against prod: the 34 `custom_fighter_types` rows with null `edition_id` still exist, but only 5 of 84,479 gangs have an unresolvable gang-level edition — the fallback resolves correctly for essentially all real gangs that own these legacy types. |
| **3 — client-side race (Low/Medium)** | **Resolved, by construction.** The entire client-side "read react-query cache → map skill_access → call `saveFighterSkillAccessOverrides`" block was deleted. Overrides are now derived and written inside the same server action call as the archetype write, so there's no longer a client-side cache in the loop for this path — the described race can't occur. |
| **4 — catalog priority order (Low, note)** | Comment expanded to assert Leader > Champion > Ganger matches the Outcasts skill-access tables. Comment-only change — still worth a ruleset-owner sign-off, no longer a silent assumption. |

### New findings

**1. [Low] Silent swallowing of override-application failures removes the user-facing error signal.**
`app/actions/edit-fighter.ts:1608-1637` — previously, a failed `saveFighterSkillAccessOverrides` call was caught client-side and surfaced as `toast.error('Fighter updated but skill access save failed...')`. Now, if the archetype fetch, override delete, or override insert fails inside `updateFighterDetails`, it's only `console.error`'d server-side — the client always shows `toast.success('Fighter updated successfully')` regardless. `selected_archetype_id` persists correctly, but the skill-access grant can silently fail to apply with no indication to the user. Narrow window (requires a DB error, or the archetype row being deleted between the eligibility check and the later re-fetch) and low severity, but a real regression in error visibility. Consider surfacing a partial-success warning (similar to `archetypeSkillAccessWarning` already used in `add-fighter.ts`) instead of only logging server-side.

**2. [Informational, not a bug, out of scope for this PR]** `saveFighterSkillAccessOverrides` (`app/actions/fighter-skill-access.ts`) is untouched by this commit and remains directly callable with an arbitrary payload, gated only by `canEdit` — it's still used by the separate "Customise Skill Set Access" modal (`skill-access-modal.tsx`), independent of archetype eligibility. This looks like a pre-existing, intentional manual-override feature (it supports `denied`, which archetypes never produce) rather than something this PR was trying to close, so not a blocker — just worth a one-line confirmation from whoever owns that feature that unrestricted manual overrides there are intended.

**3. Checked and ruled out:** `mapArchetypeSkillAccessToOverrides` can emit `access_level: 'allowed'`, which looked like it might violate the CHECK constraint in migration `20260203143002_create_fighter_skill_access_override.sql` (which only lists `primary/secondary/denied`). Confirmed against the live prod schema that the actually-deployed constraint does include `'allowed'` — the migration file is just out of sync with `schema.public.sql` (pre-existing repo drift, unrelated to this PR, not reachable today anyway since all current `skill_access_archetypes` rows only use `primary`/`secondary`).

### Verdict: **Approve**

All three substantive findings from the previous review (skill-access bypass, edition-fallback regression, client-side race) are resolved, and the fix follows the existing `edit-fighter.ts`/`add-fighter.ts` patterns consistently. The one new item (silent failure swallowing) is low severity and narrow-window — worth a follow-up, not a blocker.

---

PR #1969, opened 2026-08-10
Hide the vehicle tab if n26

joeseos:


joeseos:
## Code Review

@joeseos Reviewed the diff (`components/gang/gang-page-content.tsx`, `components/gang/gang.tsx`) against the codebase, with particular attention to whether the edition check reintroduces the "null edition_slug for custom gang/fighter types" bug that has bitten a few other recent N26 PRs (#1964, #1968).

### What's good
- Correctly uses the existing `hasVehicles(edition_slug)` capability accessor from `types/edition.ts` rather than a hand-rolled `=== 'n26'` string check — consistent with how `hasAlignment`/`hasTradePoints`/etc. already gate other N23/N26 differences elsewhere in `gang.tsx` and `gang-edit-modal.tsx`.
- **Avoids the recurring custom-gang-type edition bug.** `gangData.processedData.edition_slug` is populated via `gangEditionSlug(data)`, which already resolves through `gang_types` / `custom_gang_types` (the underlying query selects `custom_gang_type_edition:custom_gang_types!custom_gang_type_id(editions:edition_id(slug))`), so this PR inherits an edition-resolution path that's already custom-gang-type-aware. It doesn't repeat the null-edition bug found in #1964/#1968.
- Refactoring the tab list into a `tabs` array (`title`/`icon`/`content`) that's `.map()`'d into the three parallel arrays the `Tabs` component expects is safer than the old hand-maintained parallel arrays — a conditional entry can't get out of sync across title/icon/content.
- Vehicles have no dedicated route under `app/gang/[id]/*` — vehicle UI only exists inside the tab's client-rendered content — so hiding the tab doesn't leave a dangling, still-reachable URL. A pure UI-visibility change is sufficient here.
- Condition polarity checked and correct: `showVehicleRoster = !hasVehicles(edition_slug)` hides the tab only for N26, not N23 or an unresolved edition.
- The `wealthBreakdownRows` change (excluding "Vehicles (without crew)" for `hasVehicles` editions) is consistent with the data model — N26 vehicles are fighter types and never populate the separate `vehicles` table, so the excluded row is structurally always 0 for N26 and no real credit value is silently dropped.

### Findings

No High or Medium severity issues.

- **Low, non-blocking** — a gang's type (and thus edition) isn't editable post-creation through any current user-facing flow, so "a gang already has vehicles from before switching to N26" isn't reachable today. Worth remembering if a future gang-type-change action is ever added: hiding the tab would also silently hide any leftover `vehicles` rows and their value from the wealth breakdown rather than surfacing them.
- **Low, pre-existing, not introduced here** — `Tabs` keeps `activeTab` as a raw index with no reset if the tab-list length changes while mounted. Not exploitable today since the tab set is effectively static for the life of a mounted `GangPageContent`, but worth remembering if a future change ever makes the tab set change without a remount.

### Verdict: **Approve**

The edition check resolves correctly at the gang level (including custom gang types), the polarity is right, there's no dangling vehicle route, and the wealth-breakdown change matches the data model. No blockers.

---

PR #1970, opened 2026-08-10
Added edition gating to equipment selection in admin

joeseos:


joeseos:
## Code Review

@joeseos Reviewed the full diff — a single line in `components/admin/admin-edit-fighter-type.tsx`.

### Change
`AdminFighterEquipmentSelection` (the fighter-type default/optional equipment picker in the admin fighter-type modal) was passed the raw, unfiltered `equipment` array instead of `filteredEquipment` — the edition-scoped `useMemo` (`editionId ? equipment.filter(item => item.edition_id === editionId) : equipment`, `admin-edit-fighter-type.tsx:340-342`) that every other equipment selector in this same component already uses (gang-specific cost picker at line 2173, adjusted-cost picker at line 2366, etc.).

### Findings

**Correctness — confirmed fix, no issues found.**
- `filteredEquipment` is already in scope in this component (defined once via `useMemo`, depends on `[equipment, editionId]`), so this is a pure one-line swap with no new state, imports, or props needed.
- This brings the "Equipment Selection" (default/optional equipment) picker in line with the edition-gating already applied everywhere else equipment is selectable in this modal. Before this fix, admins editing an N23 fighter type could add N26-only equipment (or vice versa) as default/optional equipment — inconsistent with the rest of the form, which already blocks that for gang-specific costs and adjusted costs.
- Checked the follow-on effect in `admin-fighter-equipment-selection.tsx`: the `equipment` prop is also used to resolve display names for *already-selected* default/option equipment (`equipment.find(e => e.id === item.id)`, lines 268/409/521). Since `equipment.edition_id` can be `null` (edition-agnostic items) and the filter excludes those unless `editionId` itself is unset, a fighter type with a pre-existing default/option referencing a null-edition or other-edition equipment item would now render "Unknown Equipment" for that entry. This is the same trade-off already accepted by every other selector in this file that uses `filteredEquipment` (line 340 predates this PR) — not a new risk introduced here, so not a blocker, just worth knowing if "Unknown Equipment" shows up on an existing fighter type after this ships.

### Verdict: **Approve**

Minimal, correctly scoped, and consistent with the established filtering pattern already used throughout this component.

---

PR #1971, opened 2026-08-10
Added lasting damages to n26 vehicles

joeseos:


joeseos:

## Findings

### 🔴 High — Feature is non-functional: missing N26 seed migration
**`utils/dice.ts:1810`**

`VEHICLE_DAMAGE_TABLE_N26`'s comment states each `effect_name` must exactly match rows inserted by `supabase/migrations/20260810120000_seed_n26_vehicle_lasting_damages.sql` — but that file does not exist anywhere in `supabase/migrations`. There's no migration seeding `fighter_effect_types` rows for N26 vehicle lasting damage, and no migration adding the corresponding `fighter_effect_categories` row that `supabase/functions/add_fighter_injury.sql` now expects.

**Impact:** for an N26 vehicle-fighter, `GET /api/vehicles/lasting-damage?edition_slug=n26` returns an empty list, the D66 roller resolves a name with no matching catalog row, and `applyRolledDamage`'s `if (!match) return;` silently no-ops. The Combobox has nothing to select — the entire N26 side of this PR doesn't work until the seed migration is added.

### 🟡 Medium — Fighter status not refreshed after deleting a fatal lasting damage
**`components/fighter/vehicle-lasting-damages.tsx:352`**

`removeDamageMutation`'s `onSuccess` invalidates queries and shows a toast, but never propagates `killed_status` from `deleteFighterInjury` to `onFighterStatusUpdate` — unlike the analogous `deleteInjuryMutation` in `fighter-injury-list.tsx`, which does call `onInjuryUpdate(..., result.killed_status)`.

**Impact:** deleting a fatal lasting damage (e.g. "Catastrophic Explosion!") correctly clears `fighters.killed` in the DB, but the fighter card/page still shows the vehicle as killed/destroyed until a full page reload.

### 🟡 Medium — No confirmation step before sending fighter to Recovery
**`components/fighter/vehicle-lasting-damages.tsx:606`**

`handleAddDamage` computes `sendToRecovery` and calls `addDamageMutation.mutate` directly on the same click, submitting immediately. The equivalent flow in `fighter-injury-list.tsx` opens a Recovery confirmation modal first (`requiresRecovery && !fighterRecovery`) and only submits after the user confirms.

**Impact:** selecting an N26 damage with the Recovery flag sends the fighter to Recovery instantly with no chance to cancel — inconsistent with the established UX for the same status change elsewhere in the app.

## Verdict: Changes requested

Blocker: the missing seed migration (high) makes the feature non-functional as shipped. The two medium findings are real UX/state regressions worth fixing before merge but aren't blocking on their own.

---

PR #1972, opened 2026-08-10
Seed the N26 vehicle lasting damage catalog

joeseos:
Adds the 14 D66 rows for N26 Vehicle Lasting Damage under the existing global 'lasting damages' category, scoped by edition_id, plus the four characteristic modifiers for the 51-56 band.

Companion to #1971, which added the app-side support. Split out so the data seed lands as its own migration.

Already applied to the remote by hand; this restores the file to the repo so the seed is reproducible in a fresh environment. Verified against the applied rows: all 14 names and type_specific_data payloads match, and the NOT EXISTS guard makes a re-run a no-op.

### DIFF
--- a/supabase/migrations/20260810120000_seed_n26_vehicle_lasting_damages.sql
+++ b/supabase/migrations/20260810120000_seed_n26_vehicle_lasting_damages.sql
@@ -1,0 +1,119 @@
+-- Seed the N26 Vehicle Lasting Damage catalog.
+--
+-- Categories are global (no edition_id), so these join the same 'lasting damages'
+-- category as the six N23 damages and are distinguished by edition_id. N23 rows are
+-- left untouched.
+--
+-- N26 makes the vehicle a fighter, so it rolls D66 rather than D6 and modifies ordinary
+-- fighter characteristics, with the Recovery / Captured / destroyed outcomes of the N26
+-- injury table. The matching table is VEHICLE_DAMAGE_TABLE_N26 in utils/dice.ts;
+-- effect_name must match it exactly, since the UI looks the range up by name.
+--
+-- Names deliberately collide with N26 injuries ('Lesson Learnt', the Enmity trio,
+-- 'Captured') and with the N23 repair table ('Superficial Damage'). Separate rows.
+--
+-- Modifiers are seeded here, unlike 20260806160000_seed_n26_lasting_injuries.sql: there
+-- is no admin UI for lasting damages, so SQL is the only way in.
+--
+-- SIGN CONVENTION — BS is a target number on N26 but M/T/W are raw values, so 'reduces
+-- its Ballistic Skill by 1' is +1 while 'reduces its Movement/Toughness/Wounds by 1' is
+-- -1. See N26_FIGHTER_LIMITS in utils/characteristicLimits.ts.
+--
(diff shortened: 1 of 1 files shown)


joeseos (commented):
@joeseos Reviewed this against the codebase (and cross-checked against the companion injuries migration and the app's live consumers of `fighter_effect_types`/`fighter_effect_type_modifiers`). Everything checks out — this is a clean, well-documented seed migration. (Note: submitting as a comment, not an approval, since the review tool won't let this account approve its own PR.)

**Verified:**
- **Structure vs. companion migration** (`20260806160000_seed_n26_lasting_injuries.sql`): same columns, same `NOT EXISTS` guard pattern keyed on `(effect_name, category_id, edition_id)`, same string-boolean JSON convention. The one difference (seeding modifiers inline here vs. by-hand-in-admin there) is because there's genuinely no admin UI for the `lasting damages` category — confirmed by grepping `components/admin/*`.
- **`effect_name` vs `VEHICLE_DAMAGE_TABLE_N26`** (`utils/dice.ts:278-293`): all 14 names match exactly, including `Bitter Enmity` going through the shared `BITTER_ENMITY_EFFECT_NAME` constant so it can't drift from the N23/injuries usage.
- **`type_specific_data` keys** (`recovery`, `convalescence`, `captured`, `killed`, `hatred_target`, `effect_selection`): all are real keys with real consumers (`vehicle-lasting-damages.tsx`, `fighter-status.ts`, `injuryTarget.ts`, `fighter-effect-selection.tsx`), and the string-boolean format (`"true"`/`"false"` as JSON strings, not JSON booleans) matches what those consumers expect. The `hatred_target` values per row also match the independently-seeded `20260806170000_add_hatred_target_to_enmity_injuries.sql`.
- **Sign convention (the highest-risk claim)**: `N26_FIGHTER_LIMITS.BS` (`utils/characteristicLimits.ts:33`) is still `['2+','6+']`, i.e. still a target-number stat post-N26 (unlike Ld/Cl/Wil/Int, which became raw values). So `ballistic_skill: +1` for Busted Sights correctly worsens BS, while `movement/toughness/wounds: -1` correctly worsens those raw-value stats. `calculateAdjustedStats` in `utils/effect-modifiers.ts` applies these with plain addition, no per-stat special-casing, so the claim holds.
- **No unique-constraint collision risk**: `fighter_effect_types` has no unique constraint on `effect_name` (PK on `id` only, plus a non-unique index), so the deliberate name collisions with N26 injuries and the N23 repair table are safe.
- **`operation` column**: plain `text`, no CHECK constraint, and `m.operation || 'add'` in `utils/effect-modifiers.ts` confirms the null-defaults-to-add claim.

**Nit (low severity, non-blocking):** the comment block says "`applyModifiers` defaults a null operation to `'add'` anyway" — there's no function by that name in the codebase; the actual function is `applyNumericModifiers` in `utils/effect-modifiers.ts:39`. The described behavior is correct, just misattributed to a nonexistent identifier. Purely cosmetic, no functional impact — feel free to fix in a follow-up or ignore.

No correctness, security, or performance concerns found. Migration is idempotent (guarded `NOT EXISTS` on both inserts) and self-contained (single new file, no app code changes needed since #1971 already wired up the consuming code).

**Verdict: Approve** (functionally — this account can't cast a formal GitHub approval on its own PR, but there are no blockers here).

---

PR #1973, opened 2026-08-11
Add N26 edition gating for campaign types and territories.

Djidiouf:
Filter and rank campaign catalogs by edition, reject cross-edition territory adds, and align campaign ranking/cache invalidation with existing alliance and edition helpers.

### DIFF
--- a/app/actions/campaigns/[id]/campaign-territories.ts
+++ b/app/actions/campaigns/[id]/campaign-territories.ts
@@ -2,10 +2,11 @@
 
 import { createClient } from "@/utils/supabase/server";
 import { revalidateTag } from "next/cache";
-import { CACHE_TAGS } from "@/utils/cache-tags";
+import { CACHE_TAGS, purgePreEditionCampaignCatalogCachesOnce } from "@/utils/cache-tags";
 import { logTerritoryLost, logTerritoryClaimed } from "../../logs/gang-campaign-logs";
 import { getAuthenticatedUser } from '@/utils/auth';
 import { checkCampaignArbitrator } from '@/utils/user-permissions';
+import { editionsConflict, editionSlugFromJoin } from '@/types/edition';
 
 export interface AssignGangToTerritoryParams {
   campaignId: string;
@@ -21,7 +22,6 @@ export interface RemoveGangFromTerritoryParams {
 export interface AddTerritoryParams {
   campaignId: string;
   territoryId: string;
-  territoryName: string;
 }
 
 export interface CreateCustomCampaignTerritoryParams {
@@ -248,33 +248,67 @@ export async function removeGangFromTerritory(params: RemoveGangFromTerritoryPar
 }
 
 /**
- * Add a territory to a campaign with targeted cache invalidation
+ * Add a territory to a campaign with targeted cache invalidation.
+ * Only campaign owners, arbitrators, and system admins may call this.
  */
 export async function addTerritoryToCampaign(params: AddTerritoryParams) {
   try {
     const supabase = await createClient();
-    const { campaignId, territoryId, territoryName } = params;
+    const { campaignId, territoryId } = params;
 
     if (!territoryId) {
       throw new Error('Territory ID is required');
     }
 
+    const user = await getAuthenticatedUser(supabase);
+
+    const hasPermission = await checkCampaignArbitrator(user.id, campaignId);
+    if (!hasPermission) {
+      return { success: false, error: 'Only campaign owners and arbitrators can add territories' };
+    }
+
+    const [{ data: campaign, error: campaignError }, { data: templateTerritory, error: templateError }] =
+      await Promise.all([
+        supabase
+          .from('campaigns')
+          .select('campaign_types!campaign_type_id(editions:edition_id(slug))')
+          .eq('id', campaignId)
+          .single(),
+        supabase
+          .from('territories')
+          .select('territory_name, playing_card, editions:edition_id(slug)')
+          .eq('id', territoryId)
+          .maybeSingle(),
+      ]);
+
+    if (campaignError || !campaign) {
+      return { success: false, error: 'Campaign not found' };
+    }
+    if (templateError) {
+      console.error('Error fetching territory template:', templateError);
+      return { success: false, error: 'Failed to load territory template' };
+    }
+    if (!templateTerritory) {
+      return { success: false, error: 'Territory not found' };
+    }
+
+    const campaignEdition = editionSlugFromJoin((campaign as any).campaign_types?.editions);
+    const territoryEdition = editionSlugFromJoin((templateTerritory as any).editions);
+
+    if (editionsConflict(campaignEdition, territoryEdition)) {
+      return {
+        success: false,
+        error: 'This territory is from a different edition than the campaign',
+      };
+    }
+
     const insertData: Record<string, unknown> = {
       campaign_id: campaignId,
       territory_id: territoryId,
-      territory_name: territoryName
+      territory_name: templateTerritory.territory_name
     };
 
-    const { data: templateTerritory, error: templateError } = await supabase
-      .from('territories')
-      .select('playing_card')
-      .eq('id', territoryId)
-      .maybeSingle();
-
-    if (templateError) {
-      console.error('Error fetching territory template for playing_card:', templateError);
-    }
-    const rawCard = templateTerritory?.playing_card;
+    const rawCard = templateTerritory.playing_card;
     insertData.playing_card =
       typeof rawCard === 'string' && rawCard.trim() ? rawCard.trim() : null;
 
@@ -286,6 +320,7 @@ export async function addTerritoryToCampaign(params: AddTerritoryParams) {
 
     revalidateTag(`campaign-territories-${campaignId}`, { expire: 0 });
     revalidateTag(`campaign-${campaignId}`, { expire: 0 });
+    purgePreEditionCampaignCatalogCachesOnce();
 
     return { success: true };
   } catch (error) {
--- a/app/actions/create-campaign.ts
+++ b/app/actions/create-campaign.ts
@@ -3,7 +3,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { getAuthenticatedUser } from "@/utils/auth";
 import { revalidateTag } from 'next/cache';
-import { CACHE_TAGS, invalidateCampaignCount } from '@/utils/cache-tags';
+import { CACHE_TAGS, invalidateCampaignCount, purgePreEditionCampaignCatalogCachesOnce } from '@/utils/cache-tags';
 
 interface CreateCampaignParams {
   name: string;
@@ -54,6 +54,9 @@ export async function createCampaign({ name, campaignTypeId, trading_posts }: Cr
 
     // Invalidate global campaign count
     invalidateCampaignCount();
+
+    // Drop pre-rename catalog cache keys that share campaign-types / territories tags
+    purgePreEditionCampaignCatalogCachesOnce();
 
     // Invalidate user's campaigns list so the new campaign appears immediately
     revalidateTag(CACHE_TAGS.USER_CAMPAIGNS(user.id), { expire: 0 });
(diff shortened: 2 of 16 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed the full diff against `main`, including the SQL query shapes, the `types/edition.ts` helpers this PR builds on, and every downstream consumer of `getCampaignTypes()`/`getAllTerritories()`.

### What's good
- **`addTerritoryToCampaign` previously had no authorization check at all** — any authenticated user could call the server action directly with an arbitrary `campaignId` and insert a territory into someone else's campaign. This PR closes that gap with `checkCampaignArbitrator`, using the exact same pattern (and matching error message style) already established by the sibling `createCustomCampaignTerritory` in the same file — good consistency, and a real security fix beyond the PR's stated scope.
- **Edition-conflict gating uses the right primitive.** `editionsConflict` (fails open on an unresolved slug) is used for the server-side reject in `addTerritoryToCampaign`, while `sameEditionForDisplay` (nulls treated as `n23`) is used for client-side list filtering — exactly the split documented in `types/edition.ts`'s own comments, not a copy-paste of the wrong helper.
- **`campaign_types!campaign_type_id(...)` join hint is correct and necessary-in-spirit**: confirmed via schema that `campaigns` has only one FK to `campaign_types`, so this isn't ambiguous, but it's harmless and matches the disambiguation style used elsewhere (e.g. `fighters!gang_id` in the advancement code).
- **The cache-key rename is handled correctly.** `getCampaignTypes`/`getAllTerritories` change their `unstable_cache` key (to bust stale entries missing `edition_slug`) while keeping the same tag string (`GLOBAL_CAMPAIGN_TYPES() === 'campaign-types'`, the old literal) — so `purgePreEditionCampaignCatalogCachesOnce()`'s `revalidateTag` call actually reaches pre-deploy cache entries tagged under the old string. The one-shot module-level flag is a reasonable, low-risk way to avoid firing this on every request after the first purge in a given process.
- Traced `edition_slug` end-to-end for both new props: `getCampaignTypes`/`getAllTerritories` → `CampaignPageContent` → `TerritoryList`/`CampaignAddTerritoryModal`, and `campaignData.edition_slug` is populated from `_getCampaignBasic`'s existing `editionSlugFromJoin` resolution. No dangling `undefined` at any hop.
- `AddTerritoryParams.territoryName` removal is fully threaded through — no remaining caller still passes it, and the server now sources `territory_name` from the trusted `territories` table row instead of an unvalidated client-supplied string (a minor correctness/integrity improvement on top of the edition gating).

### Findings

**1. [Low/nit] Redundant rank-table lookup inside the sort comparator**
`components/create-campaign.tsx:249-262`
```tsx
.sort((a, b) => {
  const typeA = a.campaign_type_name.toLowerCase();
  const typeB = b.campaign_type_name.toLowerCase();
  const campaignRank = getCampaignRank(editionSlug);   // recomputed per comparison
  ...
```
`getCampaignRank(editionSlug)` is called inside the comparator, so it re-runs (and rebuilds/looks up the `Record`) on every pairwise comparison during the sort, rather than once. The sibling file this PR also touches, `campaign-add-territory-list.tsx:56`, does it correctly (hoisted once above the `.sort()` call). Harmless at this list size (a handful of campaign types), but worth hoisting for consistency with the pattern the PR itself establishes elsewhere.

### Verdict: **Approve**

No correctness or security issues found — if anything, this closes a pre-existing missing-authorization gap in `addTerritoryToCampaign`. The edition-gating logic (server reject + client-side filtering) is applied consistently and uses the right helper for each context. The one finding above is a non-blocking nit.

---

joeseos:
## Review: edition-gating conventions & caching

Two questions: does this follow the conventions in `types/edition.ts` / the rest of the app, and is the caching consistent with how the app does caching. Short version: **edition gating is on-convention with two reuse nits; the caching has one mechanism that should be removed rather than fixed.**

---

### 1. Edition gating — follows the conventions

- **Helper choice matches the contract documented in `types/edition.ts`**: `sameEditionForDisplay` for the two display filters (`campaign-add-territory-list.tsx`), `editionsConflict` for the server-side guard in `addTerritoryToCampaign`. That is exactly the split the header comments prescribe, and matches `create-gang-modal.tsx:123`, `FighterAddModal.tsx:216`, `campaign-edit-modal.tsx:113`.
- **Join flattening goes through the shared helpers** (`withEditionSlug` / `editionSlugFromJoin`) rather than hand-reading `.editions?.slug` — and it fixes an existing hand-rolled read at `app/lib/campaigns/[id]/get-campaign-data.ts:55` on the way past.
- **`getCampaignRank` + `campaignRankN23`/`campaignRankN26` is a faithful clone of the sibling registries** (`utils/allianceRank.ts`, `utils/skillSetRank.ts`) — same `Record<EditionSlug, …>` compile-error-on-new-edition trick, same `{}` fallback, same doc comment. Textbook.
- **No new `EDITION_CAPABILITIES` row, correctly** — this is content scoping (rows carry `edition_id`), not a behavioural capability. Adding a row would have been wrong.
- Worth calling out: `addTerritoryToCampaign` had **no permission check at all** before this PR, and `territory_name` is now derived server-side from the template row instead of trusted from the client. Both are real hardening fixes riding along.

#### Nit 1 — the campaign-edition lookup is inlined instead of reused

`addTerritoryToCampaign` copy-pastes the `campaigns → campaign_types → editions` select and the `editionsConflict` branch out of `app/actions/campaigns/[id]/assert-gang-campaign-edition.ts:14-46`, which is the established home for "does X conflict with this campaign's edition".

Suggest adding a `getCampaignEditionSlug(supabase, campaignId)` there and having both `assertGangMatchesCampaignEdition` and this action use it. (A sibling `assertTerritoryMatchesCampaignEdition` also works, but the territory row is needed for `territory_name`/`playing_card` anyway, so the slug-only helper composes better with the existing `Promise.all`.) If that file grows a second export, `assert-campaign-edition.ts` is a better name.

#### Nit 2 — `getCampaignRank` is called inside the sort comparator

`components/create-campaign.tsx:252` calls it inside the comparator, so it re-runs O(n log n) times per render. All seven existing call sites hoist it (`const skillSetRank = getSkillSetRank(editionSlug);`), and this PR does it correctly in `campaign-add-territory-list.tsx:56`. Just move it above the `.sort(`.

#### Note only, no action

`sameEditionForDisplay(null, …)` reads a missing slug as N23, but `getCampaignRank(null)` returns `{}`. A campaign type with `edition_id = NULL` (both admin POST handlers allow `edition_id: edition_id || null`) therefore shows in the N23 list but sorts alphabetically. This is exactly how `getSkillSetRank`/`getAllianceRank` already behave and their doc comments call it intentional, so it is on-convention.

---

### 2. Caching — admin invalidation is right, `purgePreEditionCampaignCatalogCachesOnce` should go

`invalidateCampaignCatalogLists()` follows the `invalidateUserCustomizations()` shape in `utils/cache-tags.ts:367`, and wiring it into the four admin POST/PATCH handlers **fixes a live bug**: `getCampaignTypes`/`getAllTerritories` use `revalidate: false`, and nothing in the repo ever revalidated `campaign-types` or `global-territories-list` (`invalidateUserCustomizations` currently has zero call sites). Admin catalog edits were invisible until a redeploy. This is the most valuable part of the PR.

The purge mechanism is the problem, and I'd argue for deleting it rather than fixing it:

1. **It solves a non-problem.** The `unstable_cache` *keys* changed (`campaign-types` → `campaign-types-with-edition`, `territories-list` → `territories-list-with-edition`) while the *tags* stayed byte-identical. A renamed key is already a complete migration — the old entries are orphaned, never read again, and age out on their own.
2. **It causes a real one.** Because the tags are shared, the purge doesn't only drop the orphans — it expires the **new** entries too. And the guard is a module-level boolean, i.e. once per *process*, not once globally, so every cold start / new serverless instance re-nukes the global campaign-type and territory caches for all users.
3. **The `after()` wrapper is the tell.** On `app/campaigns/[id]/page.tsx` it has to be wrapped in `after()` precisely because Next forbids `revalidateTag` during render. Cache invalidation on a read path is worth a second look regardless.
4. **Two of the four call sites invalidate data the mutation doesn't touch.** `createCampaign` and `addTerritoryToCampaign` change neither the global territory template list nor the campaign-type catalog. Every other helper in `cache-tags.ts` (`invalidateCampaignTerritory`, `invalidateFighterAddition`, …) invalidates exactly what its mutation changed.
5. **Minor:** `didPurgePreEditionCampaignCatalogCaches = true` is set *before* `invalidateCampaignCatalogLists()` runs, so a throw leaves it permanently "done".

Concretely: drop the flag + function from `utils/cache-tags.ts` and its five touch points (`campaigns/[id]/page.tsx` incl. the `after`/`next/server` import, `api/campaigns/[campaignId]/route.ts`, `create-campaign.ts`, `campaign-territories.ts`), and keep `invalidateCampaignCatalogLists()` in the admin routes.

#### Tag naming

`GLOBAL_CAMPAIGN_TYPES: () => 'campaign-types'` breaks the section-6 convention — every other global tag is `global-*` (`global-gang-types`, `global-territories-list`, `global-editions`). The old literal was only kept so the purge could reach pre-rename entries; with the purge gone it can just be `global-campaign-types`. Safe, since the cache key changed in the same PR so no live entry depends on the old tag string.

#### Small gap

Neither admin route has a DELETE handler today, so nothing is missing right now — but one added later will need `invalidateCampaignCatalogLists()` too.

---

joeseos:
## Corrections to my review above

I re-verified the claims in my previous comment against the current head (`a9b07c9`) and the Next.js 16.2.9 source. Six corrections, one of them substantive.

### Retracted — Nit 2 was already fixed

I reviewed `3d957a8`; `a9b07c9` had already hoisted `getCampaignRank` out of the sort comparator (`create-campaign.tsx:87`). **Please disregard Nit 2 entirely** — it is not an outstanding issue.

### Corrected details

| Claim | Correction |
|---|---|
| "all seven existing call sites hoist it" | **Nine** call sites in `components/` (`getSkillSetRank` ×9). They do all hoist — the count was wrong, not the pattern. |
| "its five touch points" | **Four** call sites (`campaigns/[id]/page.tsx:34`, `api/campaigns/[campaignId]/route.ts:31`, `create-campaign.ts:59`, `campaign-territories.ts:323`) plus the definition at `cache-tags.ts:389`. |
| `cache-tags.ts:367` | `invalidateUserCustomizations` is at **368** at this head. |
| `assert-gang-campaign-edition.ts:14-46` | The function spans **9-48**. |
| "orphaned … and age out on their own" | Imprecise. `revalidate: false` maps to `CACHE_ONE_YEAR_SECONDS` (`unstable-cache.ts:49`), so ~1 year, not near-term. The load-bearing half of that claim — *never read again* — is confirmed below. |

### Confirmed, with a stronger basis than I gave

- **`revalidateTag` during render genuinely throws.** `revalidate.ts:136-137`: `Route ${store.route} used "${expression}" during render which is unsupported.` So the `after()` wrapper on `campaigns/[id]/page.tsx` is load-bearing, not stylistic — which was my point about it being a tell.
- **The keyParts rename was redundant, which makes the purge doubly unnecessary.** `unstable-cache.ts:96` builds the cache key as:
  ```ts
  const fixedKey = `${cb.toString()}-${Array.isArray(keyParts) && keyParts.join(',')}`
  ```
  The key includes **the callback source**. Both callbacks changed in this PR (the `editions:edition_id (slug)` select and the `withEditionSlug` mapping), so the cache key changed on its own — renaming `campaign-types` → `campaign-types-with-edition` and `territories-list` → `territories-list-with-edition` bought nothing. The old entries were already unreachable before the rename.

Everything else in the original comment holds as written: tags are byte-identical so the purge expires the new entries too; the module-level flag makes it once-per-process rather than once-globally; two of the four call sites invalidate catalogs their mutation doesn't touch; the flag is set before the call; `invalidateUserCustomizations` still has zero call sites (`cache-tags.ts` is its only occurrence); `GLOBAL_CAMPAIGN_TYPES` is the sole non-`global-*` entry among the eleven global tags; both admin POSTs allow `edition_id: edition_id || null`; neither admin route has a DELETE handler; and `addTerritoryToCampaign` had no permission check at all on `f028e8b`.

So the outstanding items are: **remove the purge mechanism**, **rename the tag to `global-campaign-types`**, and optionally **extract `getCampaignEditionSlug` into `assert-gang-campaign-edition.ts`**. The `invalidateCampaignCatalogLists()` wiring should stay.

---

PR #1976, opened 2026-08-11
Adding Gang ranking for N26

Djidiouf:


joeseos (approved):
@Djidiouf Reviewed this PR against the codebase (diff, related component/type files, and the live `gang_types` table for edition `n26`).

## Summary
Clean, well-scoped change: splits the single `gangListRank` table into edition-specific tables (`gangListRankN23.ts` / `gangListRankN26.ts`) accessed via `getGangListRank(editionSlug)`, and swaps the native `<select>` for the existing `Combobox` component in `create-gang-modal.tsx`, preserving the header/category grouping behavior.

## Checks performed
- **Correctness of N26 rank keys**: queried the live `gang_types` table for `edition_slug = 'n26'` and confirmed every key in `gangListRankN26.ts` (`chaos helots`, `ironhead squats`, `the malstrain`, `free ogryns`, `spyre hunters`, `outcasts`, etc.) matches the actual seeded `gang_type` strings case-insensitively. No silent fallback to "Misc." for any renamed N26 gang type.
- **`getGangListRank` typing**: `GANG_LIST_RANK_BY_EDITION` is keyed by `EditionSlug` (`'n23' | 'n26'`), matching the literals in `types/edition.ts` — an unhandled new edition would be a compile error, as the comment intends.
- **No dangling references**: confirmed no other file still imports the old bare `gangListRank` export; `create-gang-modal.tsx` is the only consumer and now calls `getGangListRank(editionSlug)`.
- **`Combobox` integration**: `options`/`value`/`onValueChange`/`disabled` all match the component's prop surface; disabled category headers (`header-${category}`) are correctly skipped in `handleSelect` and can't collide with real `gang_type_id` (uuid) values.
- **CI**: `deploy-preview` check is green.

## Minor, non-blocking observations
1. **`components/create-gang-modal.tsx` (Combobox usage, ~L483-490)**: the previous `<select>` had an explicit "Select gang type" empty option, letting a user reset back to no selection. The new `Combobox` isn't passed `clearable`, so once a gang type is picked there's no UI affordance to clear it back to empty (only re-picking another type). Low severity — form validation presumably requires a non-empty selection before submit anyway — but worth a conscious call rather than an oversight.
2. Purely cosmetic: `getGangListRank` returns `{}` for an unset/unrecognized edition rather than throwing, which silently drops ranking/grouping instead of surfacing a bug — consistent with the "sibling registries" this PR references, so not asking for a change, just flagging for awareness.

## Verdict
**Approve.** No correctness, security, or performance issues found; the N26 naming is verified against real seed data and the refactor is a straightforward, low-risk extraction.

---

joeseos:
## Code Review

@Djidiouf Reviewed the diff against `main`: the `Combobox` migration in `create-gang-modal.tsx`, the edition-split of `gangListRank.ts`, and `getGangListRank`'s call sites.

### What's good
- **Edition-scoping follows the established compile-time-safety pattern.** `GANG_LIST_RANK_BY_EDITION: Record<EditionSlug, {...}>` matches the `LIMITS_BY_EDITION` / `SKILL_SET_RANK_BY_EDITION` convention referenced in the code comment — a new `EditionSlug` member is a compile error here until it gets its own table.
- **`getGangListRank` fails safe.** No slug → quiet `{}` (still resolving). Unrecognised slug → `{}` plus a dev-only, once-per-slug `console.warn` rather than silently borrowing another edition's order or throwing. Reasonable choice.
- **No leftover references to the old named export.** `gangListRank` → `getGangListRank` is a clean rename; grepped the repo and `create-gang-modal.tsx` + `gangListRank.ts` are the only two files that ever referenced it, so nothing else breaks.
- **Old inline sort-comparator nit is not present here.** `getGangListRank(editionSlug)` is called once at the top of the `gangTypeOptions` `useMemo` and closed over by `.sort()`/`.reduce()`, not re-invoked per comparison — worth calling out since a sibling PR (#1973) flagged exactly this mistake in a different file.
- **N26 table correctly drops entries that don't exist in that edition** (no "Badzone Enforcers" row) rather than reusing the N23 list with stale entries.
- **`Combobox` reuse for grouped/disabled header rows is consistent** with the component's existing "disabled = non-selectable header" support — this isn't a new pattern being introduced.

### Findings

**1. [Low] Transient sort order while `editionSlug` is still resolving**
`components/create-gang-modal.tsx` (`gangTypeOptions`)
If `editionSlug` is briefly falsy before it resolves, `getGangListRank` returns `{}`, every gang type ranks `Infinity`, and the stable sort falls back to insertion order (not alphabetical) under a single "Misc." group for one render. Cosmetic and self-corrects once the slug loads — not worth blocking on, just flagging in case a flash of unsorted/ungrouped options is visible in practice.

### Verdict: **Approve**

No correctness, security, or performance issues found. The edition split is consistent with the codebase's existing conventions, and the `Combobox` migration is a straightforward, well-scoped UI improvement over the native `<select>`.

---

PR #1977, opened 2026-08-11
Remove the N26 "Level Up" badge from the fighter cards

joeseos:
The badge counts Advancements earned but not yet taken, and it is currently reporting numbers that were never earned. The migration that introduced fighters.starting_xp backfilled every pre-existing row to a flat 0, while fighters.xp kept the recruitment value already seeded from the fighter type. A Champion recruited on 13 XP therefore reads as "recruited on 0, now on 13" and the badge counts four rank tiers crossed for a fighter that has not played a game. 15 of the 16 affected N26 fighters carry starting_xp = 0.

Pulled from both places it renders: the fighter page's details card and the gang page's fighter cards. openAdvancements had no other use in either file, so the derivation and its import go with it.

utils/advancementRanks.ts is left alone. openAdvancementsFor still caps how many Advancements the add-advancement action will accept, so removing the display does not correct that path -- affected fighters can still claim Advancements they have not earned. That needs the data backfill, which is deliberately not part of this change.

### DIFF
--- a/components/fighter/fighter-details-card.tsx
+++ b/components/fighter/fighter-details-card.tsx
@@ -9,7 +9,6 @@ import {
 } from '@/types/edition';
 import { memo } from 'react';
 import { calculateAdjustedStats } from '@/utils/effect-modifiers';
-import { countAdvancementsTaken, openAdvancementsFor } from '@/utils/advancementRanks';
 import { FighterProps, FighterEffect, Vehicle } from '@/types/fighter';
 import { TbMeatOff } from "react-icons/tb";
 import { GiHandcuffs, GiImprisoned } from "react-icons/gi";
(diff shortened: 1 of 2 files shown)


joeseos:
## Code Review

**Summary:** Clean, well-scoped removal of the N26 "Level Up" badge from `fighter-details-card.tsx` and `fighter-card.tsx`. The root cause described in the PR body checks out against `utils/advancementRanks.ts`: `advancementsEarnedFor` uses `startingXp ?? 0`, so a fighter backfilled to `starting_xp = 0` but recruited with nonzero `xp` reads as having crossed every tier between 0 and its recruitment XP, inflating the badge count.

**Verification performed:**
- Confirmed `countAdvancementsTaken` / `openAdvancementsFor` are still used server-side in `app/actions/fighter-advancement.ts` (the add-advancement cap), so removing the display-only usage doesn't touch the enforcement path — matches what the PR description claims.
- Confirmed `effects`, `skills`, `starting_xp`, and `edition_slug` remain used elsewhere in both edited files (stats table, XP display, save characteristic lookups), so no dead props or now-unused imports/variables were left behind.
- Diff is purely subtractive (removes the derived `openAdvancements` value and its JSX block plus the now-unused import) — no logic changes, no risk of regressions elsewhere.

No correctness, security, performance, or style issues found.

**Verdict: Approve.** The PR itself is honest about scope — it fixes the display symptom, not the underlying `starting_xp` backfill data or the advancement cap in `openAdvancementsFor`, which is called out explicitly as future work.

---

PR #1978, opened 2026-08-11
Scope the Advancement characteristic list to the fighter's edition

joeseos:
## The bug

The Advancements modal's "Select a Characteristic" dropdown listed every characteristic twice — Movement, Movement, Weapon Skill, Weapon Skill, and so on.

`fighter_effect_types` holds one row per characteristic **per edition**, and `effect_name` is deliberately reused across them (12 N23 rows, 13 N26 rows — the same set plus `Save`). The modal's read was unscoped, so all 25 came back.

## Cause

It was the component, not an API route. The modal skipped the app's data layer and called PostgREST straight from the browser, filtered only on a hardcoded category UUID:

```ts
fetch(`${SUPABASE_URL}/rest/v1/fighter_effect_types?fighter_effect_category_id=eq.789b2065-…`)
```

No edition filter anywhere in that path.

## Fix

No new endpoint needed. `get_fighter_available_advancements` already resolves the fighter's edition from its gang and returns only that edition's rows, and the modal **already called it twice** — once for the Ganger roll UI, once on every characteristic selection. Its `characteristics` map is keyed by `effect_name` and carries the effect type id, so it is the catalog as well as the costs.

It is now read once and feeds all three consumers: the dropdown, the cost panel behind it, and the Ganger/Exotic Beast roll UI. That removes the unscoped table read *and* a round trip per selection.

The edition filter stays in exactly one place — the RPC's `edition_id` predicate. No slug comparison enters the component and `types/edition.ts` is unchanged.

**Derived, not stored.** The list is a `useMemo` over the RPC map rather than a copy in state, so it cannot drift from the costs shown beside it, and the dropdown waits on the RPC instead of flashing empty. Deliberately uncached: `times_increased`, the escalating `xp_cost` and `has_enough_xp` all move the moment an Advancement is added.

## Save grouping

`Save` is N26-only and had no entry in `characteristicRank`, so it would have dropped into "Misc." the moment it started rendering. Ranked after Attacks to match the N26 statline (`M, WS, BS, S, T, W, I, A, Sv, Ld, Cl, Wil, Int`), with the group thresholds moved to follow.

## Dead code

`app/lib/fighter-advancements.ts` kept four exports with no callers anywhere in the repo:

| export | why it goes |
| --- | --- |
| `getAdvancementCategories` | another unfiltered copy of this same query, same hardcoded UUID |
| `getFighterAvailableAdvancements` | `unstable_cache(revalidate: false)` over this RPC with nothing revalidating its tag |
| `getAvailableSkills` | same never-invalidated cache shape, over `get_available_skills` |
| `getAvailableInjuries` | never worked — filtered on the literal string `'injury-category-id'`, `// Replace with actual ID` still in place |

Only `getGangFighters` had callers; the module is left with just that.

## Known behaviour change

Gangs whose edition does not resolve (`COALESCE(gt.edition_id, cgt.edition_id)` is NULL) now get an **empty** characteristic list rather than a duplicated one — fail-closed. Five gangs are in this state, all custom gang types with a NULL `edition_id`, 7 fighters between them. Those rows are being cleaned up separately.

## Not covered

`addCharacteristicAdvancement` (`app/actions/fighter-advancement.ts`) still looks the effect type up by the id the client sends, with no check that it belongs to the fighter's edition. Filtering the dropdown fixes the real path but leaves the action trusting its input — worth hardening separately.

## Verification

- `npx tsc --noEmit` clean project-wide; `eslint` clean on all three changed files
- The hardcoded `789b2065-…` UUID no longer appears anywhere in the codebase
- One RPC call site remains where there were three

Not yet exercised in the running app — worth a click-through: N26 fighter shows 13 characteristics with `Save` under Main Characteristics, N23 shows 12 without it, and the Ganger 2D6 roll UI still works now that its fetch is shared.


joeseos:
@joeseos Reviewed this PR against the codebase. Solid fix for the core bug (unscoped `fighter_effect_types` read causing duplicate characteristics), and the dead-code cleanup checks out — I confirmed `getAdvancementCategories`, `getFighterAvailableAdvancements`, `getAvailableSkills`, and `getAvailableInjuries` have no remaining callers anywhere in the repo.

Two findings from cross-checking against `get_fighter_available_advancements.sql` and the N26 migration:

**1. `components/fighter/fighter-advancement-list.tsx:591-595` — `plusStats` set is missing `'save'` (medium, correctness)**

```ts
const plusStats = new Set([
  'weapon_skill', 'ballistic_skill', 'initiative',
  'leadership', 'cool', 'willpower', 'intelligence'
]);
const numericValue = plusStats.has(characteristicCode) ? -1 : 1;
```

`supabase/migrations/20260810130000_n26_advancements.sql` (line 103) explicitly documents: *"Weapon Skill, Ballistic Skill, Save are target numbers -> -1"*. Since `'save'` isn't in this set, buying a Save advancement will optimistically apply `numeric_value: 1` (wrong direction) instead of `-1`, until the next server refresh corrects it. This line isn't touched by the diff, but this PR is exactly what makes Save render correctly and consistently in the main dropdown (previously it fell into "Misc." per the PR description) — so it's now much more likely to be exercised. Suggest adding `'save'` to the set.

**2. `components/fighter/fighter-advancement-list.tsx:765-809` — catalog fetch failures are now silent (medium, UX regression)**

```ts
} catch {
  // non-fatal
} finally {
  if (!cancelled) setCharacteristicsLoaded(true);
}
```

This is the single effect that now backs the characteristic dropdown itself (not just the Ganger roll UI, per `needsAdvancementCatalog`). Before this PR, a failed characteristics fetch flowed through the shared `fetchCategories` catch block and called `setError('Failed to load characteristic categories')`, which renders at line ~2743. Now `!response.ok` (line 784) just returns, and any thrown error is swallowed with no `setError` call — the user gets an empty dropdown with the spinner gone and no explanation of why (e.g. an expired session, since the RPC is `REVOKE ... FROM anon` / `GRANT ... TO authenticated` only). Worth surfacing an error at least for the `advancementType === 'characteristic'` path, even if the Ganger case stays non-fatal.

Everything else checks out: the `characteristicRank` threshold shift (8→9, 12→13) correctly matches the new `"save": 9` entry; the RPC's returned fields (`base_xp_cost`, `is_available`, `has_enough_xp`, `can_purchase`) match the new `CharacteristicAdvancement` type; the modal fully unmounts on close (`{isAdvancementModalOpen && <AdvancementModal .../>}`) so the "deliberately uncached" catalog is refetched fresh on next open — no staleness across purchases within a session as I initially suspected.

**Verdict: Changes requested** — not blocking on architecture, but recommend fixing #1 (produces a visibly wrong optimistic stat direction for the feature this PR completes) and #2 (silently hides real fetch failures from the user) before merging.


---

joeseos:
@joeseos Re-reviewed against the latest head (`12d8f6d`).

**Finding #2 from the previous review (silent catalog fetch failures) is fixed.** The `!response.ok` and `catch` branches now call `setCharacteristicsError(...)`, and `displayedError` surfaces it for the characteristic path. Good.

**Finding #1 is still open — confirming and re-flagging:**

**`components/fighter/fighter-advancement-list.tsx:602-606` — `plusStats` is missing `'save'` (medium, correctness)**

```ts
// "+" stats (lower is better) use -1; normal stats use 1
const plusStats = new Set([
  'weapon_skill', 'ballistic_skill', 'initiative',
  'leadership', 'cool', 'willpower', 'intelligence'
]);
const numericValue = plusStats.has(characteristicCode) ? -1 : 1;
```

`supabase/migrations/20260810130000_n26_advancements.sql:103` and the seeded modifier row (`('Save', 'save', -1)`, line 183) both establish Save as a target-number stat that improves with `-1`. Since `'save'` isn't in `plusStats`, the optimistic advancement created in `addCharacteristicMutation.onMutate` (used by both the standard purchase flow and the Ganger/Exotic Beast pair-purchase flow — same code path) applies `numeric_value: 1` for a Save advancement instead of `-1`. It self-corrects once the server response / cache revalidation lands, so this isn't data corruption, but it's a real, briefly-visible wrong direction on a stat where the sign matters semantically (target number, lower is better). This PR is what makes `Save` render correctly in the main dropdown for the first time (previously it silently fell into "Misc." per the PR description), so it's now the path every N26 Save purchase actually takes.

Suggested fix:
```ts
const plusStats = new Set([
  'weapon_skill', 'ballistic_skill', 'save', 'initiative',
  'leadership', 'cool', 'willpower', 'intelligence'
]);
```

**Minor, non-blocking:** `displayedError` (line 537) falls back to the shared `error` state (`characteristicsError ?? error`) when on the characteristic tab. `error` is set by the skill-set fetch/purchase paths and is never cleared on `advancementType` switch, so a prior skill-fetch failure can still surface as the displayed error after switching to Characteristic even though the characteristic catalog loaded fine. This predates the PR (same conflation existed via the old unconditional `{error && ...}` block) and isn't introduced by this change — flagging only as a nice-to-have since this PR already touches the surrounding logic.

**Verdict: Changes requested** — one correctness fix (`plusStats` + `'save'`) before merging. Everything else checks out: dead-code removal confirmed (no remaining callers of the four removed exports), `characteristicRank` thresholds match the new `"save": 9` entry, the RPC's returned fields line up with the `CharacteristicAdvancement` type, and the modal fully remounts on close so the uncached catalog fetch is fresh on every open — no cross-purchase staleness.

---

PR #1980, opened 2026-08-11
Give N26 an Advancement picker in place of the static table

joeseos:
The N26 branch of the modal offered a Roll 2D6 button and a read-only list of the nine 2-12 results, with nothing to select an outcome with. N23 has that selector but gates it to Ganger / Exotic Beast subtypes, and N26 applies one table to every subtype, so no equivalent was ever built. A player rolled, read a list, and was left with the free-choice Characteristic/Skill pickers, which are tied to nothing.

N26AdvancementEntry already carried what a picker needs -- characteristics[] and skillAcquisitionTypeIds[], the latter documented as "the ids get_available_skills returns, so a skill result feeds the existing skill picker directly". Only the UI was missing.

The row picker drives the ordinary flow rather than repeating the Ganger purchase path. Choosing a characteristic points advancementType and selectedCategory at it, and the effect already watching those fills in costs from the RPC map and feeds the existing Buy button; choosing a skill hands off to the existing skill set / acquisition type / skill flow with the acquisition types narrowed to the ones that result awards. No new mutation, no second buy path.

A result may offer characteristics, skills or both -- rolling a 2 offers +1 Leadership, +1 Intelligence or a random Primary skill -- so each kind gets its own section and choosing in one clears the other. Rolls 5, 7-8 and 12 are skills only; 11 offers three characteristics rather than the pair N23's table is limited to, which is why the Ganger row shape could not simply be reused.

The free-choice type and characteristic pickers are hidden on N26 for the same reason N23 Gangers never see them: the table result decides the type, so a second untied route to the same purchase is a way to get it wrong.

Two fixes fall out of this. gangerModalRollBuy was edition-blind, so an N26 fighter carrying a Ganger subtype routed Buy through the N23 flow and could never satisfy its preconditions -- the button was permanently disabled. And the catalog fetch was gated on the advancement type being 'characteristic', which on N26 is set by clicking a radio that needs the catalog already loaded to resolve its effect type id.

Skill results are priced by the result, not the acquisition type, and cost no XP, so the N23 per-type prices are neither applied nor shown on N26.

### DIFF
--- a/components/dice-roller.tsx
+++ b/components/dice-roller.tsx
@@ -99,8 +99,10 @@ export default function DiceRoller<T>({
   return (
     <>
       <div className="flex items-center gap-3">
+        {/* shrink-0 so a long inline result wraps rather than squeezing the
+            button until its own label breaks across lines. */}
         <button
-          className={`px-3 py-2 bg-neutral-900 text-white rounded-sm hover:bg-gray-800 disabled:opacity-50 ${className || ''}`}
+          className={`shrink-0 whitespace-nowrap px-3 py-2 bg-neutral-900 text-white rounded-sm hover:bg-gray-800 disabled:opacity-50 ${className || ''}`}
           onClick={performRoll}
           disabled={disabled || rolling}
           type="button"
--- a/components/dice-roller.tsx
+++ b/components/dice-roller.tsx
@@ -99,8 +99,7 @@ export default function DiceRoller<T>({
   return (
     <>
       <div className="flex items-center gap-3">
-        {/* shrink-0 so a long inline result wraps rather than squeezing the
-            button until its own label breaks across lines. */}
+        {/* shrink-0: a long inline result should wrap, not squeeze the button. */}
         <button
           className={`shrink-0 whitespace-nowrap px-3 py-2 bg-neutral-900 text-white rounded-sm hover:bg-gray-800 disabled:opacity-50 ${className || ''}`}
           onClick={performRoll}
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -33,7 +33,8 @@ import {
   resolveGangerExoticBeastAdvancementFromUtil,
   N26_ADVANCEMENT_TABLE,
   resolveN26AdvancementFromUtil,
-  type TableEntry
+  type TableEntry,
+  type N26AdvancementEntry
 } from '@/utils/dice';
 import { hasCumulativeXp } from '@/types/edition';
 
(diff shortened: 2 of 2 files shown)


joeseos:
## Code Review

@joeseos Reviewed the diff against `main`, plus the surrounding (unchanged) skill-set/acquisition-type machinery in `fighter-advancement-list.tsx` that the new N26 picker plugs into, to check the reuse claims in the description actually hold.

### What's good
- **The two correctness fixes described in the PR are real.** `gangerModalRollBuy` now also gates on `!isCumulativeXp` — before this, an N26 fighter carrying a `Ganger` subtype would route `Buy` through the N23 flow and the button was permanently disabled, confirmed by reading `handleAdvancementPurchase`'s `gangerModalRollBuy` branch. `needsAdvancementCatalog` now includes `isCumulativeXp` — the catalog fetch was previously gated on `advancementType === 'characteristic'`, but N26 resolves an effect type id off a radio click that happens *before* `advancementType` is set, so the catalog had to already be loaded.
- **The "any skill" row (roll of 12) correctly rides the existing "no access" fallback.** `getAllowedAcquisitionTypeIds` / `getAcquisitionTypeIdsForAccess` already return *all* acquisition type ids (including `any_random`) when the fighter has no access to the chosen Skill Set, plus a "not accessible" warning — exactly the mechanism needed to let a roll of 12 pick from any Skill Set, including gang-exclusive ones. No new bypass code was needed; traced through the `allowedIds`/`rowIds` double-filter in `acquisitionTypeComboboxOptions` for several rows and didn't find a row×access combination that produces an unexpectedly empty option list.
- **State resets are consistent.** `selectN26Row`/`selectN26Characteristic`/`selectN26SkillOutcome` each clear the fields the *other* two paths own (skill vs. characteristic vs. row change), matching the "choosing in one clears the other" behavior described in the PR body.
- **`userPermissions.canEdit` without optional chaining in the new "Take a skill instead" button initially looked like a null-deref risk** (the prop is typed `userPermissions?: UserPermissions`), but the whole block is gated by `{isCumulativeXp && userPermissions && (...)}` a few lines up, so it's narrowed safely — not an issue.
- Reuses `formatGangerAdvancementRangeLabel` → `formatAdvancementRangeLabel` cleanly instead of duplicating it for the N26 table.

### Findings

**1. [Low] Skill-only table rows require an extra click before the skill picker appears**
`components/fighter/fighter-advancement-list.tsx` (`n26SelectedRow?.skillAcquisitionTypeIds` block, ~line 2397)
For rows with *only* a skill outcome and no characteristic option (rolls 5, 7-8, 12), selecting the row still shows a "Take a skill instead" button that has to be clicked before the Skill Set combobox appears — there's no competing characteristic choice to disambiguate, so the extra step doesn't gate anything. Could auto-set `advancementType('skill')` in `selectN26Row` when the newly selected row has `skillAcquisitionTypeIds` but no `characteristics`. Not a bug, just an avoidable click.

**2. [Low] Helper text for the N26 flow duplicates the N23 line and drops the "no XP" framing**
`components/fighter/fighter-advancement-list.tsx` (~line 2318)
The old `isCumulativeXp` copy ("Advancements are earned by rank and cost no XP...") was replaced with "Rating increase is automatically calculated based on the type and number of advancements..." — nearly identical to the N23 branch's text, and no longer mentions that N26 advancements cost no XP, which is the one thing that's actually different about this flow from a player's perspective. Worth restoring that detail if it wasn't dropped intentionally.

### Verdict: **Approve**

No correctness, security, or performance issues found — the two fixes called out in the description are genuine, and the new picker reuses the existing characteristic/skill purchase machinery without a parallel buy path, matching the PR's stated design goal. The two findings above are non-blocking polish.

---

PR #1981, opened 2026-08-11
Filter admin Grants equipment options by edition

Djidiouf:
Keep the Edit Equipment grants dropdown edition-scoped like other admin pickers, and blank missing or cross-edition grant options on load and edition change.

### DIFF
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -359,13 +359,13 @@ export async function GET(request: Request) {
         console.warn('Error in fighter types fetch:', error);
       }
 
-      // Fetch all equipment for the grants dropdown
-      let allEquipment: { id: string; equipment_name: string }[] = [];
+      // Fetch all equipment for the grants dropdown; clients filter by edition_id
+      let allEquipment: { id: string; equipment_name: string; edition_id?: string | null }[] = [];
       try {
         allEquipment = await fetchAllRows((from, to) =>
           supabase
             .from('equipment')
-            .select('id, equipment_name')
+            .select('id, equipment_name, edition_id')
             .order('equipment_name')
             .order('id')
             .range(from, to)
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -36,6 +36,25 @@ interface AdminEditEquipmentModalProps {
 const EQUIPMENT_TYPES = ['wargear', 'weapon', 'vehicle_upgrade'] as const;
 type EquipmentType = typeof EQUIPMENT_TYPES[number];
 
+/** Blank grant options that are missing from the catalog or outside the given edition. */
+function sanitizeGrantsOptionsForEdition(
+  grants: EquipmentGrants,
+  catalog: Array<{ id: string; edition_id?: string | null }>,
+  editionId: string
+): EquipmentGrants {
+  return {
+    ...grants,
+    options: (grants.options || []).map(option => {
+      if (!option.equipment_id) return option;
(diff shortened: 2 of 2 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed the full diff against `main` — the route change and the new `sanitizeGrantsOptionsForEdition` helper plus its two call sites (edition-change handler, equipment-details load).

### What's good
- The `route.ts` change is minimal and correct: adding `edition_id` to the existing paginated `equipment` select for the grants dropdown, no new query, no auth changes (still gated by `checkAdmin`).
- The edition-change handler's `sanitizeGrantsOptionsForEdition(current, allEquipment, newEditionId)` call correctly passes the *unfiltered* `allEquipment` catalog (not the new `filteredAllEquipment` memo) so the lookup isn't circular.
- `filteredAllEquipment` swapped in for the dropdown's `<option>` list is the right, minimal change to stop cross-edition equipment from being selectable going forward.

### Findings

**1. [High] Sanitize-on-load silently blanks valid grants for equipment with no edition, or when the catalog fetch fails**
`components/admin/admin-edit-equipment.tsx:39-56` (`sanitizeGrantsOptionsForEdition`) and its load-time call site at `:311-319`.

```ts
if (!granted || !editionId || granted.edition_id !== editionId) {
  return { ...option, equipment_id: '' };
}
```

Two ways this fires for equipment where nothing is actually wrong:

- **Legacy/pre-edition equipment.** When `equipmentDetails.edition_id` is `null` (equipment created before the edition column existed, or intentionally shared across editions), the load call passes `editionId = equipmentDetails.edition_id ?? ''`, i.e. `''`. The `!editionId` branch then unconditionally blanks *every* configured grants option's `equipment_id` on open — even if the granted target also has `edition_id === null` (a legitimate same-edition/no-edition match). `null !== ''` also fails the direct comparison, so there's no path back to keeping the value.
- **Catalog fetch failure.** `route.ts`'s `allEquipment` fetch (the one returned as `all_equipment`) is wrapped in a try/catch that only `console.warn`s on error, leaving `all_equipment: []`. In that case `catalog.find(...)` returns `undefined` for every option, so `!granted` blanks *every* option regardless of edition — for equipment that had a perfectly valid, same-edition grant configured.

Compare with the sibling `weapon_group_id` sanitize a few lines above (`:237-245`), which only blanks when the parent **is found** and its edition **differs** — it leaves the value alone if the parent isn't in the loaded catalog, rather than treating "not found" as "wrong edition." The new grants sanitizer inverts that: "not found" (for any reason) is treated the same as "confirmed cross-edition."

**Why this matters:** the wipe happens purely from opening the edit modal, before the admin touches anything. The dropdown does re-render as `Select equipment...`, so an attentive admin editing the Grants section directly would notice — but an admin editing an unrelated field (cost, availability, category) and clicking Save has no such cue. `handleSubmit` (`:514-517`) doesn't filter or validate individual option values, so `grants_equipment` is persisted with `equipment_id: ''` for the affected options, silently corrupting a previously-working grant for any player who receives that equipment afterward.

Suggested fix: don't let "not found" or "no edition assigned" fall through to a wipe. Treat missing/null editions on both sides as equal (mirroring how the codebase already treats null editions elsewhere, e.g. `sameEditionForDisplay` defaulting a missing slug rather than treating it as a hard mismatch), and only blank when the target is definitively found in a non-empty catalog with a different, resolved edition — same posture as the `weapon_group_id` block right above it.

The edition-scoping intent and the route/filter changes are sound, but the sanitize-on-load path has a real silent-data-loss failure mode for null-edition equipment and for a transient catalog fetch failure. Worth fixing before merge given it can corrupt saved `grants_equipment` data without any hard error.

---

PR #1983, opened 2026-08-12
Fighter effects can now add and remove fighter subtypes

joeseos:


joeseos:

## Findings

### 1. Subtype grants/revocations are silently dropped for vehicle-mounted equipment (High)

Every call site of `syncSubtypeGrants` passes only the fighter-purchase id, never the vehicle's assigned fighter — even though the same functions already resolve and branch on that value for rating and cache invalidation:

- `app/actions/equipment.ts:660` — `syncSubtypeGrants(supabase, params.fighter_id, { granted: appliedEffects })`. `buyEquipmentForFighter` computes `vehicleAssignedFighterId` (line 261) and uses it at lines 523/586/649 for rating deltas, but not here.
- `app/actions/equipment.ts:1162` — `deleteEquipmentFromFighter` resolves `veh.fighter_id` for rating (~line 1120) but calls `syncSubtypeGrants(supabase, equipmentBefore.fighter_id, …)`, dropping it for vehicle equipment.
- `app/actions/sell-equipment.ts:245` — same pattern; `vehicleAssigned`/`vehicle.fighter_id` is resolved above (~line 128-147) but not passed through.
- `app/actions/move-to-stash.ts:191` — same pattern.
- `app/actions/move-from-stash.ts:308` — `syncSubtypeGrants(supabase, params.fighter_id, { granted: toApply })`; effects inserted just above are stored with `vehicle_id: params.vehicle_id || null`.

Concretely: if an effect with `fighter_subtype_ids_to_add`/`_remove` is attached to equipment bought/sold/moved for a **vehicle that is assigned to a fighter**, the grant/revocation never reaches `fighters.fighter_subtypes` — no error, just a silent no-op. Given the feature's own example ("a Dirt bike granting Mounted"), vehicle-mounted equipment looks like exactly the case this is meant to cover.

It goes deeper than the call sites: `fighter_effects` rows created for vehicle equipment are inserted with `fighter_id: params.fighter_id` (null for vehicle-only purchases) and `vehicle_id: params.vehicle_id` (see `insertEffectWithModifiers`, `app/actions/equipment.ts:166-167`). `syncSubtypeGrants`'s "still granted by another item" survivor check (`utils/fighter-subtype-grants.ts`, the `fighter_effects... .eq('fighter_id', fighterId)` query) filters strictly on `fighter_id`, so even if a caller passed the vehicle's assigned fighter id through, that query would still miss vehicle-linked effect rows. Fixing the call sites alone isn't sufficient — the survivor-check query in `fighter-subtype-grants.ts` needs to account for `vehicle_id` too (e.g. also matching on the assigned vehicle's fighter, or storing/looking up effects by the resolved target fighter consistently).

**Suggested fix:** thread the resolved target fighter id (`params.fighter_id ?? vehicleAssignedFighterId`, and equivalently in the other four files) into `syncSubtypeGrants`, and make the "remaining survivors" query in `fighter-subtype-grants.ts` reachable for vehicle-attached effect rows (e.g. join through `vehicles.fighter_id` or `OR`-match `vehicle_id` against vehicles owned by the fighter).

### 2. Read-modify-write on `fighters.fighter_subtypes` isn't atomic (Low)

`syncSubtypeGrants` reads `fighter_subtypes`, computes a new array in JS, then writes it back with a plain `update`. Two concurrent equipment mutations on the same fighter (e.g. two rapid purchases) could race and one write could clobber the other's grant. Likely low-probability given typical single-user-editing-one-gang usage, but worth a note since everything else in this PR is otherwise careful about ordering (e.g. the explicit "call revoked AFTER delete" comments).

## Verdict: Changes requested

Blocker: Finding #1 — the vehicle-equipment gap undermines the feature's own stated use case and should be fixed (both the call-site fighter-id resolution and the survivor-check query) before merge. Finding #2 is a minor note, not blocking.

Everything else looks solid: the edition-gating (`allowsMultipleSubtypes`), the "bail before touching the DB when nothing is subtype-related" fast path, the dedupe/no-op guard before the final `update`, the read-only chip treatment in the edit modal, and the `resolveFighterEditionSlugForUpdate` → `resolveFighterEditionSlug` de-duplication in `edit-fighter.ts` are all clean.

---

PR #1984, opened 2026-08-12
Add edition support to custom asset creation and filtering

joeseos:
## Summary
This PR extends edition support throughout the custom asset creation system, allowing users to create fighters, equipment, gang types, skills, and trading posts scoped to specific game editions (N23 vs N26). Assets are now filtered by edition in dropdowns and lists, and edition-specific fields (like Save characteristic, Starting XP, vehicles, trade points) are conditionally shown based on the selected edition.

### Core Edition Integration
- Added `editionSlug` prop to all custom asset components (`CustomiseFighters`, `CustomiseEquipment`, `CustomiseGangTypes`, `CustomiseSkills`, `CustomiseTradingPosts`)
- Integrated `useHomeEdition()` hook in home tabs to track the active edition context
- Edition filtering now applied to API responses for gang types, fighter subtypes, skill types, and equipment

### Fighter Customization (N23 vs N26)
- Changed fighter subtype handling from single selection to array-based to support N26's multiple subtypes per fighter
- Added conditional rendering for N26-only fields: `save`, `starting_xp`, `is_vehicle`
- Implemented multi-checkbox picker for N26 subtypes vs single dropdown for N23
- Added `toggleFighterSubtype()` utility to maintain reference list order when toggling subtypes
- Updated crew detection logic to work with subtype arrays

### Equipment & Weapon Profiles
- Added `equipment_category_id` field to resolve categories by ID instead of name (categories repeat across editions)
- Implemented edition-specific weapon profile statlines: N26 uses Lethality/SR/LR, N23 uses Damage/Accuracy/Range
- Added `profilesForEdition()` to blank unused stats before submission
- Introduced `trade_points` field for N26 equipment pricing vs N23's availability system
- Created `toProfileInput()` normalizer for stored profiles with nullable fields

### Gang Types & Trading Posts
- Added conditional `alignment` field rendering based on edition capability
- Scoped equipment lists in trading posts to active edition via context (`TradingPostEditionContext`)
- Updated cost field handling to use trade points or availability based on edition

### Data Integrity
- Edition change detection clears dropdown caches to prevent stale data from previous edition
- Form submission guards edition-specific fields so hidden values never persist
- Copy operations preserve source fighter's edition rather than forcing current tab's edition
- Profile view modal added (`renderStatView`) for read-only stat display

### API & Filtering
- Updated API calls to include `edition_slug` query parameters where applicable
- Implemented `sameEditionForDisplay()` checks throughout to filter results
- Equipment category fetches now include edition information

- Edition-specific fields are guarded at submission time rather than relying on hidden state, preventing stale typed values from being saved after edition switches
- Fighter subtype selection uses a shared `toggleFighterSubtype()` utility across admin and custom forms to maintain consistent ordering
- Weapon profile statlines are normalized to all-string form (`WeaponProfileInput`) with nullable lethality/traits fields
- Trading post edition context avoids threading edition slug through six component levels

https://claude.ai/code/session_01T4TqZcWqf1hnGxfd6tf5fu


joeseos:
## Code Review

@joeseos Reviewed the full diff against `main` (714 additions / 917 deletions, 20 files), focused on the edition-gating logic, the fighter-subtype and starting-XP handling, and the shared-state effects in `custom-fighters.tsx`.

### Findings

**1. [High] `starting_xp: data.starting_xp ?? 0` silently discards the intended NULL/"N/A" semantics — the feature this PR documents doesn't actually work**
`app/actions/customise/custom-fighters.ts:199` (create) and `:412` (edit)

The code has its own comment explaining the intent:
```ts
save: data.save ?? null,
// The column has no default, and NULL means N/A — a fighter that can never
// gain XP. Editions without the concept send an explicit 0.
starting_xp: data.starting_xp ?? 0,
```
But `??` treats `null` and `undefined` identically. `components/customise/custom-fighters.tsx:1006` correctly sends `null` when the Starting XP field is left blank (`showStartingXp ? (startingXp === '' ? null : parseInt(startingXp)) : 0`), but both server actions then coerce that `null` back to `0` before the insert/update. The same bug is repeated a third time, independently, in the Copy-fighter flow at `components/customise/custom-fighters.tsx:904` (`starting_xp: copyModalData.starting_xp ?? 0`).

Net effect: it's currently impossible to persist an N26 "N/A" starting XP through any of the three entry points (create, edit, copy) — every blank field silently becomes `0`, which is a materially different value (recruited with 0 XP but still advancement-eligible, vs. never gaining XP at all).

Suggested fix: distinguish "field omitted" from "field explicitly null" — e.g. `data.starting_xp === undefined ? 0 : data.starting_xp` (and the equivalent at the two other call sites), or have the N23 path send `0` explicitly (which it already does via the `showStartingXp` ternary) and drop the `?? 0` fallback entirely.

**2. [Medium] `handleView`'s fighter-subtype fetch bypasses `filterAllowedFighterSubtypes`, polluting the shared state used by Create/Edit**
`components/customise/custom-fighters.tsx:843-853` vs. `:588-593`

`fighterSubtypes` is a single piece of state rendered by both the read-only View panel and the Create/Edit form/checkbox-grid. The Add/Edit-modal effect curates the list correctly:
```ts
setFighterSubtypes(filterAllowedFighterSubtypes(subtypeData, editionSlug));
```
but `handleView` (touched by this PR to add the `edition_slug` query param) sets the raw response directly:
```ts
const data = await response.json();
setFighterSubtypes(data);
```
Both fetches guard on `fighterSubtypes.length === 0` before running. So: open "View" on any existing custom fighter first (before ever opening Create/Edit) → `fighterSubtypes` gets populated with the full, uncurated list → the guard in the Add/Edit effect is now `false` → the curating fetch never runs → the Create/Edit subtype picker shows placeholder rows (`*`, `Others`, `Special Terrain`) and alliance-only subtypes that should never be user-authorable.

Suggested fix: apply `filterAllowedFighterSubtypes` in `handleView` too, or use a separate state variable for the view-only list so the two code paths don't share mutable state with different invariants.

Blocker: finding #1 — the PR's own stated N26 "N/A starting XP" capability is unreachable from the UI as written, across all three entry points that set it. Finding #2 is a real but narrower correctness issue (order-dependent state pollution) worth fixing in the same pass.

Everything else — the edition-slug threading through `getCampaignTypes`/dropdowns, the `equipment_category_id` resolve-by-id change, the N23/N26 weapon-profile statline split, and the trading-post edition context — traced cleanly with no dangling `undefined` or missed call sites found.

---

PR #1986, opened 2026-08-12
Fix/fixed tp in fighters list

joeseos:


joeseos:
## Code Review

@joeseos Reviewed the full diff against `main` (57 additions / 11 deletions, 8 files), tracing the Trade Points purchase path end-to-end: `PurchaseModal` → `usePurchaseEquipment` → `buyEquipmentForFighter` → `get_equipment_detailed_data.sql`, plus the display/sync side in `WeaponList`/`VehicleEquipmentList`/`fighter-page.tsx`.

### What's good
- **Fixes a real double bug.** Before this PR, `updategangsCollection` never returned `trade_points` from `buyEquipmentForFighter` (`app/actions/equipment.ts`), so even though the server already deducted TP correctly on every purchase, the client never learned about it — the on-screen TP balance went stale until a full reload. Separately, purchases made through a fighter's own equipment list tab were being charged Trading-Post TP costs they shouldn't pay at all. Both are fixed here.
- **The "don't charge TP on fighter's-list purchases" fix is applied at every layer, not just the UI**: `chargesTradePoints`/`showTradePoints` gate the client-side affordability check and the field shown in `PurchaseModal`, the modal always sends an explicit `tradePoints` value (real cost, or `'0'` for fighter's-list) so the server never falls back to the catalog price, and `get_equipment_detailed_data.sql` independently zeroes the returned `trade_points` for the `$4=true, $5 IS NULL` (pure fighter's-list) case. Traced the SQL condition against the actual caller in `equipment.tsx` (`fighter_type_equipment: true`, `equipment_tradingpost` omitted) — it correctly does *not* match the combined `fighters-tradingpost` mode (`$5=true`), so TP is still charged there, which is correct.
- **`newGangTradePoints ?? gang.trade_points` fallback is safe**: `financialResult.newValues?.trade_points` only comes back undefined if the underlying update didn't touch it, in which case falling back to the pre-purchase value is correct.
- The toast-text fix (`parseTradePointsCost(params.manual_trade_points)` instead of `?? item.trade_points`) correctly stops purchases from displaying "and N TP" for a cost that wasn't actually charged — confirmed `parseTradePointsCost(undefined)` returns `0`, so this doesn't regress the non-TP-edition case.
- Confirmed `VehicleEquipmentList` was intentionally left without the new `onGangTradePointsUpdate` wiring — it's documented as N23-only (`vehicleEquipmentList.tsx:36`), and Trade Points is an N26-only capability (`types/edition.ts:113`), so TP can never be charged there. Not a gap.

### Findings

**1. [Low, non-blocking] Server still trusts the client's `manual_trade_points` with no cross-check, and `deleteEquipmentFromFighter` never refunds Trade Points**
`app/actions/equipment.ts:429` falls back to the raw `equipment.trade_points` catalog column (not the RPC's zeroed value) if the client omits `manual_trade_points` — currently unreachable since `PurchaseModal` always sends an explicit value, but it means the "don't charge TP on fighter's list" guarantee lives entirely in client trust, same as the existing `manual_cost` trust model noted in the surrounding comment. Separately, `deleteEquipmentFromFighter` (`app/actions/equipment.ts:1053`) has no Trade Points refund logic at all, so selling/deleting TP-costed gear never returns the spent TP. Both are pre-existing patterns/gaps this PR doesn't introduce or worsen — flagging only because this PR is the first to make Trade Points correctness a live concern; no change requested here.

### Verdict: **Approve**

No correctness, security, or performance issues in the diff itself. The fix is scoped correctly, applied consistently at both the UI and SQL layers, and the one item above is pre-existing and out of scope.

---

PR #1988, opened 2026-08-13
Fix duplicate skills in admin fighter type default skill dropdown

joeseos:
The edit-fighter-type modal keeps one `skills` state list that is never reset while the modal is open, but `fetchedSkillDetailsRef` is cleared every time a fighter type is loaded. The effect that back-fills details for a fighter type's saved default skills then appended its results without keying by id, so each fighter type opened re-appended the skills it shares with earlier ones. Opening three n26 leaders in a row left three "Inspiring" options in the default skill dropdown, even though only one such skill row exists.

Merge by id the same way the skill-set fetch already does, and mark ids as fetched even when the lookup returns nothing for them so it is not retried on every selection change.

### DIFF
--- a/components/admin/admin-edit-fighter-type.tsx
+++ b/components/admin/admin-edit-fighter-type.tsx
@@ -546,10 +546,18 @@ export function AdminEditFighterTypeModal({ onClose, onSubmit }: AdminEditFighte
         );
 
         if (missingSkills.length > 0) {
-          setSkills(prevSkills => [...prevSkills, ...missingSkills]);
-          // Mark these skills as fetched
-          missingSkillIds.forEach(id => fetchedSkillDetailsRef.current.add(id));
+          // Key by id: skills accumulate across every fighter type opened in
+          // this modal, so appending blindly repeats a skill once per fighter
+          // type that carries it as a default.
+          setSkills(previous => {
+            const merged = new Map([...previous, ...missingSkills].map(skill => [skill.id, skill]));
+            return Array.from(merged.values());
+          });
         }
+
+        // Mark these skills as fetched even when the lookup returned nothing
+        // for them, so it is not retried on every selection change.
+        missingSkillIds.forEach(id => fetchedSkillDetailsRef.current.add(id));
       } catch (error) {
         console.error('Error fetching selected skill details:', error);
       }
--- a/components/admin/admin-edit-fighter-type.tsx
+++ b/components/admin/admin-edit-fighter-type.tsx
@@ -52,6 +52,17 @@ interface Skill {
   skill_type_id: string;
 }
 
+/**
+ * Folds newly fetched skills into the ones already held, keyed by id. The
+ * skills list accumulates across every fighter type opened in the modal, so
+ * appending blindly repeats a skill once per fighter type that carries it as a
+ * default. Later entries win, keeping the freshest copy of a skill's metadata.
+ */
+function mergeSkillsById(previous: Skill[], incoming: Skill[]): Skill[] {
+  const merged = new Map([...previous, ...incoming].map(skill => [skill.id, skill]));
+  return Array.from(merged.values());
+}
+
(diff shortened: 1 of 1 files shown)


joeseos (commented):
@joeseos Reviewed this PR.

## Summary
The fix correctly addresses the root cause: `skills` state persists across every fighter type opened in the modal, while `fetchedSkillDetailsRef` is cleared on each load (line 646), so the back-fill effect kept re-appending the same skill records without de-duplication. Switching to a `Map`-keyed merge (`components/admin/admin-edit-fighter-type.tsx:552-555`) matches the pattern already used by the sibling `fetchSkills` effect (`admin-edit-fighter-type.tsx:512-515`), so `availableSkills` (derived via `.filter` over `skills`, line 212) and the dropdown built from it (line 2044) can no longer contain duplicate ids.

Moving the `fetchedSkillDetailsRef` marking outside the `if (missingSkills.length > 0)` block (line 558-560) is also a real improvement: previously, a skill id that legitimately returned no match (e.g. a stale/deleted default skill reference) would never be marked fetched and would be re-queried on every `selectedSkills`/`selectedFighterTypeId` change. Now it's marked once regardless, and only skipped on a genuine fetch failure (still inside the `try`), which correctly allows a retry later.

## Findings
No blocking issues found.

- **[low] DRY** — The `Map`-keyed merge logic (`new Map([...previous, ...x].map(s => [s.id, s])); Array.from(merged.values())`) is now duplicated in two effects (lines 512-515 and 552-555). Not introduced by this PR (the first occurrence pre-dates it) and low-risk either way, but a shared `mergeById` helper would remove the duplication if this pattern needs to be applied again elsewhere.

## Verdict
**Approve.** Small, correctly scoped fix that matches the existing merge convention in the same file, includes a clear explanation of the fix in comments, and doesn't touch unrelated code.

---

joeseos (commented):
@joeseos Automated review.

**Summary of change:** `components/admin/admin-edit-fighter-type.tsx` — the "fetch full details for selected skills" effect used to blindly append `missingSkills` onto the `skills` state, which is never reset while the modal stays open, so a skill shared by multiple fighter types got re-appended once per fighter type opened. This PR merges the addition into `skills` by `id` (via a `Map`), and moves the `fetchedSkillDetailsRef` bookkeeping outside the `missingSkills.length > 0` guard so an id that comes back empty from the lookup is still marked as fetched (previously it would be retried on every `selectedSkills` change).

**Findings:**

| # | File:Line | Severity | Issue | Suggested fix |
|---|---|---|---|---|
| 1 | `components/admin/admin-edit-fighter-type.tsx:511-514` vs `:552-555` | Low | The `new Map([...previous, ...additions].map(s => [s.id, s]))` → `Array.from(...values())` merge-by-id pattern is now duplicated verbatim in two effects in the same file (the existing `fetchSkills` effect and this newly-fixed one). | Non-blocking; consider extracting a small `mergeById(list, additions)` helper if a third use case shows up, so the two copies don't drift. |

No correctness, security, edge-case, or performance issues found:
- The fix directly matches the root cause described in the PR body, and I verified `fetchedSkillDetailsRef.current.clear()` (called on every fighter-type load, line ~646) while `skills` state persists — confirming the duplication bug as described and that keying by `id` resolves it.
- Merge order (`previous` then `missingSkills`) means freshly-fetched full skill records correctly overwrite any partial/placeholder entries with the same id, which is the desired behavior.
- Moving the "mark as fetched" loop outside the `if` block is correct and avoids a stale-id retry loop on every `selectedSkills` change; it doesn't change fetch behavior for ids that *are* found.
- Small, single-file diff (+11/-3), no new dependencies, no schema/API changes, follows the merge-by-id pattern already established elsewhere in this same file.

**Verdict:** Looks correct and ready to merge (I can't formally "Approve" my own account's PR via the API, so leaving this as a comment review — no blockers found).

---

PR #1989, opened 2026-08-13
Fix/remove global cache flush on auth

joeseos:


joeseos:
## Review

@joeseos Reviewed the diff against the full codebase, ran `tsc --noEmit` and `eslint` on the changed files (both clean), and traced the cache-tag/auth call sites end to end.

This is a well-reasoned fix. A few things worth calling out, none blocking:

### Findings

**1. Missed DRY cleanup — `components/password-change.tsx:120-126`** (low)
The PR introduces `EMPTY_PASSWORD_REQUIREMENTS` in `utils/password.ts` and uses it to replace 3 of the 4 inline `{hasLowerCase: false, ...}` literals in this file (initial state, reset-after-success). The Cancel button's `onClick` still has the old inline literal:
```tsx
setPasswordRequirements({
  hasLowerCase: false,
  hasUpperCase: false,
  hasNumber: false,
  hasSpecialChar: false,
  hasMinLength: false,
});
```
Suggest swapping this for `setPasswordRequirements(EMPTY_PASSWORD_REQUIREMENTS)` to match the other three call sites and finish the refactor.

**2. `forgotPasswordAction` now honors the request `Origin` header for the reset-link redirect — `app/actions/auth.ts:186`** (low, worth confirming)
Previously this interpolated an unset `NEXT_PUBLIC_BASE_URL`, so the redirect link was silently broken (GoTrue fell back to Site URL) — this PR fixes that, matching the existing pattern already used in `signUpAction`. Worth double-checking that Supabase's **Authentication → URL Configuration → Redirect URLs** allow-list is locked down to only the real production/preview domains, since GoTrue rejecting (falling back to Site URL for) any `redirectTo` not on that list is the actual backstop here, not the app code. Next.js's built-in Server Action origin check adds a second layer. Not a regression from this PR's logic, just flagging since this code path went from "inert" to "live."

### What looks solid
- The `revalidatePath('/', 'layout')` removal is correct given the tagging architecture in `utils/cache-tags.ts` — that call emits the implicit `_N_T_/layout` tag, which every `unstable_cache` read carries as a soft tag, so it was evicting the entire shared Data Cache rather than anything user-scoped. Confirmed no other `revalidatePath` calls remain in `app/`, `components/`, `utils/`, `hooks/`.
- `signInAction`/`signOutAction` switching from `redirect()` to returning `{ redirectTo }` + `window.location.assign()` is the right fix for the "browser Supabase client still holds the stale session" problem — verified the only two call sites (`app/sign-in/page.tsx`, `components/settings-modal.tsx`) both correctly narrow the new `AuthActionResult` union and do the full document load.
- `signOutAction` now actually revokes the refresh token via `supabase.auth.signOut()` before the manual cookie sweep, with a try/catch fallback to the sweep — sound, and consistent with how `utils/supabase/server.ts` wires cookie read/write for Server Actions.
- `reset-password/update/page.tsx`: the `signOut({ scope: 'local' })` before `verifyOtp` correctly prevents a signed-in session from swallowing someone else's recovery token; the `hasVerifiedRef` guard correctly prevents double-consuming the one-time token under StrictMode double-invocation; branching on `token_hash`/`type` vs. an already-established GoTrue session covers both recovery email template styles.
- `TurnstileWidget`: moving to the `onToken` callback + `ref`-exposed `reset()` (React 19's plain `ref`-as-prop, matches `react@19.2.7` in `package.json`) is cleaner than the old DOM-querying/hidden-input-injection approach, and correctly re-challenges on `expired-callback`/`timeout-callback`/`pageshow` (bfcache restore).
- `SubmitButton`'s `disabled={pending || disabled}` (not just `aria-disabled`) fixes a real bug: a second click during an in-flight sign-in would have resubmitted the single-use Turnstile token and always failed verification.
- `updatePasswordAction` removal confirmed dead — no remaining references anywhere in the codebase.
- `useClaims` simplification (always `loadClaims()` on non-`SIGNED_OUT` events) is behavior-preserving vs. the old branch-on-`session` logic, just removes duplication.

### Verdict
**Approve.** The two notes above are minor/informational — worth a look but not blockers.

---

PR #1990, opened 2026-08-13
Claude/turnstile session state bug y211p1

joeseos:


joeseos:
## Review

@joeseos Reviewed the diff against the full codebase (all 4 commits, `e74ed0f`→`2ca5735`), read every changed file in full (not just the patch hunks), checked React 19/Next 16 compatibility for the `ref`-as-prop pattern in `TurnstileWidget`, and cross-checked against sibling PR #1989.

### Most important finding first: this PR is a strict subset of #1989, and is missing its fix for a real cache bug

`fix/remove-global-cache-flush-on-auth` (#1989) contains these exact same 4 commits **plus one more**: `ee595ad` *"Stop flushing the entire Data Cache on every sign-in and sign-out"*. That commit is not in this branch, so `app/actions/auth.ts` here still has:

```ts
// signInAction, line ~135
revalidatePath('/', 'layout');
return { redirectTo: safePostSignInPath(nextParam) };
```
```ts
// signOutAction, line ~217
// Revalidate the root layout to clear any cached user data
revalidatePath('/', 'layout');
return { redirectTo: "/sign-in" };
```

Per #1989's commit message (verified against `next/dist/server/lib/implicit-tags.js` / `unstable-cache.js` in `node_modules`): `revalidatePath('/', 'layout')` emits the implicit `_N_T_/layout` tag, which every route carries and which `unstable_cache` submits as a soft tag on every read — so this evicts **all ~60 `unstable_cache` entries app-wide** (gang rosters, fighter data, campaigns, customisations, for every user) on every single sign-in and sign-out, not just the current user's data. That's a real, verified perf regression risk on a hot path, already fixed one commit later in #1989.

**This is very likely why GitHub reports this PR's `mergeable_state` as `blocked`** — the two branches touch identical lines with #1989 being a superset. Merging #1990 as-is would either conflict with or regress past #1989's fix.

**Recommendation:** close this PR and merge #1989 instead (it contains everything here plus the cache fix, and already has an approved review), or rebase this branch onto #1989's tip before merging. Don't merge both.

### Everything else checks out

Setting the cache-flush gap aside, the session-bug fix itself is sound:
- `signInAction`/`signOutAction` returning `{ redirectTo }` instead of calling `redirect()`, with callers doing `window.location.assign()` — correct fix for the actual bug (browser Supabase client staying stale after a client-side nav, since deleting `sb-*` cookies server-side fires no `SIGNED_OUT` event).
- `TurnstileWidget`: moving off `document.querySelector('form')` DOM hand-off to an `onToken` callback + React-19 `ref`-as-prop with `useImperativeHandle` is correct (confirmed `react@19.2.7` in `package.json`), and now actually calls `reset()` on failed sign-in, `expired-callback`/`timeout-callback`, and bfcache `pageshow` restores — fixes the real "every retry fails until hard reload" bug from resubmitting a spent single-use token.
- `reset-password/update/page.tsx`: `hasVerifiedRef` correctly guards `verifyOtp` (single-use) against StrictMode double-invocation; branching on `token_hash`+`type==='recovery'` vs. an already-established GoTrue session correctly covers both the `{{ .TokenHash }}` and `{{ .ConfirmationURL }}` email template shapes; `signOut({ scope: 'local' })` before `verifyOtp` correctly prevents a signed-in session from swallowing someone else's recovery token.
- `SubmitButton`'s `disabled={pending || disabled}` (not just `aria-disabled`) fixes a real double-submit bug (second click during sign-in burns the Turnstile token). Minor, non-blocking note: a natively-disabled button drops out of the tab order during pending state, unlike the previous aria-disabled-only pattern — an intentional and reasonable trade-off given the bug it fixes, just flagging.
- `forgotPasswordAction` deriving `redirectTo`'s origin from the request `Origin` header (`app/actions/auth.ts` ~186), falling back to `NEXT_PUBLIC_APP_URL`, with an explicit error if neither resolves — correct fix for the previous silent `"undefined/reset-password/update"` link. Same low-severity note as raised on #1989: the real backstop against a spoofed `Origin` is Supabase's Redirect URL allow-list (Next's Server Action same-origin check is a second layer) — worth confirming that allow-list is locked to real domains, not a regression from this PR.
- `updatePasswordAction` removal and `useClaims` simplification confirmed dead-code-safe / behavior-preserving via search across the repo.

Not because of a defect in this PR's own logic — it's the same fix already approved on #1989 — but because merging it as-is reintroduces the global-cache-flush bug #1989 already fixed one commit later, and the two PRs will conflict. Please consolidate onto #1989 (or rebase this branch to include `ee595ad`) before merging either one.

---

PR #1991, opened 2026-08-13
Stop the sign-in form blanking while the redirect is in flight

joeseos:
Signing in showed the form blanking, then a white flash, before the home page appeared. Both are gone.

## Cause

Two separate problems, one behind the other.

**The blanking.** Sign-in was the only `<form action={fn}>` in the app. React 19 resets uncontrolled inputs once a form action completes, and `useFormStatus` drops `pending` at the same moment. Neither fired while `signInAction` ended in `redirect()` — throwing `NEXT_REDIRECT` meant the action never completed normally. Once it returned a destination instead, the action completed as soon as `window.location.assign()` was called, blanking the form for the rest of the load.

**The white flash.** `window.location.assign()` tears down the document and builds a new one. Nothing paints the canvas in between, so the browser paints white. The gap is wide here because there is no `middleware.ts` and no `loading.tsx`, so the next paint waits on `app/page.tsx` running eight parallel Supabase queries.

That reload existed for a reason: the *server* performed the sign-in, so the browser Supabase client never learned about the session, and only a fresh document rebuilt it. That was the fix for stale auth state on account switch.

## Fix

**Sign in on the browser client.** `utils/supabase/client.ts` uses `createBrowserClient`, which stores the session in cookies — so the server still sees it on the next request, and the client that holds the session is now the one that created it. The staleness the reload was compensating for becomes structurally impossible, and the reload goes away with it. Navigation is `router.refresh()` + `router.replace()`, a client-side transition: no teardown, no flash.

Turnstile verification stays on the server, since `TURNSTILE_SECRET_KEY` can't go to the browser. `signInAction` narrows to `verifySignInChallenge`.

**Own the pending lifecycle.** The form submits via `onSubmit` + `preventDefault`, which takes the reset out of React's hands. `isSubmitting` is cleared only on the retryable paths, so the button reads "Signing In..." continuously through the navigation. There is deliberately no `finally { setIsSubmitting(false) }` — that construct is the bug.

## Also in here

- A failed sign-in now keeps the email and password. React had been clearing both on the error path too.
- `app/reset-password/update/page.tsx` had the same `finally`-clears-too-early shape, re-enabling its button during its own redirect.
- **Dead `pendingText` everywhere it was used.** `SubmitButton` read only `useFormStatus`, which reports only from inside a `<form action={fn}>`, so every other call site rendered its `pendingText` never. It now takes an explicit `pending` prop, wired at all of them, so these strings appear for the first time:
  - `app/reset-password/update/page.tsx` — "Updating..."
  - `app/reset-password/page.tsx` — "Sending..."
  - `components/create-gang-modal.tsx` — "Creating..."
  - `components/create-campaign.tsx` — "Creating..."

  create-gang-modal gains more than the text: its `disabled={!isFormValid()}` had left the button clickable mid-create, so `pending={isLoading}` also closes a double-submit hole.

  (`app/sign-up/page.tsx` is unaffected — it uses a plain `<button>` with manually-rendered text, not `SubmitButton`.)

## Note for review

Turnstile verification and the password grant are now two calls rather than one. `NEXT_PUBLIC_SUPABASE_ANON_KEY` is in the client bundle, so GoTrue's password endpoint was already reachable directly and Turnstile has never been a hard gate on authentication — it raises the cost of automated abuse through the app's own form, and it still does. Flagging it as a deliberate choice rather than burying it.

`components/settings-modal.tsx:85` still does a full document load on sign-out and so still has the white flash there. Left alone deliberately — that reload was about guaranteeing cleared cookies, which is a separate call from this one.

## Verification

`tsc --noEmit` clean, `npm run build` compiles, `npm run lint` unchanged at the 9 pre-existing problems (all in files this branch doesn't touch).

Needs a browser, since none of the above catches a repaint:

- **Account switch — the original bug this must not reintroduce.** Sign in as A, sign out, sign in as B: header, gang list and permissions must all show B, with no manual refresh anywhere.
- Valid credentials: no white flash, button reads "Signing In..." continuously.
- Wrong password: error shows, both fields retained, Turnstile re-challenges, button re-enables.
- `/sign-in?next=/gang/<id>`: still lands on that gang.
- Hard-refresh after signing in: still authenticated, confirming cookies were written rather than just held in memory.
- Forgot-password, create-gang and create-campaign buttons show their pending text and are unclickable while their request is in flight.

https://claude.ai/code/session_01KNs9k9eyranhaG6GakHvRi

### DIFF
--- a/app/actions/auth.ts
+++ b/app/actions/auth.ts
@@ -4,12 +4,10 @@ import { createClient } from "@/utils/supabase/server";
 import { headers } from "next/headers";
 import { cookies } from 'next/headers';
 import { invalidateUserCount } from '@/utils/cache-tags';
-import { safePostSignInPath } from '@/utils/auth';
 
-// Returned instead of calling redirect(), so the caller can do a full document
-// load - a client-side navigation leaves the browser Supabase client stale.
+// Returned instead of calling redirect(), so sign-out can do a full document
+// load - it is what guarantees the cleared cookies are gone from every client.
 type AuthRedirect = { redirectTo: string };
-type AuthActionResult = { error: string } | AuthRedirect;
 
 export const signUpAction = async (formData: FormData) => {
   const origin = (await headers()).get("origin");
@@ -94,46 +92,33 @@ export const signUpAction = async (formData: FormData) => {
   }
 };
 
-export const signInAction = async (formData: FormData): Promise<AuthActionResult> => {
-  const email = formData.get("email") as string;
-  const password = formData.get("password") as string;
-  const turnstileToken = formData.get("cf-turnstile-response") as string;
-  const nextParam = formData.get('next') as string | undefined;
-
-  // Check if Turnstile is configured
-  const hasTurnstileConfig = process.env.NEXT_PUBLIC_TURNSTILE_SITE_KEY && process.env.TURNSTILE_SECRET_KEY;
-
+// Only the challenge is checked here - the sign-in itself runs on the browser
+// client, so it holds the session it created and needs no page reload to see it.
+// TURNSTILE_SECRET_KEY is why this half stays on the server.
+export const verifySignInChallenge = async (
+  turnstileToken: string,
+): Promise<{ ok: true } | { error: string }> => {
   if (process.env.NODE_ENV === "development") {
     console.log("Skipping Turnstile verification in development mode.");
+    return { ok: true };
   }
-  else if (hasTurnstileConfig) {
-    if (!turnstileToken) {
-      return { error: "Please complete the security verification challenge" };
-    }
 
-    // Verify Turnstile token
-    const turnstileVerification = await verifyTurnstileToken(turnstileToken);
-    if (!turnstileVerification.success) {
-      console.error('Turnstile verification failed:', turnstileVerification);
-      return { error: "Security verification failed. Please try again." };
-    }
-  } else {
+  if (!process.env.NEXT_PUBLIC_TURNSTILE_SITE_KEY || !process.env.TURNSTILE_SECRET_KEY) {
     console.warn('Turnstile not configured - proceeding without verification');
+    return { ok: true };
   }
 
-  const supabase = await createClient();
-
-  const { error } = await supabase.auth.signInWithPassword({
-    email,
-    password,
-  });
-
-  if (error) {
-    return { error: error.message };
+  if (!turnstileToken) {
+    return { error: "Please complete the security verification challenge" };
   }
 
-  // No revalidatePath - see the note in signOutAction.
-  return { redirectTo: safePostSignInPath(nextParam) };
+  const turnstileVerification = await verifyTurnstileToken(turnstileToken);
+  if (!turnstileVerification.success) {
+    console.error('Turnstile verification failed:', turnstileVerification);
+    return { error: "Security verification failed. Please try again." };
+  }
+
+  return { ok: true };
 };
 
 async function verifyTurnstileToken(token: string) {
--- a/app/actions/auth.ts
+++ b/app/actions/auth.ts
@@ -4,10 +4,13 @@ import { createClient } from "@/utils/supabase/server";
 import { headers } from "next/headers";
 import { cookies } from 'next/headers';
 import { invalidateUserCount } from '@/utils/cache-tags';
+import { safePostSignInPath } from '@/utils/auth';
 
-// Returned instead of calling redirect(), so sign-out can do a full document
-// load - it is what guarantees the cleared cookies are gone from every client.
+// Returned instead of calling redirect(), so the caller can do a full document
+// load - the one thing that resets the router cache, the layout and the browser
+// Supabase client together, which is what an identity change should do.
 type AuthRedirect = { redirectTo: string };
+type AuthActionResult = { error: string } | AuthRedirect;
 
 export const signUpAction = async (formData: FormData) => {
   const origin = (await headers()).get("origin");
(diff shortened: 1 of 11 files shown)


joeseos:
@joeseos Reviewed the diff against the codebase (`app/sign-in/page.tsx`, `app/reset-password/update/page.tsx`, `components/submit-button.tsx`).

## Correctness

The core fix is sound. Tracing through it:
- The old `clientAction` (form `action`) completed as soon as `signInAction` resolved, and `redirect()`'d via a thrown `NEXT_REDIRECT` never let `useFormStatus().pending` or the uncontrolled-input reset fire correctly — matches the diagnosis in the description.
- Switching to `onSubmit`/`preventDefault` + local `isSubmitting` state, with no `finally` clearing it on the success path, correctly leaves the button in "Signing In..." until the scheduled `window.location.assign` actually unloads the document. Confirmed `signInAction` (`app/actions/auth.ts`) never calls `redirect()`/`notFound()`, so wrapping it in `try/catch` here is safe — no risk of swallowing Next's special redirect-control exception.
- The new `try/catch` around `signInAction(formData)` is a genuine bug fix on its own: previously an action throw (e.g. a network failure) was an unhandled rejection that left the button stuck on "Signing In..." forever with no user feedback. Now it resets state, shows an error, and resets Turnstile.
- `components/submit-button.tsx`: `pending || formPending` composes cleanly with existing `action`-based callers (`create-gang-modal.tsx`, `create-campaign.tsx`) — they don't pass `pending`, so behavior there is unchanged. Verified.
- `app/reset-password/update/page.tsx` mirrors the same pattern correctly (error paths reset `isSubmitting`, success path doesn't).

## Finding

**Medium — `app/reset-password/page.tsx:86`** (the "forgot password" request form, not touched by this PR): it still passes `disabled={isSubmitting}` to `SubmitButton` instead of the new `pending={isSubmitting}` prop, and its form already submits via `onSubmit` (not a form `action`), so `useFormStatus().pending` is always `false` there too. Net effect: `pendingText="Sending..."` never renders on this page — the identical dead-pendingText bug this PR just fixed on the sibling `reset-password/update` page. The PR description says the fix covers "the sign-up and reset-password forms," but this sibling file is the one actually named `reset-password` and still has the bug. Since it doesn't do a `window.location.assign` redirect on success, it doesn't have the blanking bug, so this isn't a regression risk, just a suggested follow-up (one-line change: `pending={isSubmitting}` instead of `disabled={isSubmitting}`).

**Low — description accuracy**: the PR body says "`pendingText` was silently dead... on the sign-up... forms," but `app/sign-up/page.tsx` doesn't use `SubmitButton`/`pendingText` at all — it's a plain `<button>` with manually-rendered `isSubmitting` text. Worth a quick correction, no code impact.

Nothing else stood out — no security/perf concerns, TS types are sound, and the redundant `if (isSubmitting) return` guard alongside the `disabled` attribute is harmless belt-and-suspenders.

## Verdict: Approve

The blanking fix and its reasoning are correct and well-verified (manual test matrix in the description is good). The one gap found is in an adjacent file this PR didn't touch and isn't a regression — happy to see it folded into this PR or filed as a fast follow, whichever you prefer.

---

joeseos:
@joeseos Reviewed the 5 commits pushed since the last review (`f71536a`…`ca8c948`) — the sign-in approach was reworked twice more in that span (moved to browser-side sign-in, then reverted back to server-side after a follow-up architect review found the earlier "no middleware / stale browser client" diagnosis was wrong).

## Previous findings

- **Medium — `app/reset-password/page.tsx` dead `pendingText`** (still `disabled={isSubmitting}` instead of `pending={isSubmitting}`): **Resolved** in `f71536a` — now wired correctly, along with `create-gang-modal.tsx` and `create-campaign.tsx`.
- **Low — PR description said `sign-up` used `SubmitButton`/`pendingText`**: **Resolved** — the description now explicitly notes `app/sign-up/page.tsx` is unaffected since it uses a plain `<button>`.

## New findings

**1. Medium — `app/loading.tsx` is the only `loading.tsx` in the app, so it becomes the fallback for every route, not just `/`**
Next's App Router uses the nearest ancestor `loading.js` as the Suspense fallback for any segment that doesn't define its own. Since this is the sole `loading.tsx` in the tree (`app/loading.tsx`, added in `053eb52`), it will also render on first navigation into `/gang/[id]`, `/campaign/[id]`, `/admin/*`, `/sign-in`, etc. while those routes' own async Server Components resolve — showing the home page's skeleton (Create Gang/Campaign button placeholders + HomeTabs shape) on completely unrelated pages. Suggest either using a neutral/generic skeleton here, or scoping the home-shaped one so it only applies to `/` (e.g. an inline `<Suspense>` around `app/page.tsx`'s own content instead of a route-level `loading.tsx`).

**2. Medium — PR description no longer matches the shipped implementation**
The "Cause"/"Fix" and "Note for review" sections describe signing in via the browser client (`createBrowserClient`) with `router.refresh()` + `router.replace()` navigation, and call out Turnstile + the password grant as "two calls rather than one." `053eb52` reverted all of that back to a single server action (`signInAction`) doing both the Turnstile check and `signInWithPassword`, followed by `window.location.assign()` — the code's own comment now reads "Challenge and grant stay in one action: splitting them costs a browser round trip." Reviewers/QA following the current description's verification checklist would be testing the wrong mental model of the fix. Worth updating the description before merge.

Nothing else stood out in the new commits — the `auth.ts` Turnstile-check extraction (`verifySignInChallenge`) is a faithful refactor of the original logic, the `color-scheme` / `app/page.tsx` Suspense-splitting changes are sound and correctly set `color-scheme: dark` in the `.dark` block too, and the `sign-in/page.tsx` revert back to `window.location.assign()` left no leftover browser-client code.

## Verdict: Changes requested

Both new findings are about the loading UI scope and stale docs rather than the core fix, but #1 is a real (if narrow) UX regression across other routes and #2 will mislead whoever runs the verification checklist, so I'd want at least #2 addressed before merge and #1 confirmed/scoped.

---

PR #1993, opened 2026-08-13
Hide Sell to Guilders on N26 fighter pages

joeseos:
Selling a fighter to the Guilders is an N23 rule with no N26 equivalent, but the fighter page rendered the button for every fighter.

Add a sellToGuilders row to EDITION_CAPABILITIES with a hasGuilderSales accessor, and gate the button on it.

FighterActions was the only child of the fighter page that did not receive editionSlug; it read the fighter-only slug smuggled through the fighter prop object. Pass the page's combined slug (gang first, fighter as fallback) as an explicit editionSlug prop, matching every sibling. This also gives beastSubtypeName the better-resolved gang slug.

### DIFF
--- a/components/fighter/fighter-actions.tsx
+++ b/components/fighter/fighter-actions.tsx
@@ -15,7 +15,7 @@ import { Tooltip } from 'react-tooltip';
 import { Checkbox } from '@/components/ui/checkbox';
 import { Combobox } from '@/components/ui/combobox';
 import { buildGangComboboxOption } from '@/utils/gang-combobox-option';
-import { beastSubtypeName } from '@/types/edition';
+import { beastSubtypeName, hasGuilderSales } from '@/types/edition';
 
 interface Fighter {
   id: string;
@@ -34,7 +34,6 @@ interface Fighter {
   is_spyrer?: boolean;
   fighter_subtypes?: string[];
   owner_name?: string;
-  edition_slug?: string | null;
   campaigns?: Array<{
     campaign_id: string;
     resources?: Array<{ resource_name: string }>;
@@ -57,6 +56,7 @@ interface FighterActionsProps {
   fighter: Fighter;
   gang: Gang;
   fighterId: string;
+  editionSlug: string | null;
   userPermissions: UserPermissions;
   onFighterUpdate?: () => void;
   onStatusMutate?: (optimistic: Partial<Fighter>, gangCreditsDelta?: number, action?: 'kill' | 'retire' | 'sell' | 'release' | 'rescue' | 'starve' | 'recover' | 'capture' | 'delete') => any;
@@ -76,9 +76,10 @@ interface ActionModals {
 }
 
 export function FighterActions({ 
-  fighter, 
-  gang, 
-  fighterId, 
+  fighter,
+  gang,
+  fighterId,
+  editionSlug,
   userPermissions,
   onFighterUpdate,
   onStatusMutate,
@@ -169,7 +170,9 @@ export function FighterActions({
     ) ?? false;
   }, [fighter.campaigns]);
 
-  const beastNoun = beastSubtypeName(fighter?.edition_slug);
+  const beastNoun = beastSubtypeName(editionSlug);
+  // Selling a fighter to the Guilders is an N23 rule; N26 has no Guilders.
+  const showGuilderSale = hasGuilderSales(editionSlug);
   const isOwnedExoticBeast = fighter?.fighter_subtypes?.some(c => c.toLowerCase().startsWith('exotic beast') || c.toLowerCase() === 'pet') && !!fighter?.owner_name;
 
   // Calculate total vehicle equipment cost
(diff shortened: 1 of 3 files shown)


joeseos (commented):
@joeseos Reviewed the diff against the surrounding codebase. (Posted as a comment, not an approval — GitHub won't let this account approve its own PR.)

**Summary**: Adds `sellToGuilders` to `EDITION_CAPABILITIES` (n23: true, n26: false), gates the "Sell to Guilders" button on `hasGuilderSales(editionSlug)`, and threads an explicit `editionSlug` prop into `FighterActions` instead of smuggling it through the `fighter` object.

**Checks performed**:
- `types/edition.ts`: `sellToGuilders` row and `hasGuilderSales` accessor follow the existing pattern exactly (matches `hasAlignment`/`hasMasterCraftedWeapons` shape). No completeness-check issues.
- `fighter-page.tsx:599`: `editionSlug = fighterData.gang.edition_slug ?? fighterData.fighter.edition_slug ?? null` — same gang-first/fighter-fallback resolution already used by every other sibling prop pass (lines 730, 781, 823, 973, 1182, 1409, 1433), so `FighterActions` is now consistent rather than an outlier. `beastSubtypeName` picks up the better-resolved slug as a side effect, as described.
- `fighter-actions.tsx`: `FighterActions` is only ever instantiated from `fighter-page.tsx`, so making `editionSlug` a required prop doesn't strand any other caller. The removed `edition_slug` field on the local `Fighter` interface has no remaining references in the file.
- Edge case considered: could a fighter already be `enslaved: true` end up on an N26 gang, hiding the "Release from Guilders" button with no way back? Gang `edition_id` is set at gang-type creation and isn't reassigned by any update path I could find, so this isn't reachable in practice — not a blocker.
- No node_modules in this environment, so I couldn't run `tsc`/build directly; reviewed types by hand instead. Everything lines up (prop types, accessor signatures, `string | null` handling).

No correctness, security, or performance issues found. Verdict: **Approve** (findings-wise; approval itself blocked by GitHub for same-account PRs).

---

joeseos:
@joeseos Reviewed the diff against the codebase (`components/fighter/fighter-actions.tsx`, `components/fighter/fighter-page.tsx`, `types/edition.ts`).

## Correctness

The `editionSlug` prop-threading is sound: `fighter-page.tsx:599` already resolves `fighterData.gang.edition_slug ?? fighterData.fighter.edition_slug ?? null` (gang first, fighter fallback) and every other sibling prop at that call site uses this same value — passing it to `FighterActions` matches convention. `EDITION_CAPABILITIES.sellToGuilders` follows the existing table format exactly (alignment, comment style, `n23`/`n26` columns).

## Finding

**High — `fighter-actions.tsx:312-322`**: `showGuilderSale` gates the *entire* Guilders button, both directions. The button also renders "Release from Guilders" when `fighter.enslaved` is true — that release path is now hidden too whenever `hasGuilderSales(editionSlug)` is false. Worse, `isStatusIncompatible` (`utils/fighter-status.ts:76,81,91,96`) treats `enslaved` as blocking Kill, Retire, Capture, and Recover. So if an N26 fighter is ever enslaved, that fighter becomes completely stuck: no kill, no retire, no capture, no recover, and now no release either — a dead end with no in-UI recovery.

That scenario is reachable because of a second gap: **`app/actions/edit-fighter.ts:521-540`** (the `sell`/enslave case) has no edition check — it unconditionally sets `enslaved: true`. This PR only hides the client button; nothing stops the server action itself from enslaving a fighter under an edition with no Guilders rule (direct API call, a future caller of this action, etc.), which is exactly what would trigger the dead end above.

Suggested fix: keep the button visible whenever the fighter is already enslaved regardless of edition, e.g. `{(showGuilderSale || fighter?.enslaved) && (...)}`, so Release always has an escape hatch — and consider adding the matching edition guard server-side so `enslaved` can't be set at all under N26.

**Low — `types/edition.ts` null-slug behavior (pre-existing pattern, not new)**: `can()` returns `false` for a missing/unresolved `editionSlug`, same as `hasAlignment`/`hasMasterCraftedWeapons`. In practice gang-derived edition should always resolve to `n23` for legacy rows (backfilled in `20260630095837_add_editions_and_root_edition_ids.sql`), so this is unlikely to bite here, just flagging since it's a slightly different risk profile than an always-rendered button.

## Verdict: Changes requested

The prop-threading and capability-table changes are clean. The high finding is the blocker — gate only the "Sell" direction, not "Release," so an already-enslaved fighter is never left with no way out.

---

joeseos:
> @joeseos Reviewed the diff against the codebase (`components/fighter/fighter-actions.tsx`, `components/fighter/fighter-page.tsx`, `types/edition.ts`).
> 
> ## Correctness
> 
> The `editionSlug` prop-threading is sound: `fighter-page.tsx:599` already resolves `fighterData.gang.edition_slug ?? fighterData.fighter.edition_slug ?? null` (gang first, fighter fallback) and every other sibling prop at that call site uses this same value — passing it to `FighterActions` matches convention. `EDITION_CAPABILITIES.sellToGuilders` follows the existing table format exactly (alignment, comment style, `n23`/`n26` columns).
> 
> ## Finding
> 
> **High — `fighter-actions.tsx:312-322`**: `showGuilderSale` gates the *entire* Guilders button, both directions. The button also renders "Release from Guilders" when `fighter.enslaved` is true — that release path is now hidden too whenever `hasGuilderSales(editionSlug)` is false. Worse, `isStatusIncompatible` (`utils/fighter-status.ts:76,81,91,96`) treats `enslaved` as blocking Kill, Retire, Capture, and Recover. So if an N26 fighter is ever enslaved, that fighter becomes completely stuck: no kill, no retire, no capture, no recover, and now no release either — a dead end with no in-UI recovery.
> 
> That scenario is reachable because of a second gap: **`app/actions/edit-fighter.ts:521-540`** (the `sell`/enslave case) has no edition check — it unconditionally sets `enslaved: true`. This PR only hides the client button; nothing stops the server action itself from enslaving a fighter under an edition with no Guilders rule (direct API call, a future caller of this action, etc.), which is exactly what would trigger the dead end above.
> 
> Suggested fix: keep the button visible whenever the fighter is already enslaved regardless of edition, e.g. `{(showGuilderSale || fighter?.enslaved) && (...)}`, so Release always has an escape hatch — and consider adding the matching edition guard server-side so `enslaved` can't be set at all under N26.
> 
> **Low — `types/edition.ts` null-slug behavior (pre-existing pattern, not new)**: `can()` returns `false` for a missing/unresolved `editionSlug`, same as `hasAlignment`/`hasMasterCraftedWeapons`. In practice gang-derived edition should always resolve to `n23` for legacy rows (backfilled in `20260630095837_add_editions_and_root_edition_ids.sql`), so this is unlikely to bite here, just flagging since it's a slightly different risk profile than an always-rendered button.
> 
> ## Verdict: Changes requested
> 
> The prop-threading and capability-table changes are clean. The high finding is the blocker — gate only the "Sell" direction, not "Release," so an already-enslaved fighter is never left with no way out.
> 
> ---
> _Generated by [Claude Code](https://claude.ai/code)_

Deliberate decision to not protect the server action

PR #1994, opened 2026-08-13
Group multi-subtype fighter variants under one catalog entry

joeseos:
N23 fighter types carried exactly one subtype, so "same fighter, different
variant" was expressed purely through fighter_specialisation and the Add/Edit
Fighter modals collapsed rows sharing fighter_type plus an identical
fighter_subtypes array.

N26 lets a variant add a subtype of its own: "Master of Shadow (Leader)" 140
sits alongside "Master of Shadow (Leader, Wyrd)" 175. The arrays differ, so
both surfaced as unrelated catalog entries and the specialisation picker never
appeared.

/api/fighter-types now tags each row with the variant family it belongs to:
same fighter_type, and one row's subtype set containing the other's. Disjoint
sets stay separate, so N23's "Gunner (Ganger)" and "Gunner (Specialist)" keep
their own entries. This fills in typeSubtypeKey, which the Edit Fighter modal
already expected but nothing ever set, and replaces the two disagreeing rules
FighterAddModal carried (the option list keyed on the joined subtype array, the
picker on fighter_subtypes[0]).

Also sync the subtype selection when the Edit Fighter specialisation changes,
keeping subtypes no row in the family owns, so switching to a Wyrd variant no
longer saves the fighter without its Wyrd subtype.

Claude-Session: https://claude.ai/code/session_01UQ5wzES1hfr5Pk3AAc3miz

### DIFF
--- a/app/api/fighter-types/route.ts
+++ b/app/api/fighter-types/route.ts
@@ -103,6 +103,10 @@ const containsAllSubtypes = (subtypes: string[], base: string[]) => {
   return base.every(name => have.has(name.toLowerCase().trim()));
 };
 
+// include_all_types returns every edition at once and is only narrowed to the gang's own
+// edition on the client, so the edition is part of the bucket: a family never spans two.
+const familyBucket = (row: any) => `${row.edition_slug ?? ''}::${row.fighter_type}`;
+
 /**
  * Tags each row with its variant family — same fighter_type, and one row's subtype set
  * containing the other's — named after the family's smallest (base) set. N26 variants can
@@ -111,16 +115,16 @@ const containsAllSubtypes = (subtypes: string[], base: string[]) => {
  * "Gunner (Ganger)" and "Gunner (Specialist)" keep their own entries.
  */
 function withVariantGroups(rows: any[]) {
-  const byFighterType = new Map<string, any[]>();
+  const byBucket = new Map<string, any[]>();
   for (const row of rows) {
-    const group = byFighterType.get(row.fighter_type);
+    const group = byBucket.get(familyBucket(row));
     if (group) group.push(row);
-    else byFighterType.set(row.fighter_type, [row]);
+    else byBucket.set(familyBucket(row), [row]);
   }
 
   const baseOfSet = new Map<string, { key: string; names: string[] }>();
 
-  for (const [fighterType, group] of byFighterType) {
+  for (const [bucket, group] of byBucket) {
     const sets = new Map<string, string[]>();
     for (const row of group) {
       if (!sets.has(subtypeKey(row))) sets.set(subtypeKey(row), row.fighter_subtypes ?? []);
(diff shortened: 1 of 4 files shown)


joeseos:
## Code Review

@joeseos Reviewed the diff against the described N23/N26 scenarios (base subtype set vs. superset variant, disjoint variants staying separate) — traced both through `withVariantGroups` and they group correctly. Two non-blocking findings from a closer look at the edge cases:

### Findings

**1. `withVariantGroups` can silently mis-group when a row's subtype set is a superset of two disjoint sibling sets — Medium**
`app/api/fighter-types/route.ts:113-156` (particularly the `bases.filter(...).sort(...)[0]` pick around line 129-141)

The algorithm processes subtype-sets smallest-first and, for each set, attaches it to whichever existing "base" it's a superset of. This works for the linear chains shown in the PR (`[]` → `[Leader]` → `[Leader, Wyrd]`), but breaks down for a diamond shape: two disjoint single-subtype variants `[A]` and `[B]` in the same `fighter_type`, plus a third variant `[A, B]` that's a superset of *both*. `[A, B]` matches both existing bases (same `names.length`, so the tie-break falls to insertion order / alphabetical key), so it's silently merged into whichever base sorts first — while the other disjoint family stays separate and loses that variant's specialisation option. No error is thrown; it just produces a plausible-looking wrong grouping.

This is a narrow edge case (not present in the N23/N26 examples given), and I can't tell from the diff whether current game data ever produces it, but there's no test or guard against it, and the failure mode is silent. Worth either a comment noting the limitation, an assertion that rejects the ambiguous case (so it fails loud instead of silently), or a small unit test capturing the intended semantics of `withVariantGroups` given it's a non-trivial set-merging algorithm with no coverage.

**2. Variant-family lookup narrowed to `filteredTypes` in the Additions flow — Low**
`components/gang/fighter-add/FighterAddModal.tsx:342-343`

```js
const familyKey = selectedType?.typeSubtypeKey ?? typeId;
const fighterTypeGroup = filteredTypes.filter(t => (t.typeSubtypeKey ?? t.id) === familyKey);
```
Previously this scanned the full `fighterTypes` list; now it's scoped to `filteredTypes`, which in the Additions tab can be narrowed by `type.fighter_subtypes?.includes(selectedSubtype)` (line ~244). Since sibling variants in a family can carry different subtype arrays (that's the whole point of this PR), a sibling that doesn't include the currently-selected subtype filter value gets excluded from `filteredTypes`, and the specialisation picker silently drops to "single variant" for that entry. `.includes()` makes this benign for the common "filter by Leader, both Leader and Leader+Wyrd match" case, but it's a real behavior change for any subtype-specific filter value (e.g. filtering specifically by "Wyrd"). Worth confirming this is intentional (filter scope should follow the active catalog view) rather than an incidental side effect of reusing `filteredTypes`.

### Other notes
- Logic in `fighter-edit-modal.tsx`'s `handleSpecialisationChange` (preserving subtypes outside the variant family) is correctly picked up at submit time via the `sameFighterSubtypes` dirty-check at line ~1078, which overrides the earlier `fighterTypeToUse.fighter_subtypes` assignment — initially looked like a bug (submit uses the catalog row's subtypes directly at line ~1054) but the later override handles it correctly.
- No existing test suite in the repo, so the lack of tests for `withVariantGroups` isn't a deviation from convention, just worth calling out given the algorithm's subtlety.

### Verdict: **Approve**

The core fix works correctly for the described scenarios. The two findings above are edge-case/behavior-nuance concerns rather than concrete regressions in the described N23/N26 scope — worth a follow-up look, not blocking.

---

PR #1995, opened 2026-08-13
N26, promotion paths from prospect to ganger, to champion, to leader

Djidiouf:


joeseos:
@Djidiouf Reviewed the diff against the codebase (`app/actions/fighter-advancement.ts`, `utils/keepTypePromotionN26.ts`, the promotion modal/list/skills-list components, `types/edition.ts`), including cross-checking against RLS/auth conventions, the cache-tag invalidation helpers used elsewhere in `app/actions/*.ts`, and (via a read-only query) live production data to confirm one hypothesis.

This is a well-organized feature that follows the codebase's existing conventions closely: cache invalidation is delegated correctly to `updateFighterDetails`/`addSkillAdvancementInternal` (both already call the right invalidation helpers), auth relies on RLS the same way every other action in this file does, and the edition-gating pattern in `types/edition.ts` is applied consistently. The "revert-on-failure" composition in `applyKeepTypePromotionWithSkillGrant` mirrors the pre-existing `applyPromotionWithSkillAdvancement`, so it isn't a new anti-pattern. However, I found one high-severity correctness bug that's already reproducible against real production data, plus a clear logic inconsistency between two of the three promotion paths, and a few lower-severity items.

## Findings

**1. (High) `deleteAdvancement`'s Prospect-promotion "undo" detection has no subtype/edition guard and mis-fires on ordinary specialist Gangers, corrupting their subtypes and specialisation.**
`app/actions/fighter-advancement.ts:1371` (`isProspectPromotionGrant = isN26ProspectPromotionSkillGrant(...)`), implemented in `utils/keepTypePromotionN26.ts:92-101`.
Unlike the Ganger→Champion and Champion→Leader variants (`utils/keepTypePromotionN26.ts:159`, `:203`, which cross-check catalog subtypes at `app/actions/fighter-advancement.ts:1403-1429`), the Prospect check only compares `fighter_specialisation_id` + skill name/id — it doesn't check current subtypes or catalog data, and isn't edition-gated (`hasProspectSpecialisationPromotion` is only checked inside `applyN26ProspectPromotion`, never inside `deleteAdvancement`). The eight specialisation IDs in `N26_PROSPECT_SPECIALISATIONS` are the same shared, general-purpose `fighter_specialisations` catalog rows (Heavy, Gunner, Gunslinger, Scout, Sniper, Brawler, Medic, Tech) used throughout the app for any specialist Ganger, in any edition, regardless of whether they ever went through this new promotion flow. I confirmed via a read-only query against the live database that there are existing production fighters (e.g. a `["Ganger","Specialist"]` fighter with a non-advance "Precision Shot" skill and `fighter_specialisation_id` = the Sniper catalog id) that exactly match this signature today. If such a player deletes that starting skill from the Skills list (already possible pre-PR — delete is enabled for any `is_advance:false` skill), `deleteAdvancement` will: skip the stash refund, silently strip their `fighter_specialisation`/`fighter_specialisation_id`, and rewrite `fighter_subtypes` via `buildN26ProspectDemotionSubtypes` — replacing `Ganger`/`Specialist` with `Prospect` on a fighter that was never promoted from Prospect at all. This is real data corruption on a common, legitimate action, not a contrived edge case. Suggested fix: require the fighter's current subtypes to actually look like a post-promotion Prospect (e.g. include `Ganger`+`Specialist` and not be a "native" specialist-Ganger catalog type), and/or gate this branch on `hasProspectSpecialisationPromotion(editionSlug)`, matching the guard already applied to the sibling functions.

**2. (High/Medium — functional bug) `applyN26GangerChampionPromotion` permanently blocks promotion for any fighter that already legitimately knows Inspiring, instead of just skipping the redundant grant.**
`app/actions/fighter-advancement.ts:1135-1147`. If `fighterAlreadyHasSkill(...)` is true, the function returns `{ success: false, error: 'This fighter already has the skill Inspiring' }` and aborts the *entire* promotion (no subtype change happens either). Compare with `applyN26ChampionLeaderPromotion` (`app/actions/fighter-advancement.ts:1249-1252`, `1273`), which computes `alreadyHasInspiring` and instead passes `skillId: alreadyHasInspiring ? undefined : skillResolved.skillId` — letting the promotion (subtype/type change) proceed and simply omitting the redundant grant. Inspiring is a normal, purchasable Advancement skill in Necromunda, so a Ganger who bought it as an Advancement before reaching Champion rank is a perfectly plausible, rules-legal state, and under this PR that fighter can never use the keep-type Ganger→Champion promotion — it will always error out. Suggested fix: make `applyN26GangerChampionPromotion` mirror the Champion→Leader handling (skip the skill grant, still update subtypes, when the skill is already known).

**3. (Low) Server-only `ratingOnly` flag on `addSkillAdvancementInternal` is only a TypeScript-level boundary, not a runtime one.**
`app/actions/fighter-advancement.ts:114-116` documents `ratingOnly` as "server-only ... clients must not set these," but the exported `'use server'` action just forwards `params` straight through with no runtime stripping/validation. Not a regression introduced by this PR specifically (the same file already lets a client set `is_advance: true` with an arbitrary `credits_increase`/`xp_cost` and no server-side recomputation), but worth a follow-up ticket.

**4. (Low) No stash/rate-limiting or DB constraint against a duplicate promotion grant from a double-submit.**
`fighter_skills` has no unique constraint on `(fighter_id, skill_id)` (confirmed against `supabase/schema/schema.public.sql:5166`), the check-then-insert in `applyKeepTypePromotionWithSkillGrant` isn't transactional, and the promotion modal's mutation is fire-and-forget rather than awaited before closing. Low real-world likelihood given the narrow window, but worth a follow-up (e.g. a partial unique index on `fighter_skills(fighter_id, skill_id) where is_advance = false`).

**5. (Low, cosmetic) Confusing error copy in the Prospect-undo ordering check.**
`app/actions/fighter-advancement.ts:1382-1386`. When a Prospect-grant skill can't be deleted because the fighter is now Champion/Leader, the message reads "Undo the Leader promotion before undoing the Ganger promotion" — but the skill being deleted is the *Prospect* promotion's specialisation skill, not "the Ganger promotion." Slightly misleading to a user trying to figure out what to click.

**6. (Informational, not a blocker) `applyN26ChampionLeaderPromotion` validates the target fighter type's edition but not that it belongs to the fighter's own gang type.**
`app/actions/fighter-advancement.ts:1195-1207`. This mirrors the exact same gap already present in the standard (non-N26) promotion flow, so it's not a regression — just flagging that this PR extends an existing app-wide trust boundary (client-side filtering only) rather than closing it.

**Positives worth calling out:** cache invalidation is handled correctly by composing existing, already-invalidating helpers rather than reinventing it; the `types/edition.ts` capability additions are clean and consistent with the existing `EDITION_CAPABILITIES` pattern; subtypes are always rebuilt server-side rather than trusted from the client; and the Ganger→Champion/Champion→Leader *demotion-detection* logic (unlike the Prospect one) does correctly cross-check catalog subtypes to avoid misfiring on a "starting Champion" fighter type — showing the author was aware of this exact false-positive class of bug, just didn't apply the same care to the Prospect path.

## Verdict: Changes requested

Blockers: Finding #1 (Prospect-undo data corruption, verified against live data) and Finding #2 (Ganger→Champion promotion permanently blocked for fighters that already know Inspiring) should be fixed before merge — both are concrete, reachable bugs, not speculative. Findings #3–#6 are non-blocking but worth addressing in this PR or a fast follow-up.

---

joeseos:
@Djidiouf Reviewed the new commit only (`27c5b1e...2bcc75e`, pushed after the previous review) against `app/actions/fighter-advancement.ts`, `utils/keepTypePromotionN26.ts`, `components/fighter/fighter-skills-list.tsx`, `components/fighter/fighter-page.tsx`, `app/fighter/[id]/page.tsx`, and `app/lib/shared/fighter-data.ts`.

## Previous findings

1. **(High) Prospect-undo mis-fires on native specialist Gangers** — **Resolved.** The new `isN26ProspectPromotionUndoCandidate()` (`utils/keepTypePromotionN26.ts:110-134`) now requires the fighter's current subtypes to look post-promotion (`Ganger`+`Specialist`) *and* the fighter's catalog type to still be `Prospect` (and not `Ganger`), plus edition-gates via `editionAllowsProspectPromotion`. This correctly excludes native specialist Gangers, whose catalog type already contains `Ganger`. Good fix, and it's used consistently in `deleteAdvancement` and both client call sites in `fighter-skills-list.tsx`.
2. **(High) Ganger→Champion permanently blocked when fighter already knows Inspiring** — **Resolved.** `applyN26GangerChampionPromotion` (`app/actions/fighter-advancement.ts:1136-1157`) now computes `alreadyHasInspiring` and passes `skillId: alreadyHasInspiring ? undefined : skillResolved.skillId`, mirroring the existing Champion→Leader handling exactly as suggested.
3. **(Low) `ratingOnly` not runtime-enforced** — **Outstanding**, not touched by this commit (pre-existing, already flagged as a follow-up item).
4. **(Low) No DB constraint / transactionality against duplicate promotion grant** — **Outstanding**, not touched by this commit.
5. **(Low, cosmetic) Confusing "undo the Ganger promotion" copy** — **Resolved.** Now reads "before undoing the Prospect promotion" (`app/actions/fighter-advancement.ts:1412-1413`), matching what's actually being undone.
6. **(Informational) Champion→Leader doesn't validate target type's gang** — **Outstanding**, acknowledged as an existing app-wide gap, not addressed (as expected — it was flagged as non-blocking).

## New finding

**(High) The "undo Champion/Leader before undoing Prospect" ordering guard silently stops firing once a fighter reaches Leader rank, allowing the specialisation-grant skill to be deleted with an incorrect stash refund.**

- Server: `app/actions/fighter-advancement.ts:1402-1408`
- Client (same condition, mirrors server): `components/fighter/fighter-skills-list.tsx:860-868`

Both guards require `catalogSubtypesForUndo` / `fighterCatalogSubtypes` — the fighter's **current** catalog `fighter_types` row — to include `'Prospect'` and not `'Ganger'`. That holds for a Prospect-origin fighter through Ganger and Champion (keep-type promotions never touch `fighter_type_id`, so the catalog type stays the original Prospect row). But `applyN26ChampionLeaderPromotion` (`fighter-advancement.ts:1258-1268`, via `fighterTypeId: leaderType.id`) rewrites `fighter_type_id` to the new Leader catalog type on promotion — whose only guaranteed subtype is `'Leader'` (checked at `fighter-advancement.ts:1232`), not `'Prospect'`.

So once a Prospect→Ganger→Champion→Leader fighter's page is loaded fresh (server-fetched catalog data, not the stale client-optimistic state), `catalogSubtypesForUndo.includes('Prospect')` is `false`, the ordering block never triggers, and neither `isProspectPromotionGrant`, `isGangerChampionPromotionGrant`, nor `isLeaderPromotionGrant` matches the specialisation skill (e.g. "Precision Shot") either — so it falls through to plain deletion with `refundStash = !skillData.is_advance` = `true`. That refunds `credits_increase` (15) to the gang stash even though those 15 credits were a rating-only grant (`ratingOnly: true`) never taken from stash — a free-credits bug — and leaves `fighter_specialisation`/`fighter_specialisation_id` stale (still set, no backing skill) since no demotion path runs.

This is reachable via the exact promotion chain this PR implements (Prospect→Ganger→Champion→Leader is the intended happy path), not a contrived edge case.

Suggested fix: for the ordering guard specifically, don't rely on the fighter's *current* catalog type (which changes on Leader promotion) — e.g. check `currentSubtypesForGrant.includes('Leader')` directly (already computed) alongside the specialisation/skill match and edition gate, instead of re-deriving it from `catalogSubtypesForUndo`. The `Champion` case is fine as-is since keep-type promotions don't change `fighter_type_id`, but the `Leader` case needs to stop depending on catalog data that gets replaced by the promotion itself.

## Other notes (non-blocking)

- `fighter-advancement.ts:1390-1392` (`if (isProspectPromotionGrant && !fighter.fighter_type_id) { isProspectPromotionGrant = false; }`) is dead code — `catalogSubtypesForUndo` is already `[]` when `fighter.fighter_type_id` is falsy, so `isN26ProspectPromotionUndoCandidate` already returns `false` via the `catalog.includes('Prospect')` check. Harmless, just redundant.
- Minor, related to the new finding rather than separate: `components/fighter/fighter-page.tsx:914-925`'s `onFighterDetailsUpdate` handler keeps the *old* `fighter_type.fighter_subtypes` when a promotion changes `fighter_type_id` (Champion→Leader). It self-corrects on refresh, but means the client's optimistic view of catalog subtypes is briefly wrong immediately after promoting — worth a follow-up if it causes visible UI flicker, not blocking.

## Verdict: Changes requested

Both previously-blocking findings are fixed correctly. One new high-severity issue (Leader-rank ordering-guard bypass) should be fixed before merge — it's in the direct happy path of the feature and has a concrete, reproducible stash/rating consequence.

---

joeseos:
@Djidiouf Reviewed the new commit only (`2bcc75e...1ea9a27`, pushed after the previous review) against `app/actions/fighter-advancement.ts`, `components/fighter/fighter-skills-list.tsx`, and `utils/keepTypePromotionN26.ts`.

## Previous findings

1. **(High) Prospect-undo mis-fires on native specialist Gangers** — Resolved (prior commit), untouched here.
2. **(High) Ganger→Champion permanently blocked when fighter already knows Inspiring** — Resolved (prior commit), untouched here.
3. **(High) Leader-rank ordering guard silently stops firing after Champion→Leader promotion, letting the Prospect specialisation skill be deleted with a wrong stash refund** — **Resolved.** The new `isN26ProspectPromotionSkillBlockedByHigherRank()` (`utils/keepTypePromotionN26.ts:147-170`) replaces the old catalog-only check with one that (a) blocks unconditionally when `currentSubtypes` includes `'Leader'`, and (b) for `'Champion'`, blocks when the catalog is either keep-type Prospect *or* still shows `'Leader'` (the post type-change/post-Leader-undo case, since Leader-undo intentionally doesn't restore `fighter_type_id` — see the comment at `fighter-advancement.ts:1497`). I traced the actual post-Leader-undo state (`fighter_type_id` stays on the Leader catalog row, subtypes get `Champion` re-added) and the new function correctly still blocks in that state via the `catalog.includes('Leader')` branch. Both the server check (`fighter-advancement.ts:1396-1412`) and the client `disabled` callback (`fighter-skills-list.tsx:859-868`) now call the exact same shared function with matching arguments, which is a nice improvement over last round's parallel-but-separate conditions — one place to keep correct going forward.
4. **(Low) `ratingOnly` not runtime-enforced** — Outstanding, not touched by this commit (pre-existing, already flagged as a follow-up item).
5. **(Low) No DB constraint / transactionality against duplicate promotion grant** — Outstanding, not touched by this commit.
6. **(Low, cosmetic) Confusing "undo the Ganger promotion" copy** — Resolved (prior commit), untouched here.
7. **(Informational) Champion→Leader doesn't validate target type's gang** — Outstanding, acknowledged non-blocking, not addressed (as expected).
8. **(Informational) Dead `!fighter.fighter_type_id` branch in the Prospect check** — Outstanding, still present at `fighter-advancement.ts:1390-1392`, still harmless/redundant.
9. **(Informational) `fighter-page.tsx` keeps stale optimistic catalog subtypes right after a Leader promotion** — Outstanding, not touched by this commit.

## New observations (non-blocking)

- `isN26ProspectPromotionSkillBlockedByHigherRank`'s `'Leader'` branch (`utils/keepTypePromotionN26.ts:165`) blocks purely on the specialisation-id/skill-name signature plus `currentSubtypes.includes('Leader')`, with no catalog cross-check — unlike its own `'Champion'` branch two lines below, and unlike `isN26ProspectPromotionUndoCandidate`, both of which verify against the catalog type before acting. In principle a native N26 Leader fighter that happens to hold one of the eight house-specialisation ids plus the matching skill (for reasons unrelated to this promotion chain) would get incorrectly blocked from deleting that skill. Necromunda's specialisation mechanic is normally Ganger/Specialist-only, so a native Leader carrying `fighter_specialisation_id` is likely rare-to-impossible in practice, and the failure mode here is just an over-cautious block (bad error message), not data corruption — so I'd call this informational rather than a blocker, but worth a one-line comment if intentional.
- The refactor to a single shared `isN26ProspectPromotionSkillBlockedByHigherRank` used identically by both server and client is a good simplification — removes the risk of the two conditions drifting apart again, which is effectively what caused finding #3 above in the first place.
- Confirmed no other references to the old `isN26ProspectPromotionSkillGrant` import were left dangling after the rename in `fighter-advancement.ts` / `fighter-skills-list.tsx`.

## Verdict: Approve

The one previously-outstanding high-severity issue (Leader-rank ordering guard bypass) is fixed correctly and the fix is applied consistently on both server and client. No new blocking issues in this commit. Remaining outstanding items are all previously-flagged low/informational items that were never required for this round.

---

joeseos:
@Djidiouf Reviewed the new commit only (`1ea9a27...c14ca37`, pushed after the last "Approve" review) — a one-line change in `app/lib/shared/fighter-data.ts`.

## Previous findings

1. **(High) Prospect-undo mis-fires on native specialist Gangers** — Resolved (prior commit), untouched here.
2. **(High) Ganger→Champion permanently blocked when fighter already knows Inspiring** — Resolved (prior commit), untouched here.
3. **(High) Leader-rank ordering guard silently stops firing after Champion→Leader promotion** — Resolved (prior commit), untouched here.
4. **(Low) `ratingOnly` not runtime-enforced** — Outstanding, not touched by this commit.
5. **(Low) No DB constraint / transactionality against duplicate promotion grant** — Outstanding, not touched by this commit.
6. **(Low, cosmetic) Confusing "undo the Ganger promotion" copy** — Resolved (prior commit), untouched here.
7. **(Informational) Champion→Leader doesn't validate target type's gang** — Outstanding, acknowledged non-blocking.
8. **(Informational) Dead `!fighter.fighter_type_id` branch in the Prospect check** — Outstanding, still present, still harmless.
9. **(Informational) `fighter-page.tsx` keeps stale optimistic catalog subtypes right after a Leader promotion** — Outstanding, not touched by this commit.

## This commit

`app/lib/shared/fighter-data.ts:1211` drops the `-v2` suffix from the `unstable_cache` key in `getFighterTypeInfo`, reverting `[\`fighter-type-${fighterTypeId}-v2\`]` back to `[\`fighter-type-${fighterTypeId}\`]`.

I checked the two claims in the commit message against the codebase:
- **Server enforcement is unaffected:** `deleteAdvancement` (`app/actions/fighter-advancement.ts:1371`) queries `fighter_types.fighter_subtypes` via a fresh, uncached Supabase call — it never goes through `getFighterTypeInfo`. So the demotion/undo-ordering guards this PR's three prior review rounds hardened are not affected by this cache-key change.
- **Nothing actively invalidates the tag:** there's no `revalidateTag(CACHE_TAGS.GLOBAL_FIGHTER_TYPES())` call anywhere in the repo, so on deploy, any pre-existing cache entry under the unsuffixed key (from before this PR's `fighter_subtypes` field was added to the select) can be served stale for up to the full `revalidate: 3600` (1 hour) window rather than being busted immediately. That's exactly the scenario the `-v2` suffix existed to avoid, and removing it reintroduces a bounded (≤1h), client-only staleness window: the promotion-undo delete button's optimistic pre-disable may not kick in until the cache naturally refreshes, but the server-side guard still catches it correctly (returns the ordering error as a toast) since it's uncached.

This is a reasonable simplification — a permanent key-naming workaround for a one-time migration concern, traded for a small, self-correcting, UI-only staleness window with no data-integrity impact. No new issues found.

## Verdict: Approve

---

PR #1996, opened 2026-08-13
Restore the gap above the N26 weapon table

joeseos:
N23 weapon tables render a two-row header: a small Rng/Acc group row above
the S/L/Str/AP/D/Am row. That group row carries no borders, so it reads as a
gap between the fighter's characteristics and the weapon profiles.

N26 has no grouped columns and so emitted a single header row, leaving
Weapon/SR/LR/Str/AP/L flush against the characteristics table.

Add an empty stand-in row to the N26 header using the same padding and font
size classes, so the gap matches N23 in every view mode and in print rather
than relying on a hardcoded margin.

Claude-Session: https://claude.ai/code/session_01PgBtMSzgToXpYniKWJaeT1

### DIFF
--- a/components/gang/fighter-card-weapon-table.tsx
+++ b/components/gang/fighter-card-weapon-table.tsx
@@ -195,17 +195,24 @@ const WeaponTable: React.FC<WeaponTableProps> = ({ weapons, entity, viewMode, ed
         </colgroup>
         <thead>
           {usesLethality ? (
-            <tr>
-              <th className={`${pClass} text-left`}>
-                {entity === 'vehicle' ? 'Vehicle Weapon' : entity === 'crew' ? 'Crew Weapon' : 'Weapon'}
-              </th>
-              <th className={`${pClass} text-center border-l border-black`}>SR</th>
-              <th className={`${pClass} text-center border-l border-black`}>LR</th>
-              <th className={`${pClass} text-center border-l border-black`}>Str</th>
-              <th className={`${pClass} text-center border-l border-black`}>AP</th>
-              <th className={`${pClass} text-center border-l border-black`}>L</th>
-              <th className={`${pClass} text-left border-l border-black`}>Traits</th>
-            </tr>
+            <>
+              {/* Empty stand-in for N23's Rng/Acc group row, so both editions leave
+                  the same gap under the characteristics table. */}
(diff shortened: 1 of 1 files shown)


joeseos:
@joeseos Reviewed the diff against `components/gang/fighter-card-weapon-table.tsx` (the only changed file), cross-checking the `<colgroup>` column count, the N23 branch it's meant to visually match, and `types/edition.ts` for what `usesLethality` gates.

## Correctness

The fix is straightforward and correct:
- `usesLethality` (N26) branch's `<colgroup>` renders exactly 7 columns when lethality is active (1 name col + 5 shared stat cols, the extra 3 N23-only cols are skipped, + 1 traits col at 35%), matching the new spacer row's `colSpan={7}` exactly.
- The spacer row deliberately carries no `border-l`/`border-black` classes, mirroring N23's own borderless colSpan=5 filler cell in its group row (`fighter-card-weapon-table.tsx:224`) — this is what produces the "gap, no line" look being restored.
- Reusing `pClass` and `rngAccHeaderSizeClass` (rather than a new hardcoded height/margin) means the row's height tracks the same small-font sizing already used for N23's Rng/Acc labels across `normal`/`4-card`/`print` view modes, so the two editions' header heights should stay in sync if those classes ever change.
- `aria-hidden="true"` on the spacer `<tr>` is appropriate since it carries no real content (`&nbsp;` only), unlike N23's group row which has visible "Rng"/"Acc" text and is correctly left un-hidden.
- Wrapping the two `<tr>`s in a fragment inside the ternary is valid JSX and doesn't affect the `!usesLethality` (N23) branch at all — that branch is untouched.

Checked `app/globals.css` for a `.table-weapons th` rule that could reintroduce a border on this cell (there's only a print-only padding override, no border), so nothing undoes the borderless intent.

## Nitpick (non-blocking)

`colSpan={7}` is a hardcoded literal with no reference back to why it's 7 (unlike the colgroup just above it, which has a comment explaining the shared 5 stat columns). Since this branch's column count is fixed by the surrounding structure, this isn't a real risk today, just a minor readability gap if someone adds an N26-only column later without noticing this row needs updating too.

## Verdict: Approve

Small, well-scoped fix that does what the description says. No correctness, security, or perf concerns.

---

PR #1997, opened 2026-08-14
Hide XP cost from advancements list in cumulative-XP editions

joeseos:
N26 fighters earn Advancements by rank rather than spending XP, so the advancements table's XP column carries no meaning and the undo dialog's promise of an XP refund is wrong — the delete action already leaves fighter XP untouched in that edition.

Derive isCumulativeXp from the editionSlug prop AdvancementsList already receives, drop the XP column (widening Name to fill the space), and reword the undo confirmation to mention only the value adjustment.

### DIFF
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -2957,6 +2957,10 @@ export function AdvancementsList({
   const [isStandalonePromotionOpen, setIsStandalonePromotionOpen] = useState(false);
   const [deleteModalData, setDeleteModalData] = useState<{ id: string; name: string; type: string } | null>(null);
 
+  // No XP is spent on Advancements in a rank-based edition, so there is no XP
+  // cost to list and nothing to refund when one is undone.
+  const isCumulativeXp = hasCumulativeXp(editionSlug);
+
   const showPromoteButton = fighterSubtypes.some(c => ['Ganger', 'Juve', 'Prospect', 'Champion', 'Specialist', 'Exotic Beast', 'Exotic Beast Specialist'].includes(c));
 
   const { data: preFetchedFighterTypes = [] } = useQuery({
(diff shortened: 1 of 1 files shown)


joeseos:
@joeseos Reviewed the diff against `components/fighter/fighter-advancement-list.tsx` (the only changed file), cross-checking `hasCumulativeXp`/`editionSlug` propagation from `fighter-page.tsx`, the `deleteAdvancement` server action's XP-refund logic, and the `List` component's column type.

## Correctness

- `editionSlug` is correctly threaded: `fighter-page.tsx:851` passes `fighterData.gang?.edition_slug ?? fighterData.fighter?.edition_slug ?? null` into `AdvancementsList`, whose prop already defaults to `null`, so `hasCumulativeXp(editionSlug)` resolves correctly for both editions.
- The premise checks out against `app/actions/fighter-advancement.ts:765`: `deleteAdvancement` computes `spendsXp = !hasCumulativeXp(editionSlugOf(fighter))` and only adds `xpToRefund` back to `fighter.xp` when `spendsXp` is true (`:966`). So for a cumulative-XP (N26) edition, undoing an advancement never refunds XP — the reworded confirmation text ("The fighter's value will be adjusted accordingly.") accurately reflects that, while the non-cumulative branch keeps the original wording verbatim.
- The `credits_increase` column remains meaningful in the N26 branch — rank-based Advancements still carry a rating-only credits cost (confirmed elsewhere in the codebase, e.g. `ratingOnly: true` grants with `credits_increase` set), so dropping only the XP column and keeping Cost is the right call, not an oversight.
- `List`'s `Column` type (`components/ui/list.tsx:9-10`) has `align`/`width` as optional, so the narrower cumulative-XP column array (`Name` 75% + `Cost`, no `width`) type-checks fine and follows the same shape already used elsewhere in this file.
- `isCumulativeXp` naming/pattern matches an existing usage earlier in the same file (line 526, a different component), so this doesn't introduce a new convention.

Nothing incorrect found — the diff is small, well-scoped, and the underlying behavioral premise (no XP is spent or refunded in N26) is verified against the actual delete/refund code path rather than just the UI.

## Verdict: Approve

No blocking issues. No correctness, security, or perf concerns.

---

PR #1998, opened 2026-08-14
Add edition_id support to fighter_gang_legacy table

joeseos:
## Summary
This PR extends edition filtering support to the `fighter_gang_legacy` table, which previously only existed on `gang_affiliation`. This enables consistent edition-scoped management of both gang affiliation and legacy lineages throughout the admin interface.

## Key Changes

- **Database Migration**: Added `edition_id` column to `fighter_gang_legacy` table with foreign key constraint and index, backfilled from associated fighter types with fallback to 'n23' edition
- **Type Definitions**: Added optional `edition_id` field to `FighterType` and `GangType` interfaces
- **Edition Filtering**: 
  - Moved edition selection to always appear (not just for affiliations) in both create and edit modals
  - Created `filteredGangTypes` derived state to filter gang types by selected edition
  - Updated `filteredFighterTypes` to filter by both gang type and edition
  - Created `accessFighterTypeOptions` derived state for edition-scoped fighter type access rules
- **Edition Change Handler**: Enhanced `handleEditionChange` to clear conflicting selections across gang types, fighter types, and access rules when edition changes
- **API Updates**: Simplified backend logic to always include `edition_id` in queries and mutations for both lineage types, removing conditional branching
- **UI Layout**: Adjusted grid layout from 3 to 4 columns to accommodate edition selector in edit view

## Implementation Details

- Edition filtering now applies uniformly to both affiliation and legacy lineages
- The edition selector is always visible and functional, providing consistent UX
- Derived state uses `useMemo` to efficiently filter options based on edition and gang type selections
- Edition conflicts are automatically resolved by clearing incompatible selections rather than blocking operations

### DIFF
--- a/app/api/admin/gang-lineages/route.ts
+++ b/app/api/admin/gang-lineages/route.ts
@@ -36,12 +36,7 @@ export async function GET(request: Request) {
 
   const type = typeParam as LineageType;
   const table = getTableName(type);
-  // edition_id only exists on gang_affiliation, not fighter_gang_legacy.
-  // Cast to the wider literal so the supabase query parser can type the result;
-  // for the legacy branch edition_id is simply absent (undefined) at runtime.
-  const lineageColumns = (type === 'affiliation'
-    ? 'id, name, fighter_type_id, created_at, updated_at, edition_id'
-    : 'id, name, fighter_type_id, created_at, updated_at') as 'id, name, fighter_type_id, created_at, updated_at, edition_id';
+  const lineageColumns = 'id, name, fighter_type_id, created_at, updated_at, edition_id';
 
   try {
     if (id) {
@@ -175,13 +170,13 @@ export async function POST(request: Request) {
 
     const table = getTableName(data.type as LineageType);
 
-    // Create the lineage in the specific table (edition_id only exists on gang_affiliation)
+    // Create the lineage in the specific table
     const { data: newLineage, error: insertError } = await supabase
       .from(table)
       .insert({
         name: data.name,
         fighter_type_id: data.fighter_type_id,
-        ...(data.type === 'affiliation' && { edition_id: data.edition_id || null })
+        edition_id: data.edition_id || null
       })
       .select()
       .single();
@@ -254,7 +249,7 @@ export async function PATCH(request: Request) {
         .update({
           name: data.name,
           fighter_type_id: data.fighter_type_id,
-          ...(currentType === 'affiliation' && { edition_id: data.edition_id || null }),
+          edition_id: data.edition_id || null,
           updated_at: new Date().toISOString()
         })
         .eq('id', id);
@@ -288,13 +283,13 @@ export async function PATCH(request: Request) {
     const oldFightersFk = currentType === 'legacy' ? 'fighter_gang_legacy_id' : 'gang_affiliation_id';
     const newFightersFk = newType === 'legacy' ? 'fighter_gang_legacy_id' : 'gang_affiliation_id';
 
-    // Create new record in new table (edition_id only exists on gang_affiliation)
+    // Create new record in new table
     const { data: insertedNew, error: insertNewErr } = await supabase
       .from(newTable)
       .insert({
         name: data.name,
         fighter_type_id: data.fighter_type_id,
-        ...(newType === 'affiliation' && { edition_id: data.edition_id || null })
+        edition_id: data.edition_id || null
       })
       .select()
       .single();
--- a/app/api/admin/gang-lineages/route.ts
+++ b/app/api/admin/gang-lineages/route.ts
@@ -17,6 +17,22 @@ function getTableName(type: LineageType) {
 // Always uses fighter_gang_legacy_id as the foreign key
 const JUNCTION_FK_COLUMN = 'fighter_gang_legacy_id';
 
+// edition_id is derived from the associated fighter type, never taken from the
+// client, so a lineage can never disagree with the edition it belongs to
+async function getFighterTypeEditionId(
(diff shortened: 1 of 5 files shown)


joeseos:

## Summary
Extends `edition_id` support from `gang_affiliation` to `fighter_gang_legacy`, unifying edition filtering across both lineage types. The diff is clean and the migration is safe (`IF NOT EXISTS` everywhere, `ON DELETE RESTRICT`, indexed). No crashes or correctness regressions found. A few items are worth a look, but they're mostly extensions of patterns that already existed for `gang_affiliation` rather than new problems introduced by this PR.

## Findings

**1. No server-side check that `edition_id` matches the fighter type's actual edition (Low-Medium)**
`app/api/admin/gang-lineages/route.ts:179`, `:252`, `:292` — `POST`/`PATCH` insert/update `edition_id: data.edition_id || null` straight from the client body, with no cross-check against `fighter_type_id`'s own `edition_id`. The admin UI keeps them in sync via `filteredFighterTypes`/`filteredGangTypes`, but the API itself doesn't enforce it, so a stale client or manual API call can save a `fighter_gang_legacy`/`gang_affiliation` row whose edition disagrees with its associated fighter type. This gap already existed for `gang_affiliation`; this PR doubles the surface by extending the same unchecked write to `fighter_gang_legacy`. Suggest deriving `edition_id` server-side from `fighter_type_id` (or at least validating it matches) rather than trusting the client value.

**2. Top-level lineage-browse `EditionSelect` loses its "all editions" option as soon as it auto-defaults (Low)**
`components/admin/admin-gang-lineage.tsx:600` — this instance is used to *filter the list of existing lineages* to pick one for editing, not just to set a value on the record being created/edited. `components/edition-select.tsx:61-64` documents that "filter contexts keep a blank 'all editions' choice," but this instance is passed `defaultToCurrent`, so the moment editions load, `handleEditionChange` fires automatically, `value` becomes non-empty, and `showBlankOption` (edition-select.tsx:64) goes false — removing the blank `<option>` from the DOM for the rest of that modal session. Any lineage with a `null` or non-current `edition_id` becomes unreachable through this dropdown until the modal is closed and reopened. This behavior already existed for the affiliation-only selector before this PR; the PR extends it to legacies too (and now also drives the new `selectedGangTypeId`/`accessRuleGangTypeId`/`fighterTypeAccess` clearing side effects in `handleEditionChange`, lines 215-242, on that same automatic first fire). Worth giving this particular usage a real "All editions" affordance distinct from the create/edit-form default behavior.

**3. Migration's `n23` fallback fails silently if that edition is missing (Low)**
`supabase/migrations/20260814120000_add_edition_id_to_fighter_gang_legacy.sql:17-19` — `(SELECT id FROM public.editions WHERE slug = 'n23')` evaluates to `NULL` if no such row exists, so the `UPDATE` becomes a no-op and rows are silently left with `edition_id IS NULL`, contrary to the migration's stated intent. The original editions migration (`supabase/migrations/20260630095837_add_editions_and_root_edition_ids.sql`) uses `SELECT id INTO STRICT current_edition_id ... WHERE slug = 'n23'`, which raises a clear error instead of silently degrading. Consider following that same STRICT pattern here for consistency and safety.

## Verdict
**Approve.** No blocking correctness, security, or data-loss issues. The findings above are pre-existing patterns being extended rather than new bugs, but items 1 and 2 are worth addressing in a follow-up given they now apply to two catalog tables instead of one.

---

joeseos:
@joeseos Follow-up review — covers the two commits since the last review (`49c8c74`, `628705e`).

## Previous findings

1. **No server-side check that `edition_id` matches the fighter type's actual edition** — **Resolved.** `628705e` adds `getFighterTypeEditionId()` in `app/api/admin/gang-lineages/route.ts` and both `POST`/`PATCH` now derive `edition_id` from `fighter_type_id` server-side instead of trusting the client body; the client (`admin-gang-lineage.tsx`) no longer sends `edition_id` at all. A lineage can no longer disagree with its fighter type's edition.

2. **Top-level lineage-browse `EditionSelect` loses its "all editions" option on auto-default** — **Outstanding.** `admin-gang-lineage.tsx:598` still passes `defaultToCurrent` to the browse-filter `EditionSelect`, unchanged by these commits. Not a regression, just not addressed yet.

3. **Migration's `n23` fallback fails silently if that edition is missing** — **Resolved.** The migration now uses `SELECT id INTO STRICT base_edition_id ... WHERE slug = 'n23'` inside a `DO $$` block, matching the original editions migration's pattern — a missing base edition now raises instead of silently leaving rows `NULL`.

## New changes in this round

`49c8c74` is an unrelated but related fix bundled into the same push: `/api/gang-types` previously attached the *entire* `gang_affiliation` table to every gang type with `affiliation` enabled, so an N26 gang type offered all 38 affiliations instead of its own edition's subset. The fix joins `editions:edition_id (slug)` onto both the `gang_types` and `gang_affiliation` queries and filters `available_affiliations` per gang type via `sameEditionForDisplay`. `gang-edit-modal.tsx`'s affiliation-list fallback is scoped the same way (previously picked the first gang type with *any* affiliations, ignoring edition).

No new correctness issues found in either commit — the edition derivation and the affiliation scoping both use the existing `types/edition.ts` helpers consistently with how they're used elsewhere in the codebase, error paths are still caught by the existing try/catch in the API routes, and null/missing-edition cases fall back to the documented n23-default behavior rather than crashing.

## Verdict
**Approve.** Findings 1 and 3 are resolved cleanly. Finding 2 remains open but is low severity and pre-existing (not something these commits touched or worsened) — fine to land as a follow-up.

---

PR #2000, opened 2026-08-14
Fix equipment modal cache tracking and loading state

joeseos:
## Summary
This PR fixes critical bugs in the equipment modal's cache management and loading state handling. The main issue is that the cache restoration logic couldn't distinguish between "legitimately empty results" and "never fetched", causing the modal to fail when equipment lists were empty or when loading errors occurred.

## Key Changes

- **Added `loadedBuckets` state**: Introduced a separate boolean state to track whether each equipment bucket (all, fighter, tradingpost) has been fetched, independent of its contents. This allows the code to differentiate between an empty result and a never-fetched state.

- **Fixed loading state cleanup**: Added `setIsLoadingAllEquipment(false)` in the error path to prevent the modal from getting stuck in a loading state when fighter/vehicle type information is missing.

- **Simplified cache restoration logic**: Replaced complex emptiness checks (`cachedAllCategories.length > 0 && cachedEquipment.all && Object.keys(cachedEquipment.all).length > 0`) with simple boolean checks (`loadedBuckets.all`), making the code more maintainable and correct.

- **Fixed context-based fetching**: Changed `fetchedContextsRef` check from `Object.keys(equipment).length > 0` to just checking if the context key exists, since empty results are still valid fetched results. This is especially important for campaign-filtered Trading Post results which have no cache bucket fallback.

- **Updated dependency array**: Added `loadedBuckets` properties and `campaignCustomTPKey` to the effect dependency array to ensure proper re-fetching when these values change.

- **Added null-coalescing operators**: Used `?? {}` when restoring equipment from cache to safely handle potentially undefined values.

The core insight is that an empty equipment list is a legitimate, cacheable result (e.g., a fighter with no equipment). The previous logic couldn't distinguish this from "never fetched", causing incorrect behavior. By tracking fetch status separately, the modal now correctly handles all cases including empty results and loading errors.

### DIFF
--- a/components/equipment/equipment.tsx
+++ b/components/equipment/equipment.tsx
@@ -143,6 +143,14 @@ const ItemModal: React.FC<ItemModalProps> = ({
     fighter: {},
     all: {}
   });
+  // Whether a bucket has been fetched, tracked separately from its contents: a
+  // fighter with no equipment list caches a legitimately empty result, and an
+  // emptiness test cannot tell that apart from "never fetched".
+  const [loadedBuckets, setLoadedBuckets] = useState<{ all: boolean; fighter: boolean; tradingpost: boolean }>({
+    all: false,
+    fighter: false,
+    tradingpost: false
+  });
   const [isLoadingAllEquipment, setIsLoadingAllEquipment] = useState(false);
   const [costRange, setCostRange] = useState<[number, number]>([10, 160]);
   const [availabilityRange, setAvailabilityRange] = useState<[number, number]>([6, 12]);
@@ -263,6 +271,9 @@ const ItemModal: React.FC<ItemModalProps> = ({
         ? `Vehicle type information is missing. Vehicle: ${vehicleType || 'unknown'}`
         : 'Fighter type information is missing';
       setError(errorMessage);
+      // Returns before the try below, so release the in-flight flag here or the
+      // modal stays wedged with no way to retry.
+      setIsLoadingAllEquipment(false);
       return;
     }
 
@@ -404,14 +415,17 @@ const ItemModal: React.FC<ItemModalProps> = ({
       if (equipmentListType === 'unrestricted') {
         setCachedAllCategories(uniqueCategories);
         setCachedEquipment(prev => ({ ...prev, all: equipmentByCategory }));
+        setLoadedBuckets(prev => ({ ...prev, all: true }));
       } else if (equipmentListType === 'fighters-list') {
         setCachedFighterCategories(uniqueCategories);
         setCachedEquipment(prev => ({ ...prev, fighter: equipmentByCategory }));
+        setLoadedBuckets(prev => ({ ...prev, fighter: true }));
       } else if (equipmentListType === 'fighters-tradingpost') {
         // Only cache when not using campaign filter (campaign-filtered results would overwrite unfiltered cache)
         if (campaignTradingPostIds === undefined) {
           setCachedFighterTPCategories(uniqueCategories);
           setCachedEquipment(prev => ({ ...prev, tradingpost: equipmentByCategory }));
+          setLoadedBuckets(prev => ({ ...prev, tradingpost: true }));
         }
       }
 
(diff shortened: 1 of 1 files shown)


joeseos (commented):

## Summary
The change replaces emptiness-based cache checks (`cachedX.length > 0 && Object.keys(...).length > 0`) with an explicit `loadedBuckets` flag per equipment bucket (`all`/`fighter`/`tradingpost`), and stops treating "empty fetch result" as "never fetched" in `fetchedContextsRef`. This is a real fix for the RPC loop: previously, when a fetch legitimately returned an empty category map (e.g. fighter with no equipment list), `Object.keys(equipment).length > 0` was `false`, so `fetchedContextsRef.current.has(contextKey) && Object.keys(equipment).length > 0` never short-circuited, and the effect kept re-invoking `fetchAllCategories()` on every dependency change (including `isLoadingAllEquipment` toggling true→false), causing a loop against the RPC. Marking the context as fetched unconditionally (`fetchedContextsRef.current.has(contextKey)` alone) closes that loop for both the normal buckets and the campaign-filtered Trading Post path (which has no cache bucket at all).

## Findings

**1. (Low, non-blocking, pre-existing — not introduced by this PR)** `components/equipment/equipment.tsx`, in the `useEffect` around the bottom: `fetchedContextsRef.current.add(contextKey)` runs unconditionally right before calling `fetchAllCategories()`, but `fetchAllCategories` itself early-returns via `if (!session || isLoadingAllEquipment) return;` if another fetch is already in flight. If two contexts become eligible to fetch in quick succession (e.g. rapid tab switching), the second context can get marked as "fetched" in the ref without ever actually being fetched, permanently skipping it until the ref is cleared (tab switch clears it, so impact is limited, but a same-tab prop change wouldn't). This behavior predates this diff (same ordering existed before), so it's not something this PR needs to fix, but worth a follow-up ticket since the PR is already hardening this exact code path.

**2. (Nit)** The `?? {}` fallbacks added on `cachedEquipment.all/.fighter/.tradingpost` are defensive but currently unreachable dead code — `cachedEquipment` is only ever written via `setCachedEquipment(prev => ({ ...prev, X: equipmentByCategory }))`, never to `undefined`, and the initial state already seeds `all`/`fighter` as `{}`. Harmless, and reasonable as a guard against the (missing) `tradingpost` key in the initial state object, so no change requested.

No correctness bugs, security issues, or performance regressions found in the diff itself. The dependency-array fix (adding `campaignCustomTPKey`, which was already used inside the effect body but missing from deps) is a legitimate stale-closure fix bundled into this PR. The loading-state cleanup on the early-return error path (`setIsLoadingAllEquipment(false)`) correctly prevents the modal getting stuck in a loading state, since that return bypassed the `finally` block before.

## Verdict
**Approve.** The core fix is sound and directly addresses the described RPC loop / cache correctness bug. Only a pre-existing, out-of-scope low-severity note above; no blockers.

---

joeseos:
## Review

No prior review comments exist on this PR, so this covers the single commit (6d76b4a) in full — there's nothing outstanding from an earlier pass.

### Previous findings
N/A — first review of this PR.

### Analysis

The root-cause diagnosis checks out: the old guards (`cachedXCategories.length > 0 && ... Object.keys(...).length > 0`) were emptiness tests standing in for "already fetched," so a legitimately-empty RPC result (fighter type with no `fighter_type_equipment` rows) was indistinguishable from "never fetched." That kept re-triggering `fetchAllCategories`.

Traced through the fix's edge cases:
- **`loadedBuckets` vs cache-restore/skip checks** (`equipment.tsx:507-517`, `527-529`): consistent — restoration and the fetch-skip guard now key off the same boolean, so an empty result correctly short-circuits both instead of falling through.
- **`fetchedContextsRef` check** (`equipment.tsx:534`): dropping the `Object.keys(equipment).length > 0` AND-clause is the actual loop fix — previously, a zero-result fetch left `equipment` at `{}`, so the very next effect run re-added and re-fetched. Now membership alone gates it.
- **Failed fetch → no infinite retry**: since `loadedBuckets` isn't set on the `catch` path but the `contextKey` was already added to the ref before the async call, a failed fetch surfaces the error and stays put rather than retrying every render. Confirmed this is intentional per the commit message (tab switches clear the ref).
- **Missing-fighter-type early return** (`equipment.tsx:273-277`): now releases `isLoadingAllEquipment` before returning — previously this path could strand the modal in a loading state with no retry path. Good catch.
- **Campaign-filtered Trading Post**: correctly never sets `loadedBuckets.tradingpost` (write is gated on `campaignTradingPostIds === undefined`), so it always falls through to the `fetchedContextsRef` context-key check, which is the only thing guarding it — matches the stated intent since it has no cache bucket.
- **Gang Legacy toggle** (`equipment.tsx:817-822`): calls `fetchAllCategories` directly, bypassing the ref-gated effect. On completion `isLoadingAllEquipment` flips false, re-running the effect; with `loadedBuckets.fighter` now correctly `true` even for an empty result, the effect's early-return prevents the duplicate fetch the commit message describes.
- **`?? {}` fallbacks** on cache restoration (`equipment.tsx:509,512,515`): harmless defensive addition, doesn't change behavior since the cache write always stores an object.
- **Dependency array**: adding `campaignCustomTPKey` is a genuine fix in its own right — this key was previously excluded, so changes to `campaignCustomTradingPostIds` alone (without an `equipmentListType` change) wouldn't have re-run the effect.

### New findings
None found. No new bugs, regressions, or null-handling gaps introduced by this diff — I didn't find a case where the new `loadedBuckets` state and the `fetchedContextsRef` membership check disagree in a way that reintroduces looping or drops a legitimate refetch.

### Nit (non-blocking)
`equipment.tsx:503-517` and `527-529` still use the render-time "compute key, compare to previous, setState if changed" pattern for cache restoration — this predates the PR and isn't something this change needs to fix, just flagging it's still doing setState during render, which is the kind of pattern that made the original bug easy to introduce.

### Verdict
**Approve.** The diagnosis is well-supported by the diff, the fix directly targets the emptiness-vs-fetched conflation that caused the loop, and I didn't find edge cases (error path, campaign filter, legacy toggle) where the new logic regresses.

@joeseos

---

joeseos:
**(2)** Agreed, the three `?? {}` are unreachable. Two extra data points: `tsconfig.json` sets `strict: true` but not `noUncheckedIndexedAccess`, so `cachedEquipment.all` types as `Record<string, Equipment[]>` rather than `... | undefined` — the `??` is redundant to the compiler too. And the "guards the missing `tradingpost` key" reading doesn't quite hold either, since that branch is gated on `loadedBuckets.tradingpost`, which is only set in the same batch that writes the key. Removed them, and seeded `tradingpost: {}` in the initial state so the record shape is uniform.

**(1)** The pre-existing / non-blocking call is right, but I don't think the described mechanism is reachable — the effect's first line is the same guard `fetchAllCategories` re-checks:

```
522:  if (!session || isLoadingAllEquipment) return;   // effect
536:  fetchedContextsRef.current.add(contextKey);
537:  fetchAllCategories();
227:    if (!session || isLoadingAllEquipment) return;  // fetchAllCategories
```

Both read the same render's closure and 537 calls 227 synchronously, so if the effect passed 522, 227 cannot early-return. Rapid tab switching resolves the same way: with a fetch in flight the flag is `true`, so the effect returns at 522 without marking the context, then re-runs and fetches the new tab once the flag clears.

There is a narrower real version of the concern — `fetchAllCategories`'s *other* early return (missing fighter/vehicle type) does return after the ref was marked. Low impact, and currently intended: the error renders instead of retrying.

Worth noting this PR hardens that line rather than loosening it. Under StrictMode's dev double-invoke the old `&& Object.keys(equipment).length > 0` let the second invocation refetch; the ref now catches it.

PR #2001, opened 2026-08-14
Add N26 Post-cycle Actions to the gang Campaign tab

joeseos:
N26 introduces a Post-cycle Sequence between battles, where each model may perform one Post-cycle Action. The app had no representation of it, so players applied every consequence by hand across the Gang, Stash and Vehicles tabs.

Adds a Post-Cycle Actions panel to the gang page's Campaign tab. The player assigns an action per fighter and confirms once; the server applies every consequence and logs what each fighter did. Shown only for N26 gangs that are in a campaign.

Supported actions:
  - Medical Escort (Leader/Champion) - 30cr plus optional 50cr steps that each add +1 to a server-rolled D6. 1 kills the target, 2-3 stabilises them with a 51-56 Lasting Injury, 4+ is a clean Recovery. Declining to pay kills them with no roll.
  - Fit Bionics (Leader/Champion) - 50cr per Lasting Injury removed; a Critical Injury cannot be removed this way.
  - Visit Chop Shop (vehicles) - flat per-damage repair, delegated to the existing N26 path in repairVehicleDamage.
  - Work Territory (Leader/Champion/Ganger/Prospect, max five) - 15cr each.
  - Train (any model) - 2 XP.
  - Develop Tactics and Visit Trading Post - logged only; there is no gang tactics model, and equipment is bought from the Stash tab.

Implementation notes:
  - Gated by a new postCycleActions row in EDITION_CAPABILITIES rather than by comparing edition slugs.
  - utils/postCycleActions.ts owns the catalog, eligibility and the cross-fighter caps, so the form and the server agree on one set of rules. The server re-runs validatePostCycleAssignments against its own reads, because the assignment list arrives from the browser.
  - The server action composes the existing per-fighter helpers so rating, wealth and cache invalidation stay in one place. It is deliberately not atomic, matching clearRigGlitchesDowntime and repairVehicleDamage, and returns per-action results so a partial application is reported precisely.
  - Credits move once via updateGangFinancials rather than per fighter.
  - Outcomes carry a structured change list so the gang page can patch its own fighter state; router.refresh() would not reach it, since the page copies its server payload into useState.


joeseos:
## Code Review

@joeseos Reviewed the N26 Post-cycle Actions feature (`app/actions/post-cycle-actions.ts`, `app/actions/logs/gang-post-cycle-logs.ts`, `utils/postCycleActions.ts`, `utils/dice.ts`, plus the `types/edition.ts` and `gang-page-content.tsx` wiring). Traced the full request lifecycle: client assignment → server re-validation → per-action apply loop → credit settlement → logging. No CLAUDE.md in the repo to check conventions against.

### Findings

**1. `utils/postCycleActions.ts:405` (`validatePostCycleAssignments`) + `app/actions/post-cycle-actions.ts:279-313` — two performers can target the same fighter with Medical Escort, causing a resurrection (High)**

`medicalEscortTargets` is a `Set`, so it only detects overlap between Medical Escort and Fit Bionics targets — it never rejects two *different* Medical Escort assignments aimed at the same target. In the apply loop, `target` is looked up once via `byId.get(assignment.targetFighterId)!` from a snapshot read at the top of `applyPostCycleActions`, and that object is never updated as earlier assignments in the same loop mutate the DB.

Concrete case: Leader A and Champion B both assign Medical Escort to the same Critically Injured fighter X. Validation passes for both (Set dedup). A's assignment resolves first — say `Complications`, which kills X via `editFighterStatus('kill')` (a toggle). B's assignment then runs `killTarget()`, which reads the stale `target!.killed` (still `false` from the pre-loop snapshot), calls `editFighterStatus('kill')` again, and **toggles X back to alive** — resurrecting a fighter this same request just killed — before applying B's own (possibly conflicting) roll outcome on top.

Fix: track counts, not just membership, in `validatePostCycleAssignments` (reject a target hit by more than one Medical Escort *or* more than one Fit Bionics assignment), and/or re-read `target.killed` from the DB (or update the in-memory map) before each `killTarget()` call in the apply loop.

**2. `app/actions/post-cycle-actions.ts:254` & `:565` — the gang is charged the full cost even when an action fails or partially completes (High)**

`credits.aggregate` is computed once from the raw, unvalidated-against-outcomes `assignments` list before the apply loop runs, then deducted unconditionally via `updateGangFinancials(..., creditsDelta: credits.aggregate)` after the loop — regardless of each result's `failed` flag.

Concrete case A: Fit Bionics removes 3 Lasting Injuries at 50cr each (150cr charged via `assignmentCreditsDelta`, based on `assignment.injuryIds.length`). The first `deleteFighterInjury` succeeds, the second fails — the loop breaks with `failed: true` and only 1 of 3 actually removed. The gang is still charged the full 150cr.

Concrete case B: a Medical Escort assignment fails outright (e.g. "Could not resolve a Medical Escort roll", or the Stabilised injury type lookup fails). `results` marks it `failed: true`, but its cost was already baked into `credits.aggregate` before the loop ran, so the gang pays 30cr+ for nothing.

Fix: compute the aggregate credits delta from what actually succeeded (e.g. sum `assignmentCreditsDelta` only over non-failed results, or recompute Fit Bionics' cost from `removedIds.length` rather than `injuryIds.length`), not from the pre-apply assignment list.

**3. `app/actions/post-cycle-actions.ts:375-379` — the Full Recovery DB write's error is silently discarded (Medium)**

```ts
await supabase
  .from('fighters')
  .update({ recovery: true, updated_at: new Date().toISOString() })
  .eq('id', target!.id);
```

The result isn't even destructured, so a failed write (RLS hiccup, transient error, concurrent update) is invisible — the code unconditionally pushes a success outcome (`failed` left falsy) and reports `changes: [{ recovery: true }]`. The client patches its local state to show the fighter went into Recovery while the DB row was never updated, so the fighter silently stays flagged as available to act/be targeted next cycle despite the UI saying otherwise.

Fix: capture `{ error }` from the update and push a `failed: true` outcome (mirroring how every other branch in this switch already handles its write failures) when it's set.

**4. `utils/postCycleActions.ts:317` — `assignmentTargetId` is dead code (Low)**

Exported but has no callers anywhere in the codebase. Either wire it in where `assignment.action === 'medical_escort' || assignment.action === 'fit_bionics'` is currently duplicated inline (e.g. `app/actions/post-cycle-actions.ts:235`, `:279-282`), or drop it.

### What looks solid
- `resolveMedicalEscort` / `medicalEscortStabilisedRoll` in `utils/dice.ts` correctly handle the open-ended "Good Stuff" band and the fixed-first-die D66 stabilised roll.
- The affordability pre-check (`totalCost = -credits.total`) correctly includes both the aggregate and Chop-Shop-delegated costs before any writes happen.
- Cache invalidation and the re-read-after-write for `rating`/`wealth`/`credits` (since the per-fighter helpers mutate them independently) are handled correctly.
- The non-atomicity is explicitly documented and matches the existing `clearRigGlitchesDowntime`/`repairVehicleDamage` precedent — a reasonable, consistent trade-off given the per-fighter-helper architecture.

Findings 1 and 2 are data-integrity bugs (a resurrection path and a real-money overcharge on failure) that should be fixed before merge; finding 3 should be fixed alongside them since it's the same class of issue. Finding 4 is optional cleanup.

---

joeseos:
## Follow-up Code Review

@joeseos Reviewed only what's changed since my last review (commit `376c0f8`/`91b0e2d`): `0099a27` (log-type registration), `622e0ce` (dead-code removal), `61685b2` (billing/duplicate-target fixes), and `408d2b7` (wiring Develop Tactics to real Gang Tactics cards, after the `1233e19` merge from `main`).

### Previous findings

1. **Two Medical Escorts on one patient → resurrection (High) — Resolved.** `medicalEscortTargets`/`fitBionicsTargets` are now `Map<string, number>` and a target hit more than once is rejected (`utils/postCycleActions.ts` ~L505-520). As defense in depth, a successful kill now also updates the in-memory snapshot (`if (killed.success) target!.killed = true;`), so even a future validation gap couldn't resurrect a fighter. Nice bonus catch alongside this: the Medical Escort reorder so the Stabilised injury type is resolved *before* the Critical Injury is deleted, closing a free-heal path on N26 "Eye Injury" that wasn't part of my original findings but is the same class of bug.
2. **Full charge on a failed/partial action (High) — Resolved.** Every branch now reports its own `creditsDelta` derived from what actually happened (e.g. `-(removedIds.length * FIT_BIONICS_COST_PER_INJURY)`, not the requested count), and settlement sums `results[].creditsDelta` instead of the pre-loop planned total. Chop Shop is correctly excluded from the sum since `repairVehicleDamage` settles its own cost.
3. **Full Recovery write's error silently discarded (Medium) — Resolved.** Now destructures `{ error: recoveryError }` and pushes a `failed: true` outcome with the DB error surfaced, matching every other branch.
4. **`assignmentTargetId` dead code (Low) — Resolved.** Removed, along with `PostCycleActionTarget`, `.target`, and `.maxPerSequence`, none of which had a reader.

All four are solidly closed — this is a thorough fix, not a minimal patch.

### New findings

**1. Develop Tactics assignments are unconditionally rejected by validation — the new feature cannot be submitted at all (High), `utils/postCycleActions.ts:443`**

```ts
if (!canPerformPostCycleAction(performer, assignment.action)) {
```

`canPerformPostCycleAction`'s signature is `(fighter, actionId, availability: PostCycleAvailability = {})`, and it now gates Develop Tactics on `availability.tacticsCardsAvailable` (L219: `if (actionId === 'develop_tactics' && !availability.tacticsCardsAvailable) return false;`). This call site never passes a third argument, so `availability` is always `{}` and `tacticsCardsAvailable` is always `undefined` — **every** Develop Tactics assignment fails this check, regardless of whether the gang's edition actually has a tactics catalogue.

`validatePostCycleAssignments` is called with the same two-arg-plus-context shape from both sides, and neither threads availability through:
- Client: `components/gang/post-cycle-actions.tsx:432-434` passes `{ ownedTacticsCardIds }` only — it has `availability.tacticsCardsAvailable` in scope two lines above (used for `eligiblePostCycleActions`) but doesn't forward it.
- Server: `app/actions/post-cycle-actions.ts` (re-validation call) passes `{ ownedTacticsCardIds }` only, and never computes `tacticsCardsAvailable` at all — though it already resolves `editionSlug` at L236 for the `EDITION_N26` check, so `hasGangTacticsCards(editionSlug)` is one line away.

The visible effect: a player opens the Develop Tactics picker, selects cards, but the "Resolve N Post-Cycle Actions" button stays disabled (`post-cycle-actions.tsx:601`, `issues.length > 0`) with the message "`{fighter}` cannot perform Develop Tactics" — for every gang, on every edition, including N26 where tactics cards do exist. If that check were ever bypassed client-side, the server's identical re-validation would reject it the same way. This is the entire point of the last commit, so it's worth catching before merge — there's no test suite here to have caught it, and it's a pure logic gap `tsc`/`next build` can't see.

Fix: add `tacticsCardsAvailable?: boolean` to `PostCycleValidationContext`, pass it through to the `canPerformPostCycleAction` call at L443, and supply it from both call sites (client: reuse the existing `availability` object; server: `hasGangTacticsCards(editionSlug)`).

### What looks solid
- `TacticsCardPickerModal` extraction is clean — `unavailableCardIds = owned ∪ reserved` correctly keeps the D66 roller off cards another row in the same batch already claimed, and the checkbox list disables/labels both cases distinctly.
- Cross-fighter duplicate-card validation (`claimedTacticsCards`) correctly catches two performers picking the same unowned card in one sequence, which matters because the DB insert ignores duplicates and would otherwise silently drop the second fighter's pick.
- `addGangTacticsCards` (reused, not changed here) validates catalogue membership server-side and upserts with `ignoreDuplicates`, so it doesn't trust the picker's ids.
- Outcome plumbing (`addedTacticsCards` on the result, merged into page state via `onTacticsCardsUpdate`) is consistent with how the rest of this PR reports partial/authoritative state back to the client.

Finding 1 is a straightforward wiring bug but it's blocking — as written, Develop Tactics cannot be resolved from the UI at all, on any edition. Everything else in this round of changes is a clean, verified fix of the prior findings.

---

joeseos:
## Second Follow-up Code Review

@joeseos One new commit since my last pass: `49f262a` — "Fix Develop Tactics being rejected on every gang."

### Previous finding

**Develop Tactics unconditionally rejected by validation (High) — Resolved, and hardened beyond the minimal fix.**

Rather than just adding the missing third argument at the one call site, this removes the shape that let the bug happen at all:
- `PostCycleAvailability.tacticsCardsAvailable` and `PostCycleValidationContext.ownedTacticsCardIds` are now required, not optional, and both functions dropped their `= {}` / `= new Set()` defaults.
- `PostCycleValidationContext` now `extends PostCycleAvailability`, so the one context object callers already build carries the availability flag into `canPerformPostCycleAction(performer, assignment.action, context)` (`utils/postCycleActions.ts:452`) automatically — no separate plumbing to forget next time.
- Both real call sites now supply it: the client spreads its existing `availability` memo (`components/gang/post-cycle-actions.tsx:230`), and the server derives it from the `editionSlug` it already resolves for the N26 check (`hasGangTacticsCards(editionSlug)` in `app/actions/post-cycle-actions.ts:274`).

I checked every remaining call site (`eligiblePostCycleActions`, `canPerformPostCycleAction`, `validatePostCycleAssignments`) and all of them supply the now-required fields — nothing left defaulting silently. `tsc --noEmit` on this file/component shows only pre-existing environment noise (missing `node_modules`), no argument-count or type errors from this change. Diff is exactly the 3 files the commit claims, nothing incidental.

Making the previously-optional fields required so a future omission is a compile error rather than a silent runtime "action always denied" is the right fix here — better than the narrower one I suggested last round.

### New findings
None. This is a small, self-contained fix with no new surface.

### Verdict: **Approve**
Both rounds of review findings are now fully addressed. Nothing outstanding from my side — worth the manual "click through an N26 gang" check the original PR description flagged, since none of this has automated test coverage, but I found no further code-level issues.

---

PR #2007, opened 2026-08-15
Don't let Admin save a grants option with no equipment picked

joeseos:
"Add Option" in the equipment editor seeds { equipment_id: '', additional_cost: 0 }, and the save-time normaliser only checked options.length — so saving before picking an item stored an empty string where a uuid belongs.

get_equipment_detailed_data casts that value to uuid while attaching each option's equipment_name. One blank id made the whole call fail with 22P02, which surfaced as "Failed to load equipment categories" across the entire list. It reached only the gang stash on an N26 gang: the item involved is core_equipment, which the fighter's-list and Trading Post filters exclude and the stash's Unrestricted list (no fighter_type_id) does not, and Unrestricted also filters by the gang's edition.

Drop options with no equipment_id before sending, and treat a grants block left with none as no grants at all — the same rule the normaliser already applied to an empty options array.

### DIFF
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -520,9 +520,14 @@ export function AdminEditEquipmentModal({ onClose, onSubmit }: AdminEditEquipmen
 
       const hasEditedFighterTypes = selectedFighterTypes.length !== fighterTypes.filter(ft => selectedFighterTypes.includes(ft.id)).length;
 
-      // Validate and normalize grants_equipment - treat empty options as no grants
-      const normalizedGrantsEquipment = grantsEquipment?.options?.length
-        ? grantsEquipment
+      // Validate and normalize grants_equipment - treat empty options as no grants.
+      // An option with no equipment picked is dropped: a blank equipment_id can never
+      // be granted, and it is not a uuid, so anything casting it downstream breaks.
+      const grantsEquipmentOptions = (grantsEquipment?.options || []).filter(
+        option => Boolean(option.equipment_id)
+      );
+      const normalizedGrantsEquipment = grantsEquipment && grantsEquipmentOptions.length
+        ? { ...grantsEquipment, options: grantsEquipmentOptions }
         : null;
 
       const requestBody = {


joeseos (commented):

## Summary

The fix itself is correct and well-targeted: `grantsEquipmentOptions` filters out any option whose `equipment_id` is `''` before it's ever sent, and `normalizedGrantsEquipment` correctly collapses to `null` when nothing survives the filter — matching the existing "empty options array → no grants" rule. I traced the UI: `equipment_id` on an option can only ever be `''` (the placeholder `<option value="">`) or a real uuid from the `<select>` (admin-edit-equipment.tsx:959-967), so `Boolean(option.equipment_id)` is a safe, sufficient filter with no whitespace/edge-case gap. I also checked `admin-create-equipment.tsx`, which has no `grants_equipment` UI at all, so there's no sibling copy of this bug to fix elsewhere.

## Findings

**1. [Medium] No server-side guard against the same bad state (defense-in-depth gap)**
- File: `app/api/admin/equipment/route.ts:606-649`
- The PATCH handler destructures `grants_equipment` straight from the request body and writes it with only `grants_equipment: grants_equipment || null` — no validation that every `options[].equipment_id` is a non-empty uuid.
- `supabase/functions/get_equipment_detailed_data.sql:365-366` still does the unguarded cast: `LEFT JOIN equipment eq ON eq.id = (opt->>'equipment_id')::uuid`. That's the exact line that threw `22P02` and broke "Failed to load equipment categories" for the whole equipment list, per this PR's own description.
- Since the fix lives entirely in `admin-edit-equipment.tsx`, it only closes the path through this one modal. Any other caller of `PATCH /api/admin/equipment` (a future UI, a direct API call, a retried/replayed request) can still persist a blank `equipment_id` and reproduce the original outage.
- Suggested fix: apply the same filter (or reject the request) server-side in the route handler before the `update()` call — that's the one place that guarantees the invariant regardless of which client calls it. Given how severe the blast radius was here (crashed equipment loading for an entire list from one bad row), I'd treat this as worth closing now rather than after a recurrence.

Not a blocker for merging this specific fix, but I'd like to see the server-side validation land as a fast follow (or in this PR, if you'd rather not split it).

## Verdict

**Approve** — the change fixes the described bug correctly and safely. The one finding above is about closing the root cause more durably, not about a defect in what's here.

---

joeseos:
@joeseos Reviewed the current state of this PR (single commit, no prior review threads to reconcile).

**Previous findings:** None — this is the initial review, no earlier comments on this PR.

**Change assessed** (`components/admin/admin-edit-equipment.tsx`, `handleSubmit`):
```js
const grantsEquipmentOptions = (grantsEquipment?.options || []).filter(
  option => Boolean(option.equipment_id)
);
const normalizedGrantsEquipment = grantsEquipment && grantsEquipmentOptions.length
  ? { ...grantsEquipment, options: grantsEquipmentOptions }
  : null;
```
This correctly fixes the described bug: options with a blank `equipment_id` are dropped before submit, and a grants block left with zero valid options normalizes to `null`, matching the existing rule for an empty `options` array. `EquipmentGrants.options[].equipment_id` is typed `string`, and `Boolean('')` is `false`, so this closes the reported path cleanly. No regressions in the surrounding logic — the `grantsEquipment && …` guard is equivalent to the old `grantsEquipment?.options?.length` check when `grantsEquipment` is null.

**New finding (medium, defense-in-depth):**
The fix is applied only at this one client call site. `app/api/admin/equipment/route.ts` (~line 649) still does `grants_equipment: grants_equipment || null` with no filtering of `options`, and the actual crash lives one layer down in `supabase/functions/get_equipment_detailed_data.sql:366`:
```sql
LEFT JOIN equipment eq ON eq.id = (opt->>'equipment_id')::uuid
```
which unconditionally casts every option's `equipment_id` to `uuid`. So the 22P02 failure this PR fixes could still be triggered by any other caller of that API route (script, future admin UI, direct API call) that submits a grants option with a blank/non-uuid `equipment_id` — the guard rail exists only in this one form's submit handler, not in the API or the SQL function that actually breaks. Worth adding the same filter (or a uuid-format check) server-side, or a `WHERE opt->>'equipment_id' ~ '^[0-9a-f-]{36}$'`-style guard in the SQL function, so this class of bug can't resurface through a different entry point.

I checked the other admin components that call this same API route (`admin-create-equipment.tsx`, `admin-edit-fighter-type.tsx`, etc.) — none of them currently build `grants_equipment` themselves, so there's no second live instance of the bug today. This is about hardening the actual failure point, not an active regression.

**Verdict: Approve**, with the suggestion above as a good follow-up (either in this PR or a fast-follow) rather than a blocker — the immediate crash reported is fixed.

---

PR #2008, opened 2026-08-15
Show XP progress toward the next rank on N26 Advancements

joeseos:
The Advancements header counted Advancements taken, which is the right
summary for N23 where they are bought with XP. N26 earns them by rank, so
what matters there is how close the fighter is to the next tier: a fighter
on 37 XP reads 37/49, on 73 XP reads 73/85.

nextTierStartFor inherits the per-edition gating already in tierStartsFor,
and the header branches on the isCumulativeXp local the component already
computes, so N23 keeps its count unchanged. A Legend of the Underhive has
no tier above it and reads 235/–.

### DIFF
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -9,6 +9,7 @@ import { TypeSpecificData } from '@/types/fighter-effect';
 import { createClient } from '@/utils/supabase/client';
 import { getSkillSetRank } from "@/utils/skillSetRank";
 import { characteristicRank } from "@/utils/characteristicRank";
+import { nextTierStartFor } from "@/utils/advancementRanks";
 import { List } from "@/components/ui/list";
 import { UserPermissions } from '@/types/user-permissions';
 import { useMutation, useQuery } from '@tanstack/react-query';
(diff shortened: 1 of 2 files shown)


joeseos:

## Summary
Small, well-scoped change. `nextTierStartFor` reuses the existing `tierStartsFor` helper (no logic duplication with `advancementsEarnedFor`), and the header branch correctly reuses the `isCumulativeXp` local already computed in the component instead of re-deriving edition capability.

## Verification
- `isCumulativeXp` (`hasCumulativeXp`) is `true` only for `n26` (`types/edition.ts:130`), and `TIER_STARTS_BY_EDITION.n26` is the only non-empty tier list, so the `isCumulativeXp` branch and `nextTierStartFor` returning `null` only for N26 above the top tier are consistent — the `'–'` fallback is reachable only for "no tier above," never for a misconfigured/unknown edition slug reaching this render path.
- `fighterXp` is passed from `fighterData.fighter?.xp` (`components/fighter/fighter-page.tsx:824,894`) and is the same value already passed as `currentXp` to the advancement-earning calculation a few lines below (`fighter-advancement-list.tsx:3480`), so the header's XP reads off the same source of truth used for the actual advancement math — no drift risk.
- Tier boundary semantics check out: `.find(tierStart => tierStart > currentXp)` picks the *next* threshold strictly above current XP, matching the PR description's examples (37 → 49, 73 → 85) and the "already on this tier" case at an exact boundary.
- `fighterXp: number` is non-nullable in `AdvancementsListProps`, so no null/undefined handling is needed there.
- Markup mirrors the existing `advancementCount` branch's responsive spans (`sm:hidden` / `hidden sm:inline`) exactly, so no layout inconsistency.

## Findings
None that block. No high/medium issues found — didn't find edge cases, security, or performance concerns in this diff (it's a pure read-only derivation added to `utils/advancementRanks.ts` plus a header render branch).

Two very minor, non-blocking notes:
- **(nit, low)** `utils/advancementRanks.ts` has no test file today (`countAdvancementsTaken`, `advancementsEarnedFor`, etc. are all untested), so this isn't a regression — just flagging that `nextTierStartFor`'s boundary behavior (exact-tier-value input, above-top-tier input, empty-tiers input) would be a good candidate if/when tests are added for this file.
- **(nit, low)** The "Add Advancement" modal header (`fighter-advancement-list.tsx:2298-2299`) shows raw `{currentXp}` without tier context — out of scope for this PR, just noting for consistency if a follow-up wants to unify the two XP displays.

## Verdict: Approve

---

PR #2009, opened 2026-08-15
Stop N26 refusing Advancements, and stop its log claiming XP was spent

joeseos:
Adding an Advancement failed with "No advancements available" once a
fighter had taken as many as its XP had earned. Playing groups house-rule
when a model may advance, and the app records what happened at the table
rather than refereeing it, so the rank is no longer enforced. This was the
only gate: the DB function already leaves the question to its caller, and
every client guard is an XP-affordability check that a zero cost passes.

advancementBlockedReason keeps just the N23 affordability branch, which
makes it synchronous and retires countAdvancementsTakenFor along with the
two count queries it ran on every add. N23 is unchanged — its branch
returned before the removed code was reachable.

The three advancement log lines take spends_xp and drop their cost and
remaining/refunded-XP wording when it is false, so a rank-based edition no
longer reads "for 0 XP" or "(refunded 0 XP)". The flag defaults to true, so
wording outside this flow is untouched.

### DIFF
--- a/app/actions/fighter-advancement.ts
+++ b/app/actions/fighter-advancement.ts
@@ -12,7 +12,6 @@ import {
   hasGangerChampionKeepTypePromotion,
   hasProspectSpecialisationPromotion,
 } from '@/types/edition';
-import { openAdvancementsFor } from '@/utils/advancementRanks';
 import {
   N26_CHAMPION_PROMOTION_SKILL_ID,
   N26_CHAMPION_PROMOTION_SKILL_NAME,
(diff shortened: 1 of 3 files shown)


joeseos:
@joeseos Review of the current diff (this PR's first commit — no prior review comments to reconcile).

**Previous findings:** none — this is the first review pass on this PR.

**Summary of the change:** Removes the N26 "no advancements available" rank gate from `advancementBlockedReason` (now synchronous, N23-affordability-only), deletes the now-dead `countAdvancementsTakenFor` helper and its two count queries, and threads a `spends_xp` flag through the three advancement gang-log functions so N26 log lines drop XP-cost/refund wording instead of reading "for 0 XP" / "(refunded 0 XP)".

**Correctness check:**
- Both `advancementBlockedReason` call sites correctly drop `await` and the removed `supabase`/`isAdvance` args — no dangling references, no orphaned `await` on a non-promise.
- `spends_xp` is threaded consistently at all three log call sites (`addCharacteristicAdvancement`, `addSkillAdvancementInternal`, `deleteAdvancement`), always derived from the same `!hasCumulativeXp(editionSlugOf(fighter))` computation used for the XP gate itself, so the log wording and the actual XP mutation can't disagree.
- `spends_xp ?? true` defaults preserve old wording for any caller that omits the flag — verified there are no other callers of the three log functions besides the three updated in this diff, so this is just a safety net, not load-bearing.
- Confirmed the N23 branch is untouched: `advancementBlockedReason` still returns `'Insufficient XP'` under the same condition it always did.
- Checked the frontend (`fighter-advancement-list.tsx`) for a client-side mirror of the removed rank gate that might now disable the Add button while the server silently allows it — found none; the header only shows an XP/tier progress readout (`nextTierStartFor`), not a blocking check.

**New finding:**

| File | Issue | Severity |
|---|---|---|
| `utils/advancementRanks.ts:92,106` | `countAdvancementsTaken` and `openAdvancementsFor` are now dead code. This PR removes their only caller (`countAdvancementsTakenFor` in `fighter-advancement.ts`, which duplicated rather than called them, but served the same purpose). A repo-wide grep turns up zero remaining references to either export. Since they're the two functions that implemented the exact rank-gate logic this PR intentionally retires, worth either deleting them in this PR or leaving a one-line note if they're kept for a near-term reuse (e.g. a future informational "advancements remaining" display) — otherwise this is orphaned exported code from day one. | low |

Nothing else stood out — no null-handling gaps in the new code (`fighter.xp`/`xpCost` are always numbers by the time they reach `advancementBlockedReason`), and the log-string edits read cleanly for both the `spends_xp` true/false cases including the punctuation/trailing-space boundaries.

**Verdict: approve** (the dead-code note above is optional cleanup, not a blocker).

---

joeseos:
@joeseos Reviewed this PR (removing the N26 advancement-count gate, and quieting the XP-spent log wording for zero-cost advancements).

## Findings

**1. (High) `app/actions/logs/gang-fighter-logs.ts:155` — the log still fabricates an XP deduction for two N26 promotion flows**

The PR's fix correctly gates the first description line on `spends_xp`, but a second "summary" line a few statements later is only gated on `xp_cost > 0`, not on `spends_xp`. That block computes `xpBefore = fighter.xp + xp_cost` and appends `XP: {xpBefore} → {fighter.xp}` whenever `xp_cost` is nonzero — regardless of whether XP was actually spent.

Two real N26 call paths hit this:
- Champion promotion — `xp_cost` is hardcoded to `12` in `components/fighter/fighter-advancement-list.tsx:418`.
- Ganger → Specialist promotion — defaults to `xp_cost: 6` in the same file.

Both route through `applyPromotionWithSkillAdvancement` → `addSkillAdvancement` → `logSkillAdvancement` without zeroing `xp_cost` for N26 (unlike the plain-advancement and Prospect-promotion paths, which do pass `xp_cost: 0`). So for an N26 fighter, e.g. at 50 XP taking the Champion promotion: `spendsXp` is false, `fighter.xp` correctly stays 50, and the first line omits "for 12 XP" — but this second block still computes `xpBefore = 62` and logs `XP: 62 → 50`, falsely implying 12 XP was spent when nothing changed. This directly contradicts the PR's stated goal.

Suggested fix: gate this block on `spends_xp` as well (or zero `xp_cost` before it reaches `logSkillAdvancement` for these two N26 flows, matching the pattern already used for plain advancements/Prospect promotion).

**2. (Low) `utils/advancementRanks.ts:106` — `openAdvancementsFor` is now dead code**

This diff deletes `countAdvancementsTakenFor`, which was `openAdvancementsFor`'s only caller in the codebase. The function itself (and its docstring, which still describes it as "the cap the add-advancement actions enforce") is left behind. Since the cap is intentionally no longer enforced per this PR's description, this function should probably be removed too — otherwise a future reader trusts the stale docstring and assumes some code path still gates on it.

## Verdict: Changes requested

Blocker: finding #1 — the PR's own description says the log should stop claiming XP was spent for zero-cost advancements, but the Champion/Ganger→Specialist N26 promotion paths still show a false XP deduction via the second log line. Finding #2 is a minor cleanup, non-blocking.


---

PR #2010, opened 2026-08-15
Stop caching permission denials, which locked owners out of their own gangs

joeseos:
checkPermissionCached wrapped the check_permission RPC in unstable_cache with
revalidate: false, and both failure branches returned a deny result instead of
throwing:

  if (error) {
    console.error('Error in checkPermissionCached RPC:', error);
    return { isOwner: false, canEdit: false, ... };
  }

unstable_cache cannot tell a failure from an answer, so that deny was written to
the shared Data Cache and, with no TTL, stayed there. One transient RPC error -
an expired JWT, a connection blip - permanently stripped the owner of canEdit on
that gang. getGangBasic gets this right and throws.

Nothing cleared it. CHECK_PERMISSION(userId, gangId) is only revalidated by
invalidatePermissionForUser, called from campaign membership changes and
delete-gang - never from an owner simply loading their gang.

Until ee595ad, signInAction and signOutAction called revalidatePath('/', 'layout'),
which evicted the entire Data Cache and took poisoned permission entries with it.
That flush was wasteful and removing it was right, but it was also the only thing
making this bug self-healing, which is why re-logging-in used to fix it and why
the bug went unnoticed.

Where the failures come from: proxy.ts returns early on any Next-Action header
and its matcher excludes api/, so server actions and API routes get no session
refresh. Two of the four checkPermissionCached call sites are on those paths
(fighter-skill-access.ts, api/fighters/skill-access). An expired token there
reaches PostgREST, the RPC 401s, and the deny is cached under a key of
(userId, gangId, gangOwnerId) - which the gang page then reads back even though
its own token is fresh. Page renders, edit controls gone.

Four changes:

- Ownership is applied outside the cache. gangOwnerId === userId is a plain id
  comparison needing no RPC, so an owner keeps canEdit even if the cached entry,
  the RPC, or Supabase is broken. isAdmin still comes from the RPC, since an owner
  who is also an admin sees extra fields in the gang edit modal.
- The cached callback throws on RPC error. The caller catches outside the cache
  and degrades for that request only via the uncached checkPermission, so nothing
  bad is ever persisted.
- Permission entries carry a new USER_PERMISSIONS(userId) tag alongside the
  per-gang one, giving a handle on all of a user's entries at once.
- signInAction calls invalidateUserPermissions(data.user.id), restoring the
  "log out and back in fixes it" escape hatch ee595ad removed - scoped to one
  user's tag rather than the whole shared cache.

revalidate: false is kept. It matches the convention used by the other 60
unstable_cache sites and adds nothing to page-load cost; recoverability comes
from the sign-in invalidation instead. The one uncovered case is a
profiles.user_role change, which no server action writes, so a newly promoted
admin sees it after signing in again.

Also fixes the same pattern in three other cached callbacks, where a transient
error would have cached an empty gang list, an empty campaign list, or a roster
silently missing a fighter.

No manual cache purge is needed for the users stuck today: unstable_cache derives
its key from cb.toString() (next/dist/server/web/spec-extension/unstable-cache.js:55),
so changing these callbacks orphans every existing check-permission-* entry.

Claude-Session: https://claude.ai/code/session_015Q2vJsvkkW8of6CfZmuivN

### DIFF
--- a/app/actions/auth.ts
+++ b/app/actions/auth.ts
@@ -3,7 +3,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { headers } from "next/headers";
 import { cookies } from 'next/headers';
-import { invalidateUserCount } from '@/utils/cache-tags';
+import { invalidateUserCount, invalidateUserPermissions } from '@/utils/cache-tags';
 import { safePostSignInPath } from '@/utils/auth';
 
 // Returned instead of calling redirect(), so the caller can do a full document
@@ -123,7 +123,7 @@ export const signInAction = async (formData: FormData): Promise<AuthActionResult
 
   const supabase = await createClient();
 
-  const { error } = await supabase.auth.signInWithPassword({
+  const { data, error } = await supabase.auth.signInWithPassword({
     email,
     password,
   });
@@ -132,7 +132,11 @@ export const signInAction = async (formData: FormData): Promise<AuthActionResult
     return { error: error.message };
   }
 
-  // No revalidatePath - see the note in signOutAction.
+  // Clear this user's permission entries - they are cached with revalidate: false.
+  if (data.user) {
+    invalidateUserPermissions(data.user.id);
+  }
+
   return { redirectTo: safePostSignInPath(nextParam) };
 };
 
--- a/app/auth/callback/route.ts
+++ b/app/auth/callback/route.ts
@@ -1,6 +1,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { NextResponse } from "next/server";
 import { safePostSignInPath } from "@/utils/auth";
+import { invalidateUserPermissions } from "@/utils/cache-tags";
 
 export async function GET(request: Request) {
   const requestUrl = new URL(request.url);
@@ -26,7 +27,13 @@ export async function GET(request: Request) {
 
   if (code) {
     const supabase = await createClient();
-    await supabase.auth.exchangeCodeForSession(code);
+    const { data, error } = await supabase.auth.exchangeCodeForSession(code);
+
+    // Same reason as signInAction: any path that establishes a session clears
+    // that user's permission entries, which are cached with revalidate: false.
+    if (!error && data.user) {
+      invalidateUserPermissions(data.user.id);
+    }
   }
 
   const destination = safePostSignInPath(requestUrl.searchParams.get("next"));
--- a/app/lib/get-user-campaigns.ts
+++ b/app/lib/get-user-campaigns.ts
@@ -137,7 +137,8 @@ export const getUserCampaigns = async (userId: string, supabase: any): Promise<C
           // captureException(error)
         }
 
-        return [];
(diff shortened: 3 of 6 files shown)


joeseos:
@joeseos Review of the current diff (single commit, 835b0bc). No prior review comments on this PR, so there's nothing to reconcile against.

**Previous findings:** N/A — this is the first review.

**What's solid**
- `checkPermissionCached`: throwing on RPC error instead of returning a deny result is the right fix for the `revalidate: false` cache-poisoning bug, and the try/catch fallback to uncached `checkPermission` means a broken RPC/cache degrades for one request instead of crashing the page.
- Applying `gangOwnerId === userId` ownership outside the cache is a good belt-and-suspenders move — an owner keeps `canEdit` even if the fallback path also fails.
- `USER_PERMISSIONS(userId)` tag + `invalidateUserPermissions` on sign-in correctly restores the "log out/in fixes it" escape hatch, scoped per-user instead of nuking the whole Data Cache like the old `revalidatePath('/', 'layout')` did.
- Root-cause writeup in the PR description is accurate against the code (verified `proxy.ts`'s Next-Action early-return and the `unstable_cache` key-derives-from-`cb.toString()` claim).

**New finding**

*Medium — `get-user-campaigns.ts`, `get-user-gangs.ts`, `gang-data.ts:getGangFightersList` rethrow with no fallback or error boundary, unlike `checkPermissionCached`.*

These three call sites got the same "must throw, not return `[]`" treatment to stop `unstable_cache` from persisting bad data forever — correct as far as it goes. But `checkPermissionCached` is the only one of the four that also added a catch-and-degrade path. The other three just propagate:

- `getUserGangs`/`getUserCampaigns` are awaited directly in `Promise.all(...)` in `app/page.tsx:45-47` with no try/catch around them.
- `getGangFightersList`'s per-fighter `catch` (`app/lib/shared/gang-data.ts:2052-2056`) used to log-and-skip one bad fighter and keep rendering the rest of the roster; now it rethrows and aborts the whole roster fetch. The caller in `app/gang/[id]/page.tsx:157-160` catches only to log, then rethrows.
- There is no `error.tsx`/`global-error.tsx` anywhere in `app/` (confirmed via search), so any of these throws hits Next's bare default error page — no retry UI, no partial content.

Net effect: a transient RPC error (the same failure class this PR is designed to fix) now takes down the whole home page or gang page instead of degrading to an empty list or a partial roster. That may be an acceptable trade (fail loud + self-heal on next request, vs. silently-wrong-forever), but it's inconsistent with the fallback pattern applied to `checkPermissionCached` in the same PR, and there's currently no safety net catching it. Worth either mirroring the try/catch-and-degrade approach here, adding an `error.tsx` boundary, or an explicit note that this trade-off is intentional.

**Verdict: changes requested** — the core cache-poisoning fix is correct and well-reasoned; flagging the missing fallback/error-boundary on the three sibling call sites before merge.

---

PR #2011, opened 2026-08-15
Show outstanding Advancements on the N26 Advancements header

joeseos:
Crossing an XP rank earns an Advancement, but nothing said so: the header's
progress readout simply slid to the next range, so a fighter could sit on
unspent Advancements indefinitely without the roster hinting at it. An amber
count now sits beside the readout whenever any are outstanding, and is absent
otherwise, so its presence is the whole message.

Amber rather than the green already used for XP pills, which would blur into
the readout it sits next to. Rendered inside the existing isCumulativeXp
branch, so N23 — where Advancements are bought, not earned — is untouched.

advancementsEarnedFor and openAdvancementsFor come back from history to do
the arithmetic, the latter reworded around this use: the count is floored at
zero because taking an Advancement is no longer gated on having earned one,
so a house-ruling group can hold more than its XP accounts for. The taken
count is the header's existing advancementCount, so countAdvancementsTaken
stays deleted rather than computing the same number twice.

### DIFF
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -9,7 +9,7 @@ import { TypeSpecificData } from '@/types/fighter-effect';
 import { createClient } from '@/utils/supabase/client';
 import { getSkillSetRank } from "@/utils/skillSetRank";
 import { characteristicRank } from "@/utils/characteristicRank";
-import { nextTierStartFor } from "@/utils/advancementRanks";
+import { nextTierStartFor, openAdvancementsFor } from "@/utils/advancementRanks";
 import { List } from "@/components/ui/list";
 import { UserPermissions } from '@/types/user-permissions';
 import { useMutation, useQuery } from '@tanstack/react-query';
@@ -180,6 +180,7 @@ interface FighterChanges {
 
 interface AdvancementsListProps {
   fighterXp: number;
+  fighterStartingXp?: number | null;
   fighterChanges?: FighterChanges;
   fighterId: string;
   editionSlug?: string | null;
@@ -2942,6 +2943,7 @@ export function AdvancementModal({
 // AdvancementsList Component
 export function AdvancementsList({
   fighterXp,
+  fighterStartingXp = null,
   fighterChanges = { advancement: [], characteristics: [], skills: [] },
   fighterId,
   editionSlug = null,
@@ -3371,6 +3373,14 @@ export function AdvancementsList({
   const nextTierStart = nextTierStartFor(editionSlug, fighterXp);
   const xpProgress = `${fighterXp}/${nextTierStart ?? '–'}`;
 
+  // advancementCount is the taken count: effects plus is_advance skills.
+  const openAdvancements = openAdvancementsFor(
+    editionSlug,
+    fighterStartingXp,
+    fighterXp,
+    advancementCount,
+  );
+
   const title = (
     <>
       <span className="sm:hidden">Advanc.</span>
@@ -3380,6 +3390,14 @@ export function AdvancementsList({
           {' '}
           <span className="text-sm sm:hidden">({xpProgress})</span>
           <span className="text-sm hidden sm:inline">(XP: {xpProgress})</span>
+          {openAdvancements > 0 && (
+            <>
+              {' '}
+              <span className="align-middle inline-flex items-center rounded-full bg-amber-500 px-2 py-0.5 text-xs font-semibold text-white">
+                {openAdvancements}
+              </span>
+            </>
+          )}
         </>
       ) : advancementCount > 0 ? (
         <>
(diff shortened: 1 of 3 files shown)


joeseos:
@joeseos Reviewed the diff, the migration it depends on, and the surrounding N26/N23 XP-capability code. This is a well-reasoned, well-documented change — no correctness bugs found. A couple of low-severity notes below; nothing blocking.

## Findings

**1. `bg-amber-500` badge duplicates `components/ui/badge.tsx` instead of extending it** — `components/fighter/fighter-advancement-list.tsx:3396` (low, `dry`)
The existing `Badge` component (`components/ui/badge.tsx`) already centralizes pill styling via `cva` variants (`default`, `secondary`, `destructive`, `outline`, `discreet`), but none of them are amber/warning-colored, so this PR hand-rolls a new `<span className="... rounded-full bg-amber-500 ...">` instead. Not a bug — the file doesn't use `Badge` anywhere else either, and a bare `<span>` fits the surrounding sibling spans better than `Badge`'s `<div>` — but if an amber/warning treatment shows up again, it'd be worth promoting to a `warning` variant on `Badge` rather than a third hand-rolled pill.

**2. `advancementsEarnedFor` re-scans both tier arrays on every render, no memoization** — `utils/advancementRanks.ts:61-71`, called from `components/fighter/fighter-advancement-list.tsx:3377-3382` (low, `efficiency`)
`tiersAtOrBelow` does a linear `.filter()` over a ≤20-element array, twice, on every render of `AdvancementsList` (not wrapped in `useMemo` like `transformedAdvancements` above it). Given the array size this is negligible in practice — not worth blocking on — just flagging since the surrounding derived values in this component are otherwise memoized.

## Correctness spot-check
- Verified the tier-range math (`tiersAtOrBelow(currentXp) - tiersAtOrBelow(startingXp)`) against the "Ganger recruited on 13 XP" example in the docstring, and against a fighter recruited mid-tier (e.g. 5 XP) — both produce the intended count, avoiding the widening-tier double-count that a plain `currentXp - startingXp` subtraction would cause.
- Confirmed `starting_xp` nullability semantics match the `20260810130000_n26_advancements.sql` migration (`NULL` = type can't gain XP), and that treating `null` as `0` in `advancementsEarnedFor` is safe given `xp` is also seeded from `starting_xp ?? 0` at fighter creation (`app/actions/add-fighter.ts`).
- Confirmed the amber badge only renders inside the existing `isCumulativeXp` branch, so N23 (which spends XP rather than earning Advancements by rank) is unaffected — `TIER_STARTS_BY_EDITION.n23` stays `[]` and `hasCumulativeXp` gates it out regardless.
- `openAdvancementsFor` argument order at the call site matches its signature (`editionSlug, startingXp, currentXp, advancementsTaken`).
- `AdvancementsList` has exactly one call site (`fighter-page.tsx`), so making `fighterStartingXp` optional with a `null` default doesn't leave any other caller in a stale state.
- Ran `tsc --noEmit`; pre-existing, unrelated failures only (`react-icons/lu`, `next/link` — `node_modules` isn't installed in this review environment), nothing implicating the changed files.

## Verdict
**Approve.** No blockers — the two notes above are optional polish.

---

joeseos:
## Review

@joeseos This is the first review on this PR (no prior review comments to reconcile), so I reviewed the full diff — the single commit is small (67/-4 across 3 files) and self-contained.

### Previous findings
N/A — no prior review comments on this PR.

### New findings
None that rise to reportable severity. A few things I specifically checked and found correct:

- **`advancementsEarnedFor` tier math.** `tiersAtOrBelow(current) - tiersAtOrBelow(startingXp ?? 0)`, both floored via the outer `Math.max(0, …)`. Verified against the "recruited mid-tier" case the docstring calls out (e.g. recruited at 13 XP, which is itself a tier start, correctly yields 0 earned until XP passes 19) and against XP dropping below `starting_xp` (clamped to 0, never negative).
- **Null `starting_xp` semantics.** Confirmed against `supabase/migrations/20260810130000_n26_advancements.sql`: NULL genuinely means "type cannot gain XP," and defaulting it to 0 in `advancementsEarnedFor` is the documented, deliberate choice (lets house-ruled XP on such a fighter still rank up) rather than an oversight.
- **Prop threading.** `AdvancementsList` has exactly one call site (`fighter-page.tsx`), and it's updated in the same commit — no stale caller missing `fighterStartingXp`.
- **Gating.** The new badge only renders inside the existing `isCumulativeXp` branch, and `openAdvancementsFor` independently returns 0 for N23 anyway (empty tier list), so there's no path to the badge appearing outside N26.
- **Styling.** `bg-amber-500 ... text-white` for a count badge matches the existing pattern used in `fighter-equipment-list.tsx`, `vehicle-equipment-list.tsx`, and `stash-tab.tsx`.

One minor, non-blocking observation: the amber badge has no title/tooltip, unlike the other `bg-amber-500` badges in the codebase which use `data-tooltip-content`. The PR description explicitly frames "presence is the whole message" as the intended design, so I'm not flagging this as a fix-it — just noting it in case a hover label is worth adding later for first-time users.

No test suite exists in this repo to run against the new pure functions (`advancementsEarnedFor`/`openAdvancementsFor`), consistent with the rest of the codebase.

### Verdict
**Approve.**

---

PR #2012, opened 2026-08-15
fix: equipment picker silently dropping items past PostgREST's row cap

FlorianWeissDev:
## Issue

Some weapons (e.g. **Stub gun**) were missing from the Default Equipment / Equipment List pickers when creating or editing a custom fighter type, on both N23 and N26. Other items in the same category showed up fine.

## Root cause

`app/api/equipment/route.ts` queried the `equipment` and `custom_equipment` tables directly, with no pagination. PostgREST silently caps unpaginated responses at its default max-rows limit (1000) — with no error. Once the combined catalog exceeded that cap, anything sorting alphabetically after the cutoff was dropped from the response. "Stub gun" happened to fall right at that boundary.

## Fix

Page through both queries with the existing `fetchAllRows` helper (`utils/supabase/fetch-all-rows.ts`), matching the pattern already used in the admin equipment route. No changes needed on the client side — the picker logic was correct, it just never received the full catalog.

### DIFF
--- a/app/api/equipment/route.ts
+++ b/app/api/equipment/route.ts
@@ -1,7 +1,25 @@
 import { NextRequest, NextResponse } from 'next/server'
 import { createClient } from "@/utils/supabase/server";
 import { getUserIdFromClaims } from "@/utils/auth";
-import { editionSlugFromJoin } from "@/types/edition";
+import { editionSlugFromJoin, EditionJoin } from "@/types/edition";
+import { fetchAllRows } from "@/utils/supabase/fetch-all-rows";
+
+interface RegularEquipmentRow {
+  id: string;
+  equipment_name: string;
+  equipment_category: string;
+  equipment_type: string;
+  core_equipment: boolean | null;
+  editions: EditionJoin;
+}
+
+interface CustomEquipmentRow {
+  id: string;
+  equipment_name: string;
+  equipment_category: string;
+  equipment_type: string;
+  editions: EditionJoin;
+}
 
 export async function GET(request: NextRequest) {
   const supabase = await createClient();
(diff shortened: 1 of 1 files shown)


joeseos:
## Review

No prior review comments exist on this PR, so this covers the full (single-commit) diff.

**Correctness**
- The fix directly addresses the stated root cause: `app/api/equipment/route.ts` now pages through both the `equipment` and `custom_equipment` queries via `fetchAllRows`, matching the pattern already established in `app/api/admin/equipment/route.ts`.
- Good catch adding `.order('id')` as a tiebreaker after `.order('equipment_name')` on both queries — without a unique secondary sort key, rows with duplicate `equipment_name` values could shift between pages and either be skipped or duplicated during pagination. This is required for the fix to be correct, not just a style choice.
- Filters (`equipment_type`, `core_equipment`) are re-applied per page inside each `buildPage` callback, which is correct since Supabase query builders are single-use and must be rebuilt on every call.
- Error handling is preserved: `fetchAllRows` throws on a Postgrest error, which propagates to the existing outer `try/catch` and returns the same 500 response as before.
- Response shape and filtering semantics are unchanged from the client's perspective — only the server-side fetch strategy changed.

**Edge cases**
- Empty results, unauthenticated requests, and the `core_equipment === 'false'` OR-filter are all preserved from the original logic.
- Pagination is bounded by actual row count (loop exits on an empty page), so no risk of an infinite loop or off-by-one at the boundary.

**Nothing else stood out.** No new bugs, regressions, or edge-case gaps introduced by this change. The implementation is minimal and consistent with existing conventions in the codebase.

**Verdict: ✅ Approve**

cc @FlorianWeissDev

---

joeseos (approved):
@FlorianWeissDev Reviewed this PR (single file, `app/api/equipment/route.ts`, 53/-26).

## Summary
The fix replaces two single-shot Supabase queries (implicitly capped at PostgREST's default 1000-row limit) with `fetchAllRows`-driven pagination for both `equipment` and `custom_equipment`. This matches the root cause described in the PR body exactly.

## What I checked
- **Correctness**: `fetchAllRows` (`utils/supabase/fetch-all-rows.ts`) is a pre-existing helper already used identically in `app/api/admin/notifications/route.ts`, `app/api/admin/equipment/route.ts`, and `app/api/admin/fighter-types/route.ts` — this PR follows that established convention exactly (order-by-name + order-by-id tiebreaker + `.range(from, to)`, rebuilding the query builder per page).
- **Pagination key**: `id` is a `uuid NOT NULL` primary key on both `equipment` and `custom_equipment`, so the added `.order('id')` secondary sort is a valid, deterministic tiebreaker. This actually closes a latent gap — without it, ties on `equipment_name` would have had no stable secondary order, risking skipped/duplicated rows across page boundaries.
- **Error handling**: the removed inline `if (...Result.error) throw ...` checks are correctly re-established inside `fetchAllRows` (`if (error) throw error;`), so errors still propagate into the route's outer `try/catch`.
- **Filter semantics**: `equipment_type` eq filter and the `core_equipment` null-or-false or-clause are applied identically to before, just relocated into the `buildPage` closures.
- **Call sites**: traced `types/equipment.ts`, `custom-equipment.tsx`, `custom-fighters.tsx`, `custom-trading-posts.tsx` — response shape (flat JSON array) is unchanged, so no breakage downstream.
- No repo/user `CLAUDE.md` conventions apply here (none exists in the repo).

No correctness bugs, edge-case gaps, security issues, or DRY/YAGNI concerns found. The one cosmetic nit — `import { EditionJoin }` vs. the `import { type EditionJoin }` style used elsewhere (e.g. `app/actions/add-fighter.ts`) — has no compile/runtime effect since `verbatimModuleSyntax` isn't enabled, so I'm not blocking on it.

## Verdict: **Approve**
Straightforward, well-scoped fix that follows an existing pattern already proven in three other routes. No blockers.

---

PR #2013, opened 2026-08-15
Fix email button styling for dark mode across email clients

joeseos:
## Summary
Updated the email notification layout to properly support dark mode across various email clients by switching from a light button design to a dark button design that survives client-side dark mode transformations.

## Key Changes
- **Button color scheme**: Changed from white background with dark text (`#ffffff` bg, `#0a0a0a` text) to dark background with light text (`#404040` bg, `#ffffff` text)
- **Button border**: Added explicit border color (`#737252`) to maintain visibility in dark mode
- **Color scheme declaration**: Updated meta tags from `"light dark"` to `"dark"` only to prevent clients from applying their own dark mode transforms
- **CSS styling**: Added comprehensive style block with:
  - Root-level `color-scheme` and `supported-color-schemes` declarations for consistency
  - Outlook.com dark mode compatibility using `data-ogsc` and `data-ogsb` attribute selectors to restore button colors after Outlook's recoloring
- **HTML classes**: Added `mm-button` and `mm-button-cell` classes to button elements for targeted CSS styling

## Implementation Details
The changes address a critical issue where Android's Gmail dark mode (via WebView force-dark) was repainting light backgrounds while leaving text colors unchanged, resulting in invisible buttons. The new dark-on-light design matches the app's primary button styling and uses mid-tone colors (`#404040`) that dark mode transforms typically leave untouched. Outlook.com-specific CSS rules ensure the button survives that client's aggressive recoloring while other email clients that respect the `color-scheme` declaration leave the palette alone.

### DIFF
--- a/supabase/edge-functions/send-notification-email/_email-layout.ts
+++ b/supabase/edge-functions/send-notification-email/_email-layout.ts
@@ -12,6 +12,15 @@
 // the Next files are extensionless, so the worker carries its own copy rather than
 // importing across the runtime boundary.
 
+// The CTA is deliberately a DARK fill with a LIGHT label — do not "restore" a white button.
+// Android dark-mode recolouring (Gmail's own transform, and WebView force-dark) repaints light
+// *backgrounds* but leaves authored *text* colours alone. A white button therefore arrived as a
+// darkened pill still carrying its near-black label, i.e. an empty-looking outline, while the
+// white wordmark and headings on the same dark card came through untouched. So nothing here may
+// depend on a light background for its contrast: the button gets light-on-dark like every other
+// element in this layout, which is also the app's own primary button (bg-neutral-900 text-white,
+// components/ui/button.tsx). Mid-tones are additionally the range dark-mode transforms leave
+// alone, so #404040 is stable whether or not a client repaints it.
 const BRAND = {
   name: "Munda Manager",
   bg: "#0a0a0a", // page background
@@ -21,8 +30,9 @@ const BRAND = {
   body: "#a1a1a1", // body copy
   footer: "#525252", // footer copy
   link: "#a1a1a1", // footer links
-  buttonBg: "#ffffff",
-  buttonText: "#0a0a0a",
+  buttonBg: "#404040",
+  buttonBorder: "#737373",
+  buttonText: "#ffffff",
   font:
     "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif",
 };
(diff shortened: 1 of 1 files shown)


joeseos:
@joeseos Review of the current commit (`4dc19f4`) — this is the PR's first review, no prior comments/reviews to reconcile.

**Previous findings:** none — first review pass on this PR.

**New findings**

1. **Low — `supported-color-schemes` isn't a real CSS property.** `_email-layout.ts:120` (`:root { color-scheme: dark; supported-color-schemes: dark; }`). `supported-color-schemes` only exists as a `<meta name="supported-color-schemes">` value (used by Apple Mail/iOS Mail); it isn't a CSS declaration, so this line is inert everywhere — parsers will just drop the unknown property. Harmless, but it doesn't do what the comment above it implies. Suggest dropping it from the `:root` block (the `<meta>` tag above already covers the same clients) or leaving a note that it's meta-only.

2. **Low — top-of-file comment is now slightly stale.** `_email-layout.ts:7` still says "Email clients strip `<head>`/external CSS... so this uses a table-based layout with inline styles only," but the new `<style>` block in `<head>` (relied on for Apple Mail/Outlook.com) means that's no longer strictly true. Worth a one-line update so the invariant doesn't mislead the next editor, since the rest of the layout still is inline-only by design.

Everything else checks out: the dark-fill/light-label CTA and the `data-ogsc`/`data-ogsb` Outlook.com overrides are consistent with the stated Android/Gmail force-dark failure mode, `escapeHtml`/`safeUrl` handling for user-controlled content (subject, ctaLabel, ctaUrl) is untouched, and the new `BRAND` fields are plain constants with no injection or null-handling surface. Diff is small, self-contained, and well-commented.

**Verdict: approve**

---

joeseos:
@joeseos Reviewed this PR (single file, `supabase/edge-functions/send-notification-email/_email-layout.ts`).

## Summary
This is a well-scoped, well-reasoned fix. Switching the CTA to a dark-fill/light-text button and scoping `color-scheme` to `dark` only (rather than the previous, inaccurate `light dark`) correctly addresses the Android/Gmail invisible-button issue, and the `data-ogsc`/`data-ogsb` rules are a known, correct technique for Outlook.com's forced recoloring. Contrast between `buttonBg` (#404040) and `buttonText` (#ffffff) is solid, and the border color is now distinct from the fill so the button has a visible edge on clients that don't apply any dark-mode transform at all. No correctness, security, or performance issues found.

## Findings

| # | File:Line | Issue | Suggested fix | Severity |
|---|---|---|---|---|
| 1 | `_email-layout.ts:118` | `supported-color-schemes: dark;` is used as a CSS property inside the `:root { ... }` rule. That's not a real CSS property (it only exists as a `<meta name="supported-color-schemes">` value/older Apple Mail extension) — as written it's a no-op that any CSS parser will silently ignore. Harmless, but doesn't do what the comment implies. | Drop it from the `:root` rule (the `<meta name="supported-color-schemes" content="dark">` tag already covers this), keeping just `color-scheme: dark;`. | Low |
| 2 | PR description vs. `_email-layout.ts:34` | Description states the added border color is `#737252`, but the code uses `#737373`. Doesn't affect functionality, just a documentation mismatch. | Update the PR description to match, or vice versa if `#737252` was intended. | Low |

Neither finding blocks merge — both are cosmetic/documentation nits, not code defects.

## Verdict: **Approve**

---

PR #2014, opened 2026-08-16
Make custom-asset sharing edition-aware

joeseos:
The Custom Assets tab is edition-gated end to end, but the share flow was not:
ShareToCampaignsModal listed every campaign the user arbitrates regardless of
edition, UserCampaign carried no edition, and no share action compared the item's
edition against the campaign's. An arbitrator on the N26 toggle could tick an N23
campaign, for any custom asset -- not just Asset Collections.

Cross-edition sharing is now invalid at every layer:

- custom_shared gains a NOT NULL edition_id FK, backfilled from the campaign's
  type, with a BEFORE INSERT/UPDATE trigger that derives it and rejects a definite
  mismatch. The edition comes from the campaign and is never defaulted:
  types/edition.ts reads a missing slug as n23 for display filtering only, so an
  unresolvable campaign edition raises rather than recording an invented one. A
  null on the item side still makes no claim, so legacy rows stay writable.
- UserCampaign carries edition_slug, resolved via campaign_types, and the share
  modal offers only campaigns of the item's own edition.
- assertCampaignsMatchEdition guards all six share actions, mirroring
  assertGangMatchesCampaignEdition.

Also fixes a related edition gap: copy_custom_collection dropped edition_id on
every clone except custom_fighter_types, so copying an N26 collection landed it
as NULL and it resurfaced under N23. Existing NULL rows are left alone -- their
true edition is unrecoverable, and they render as N23 today either way.

ShareCustomSkillModal was a near-duplicate of the shared modal; skills now use
ShareToCampaignsModal, removing ~130 lines and one path that could drift again.

Verified against production data in rolled-back transactions: 665 custom_shared
rows backfill with 0 conflicts and 0 remaining nulls; the trigger stamps a
matching share and rejects an N26 item into an N23 campaign; all five rewritten
copy inserts accept their new column lists.

### DIFF
--- a/app/actions/customise/custom-share.ts
+++ b/app/actions/customise/custom-share.ts
@@ -147,6 +147,9 @@ export async function shareCustomFighter(customFighterTypeId: string, campaignId
               .from('custom_shared')
               .insert(newSkillShares);
 
+            // Non-fatal by design: the fighter is already shared. Note that a skill whose type
+            // is from another edition would be rejected by the custom_shared trigger, and one
+            // bad row aborts the whole insert, so every skill here fails together.
             if (shareSkillsError) {
               console.error('Error auto-sharing custom skills for fighter:', shareSkillsError);
             }
--- a/app/actions/customise/custom-share.ts
+++ b/app/actions/customise/custom-share.ts
@@ -4,6 +4,45 @@ import { createClient } from '@/utils/supabase/server';
 import { getAuthenticatedUser } from '@/utils/auth';
 import { revalidateTag } from 'next/cache';
 import { CACHE_TAGS } from '@/utils/cache-tags';
+import { editionsConflict, editionSlugFromJoin } from '@/types/edition';
+
+/**
+ * A custom asset may only be shared to campaigns of its own edition. Uses
+ * editionsConflict, not sameEditionForDisplay: this rejects a user action, so an
+ * unresolved edition must not be read as N23 and used to turn someone away.
+ */
+async function assertCampaignsMatchEdition(
+  supabase: Awaited<ReturnType<typeof createClient>>,
+  itemEditionSlug: string | null,
+  campaignIds: string[]
+): Promise<{ ok: true } | { ok: false; error: string }> {
+  if (campaignIds.length === 0 || itemEditionSlug == null) return { ok: true };
+
+  const { data: campaigns, error } = await supabase
+    .from('campaigns')
+    .select('id, campaign_types!campaign_type_id (editions:edition_id (slug))')
+    .in('id', campaignIds);
+
+  if (error) {
+    return { ok: false, error: 'Failed to verify campaign editions' };
+  }
+  if ((campaigns?.length ?? 0) !== campaignIds.length) {
+    return { ok: false, error: 'Campaign not found' };
+  }
+
+  const mismatch = (campaigns ?? []).some((campaign: any) =>
+    editionsConflict(itemEditionSlug, editionSlugFromJoin(campaign.campaign_types?.editions))
+  );
+
+  if (mismatch) {
+    return {
+      ok: false,
+      error: 'This custom asset is from a different edition than the campaign'
+    };
+  }
+
+  return { ok: true };
+}
 
 /**
  * Share a custom fighter to selected campaigns
@@ -16,7 +55,7 @@ export async function shareCustomFighter(customFighterTypeId: string, campaignId
     // Verify the custom fighter belongs to the user
     const { data: customFighter, error: fighterError } = await supabase
       .from('custom_fighter_types')
-      .select('id, user_id')
+      .select('id, user_id, editions:edition_id (slug)')
       .eq('id', customFighterTypeId)
       .eq('user_id', user.id)
       .single();
@@ -24,6 +63,11 @@ export async function shareCustomFighter(customFighterTypeId: string, campaignId
     if (fighterError || !customFighter) {
       return { success: false, error: 'Custom fighter not found or not owned by user' };
     }
+
+    const editionCheck = await assertCampaignsMatchEdition(
+      supabase, editionSlugFromJoin((customFighter as any).editions), campaignIds
+    );
+    if (!editionCheck.ok) return { success: false, error: editionCheck.error };
 
     // Delete existing shares for this fighter
     const { error: deleteError } = await supabase
@@ -134,7 +178,7 @@ export async function shareCustomGangType(customGangTypeId: string, campaignIds:
     // Verify the custom gang type belongs to the user
     const { data: customGangType, error: gangTypeError } = await supabase
       .from('custom_gang_types')
-      .select('id, user_id')
+      .select('id, user_id, editions:edition_id (slug)')
       .eq('id', customGangTypeId)
       .eq('user_id', user.id)
       .single();
@@ -142,6 +186,11 @@ export async function shareCustomGangType(customGangTypeId: string, campaignIds:
     if (gangTypeError || !customGangType) {
       return { success: false, error: 'Custom gang type not found or not owned by user' };
     }
+
+    const editionCheck = await assertCampaignsMatchEdition(
+      supabase, editionSlugFromJoin((customGangType as any).editions), campaignIds
+    );
+    if (!editionCheck.ok) return { success: false, error: editionCheck.error };
 
     // Delete existing shares for this gang type
     const { error: deleteError } = await supabase
@@ -293,7 +342,7 @@ export async function shareCustomEquipment(customEquipmentId: string, campaignId
     // Verify the custom equipment belongs to the user
     const { data: customEquipment, error: equipmentError } = await supabase
       .from('custom_equipment')
-      .select('id, user_id')
+      .select('id, user_id, editions:edition_id (slug)')
       .eq('id', customEquipmentId)
       .eq('user_id', user.id)
       .single();
@@ -301,6 +350,11 @@ export async function shareCustomEquipment(customEquipmentId: string, campaignId
     if (equipmentError || !customEquipment) {
       return { success: false, error: 'Custom equipment not found or not owned by user' };
     }
+
+    const editionCheck = await assertCampaignsMatchEdition(
+      supabase, editionSlugFromJoin((customEquipment as any).editions), campaignIds
+    );
+    if (!editionCheck.ok) return { success: false, error: editionCheck.error };
 
     // Delete existing shares for this equipment
     const { error: deleteError } = await supabase
@@ -353,7 +407,7 @@ export async function shareCustomSkill(customSkillId: string, campaignIds: strin
     // Verify the custom skill belongs to the user
     const { data: customSkill, error: skillError } = await supabase
       .from('custom_skills')
-      .select('id, user_id')
+      .select('id, user_id, custom_skill_types:custom_skill_type_id (editions:edition_id (slug))')
       .eq('id', customSkillId)
       .eq('user_id', user.id)
       .single();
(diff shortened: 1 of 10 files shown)


joeseos:
@joeseos Reviewed the diff, the DB trigger/migration, and the `types/edition.ts` helpers this builds on. Solid design overall (defense-in-depth: TS-level `assertCampaignsMatchEdition` + DB trigger, careful null-as-"no claim" semantics, backfill verified against prod). One high-severity bug found in the share modal that I think should block merge.

### Findings

**1. [High] Legacy (null-edition) items silently lose their N26 shares when the share modal is confirmed**
`components/customise/custom-shared.tsx:77-79` and `:98-104`

```js
const eligibleCampaigns = useMemo(
  () => userCampaigns.filter(c => sameEditionForDisplay(c.edition_slug, itemEditionSlug)),
  [userCampaigns, itemEditionSlug]
);
...
const eligibleIds = new Set(eligibleCampaigns.map(c => c.id));
setSelectedCampaigns(sharedCampaignIds.filter(id => eligibleIds.has(id)));
```

`sameEditionForDisplay` reads a missing/null edition as `n23` (that's its documented purpose — display filtering). But `itemEditionSlug` can genuinely be `null` here: `edition_id` is nullable on every one of the shareable tables (`custom_equipment`, `custom_fighter_types`, `custom_gang_types`, `custom_skill_types`, `custom_trading_posts`, `custom_collections`), and the PR description itself notes such rows exist in prod today ("Existing NULL rows are left alone — their true edition is unrecoverable").

For such an item, `eligibleCampaigns` collapses to N23-only campaigns, so if the item was previously shared to an N26 campaign, that campaign silently disappears from `eligibleCampaigns` — and therefore from the `selectedCampaigns` state seeded at line 103. The user never sees it, never unchecks it, but the moment they confirm the modal for *any* reason (e.g. just adding one more N23 campaign), the share action does a delete-then-reinsert of `selectedCampaigns`, permanently dropping the N26 share.

This is exactly the failure mode `assertCampaignsMatchEdition` was written to avoid — its own doc comment says: *"Uses editionsConflict, not sameEditionForDisplay: this rejects a user action, so an unresolved edition must not be read as N23 and used to turn someone away."* The backend/trigger correctly let null-edition items through to any campaign; the modal's eligibility filter doesn't follow the same rule.

Suggested fix: don't apply the edition filter at all when `itemEditionSlug == null` (or filter with `!editionsConflict(c.edition_slug, itemEditionSlug)` instead of `sameEditionForDisplay`), so a null item shows — and keeps — every arbitrated campaign, matching the server-side semantics. The empty-state copy at line 159 (`None of your campaigns are {(itemEditionSlug ?? EDITION_N23).toUpperCase()} campaigns.`) has the same null→N23 assumption baked in and should be revisited alongside the fix.

**2. [Low] Cross-edition auto-share failures are now silently swallowed**
`app/actions/customise/custom-share.ts:151` (in `shareCustomFighter`'s skill-access cascade)

```js
if (shareSkillsError) {
  console.error('Error auto-sharing custom skills for fighter:', shareSkillsError);
}
```

This auto-shares a fighter's referenced custom skills to the same campaigns, but doesn't run `assertCampaignsMatchEdition` for those skill rows first. If a fighter's skill access ever points at a skill type from a different edition than the fighter (an existing data-consistency assumption, not enforced elsewhere), the new DB trigger will now reject that insert — and the function still returns `{ success: true }` with the failure only in server logs. Pre-existing swallow pattern, but the trigger gives it a new, edition-specific way to fire. Not blocking, just worth a comment or a surfaced warning if it's expected to matter in practice.

### Verdict: Changes requested

Blocker is #1 — it's a real, silent data-loss path for existing production rows (null `edition_id` is documented as a known, present-day state), not just a hypothetical. Everything else — the trigger, the backfill, the `assertCampaignsMatchEdition` guard, the `copy_custom_collection` edition_id fixes, and the skill-modal consolidation — looks correct and well-reasoned.


---

joeseos:
## Review: Make custom-asset sharing edition-aware

This PR has no prior review comments, so the full diff (the single commit `4e3e4e6`) was reviewed as "the latest changes."

### Previous findings
None — this is the first review of this PR.

### New findings

**1. [Medium] `components/customise/custom-shared.tsx:77-80` (and the empty-state copy at ~159) — the eligible-campaign filter contradicts the server-side edition semantics for legacy (null-edition) items**

```ts
const eligibleCampaigns = useMemo(
  () => userCampaigns.filter(c => sameEditionForDisplay(c.edition_slug, itemEditionSlug)),
  [userCampaigns, itemEditionSlug]
);
```

`sameEditionForDisplay` defaults a missing/null slug to `n23` — by its own doc comment in `types/edition.ts`, that's meant for **display** filtering only. `assertCampaignsMatchEdition` in `custom-share.ts` deliberately avoids this exact function, with a comment explaining why: *"an unresolved edition must not be read as N23 and used to turn someone away."* It uses `editionsConflict` instead, which treats `itemEditionSlug == null` as "no claim" — a legacy custom asset with no `edition_id` may be shared to **any** campaign. The DB trigger (`custom_shared_set_edition`) agrees: a null item edition never blocks the insert.

The modal's `eligibleCampaigns` doesn't follow that same rule. For a legacy item (`itemEditionSlug === null`) whose owner only arbitrates N26 campaigns, `eligibleCampaigns` comes back empty, and the modal shows "None of your campaigns are N23 campaigns" — blocking a share the server would actually accept. Before this PR the modal listed every `userCampaigns` entry unconditionally, so this restriction is newly introduced here.

Suggested fix: skip the filter when `itemEditionSlug == null` (show all `userCampaigns`), or filter on `!editionsConflict(c.edition_slug, itemEditionSlug)` instead of `sameEditionForDisplay`, matching the "don't turn someone away over an edition that failed to resolve" rule already applied server-side.

**2. [Low] Cascaded auto-shares can now silently fail on an edition mismatch without surfacing an error**

`shareCustomFighter`/`shareCustomGangType` auto-share referenced custom skills (and, for gang types, fighters) to the same `campaignIds` after the primary share succeeds, and `shareCollection` does the equivalent cascade explicitly. These secondary inserts only `console.error` on failure — the user still sees a success toast. Previously these inserts couldn't fail on edition grounds; now the new `custom_shared_set_edition` trigger can reject a cascaded insert if e.g. a fighter's referenced custom skill type has a different `edition_id` than the fighter itself (`assertCampaignsMatchEdition` only validates the top-level item, not cascaded items). Likely rare given the audited production data, but worth a quick sanity check that this failure mode is acceptable — it fits the existing silent-cascade-failure pattern, just with a new way to trigger it.

### Edge cases / null handling
Otherwise handled consistently: `assertCampaignsMatchEdition`'s early return on `itemEditionSlug == null`, the trigger's "either side null → no claim" logic, and `editionSlugFromJoin`'s object/array normalization all agree with each other — finding #1 is the one place that consistency breaks.

### `copy_custom_collection.sql` / migration
Cross-checked against `schema.public.sql`: all six tables that actually have an `edition_id` FK (`custom_collections`, `custom_equipment`, `custom_fighter_types`, `custom_gang_types`, `custom_skill_types`, `custom_trading_posts`) are now carried through the clone — nothing missed, nothing over-added (`custom_skills` correctly has no `edition_id` column of its own). The migration's statement ordering (backfill via plain `UPDATE` before the trigger exists → install the trigger → `SET NOT NULL`) is correct and matches the documented audit (665 rows, 0 remaining nulls).

### Readability / conventions
No concerns. The shared `assertCampaignsMatchEdition` helper and folding `ShareCustomSkillModal` into `ShareToCampaignsModal` are clean wins, and the comments explain the non-obvious null-handling decisions well.

Finding #1 is a real (if narrow) regression in the share modal's campaign picker for legacy items; everything else is solid. Recommend fixing #1 before merge and taking a look at #2.

@joeseos

---

joeseos:
@joeseos Reviewed the new commit only (`4e3e4e6...93cb760`, pushed after both prior reviews) against `components/customise/custom-shared.tsx`, `app/actions/customise/custom-share.ts`, `types/edition.ts`, `supabase/migrations/20260816120000_add_edition_id_to_custom_shared.sql`, and `supabase/schema/schema.public.sql`.

## Previous findings

1. **[High] Modal's eligible-campaign filter read a null item edition as N23, silently dropping N26 shares on confirm** (`custom-shared.tsx:77-79`) — **Resolved.** The filter now uses `!editionsConflict(c.edition_slug, itemEditionSlug)` instead of `sameEditionForDisplay`, matching the "null makes no claim" rule the doc comment on `editionsConflict` describes and that `assertCampaignsMatchEdition`/the DB trigger already followed. Checked the empty-state copy too (`{itemEditionSlug?.toUpperCase()}`, now without the `?? EDITION_N23` fallback): that branch only renders when `eligibleCampaigns.length === 0 && userCampaigns.length > 0`, and with the new filter that's only reachable when `itemEditionSlug` is non-null (a null item is never excluded by `editionsConflict`), so the removed fallback can't actually render as `null`/`undefined` — safe.
2. **[Low] Cascaded skill auto-share can now silently fail on an edition mismatch** (`custom-share.ts` in `shareCustomFighter`) — **Acknowledged, not changed.** A comment was added at the `console.error` explaining the trigger can reject a cross-edition row here and that one bad row aborts the whole cascade insert. This was flagged as non-blocking last round and the behavior is unchanged, just documented — reasonable given it's a pre-existing silent-cascade pattern the PR didn't introduce.

## New changes reviewed

**`supabase/migrations/20260816130000_backfill_null_custom_edition_ids.sql`** (new): backfills `edition_id` on the six `custom_*` root tables for the 385 rows left null by `copy_custom_collection`'s pre-existing bug (fixed in the first commit). I checked this migration on its own merits since it's new, security/data-integrity-relevant SQL:
- Ordering is correct and dependency-aware: fighter types → gang types (reads fighter types) → equipment (reads fighter types via the join table) → skill types/trading posts (no dependencies) → collections (reads everything else via the `items` jsonb). Each step only consumes tables already resolved by an earlier step.
- Verified every table/column referenced (`custom_fighter_type_equipment`, `equipment_categories.edition_id`, `custom_gang_types.custom_gang_type_id` back-reference via `custom_fighter_types`, etc.) exists in `schema.public.sql` — no typos.
- The collections step derives edition from `items` jsonb entries by id, without discriminating on the `type` field also present in each entry (`copy_custom_collection.sql` confirms entries are tagged `equipment`/`fighter_type`/`gang_type`/`skill`/`trading_post`). Functionally fine since ids are UUIDs unique per table, just noting it relies on that rather than the discriminator already available.
- The closing `DO $$ ... RAISE EXCEPTION` if any row is left null is a good belt-and-suspenders check — it makes the migration fail loudly rather than silently leaving a gap the trigger/modal would then have to reason about.
- Runs after `20260816120000_add_edition_id_to_custom_shared.sql` (which derives `custom_shared.edition_id` purely from the *campaign's* type, not the item's). Confirmed there's no ordering hazard: backfilling item `edition_id` afterward doesn't retroactively touch already-recorded `custom_shared` rows, so the two migrations are independent.

I also traced `shareCollection` (unchanged by this commit) against my initial suspicion that the DB trigger's `v_item_edition` `COALESCE` list doesn't reference `custom_collection_id` — turns out that's fine: `shareCollection` expands a collection into per-item `custom_shared` rows keyed by the item's own column (`custom_equipment_id`, etc.), with `custom_collection_id` only riding along as a tag, so the trigger still resolves each row's real item edition correctly. No bug there.

## Edge cases / null handling
Consistent throughout: the modal, `assertCampaignsMatchEdition`, the trigger, and now the backfill's own "unresolved → fail loudly" step all agree on the same null-handling contract. No gaps found in this commit.

## Readability / conventions
Migration follows the repo's existing `BEGIN ... COMMIT` convention for multi-statement migrations, and the inline comments clearly explain why each heuristic was chosen (and which "obvious" ones were rejected and why) — good trail for anyone auditing this later.

## Verdict: Approve

Both previously-blocking/flagged items are resolved or reasonably acknowledged, and the new migration is correct, well-guarded, and consistent with the rest of the feature. No new issues found.

---

PR #2016, opened 2026-08-16
Show grouped weapon profiles whose parent weapon has no group id

joeseos:
Ammo sold as its own equipment ("Krak grenades (Subjugator grenade launcher)") carries a weapon_group_id pointing at the parent weapon, and the server attaches those profiles onto the parent's profile list. The card table then re-derived a group key per profile as `weapon_group_id || fighter_weapon_id`, so the parent's own profiles (group id null) and the attached ammo (group id = parent's catalog id) landed in different blocks. The ammo block had no base profile and was dropped as an orphan, which is why the grenades were costed and listed under Equipment but never reached the fighter card.

Key every profile that belongs to the weapon -- null group id or one pointing at the weapon's own catalog id -- to the same anchor, keeping that anchor as whatever the weapon already used so self-grouping weapons still collapse duplicates into one "(xN)" row.

The mismatch comes from the admin API writing two different defaults: POST stored null for an ungrouped profile while PATCH stored the weapon's own id, so a weapon created and never re-saved lost its ammo rows. Align POST with PATCH and backfill the weapons already affected.

Also stop the fighter equipment fetch from indexing a self-grouping profile twice under the same weapon.

Claude-Session: https://claude.ai/code/session_01J5yfpDfyQ43xiV7wTrq8Ph

### DIFF
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -501,8 +501,10 @@ export async function POST(request: Request) {
         ...profile,
         weapon_id: weaponId,
         profile_name: profile.profile_name.trimEnd(),
-        // Set weapon_group_id to either the selected weapon's ID or this weapon's ID
-        weapon_group_id: profile.weapon_group_id || null,
+        // Set weapon_group_id to either the selected weapon's ID or this weapon's ID.
+        // Must match the PATCH branch below: a profile left ungrouped self-groups,
+        // so ammo attached to this weapon later lands in the same card block.
+        weapon_group_id: profile.weapon_group_id || weaponId,
         // Properly handle explicit null values
         range_short: profile.range_short === null ? '' : profile.range_short || '',
         range_long: profile.range_long === null ? '' : profile.range_long || '',
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -394,8 +394,11 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
         standardProfilesMap.get(profile.weapon_id)!.push(profile);
         
         // Also add profile to the map under weapon_group_id if it exists (for grouped weapons)
-        // But only if the fighter owns the weapon that owns this profile
-        if (profile.weapon_group_id && standardEquipmentIds.includes(profile.weapon_id)) {
+        // But only if the fighter owns the weapon that owns this profile.
+        // A self-grouping profile is already under this key -- re-adding it would
+        // list the same profile twice on the weapon.
+        if (profile.weapon_group_id && profile.weapon_group_id !== profile.weapon_id
+            && standardEquipmentIds.includes(profile.weapon_id)) {
           if (!standardProfilesMap.has(profile.weapon_group_id)) {
             standardProfilesMap.set(profile.weapon_group_id, []);
           }
--- a/components/gang/fighter-card-weapon-table.tsx
+++ b/components/gang/fighter-card-weapon-table.tsx
@@ -134,8 +134,28 @@ const WeaponTable: React.FC<WeaponTableProps> = ({ weapons, entity, viewMode, ed
     
     const hardpointKey = weapon.hardpoint_location || 'no-hp';
 
+    // The ids that mean "this weapon". A weapon's own profiles spell that two
+    // ways: weapons saved through the admin edit form group under their own
+    // catalog id, weapons saved through the create form leave it null. The
+    // server attaches grouped ammo (separately owned equipment whose profiles
+    // point at this weapon) onto this weapon's profile list, and those always
+    // carry the catalog id -- so both spellings have to resolve to one block,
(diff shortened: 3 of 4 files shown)


joeseos:
@joeseos Review of the latest (and only) commit — no prior review comments/threads exist on this PR, so there's nothing to mark resolved/outstanding.

**Previous findings:** N/A — this is the first review pass.

**Trace-through of the fix:**
- `app/api/admin/equipment/route.ts`: POST now defaults `weapon_group_id` to `weaponId`, matching the existing PATCH branch (confirmed at the `id`-based default further down in the same file). Aligns the two write paths as described.
- `app/lib/shared/fighter-data.ts`: the `weapon_group_id !== weapon_id` guard correctly stops a self-grouping profile from being indexed a second time under `standardProfilesMap`.
- `components/gang/fighter-card-weapon-table.tsx`: walked through the `ownGroupIds`/`ownAnchor` logic against several cases — ungrouped weapon (no ammo), self-grouping weapon with attached ammo (the Krak grenades case), and an explicit cross-weapon group (`weapon_group_id` pointing at a *different* weapon's catalog id, used to merge two purchased weapon entries into one block). All three land on the same key the old code produced when data was already consistent, and correctly merge the previously-orphaned ammo case. No regression found for the pre-existing "ungrouped weapon keyed by `fighter_weapon_id`" behavior — that fallback is unchanged.
- Migration: `UPDATE ... WHERE weapon_group_id IS NULL AND EXISTS (ammo grouped onto this weapon)` is idempotent (re-running matches zero rows) and correctly scoped to only the native profiles of weapons that actually have ammo grouped onto them.

**New findings:** none that block. One low-severity observation:
- `fighter-card-weapon-table.tsx`: `ownAnchor` picks the *first* base profile with a non-null `weapon_group_id` via `.find()`. If a weapon ever had base profiles with inconsistent/differing `weapon_group_id` values (shouldn't happen given POST/PATCH now always write a consistent value), the anchor choice would be array-order-dependent. Not worth guarding against given the write-path fix, just flagging for awareness.

**Verdict: Approve.** The root-cause analysis holds up under a manual walkthrough of the three affected read/write paths, and the migration is safe to re-run.

(Note: I see this PR shows as closed/unmerged — posting this in case the review is still useful for a follow-up PR carrying the same changes.)

---

joeseos:
## Code Review

@joeseos Reviewed the diff against the full codebase (this PR shows as closed at the time of review, but posting findings for the record).

### Summary

Root cause is correctly diagnosed and the fix is sound. I traced the data flow end-to-end:

- `types/equipment.ts`'s `Weapon.weapon_id` is consistently populated as the equipment catalog id everywhere `Weapon[]` is built (`gang-data.ts:1944`, `fighter-card.tsx:267`), so the new `weapon.weapon_id`-based anchor in `fighter-card-weapon-table.tsx` is safe to rely on for both standard and custom weapons.
- The client-side fix in `fighter-card-weapon-table.tsx` (treating any profile whose `weapon_group_id` equals `weapon.weapon_id` as "own") is robust even *without* the DB backfill/POST fix — it recovers correctly whether the weapon's own profile is self-grouped (`weapon_group_id = weapon_id`) or ungrouped (`null`), since `weapon.weapon_id` is unconditionally added to `ownGroupIds`.
- The `fighter-data.ts` dedup fix (`weapon_group_id !== weapon_id`) matches the same pattern already used a few hundred lines down for vehicle equipment (lines ~839, ~859), so it's consistent with existing convention rather than a new pattern.
- The migration is idempotent (`WHERE weapon_group_id IS NULL AND EXISTS(...)`) and correctly scoped (only backfills weapons that something actually groups under; `ammo.weapon_id <> wp.weapon_id` avoids self-matching).
- POST/PATCH defaults for `weapon_group_id` are now aligned (`profile.weapon_group_id || weaponId` in both).

I did not find any correctness bugs, edge-case gaps, or security/performance issues in the diff.

### Minor notes (non-blocking, low severity)

1. **`custom_weapon_profiles` backfill** — the migration only touches `weapon_profiles` (standard equipment). If custom weapons can have the same POST-vs-PATCH split (i.e., there's a separate creation path for custom weapon profiles that also defaulted to `null`), those wouldn't get backfilled. Worth a quick check that custom equipment doesn't have an equivalent historical gap — if it's genuinely standard-equipment-only (per the admin route being the only place with this bug), no action needed.
2. **`ownAnchor` selection** (`fighter-card-weapon-table.tsx`) — picks the *first* base profile with a truthy `weapon_group_id` via `.find()`. If a single weapon ever had multiple own base profiles with genuinely divergent `weapon_group_id` values (data inconsistency), the anchor choice would depend on array order. This is a pre-existing data-integrity assumption, not something this PR introduces, and is very unlikely in practice.
3. No automated test coverage was added for the grouping logic, but the repo has no existing test infrastructure, so this isn't a deviation from convention.

### Verdict: **Approve**

No blockers. The comments above are informational/food-for-thought, not required changes.

---

PR #2017, opened 2026-08-16
Store a weapon's own id on its ungrouped weapon profiles

joeseos:
Ammo sold as its own equipment ("Krak grenades (Subjugator grenade
launcher)") carries weapon_group_id pointing at the parent weapon, and
the fighter card groups a weapon's rows by that column. The parent's own
profiles must therefore carry the parent's id as well, or the two end up
in different groups and the ammo rows -- having no base profile of their
own -- are discarded before the card renders. The grenades were costed
and listed under Equipment but appeared nowhere on the card.

The admin API wrote two different defaults for an ungrouped profile:
PATCH stored the weapon's own id, POST stored null. A weapon was only
correct once it had been re-saved through the edit form, and four
weapons had been created and never edited: Subjugator grenade launcher,
Assault grenade launcher*, Grenade launcher array and Twin grenade
launcher. The create modal cannot supply the id itself -- the row does
not have one at submit time -- so the default belongs on the server.

Align POST with PATCH and backfill the nine profile rows already stored
without it, scoped to weapons something actually groups onto so an
ungrouped weapon's inert null is left alone.

The edit modal's Weapon Group select excludes the weapon being edited
from its options, so a self-referencing profile matched no option and
rendered blank instead of "Use This Weapon (Default)". Show it as the
default it is.

Also stop the fighter equipment fetch indexing a profile twice under the
same weapon when its group id is the weapon's own.

### DIFF
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -502,7 +502,7 @@ export async function POST(request: Request) {
         weapon_id: weaponId,
         profile_name: profile.profile_name.trimEnd(),
         // Set weapon_group_id to either the selected weapon's ID or this weapon's ID
-        weapon_group_id: profile.weapon_group_id || null,
+        weapon_group_id: profile.weapon_group_id || weaponId,
         // Properly handle explicit null values
         range_short: profile.range_short === null ? '' : profile.range_short || '',
         range_long: profile.range_long === null ? '' : profile.range_long || '',
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -395,7 +395,8 @@ export const getFighterEquipment = async (fighterId: string, supabase: any): Pro
         
         // Also add profile to the map under weapon_group_id if it exists (for grouped weapons)
         // But only if the fighter owns the weapon that owns this profile
-        if (profile.weapon_group_id && standardEquipmentIds.includes(profile.weapon_id)) {
+        if (profile.weapon_group_id && profile.weapon_group_id !== profile.weapon_id
+            && standardEquipmentIds.includes(profile.weapon_id)) {
           if (!standardProfilesMap.has(profile.weapon_group_id)) {
             standardProfilesMap.set(profile.weapon_group_id, []);
           }
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -1830,7 +1830,11 @@ export function AdminEditEquipmentModal({ onClose, onSubmit }: AdminEditEquipmen
                             Attach this profile to an existing weapon, or leave it as is to use with this weapon.
                           </p>
                           <select
-                            value={profile.weapon_group_id || ''}
+                            value={
+                              profile.weapon_group_id && profile.weapon_group_id !== selectedEquipmentId
+                                ? profile.weapon_group_id
+                                : ''
+                            }
                             onChange={(e) => handleProfileChange(index, 'weapon_group_id', e.target.value)}
                             className="w-full p-2 border rounded-md"
                             disabled={!selectedEquipmentId}


joeseos:
## Review

@joeseos Reviewed the diff against the full codebase. This is a clean, well-scoped fix — verdict below.

### Findings

None that block. A couple of confirmations rather than issues:

- **`app/api/admin/equipment/route.ts:505`** (POST) now matches the PATCH handler at `route.ts:687` (`profile.weapon_group_id || weaponId`/`id`), and also matches the existing custom-equipment path in `app/actions/customise/custom-weapon-profiles.ts:62`. Good consistency — this was the actual source of the four un-edited weapons having `NULL` self-references.
- **`app/lib/shared/fighter-data.ts:398`** — the added `profile.weapon_group_id !== profile.weapon_id` guard mirrors the same pattern already used in `app/lib/shared/gang-data.ts` (lines 1306, 1318, 1444, 1456) for the standard/custom, fighter/vehicle equipment maps. Correct fix for the double-indexing bug and consistent with prior art in the same file family.
- **`components/admin/admin-edit-equipment.tsx:1832-1837`** — the select's displayed value now correctly falls back to `''` (→ "Use This Weapon (Default)") when `weapon_group_id === selectedEquipmentId`, matching the submit-time fallback already at line 553. No behavior change on save, only display.
- **Migration `20260816140000_self_group_parent_weapon_profiles.sql`** — correctly scoped (only touches weapons that something groups onto, via the `EXISTS` with `ammo.weapon_id <> wp.weapon_id`), idempotent (re-run is a no-op since the `WHERE weapon_group_id IS NULL` guard excludes already-backfilled rows), and safe (no FK constraint on `weapon_group_id` in `schema.public.sql`, so no ordering/constraint concerns). Naming and timestamp match the repo's migration convention.

### Not in scope, worth noting for awareness only

`supabase/functions/get_gang_details.sql` (`weapon_profiles_deduplicated` CTE, ~line 326) joins profiles by `wp.weapon_id = fe.equipment_id` only, so it never picks up ammo rows grouped via `weapon_group_id` onto a different weapon in the first place — it doesn't have the double-count bug this PR fixes, but it also doesn't render grouped ammo. I couldn't find any current caller of this RPC in `app/`, so it looks currently unused/dead; flagging only in case it gets wired up later, no action needed for this PR.

### Verdict: **Approve**

No blockers. Logic is correct, consistent with existing patterns in the codebase, and the migration is safe to run.

---

joeseos:
@joeseos Reviewed the current changes (single commit, no prior review comments on this PR to reconcile against).

**Previous findings:** None — this is the first review of this PR.

**Review of the diff**

- `app/api/admin/equipment/route.ts` (POST): `weapon_group_id: profile.weapon_group_id || weaponId` now matches the PATCH handler's existing `profile.weapon_group_id || id` (line ~687) exactly. Root cause (POST/PATCH divergence) is fixed correctly.
- `app/lib/shared/fighter-data.ts:398`: the added `profile.weapon_group_id !== profile.weapon_id` guard prevents double-indexing a self-referencing profile into `standardProfilesMap`. This matches the identical guard pattern already used elsewhere in the same file (lines 837, 857) and in `app/lib/shared/gang-data.ts` (lines 1306, 1318, 1444, 1456) — good consistency, this was the one remaining spot missing it.
- `components/admin/admin-edit-equipment.tsx`: the `<select>` value now falls back to `''` when `weapon_group_id === selectedEquipmentId`, correctly rendering "Use This Weapon (Default)" for self-referencing profiles instead of a blank/unmatched select. Consistent with the options list already filtering out `selectedEquipmentId` from `filteredWeapons`.
- Migration `20260816140000_self_group_parent_weapon_profiles.sql`: correctly scoped (only backfills weapons that something actually groups onto via `EXISTS`), idempotent (`weapon_group_id IS NULL` guard), and won't touch genuinely ungrouped weapons. Matches the PR description's "nine profile rows / four weapons" claim structurally.

**New findings:** None of note. No null-handling or edge-case issues found — the `|| weaponId` / `!== profile.weapon_id` treatments correctly handle both the empty-string-from-select and null-from-DB cases used across the codebase.

One non-blocking observation (not a defect in this diff, just flagging for awareness): `supabase/functions/get_gang_details.sql`'s `weapon_profiles_deduplicated` CTE joins only on `wp.weapon_id = fe.equipment_id` and doesn't group by `weapon_group_id` at all, so it wouldn't merge grouped ammo profiles under their parent weapon the way `fighter-data.ts`/`gang-data.ts` now do. I couldn't find any live caller of this RPC in the app code, so it looks unused/stale rather than a live regression — worth confirming it's dead before it bit-rots further, but out of scope for this fix.

**Verdict: approve**


---

PR #2018, opened 2026-08-16
Fix gang variant fighters not showing for N26 Malstrain Corrupted

joeseos:
An admin ticked **Gang Variants → Malstrain Corrupted** on an N26 gang and the Malstrain fighters never appeared in Add Fighter.

**Nothing was missed on the data side.** The variant row, its hidden `Variant: Malstrain Corrupted` gang type and its ten fighters were all entered correctly, and `get_fighter_types_with_cost` returns them. The break was in code.

`app/api/fighter-types/route.ts` resolved a variant's fighter pool through `utils/gangVariantMap.ts`, a hand-written table of four variant UUIDs:

```ts
const variantModifier = gangVariantFighterModifiers[variant.id];
if (!variantModifier) continue;   // ← silently skips
```

The N26 variant was never added to that table, so the merge was skipped without a trace — invisible to the 38 N26 gangs that had already selected it. N23's "Aranthian-aligned" was missing from it too, unnoticed only because its gang type has no fighters yet.

## The fix

The pairing is already in the data: a variant's pool lives in the hidden gang type named `'Variant: ' || variant` within the same edition. That holds for all six variants that have one, so the lookup is resolved from the database and the table is deleted. Adding a gang variant no longer needs a code change.

The edition has to be part of the match — "Malstrain Corrupted" exists in both N23 and N26.

Secundan Incursion's "no Leaders" rule isn't derivable from data, so it stays in code as a named set, keyed by variant name so a later edition's re-issue inherits it. That left `utils/gangVariantMap.ts` with nothing to hold, so it's removed.

## Verification

Checked against production data rather than spot-checked. Every variant currently selected by at least one gang:

| Variant | Edition | Gangs | In old map? | Fighters before | Fighters after |
|---|---|--:|---|--:|--:|
| Outlaw | n23 | 5320 | no | 0 | 0 |
| Wasteland | n23 | 1653 | no | 0 | 0 |
| Chaos Corrupted | n23 | 1715 | yes | 1 | 1 |
| Secundan Incursion | n23 | 1130 | yes | 4 | 4 |
| Malstrain Corrupted | n23 | 1083 | yes | 4 | 4 |
| Skirmish | n23 | 892 | no | 0 | 0 |
| Crusading | n23 | 566 | no | 0 | 0 |
| Genestealer Infected | n23 | 527 | yes | 9 | 9 |
| Aranthian-aligned | n23 | 228 | no | 0 | 0 |
| **Malstrain Corrupted** | **n26** | **38** | **no** | **0** | **10** |

The N26 row is the only change across the whole dataset. All four entries of the deleted table resolve to byte-identical gang type UUIDs, so N23 is untouched.

Two cases worth calling out:

- **Aranthian-aligned** now resolves where it previously bailed out. Its gang type is empty so nothing changes today; if fighters are added to it later they'll simply start working.
- **Secundan's Leader filter** now runs before the pool lookup rather than after. Only reachable for Secundan Incursion, which resolves fine, so behaviour is identical.

`tsc --noEmit` clean, `next build` compiles, lint output matches `main` (6 pre-existing errors, none in changed files).

Verified at the data layer via SQL against production — not clicked through in the running app, so worth a quick manual check on an N26 gang before release.

**No migration and no database change ships with this PR.**


joeseos:
## Review

Reviewed the diff (`app/api/fighter-types/route.ts`, `docs/refactor-fighter-type-parents.md`, deletion of `utils/gangVariantMap.ts`) against the rest of the file and the schema.

The core fix is sound: resolving the variant→fighters link from `gang_types` by name+edition instead of a hand-maintained UUID table is the right move, and the verification table against production data (byte-identical UUIDs for the four previously-mapped variants, N26 Malstrain now resolving) is convincing. No issues with the Leader-filter reordering — as noted in the description, it's only reachable for Secundan Incursion, which always resolves.

One finding worth addressing:

**`app/api/fighter-types/route.ts:399-402` — silent skip on query failure (medium)**
```ts
const { data: variantGangTypes } = await supabase
  .from('gang_types')
  .select('gang_type_id, gang_type, edition_id')
  .in('gang_type', gangVariants.map(v => variantGangTypeName(v.variant)));
```
The `error` from this call isn't checked or logged. If it fails (RLS hiccup, network blip, transient Supabase error), `variantGangTypes` is `undefined`, `(variantGangTypes ?? [])` becomes `[]`, every variant fails to resolve a `variantGangTypeId`, and the loop just `continue`s for all of them — fighters silently disappear from Add Fighter with nothing in the logs. That's the same failure shape this PR is fixing (a variant's fighters vanishing with no trace), just triggered by a transient query error instead of a missing map entry. Every other Supabase call in this file (including the `gang_variant_types` lookup two blocks above) checks and logs its error — this one should too, e.g.:
```ts
const { data: variantGangTypes, error: variantGangTypesError } = await supabase...
if (variantGangTypesError) {
  console.error('Error fetching variant gang types:', variantGangTypesError);
}
```
Doesn't need to fail the request (falling through to "no variant fighters" is a reasonable degradation), but it should be visible when it happens.

**Minor / non-blocking:**
- `gt.edition_id === variant.edition_id` (line 410-411): both `gang_types.edition_id` and `gang_variant_types.edition_id` are nullable in the schema. If two rows ever had `edition_id IS NULL`, `null === null` would incorrectly match them. Not an issue with current data (every variant in the verification table has an edition), just a latent trap if a variant is ever entered without one.
- The `gang_type` string match (`'Variant: ' + variant.variant`) is still a naming-convention coupling — rename a variant without renaming its gang type and fighters vanish again, silently (no error to log here, since the query itself succeeds). This is already called out explicitly in `docs/refactor-fighter-type-parents.md`, so it's a known, accepted tradeoff rather than an oversight — just flagging that it's the same fragility class as the bug being fixed, one level up.

Confirmed the deleted `utils/gangVariantMap.ts` has no remaining references anywhere in the codebase.

**Verdict: Approve.** The medium finding above is worth a quick follow-up (either in this PR or immediately after) since it reintroduces a silent-failure path in the exact function this PR was written to make loud, but it doesn't block merging — the happy path is well-verified and the degradation on error is safe (no fighters shown, not wrong fighters shown).

cc @joeseos

---

joeseos:
## Review

@joeseos This is the first review on this PR, so there's nothing prior to reconcile — reviewed both commits (`d89ed7b`, `5457851`) in full.

The core fix is sound: resolving the variant→fighter-pool link from `gang_types` (name + edition) instead of a hand-maintained UUID map correctly closes the N26 gap, and the production-data verification table in the description is a good level of rigor for a change like this.

### New findings

**1. `variant.variant.toLowerCase()` can throw if the DB value is null — medium/low, `app/api/fighter-types/route.ts` ~L406**
```ts
if (variantsWithoutLeaders.has(variant.variant.toLowerCase())) {
```
`gang_variant_types.variant` is nullable in the schema (`variant text`, no `NOT NULL`), but the local type narrows it to `Array<{..., variant: string, ...}>`. If a variant row ever has a null `variant` (bad data entry, admin mistake — there's no dedicated admin UI for this table in the codebase, so it'd be a direct SQL insert), this throws and 500s the whole `/api/fighter-types` route for that gang, which is worse than the old code's behavior (silent skip on an unmapped id). Cheap fix: `(variant.variant ?? '').toLowerCase()`, or filter out variants with a null name when fetching `variantDetails`.

**2. New `gang_types` lookup swallows its error — low, ~L399-403**
```ts
const { data: variantGangTypes } = await supabase
  .from('gang_types')
  .select('gang_type_id, gang_type, edition_id')
  .in('gang_type', gangVariants.map(v => variantGangTypeName(v.variant)));
```
Unlike the two sibling queries just above it in the same function (`gangError`, `variantError`, both logged), this one doesn't check or log its error. If it fails, `variantGangTypeId` resolves to `undefined` for every variant and the merge silently no-ops — which is the exact "skipped without a trace" failure mode this PR is fixing elsewhere. Worth a `console.error` on the destructured error for consistency and so a future regression is at least visible in logs.

**3. Nit (non-blocking): `gt.edition_id === variant.edition_id` would match if both happen to be null.** Very unlikely given `edition_id` looks consistently populated on real rows, just flagging as a theoretical edge case, not asking for a change.

### Other checks
- No leftover references to the deleted `utils/gangVariantMap.ts` anywhere in the tree.
- `gang_types` has an `Allow authenticated users to view gang_types` RLS policy (`USING (true)`), so the new lookup query works fine as a regular authenticated user despite `is_hidden` rows — no RLS gap.
- Doc-only addition (`docs/refactor-fighter-type-parents.md`) is out of scope for correctness review; reads as a reasonable write-up of the follow-on work.

### Verdict: **Approve**, with 1 and 2 as suggested (non-blocking) follow-ups — worth a quick fix before merge if convenient, but they don't undermine the core fix, which is well-verified and correctly scoped.

---

PR #2019, opened 2026-08-17
Carry edition_id when cloning a shared custom gang type

joeseos:
A user copied an N26 gang and got an N23 one back. The gang used a custom gang type.

Copying an *official* gang is fine — `copy-gang.ts` carries `gang_type_id` across and the edition is derived from it. The break is one branch: when the custom gang type belongs to another user, the copy clones the whole asset tree via `utils/duplicate-custom-gang-type.ts` so it survives the owner deleting their originals. That `custom_gang_types` insert enumerated its columns and omitted `edition_id`, so the clone landed null. Gangs have no edition of their own, and a null slug is read as n23 rather than raising — so the copy came back N23, filtered under the N23 toggle on the home page with every N26 capability switched off. `createGang` shares the helper, so building a gang on someone else's shared custom type had the same result.

This is `20260816130000`'s defect one layer over. That migration fixed `copy_custom_collection`, "which dropped edition_id on every clone except custom_fighter_types", and backfilled the 385 rows it had produced. The TypeScript clone has the identical asymmetry for the identical reason — the `custom_fighter_types` insert beside it spreads its row, the gang-type insert enumerated. `custom_skill_types` dropped `edition_id` the same way.

## Changes

- `utils/duplicate-custom-gang-type.ts` — both inserts spread and override, matching the sibling clones, so a column added to either table is carried instead of silently disappearing. Also restores `description`, which the gang-type insert was dropping too.
- `supabase/migrations/20260817120000_backfill_custom_gang_type_edition_from_children.sql` — recovers the parent's edition from the children that kept theirs, reusing rule 2 of the earlier backfill unchanged. `edition_id` stays nullable.

## Verification

`tsc --noEmit` and eslint clean.

One row is affected in production: a copy of an n26 gang type whose fighter types are all n26, with no gang attached — the reporter appears to have deleted the bad copy. Ran the migration's derivation as a plain `SELECT` rather than applying it; it resolves that row to n26. No gang currently resolves to a null edition.

Copying an official gang, or a custom gang type you own yourself, never reached this helper and was never affected.

## Noted, not fixed

- `app/actions/chem-alchemy.ts:56-69` inserts `custom_equipment` with no `edition_id` — the source of all 17 null-edition equipment rows in production. Same bug class, left for a follow-up.
- `gang_types` is unique on `(gang_type_id, edition_id)`, so one `gang_type_id` could exist in both editions and make `gangEditionJoin`'s "first row wins" nondeterministic. Zero such duplicates today.

### DIFF
--- a/supabase/migrations/20260817120000_backfill_custom_gang_type_edition_from_children.sql
+++ b/supabase/migrations/20260817120000_backfill_custom_gang_type_edition_from_children.sql
@@ -1,0 +1,23 @@
+-- Repair custom gang types whose edition_id was dropped by duplicateCustomGangType (fixed
+-- alongside this). Same defect as 20260816130000, one layer over, and recoverable the same way:
+-- the sibling custom_fighter_types clone kept its edition, so rule 2 of that migration -- reused
+-- unchanged below -- reads the parent's edition back off its children.
+
+UPDATE public.custom_gang_types g
+SET edition_id = (
+  SELECT id FROM public.editions
+  WHERE slug = CASE
+    WHEN EXISTS (
+           SELECT 1 FROM public.custom_fighter_types f
+           WHERE f.custom_gang_type_id = g.id AND f.edition_id IS NOT NULL
+         )
+     AND NOT EXISTS (
+           SELECT 1 FROM public.custom_fighter_types f
+           JOIN public.editions e ON e.id = f.edition_id
+           WHERE f.custom_gang_type_id = g.id AND e.slug <> 'n26'
+         )
+    THEN 'n26'
+    ELSE 'n23'
+  END
+)
+WHERE g.edition_id IS NULL;
--- a/utils/duplicate-custom-gang-type.ts
+++ b/utils/duplicate-custom-gang-type.ts
@@ -32,14 +32,14 @@ export async function duplicateCustomGangType(
     throw new Error('Source custom gang type not found');
   }
 
+  // Spread, don't enumerate: a dropped edition_id reads as n23 rather than erroring.
+  const { id: _cgtId, created_at: _cgtCa, updated_at: _cgtUa, user_id: _cgtUid, ...cgtData } = sourceCustomGangType;
+
   const { data: newCgt, error: newCgtError } = await supabase
     .from('custom_gang_types')
     .insert({
+      ...cgtData,
       user_id: targetUserId,
-      gang_type: sourceCustomGangType.gang_type,
-      alignment: sourceCustomGangType.alignment,
-      trading_post_type_id: sourceCustomGangType.trading_post_type_id,
-      default_image_urls: sourceCustomGangType.default_image_urls,
     })
     .select('id')
     .single();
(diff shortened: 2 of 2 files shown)


joeseos:
## Review

Reviewed the diff (`utils/duplicate-custom-gang-type.ts`, new migration `20260817120000_backfill_custom_gang_type_edition_from_children.sql`) against the full function and the schema.

**Code fix is correct.** Traced both callers (`copy-gang.ts`, `create-gang.ts`): `newCustomGangTypeId` from `duplicateCustomGangType` flows straight into the new gang's `custom_gang_type_id`, and edition is derived from that row downstream, so a null `edition_id` on the clone really does produce the wrong-edition gang described. Checked the `custom_gang_types` and `custom_skill_types` schemas — the columns picked up by the new spreads (`edition_id`, `description`, `alignment`, `trading_post_type_id`, `default_image_urls` / `name`) are exactly the ones that should carry over, and the destructured-out fields (`id`, `created_at`, `updated_at`, `user_id`) are exactly the ones that shouldn't. No stray columns (no "shared"/visibility flag) that would leak unintended state into the copy.

**Migration logic:** confirmed it's a verbatim reuse of rule 2 from `20260816130000`, which makes sense — same derivation, same table. The `ELSE 'n23'` branch (no fighter-type children, or a mixed/empty edition signal) matches the documented "null reads as n23" fallback, consistent with the rest of this migration family.

One thing worth double-checking before merge, not necessarily a blocker:

**Backfill scope — `custom_skill_types` (medium, verify then decide)**
The PR description calls out that `custom_skill_types` had the identical dropped-`edition_id` bug in the same function (fixed in this diff via the same spread pattern as the sibling clones). `20260816130000` did a one-time blanket backfill of every null-edition row across all `custom_*` tables as of 2026-08-16, including `custom_skill_types`. But the `duplicateCustomGangType` code bug wasn't fixed until this PR — so any custom gang type copy that included skill types and happened *between* 2026-08-16 and now would have produced new null-`edition_id` `custom_skill_types` rows that the old blanket backfill never saw and this PR's migration doesn't touch (it only backfills `custom_gang_types`). `custom_skill_types.edition_id` is actively read (`app/lib/customise/custom-skills.ts:59`, `app/actions/customise/custom-share.ts:413`) so a stray null there isn't inert — it'd hit the same "reads as n23" path.

The verification section confirms one affected `custom_gang_types` row and derives it correctly, but doesn't mention checking `custom_skill_types` for null-edition rows in that same window. Worth a quick `SELECT count(*) FROM custom_skill_types WHERE edition_id IS NULL` (or scoped to rows created after 2026-08-16) before merging — if it's zero, nothing further is needed; if not, the same rule-2-style backfill (this time keyed off `custom_skills`/`fighter_type_skill_access` → back to the gang type, or more simply "n23 by elimination" per the original migration's rule 4, if these predate n26 skill content) should probably ride along in this migration rather than a follow-up, since the migration is already touching this exact defect class.

Everything else — the `custom_equipment` gap in `chem-alchemy.ts` and the `gang_types (gang_type_id, edition_id)` uniqueness caveat — is explicitly flagged as known/out-of-scope in the description, which is the right call for this PR's size.

**Verdict: Approve**, contingent on confirming the `custom_skill_types` question above (either it's a non-issue and can be merged as-is, or it needs one more backfill statement). Everything else checks out.

cc @joeseos

---

PR #2020, opened 2026-08-17
Added sell logic to n26

joeseos:


joeseos:
## Code Review

@joeseos Reviewed against the full codebase (`sell-confirm-modal.tsx`, `fighter-equipment-list.tsx`, `stash-tab.tsx`, `vehicle-equipment-list.tsx`, `vehicles-tab.tsx`, `types/edition.ts`). `npx tsc --noEmit` passes clean on the branch.

### Findings

**1. `components/fighter/vehicle-equipment-list.tsx:964-979` — missing `editionSlug` wiring (High)**

This is the one `SellConfirmModal` call site that sells credit-cost items (`showD6Roll={!sellModalData.cost_resource_name}`) but doesn't forward `editionSlug`, unlike the two other call sites this PR updates (`fighter-equipment-list.tsx`, `stash-tab.tsx`). The component already receives `editionSlug` as a prop (it's used a few lines above for `EquipmentTooltipTrigger` at line 863, and is passed in from `fighter-page.tsx:734`), so this looks like an oversight rather than an intentional gap.

Effect: for N26 gangs selling vehicle-mounted weapons/wargear, `editionSlug` is `undefined` in the modal, so `offersHalfValue` short-circuits to `false` and the modal falls back to the legacy D6-roll flow — the same item sold from the fighter's main equipment list or the stash correctly offers the half-value checkbox instead. That's an inconsistent, edition-incorrect discount depending on which list the item is sold from.

Fix: add `editionSlug={editionSlug}` to the `SellConfirmModal` call in `vehicle-equipment-list.tsx`, matching the other two sites.

### Other things checked, no issues found
- `offersHalfValue` / `showRoll` mutual exclusivity in `sell-confirm-modal.tsx` is correct for all resolved/unresolved `editionSlug` and N23/N26 cases, including the legacy-row (`editionSlug == null`) fallback to the roll, matching the documented N23 default.
- `halfPrice = Math.max(5, Math.ceil(initialCost / 2))` matches the "floor at 5 credits" behavior described in the new `types/edition.ts` capability doc comment.
- `Checkbox`'s `onCheckedChange={handleHalfValueChange}` (typed `(checked: boolean) => void`) type-checks fine under `strict: true` — Radix declares `onCheckedChange` with method-shorthand syntax, which TS checks bivariantly, so this isn't a real type hole despite diverging from the `checked === true` pattern used everywhere else in the codebase for this prop. Not blocking, but for consistency with the rest of the codebase it'd read more defensively as `(checked) => handleHalfValueChange(checked === true)`.
- Minor UX nit (pre-existing pattern, not introduced by this PR): unchecking "Half value" resets `manualCost` back to `initialCost` unconditionally, discarding any manual edit the user made to the input while the box was checked. Low severity, same class of behavior as the existing roll button.

The vehicle-equipment sell path needs `editionSlug` wired through for the N26 half-value feature to actually apply there — everything else looks solid.

---

PR #2023, opened 2026-08-17
Fixed skill selection in advancement

joeseos:
Selecting a Skill Set in the Advancement modal showed no acquisition type picker and no skill list, leaving no way to complete the advancement. Two separate causes, both N26:

* get_available_skills gated available_acquisition_types on an N23 subtype list, returning [] for every subtype outside it. N26 applies one Advancement table to every subtype and prices the result itself, so the gate only stopped Pets, Gangers, Hangers-on and vehicles from taking any skill at all. Now resolves the gang's edition and skips the N23 gate there.
* The modal also intersected the Skill Set's access level with the selected result's acquisition types, which is empty for valid combinations — a "new Secondary skill" result against a Primary set, and any result against an Allowed or Legendary Name set. The table is an upper bound the player chooses from, not a restriction, so that second filter is dropped; access level still applies, as in N23.

N23 is unaffected: Exotic Beasts and Gangers keep their flat-cost table.

Note: [supabase/functions/get_available_skills.sql](https://github.com/joeseos/mundamanager/blob/fix/fixed-skill-selection-in-advancement/supabase/functions/get_available_skills.sql) is applied manually — the RPC needs deploying for the first fix to take effect.

### DIFF
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -1854,13 +1854,11 @@ export function AdvancementModal({
     );
     const allTypes = sample?.available_acquisition_types ?? [];
     if (allTypes.length === 0) return [];
+    // Access level is the only filter. An N26 result deliberately does not narrow this
+    // further: its table is an upper bound the player picks from, not a restriction.
     const allowedIds = new Set(getAllowedAcquisitionTypeIds(selectedSkillSetAccess, allTypes));
-    // An N26 result names the acquisition types it may award, so the row narrows
-    // the list; it prices them itself, so the N23 per-type prices are hidden.
-    const rowIds = n26SelectedRow?.skillAcquisitionTypeIds;
     return allTypes
       .filter((t) => allowedIds.has(t.type_id))
-      .filter((t) => !rowIds || rowIds.includes(t.type_id))
       .sort((a, b) => a.xp_cost - b.xp_cost)
       .map((t) => {
         const label = isCumulativeXp
@@ -1873,7 +1871,6 @@ export function AdvancementModal({
     selectedCategory,
     availableAdvancements,
     selectedSkillSetAccess,
-    n26SelectedRow,
     isCumulativeXp
   ]);
 
--- a/supabase/functions/get_available_skills.sql
+++ b/supabase/functions/get_available_skills.sql
@@ -14,17 +14,28 @@ DECLARE
     v_fighter_type_id uuid;
     v_custom_fighter_type_id uuid;
     v_origin_skill_type_id uuid;
+    v_edition_id uuid;
+    v_cumulative_xp boolean; -- Edition earns Advancements by rank instead of buying them
 BEGIN
-    -- Get fighter subtypes, gang origin ID, gang ID, fighter type IDs, and verify fighter exists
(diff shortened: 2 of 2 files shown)


joeseos:
## Code Review

@joeseos Reviewed against the full codebase (`components/fighter/fighter-advancement-list.tsx`, `supabase/functions/get_available_skills.sql`, and the sibling `get_fighter_available_advancements.sql` this mirrors). `npx tsc --noEmit` passes clean on the branch.

### Summary
Two independent fixes for the same symptom (no skill picker for a Pet/Ganger/vehicle taking an N26 skill Advancement):

1. **`get_available_skills.sql`** — `available_acquisition_types` was gated on an N23-only subtype allowlist (`Leader, Champion, Juve, Specialist, Crew, Prospect, Brute, Exotic Beast Specialist`), which silently returned `[]` for every subtype outside it (Pets, Gangers, Hangers-on, vehicles). Since N26 prices one table for every subtype, the fix resolves the gang's edition (via the same `COALESCE(gt.edition_id, cgt.edition_id)` + `editions.slug = 'n26'` pattern already used in `get_fighter_available_advancements.sql`, right down to the "resolved once here" comment cross-reference) and lets a cumulative-XP edition skip the subtype gate entirely.
2. **`fighter-advancement-list.tsx`** — the modal additionally intersected the Skill Set access level's allowed types with the selected N26 table row's `skillAcquisitionTypeIds`, which is empty for legitimate combinations (e.g. a "new Secondary skill" result against a Primary-access set, or any result against an Allowed/Legendary set). Per `N26_ADVANCEMENT_TABLE`'s existing doc comment ("a roll sets an upper bound rather than picking a row"), the row only decides *whether* a skill is on offer at all (still used for that at `skillAcquisitionTypeIds?.length` in the "Choose a skill" UI and the skill-only fast path in `selectN26Row`) — access level alone now gates which specific acquisition types are offered.

### Verification
- Confirmed `n26SelectedRow` isn't now dead: still referenced at lines 1908/1920 (default credits) and 2360/2364/2380 (characteristic/skill UI), so dropping it from just this one `useMemo`'s deps is correct, not a leftover.
- Confirmed no orphaned `rowIds` references remain after removal.
- N23 behavior is unchanged: `v_cumulative_xp` is false there, so the original subtype allowlist still applies, and Ganger/Exotic Beast still correctly get `[]` (they use the flat-cost table instead, per `v_uses_flat_cost` in the sibling function).
- `supabase/functions/*.sql` is deployed directly by `.github/workflows/deploy_supabase_functions.yml` on merge — no separate migration needed, consistent with repo convention.

### One thing worth a second look (not blocking)
Dropping the row-level `skillAcquisitionTypeIds` intersection is a rules judgment call, not just a code-correctness one — it changes what's *purchasable* for a given roll, not just what's displayed (e.g. a fighter who rolled "7-8: A new Secondary skill" can now buy a Primary-tier skill from an Allowed-access set too, where before it was blocked entirely). The commit message reasons through this convincingly and it's internally consistent with the table's own "upper bound, not a restriction" comment, but since I can't check it against the N26 rulebook myself, it'd be worth a quick sanity check against the physical rules before merge if that hasn't already happened.

### Verdict: **Approve**
No correctness bugs found in the code itself; the one open question above is a rules-accuracy check rather than an implementation issue.

---

joeseos:
Will be resolved in another PR

PR #2028, opened 2026-08-19
Handle cross-edition assets in sharing and cascading operations

joeseos:
## Summary
This PR adds robust handling for cross-edition assets when sharing custom content and cascading related items. When a user attempts to share content that references assets from a different edition (e.g., a fighter from one edition referencing a skill type from another), those mismatched assets are now gracefully filtered out with user-friendly warnings instead of failing the entire operation.

## Key Changes

- **New utility functions for edition-aware filtering:**
  - `partitionByEdition()`: Separates rows into kept/dropped based on edition compatibility
  - `droppedWarning()`: Generates user-friendly messages about skipped cross-edition assets
  - `friendlyShareError()`: Converts database trigger errors into readable messages

- **Enhanced sharing functions with warning support:**
  - `shareCustomFighter()`, `shareCustomGangType()`, `shareCollection()` now return optional `warning` field
  - Cascaded skill types and fighter types are filtered by edition before sharing
  - Non-fatal errors (e.g., skill sharing failures) are captured as warnings rather than aborting

- **Edition resolution improvements:**
  - `resolveFighterEditionId()`: Fighters now inherit their gang type's edition instead of using the current tab's edition
  - User profile endpoint now includes `edition_slug` on all custom assets via `withEditionSlug()` helper
  - Chem-Alchemy creation validates edition support and uses gang's edition for created elixirs

- **SQL fix for collection copying:**
  - Scoped fighter type cloning to the collection owner to prevent unintended cross-user asset inclusion

- **Edition capability tracking:**
  - Added `chemAlchemy` capability flag (N23 only) with `hasChemAlchemy()` helper

- Cross-edition asset filtering happens at the application level before database inserts, preventing trigger failures
- Warnings are accumulated and joined with periods for multi-step cascading operations
- Edition resolution follows a hierarchy: explicit gang type > parent gang type > edition slug parameter
- The `withEditionSlug()` helper normalizes edition data across all asset types for consistent handling

### DIFF
--- a/app/actions/chem-alchemy.ts
+++ b/app/actions/chem-alchemy.ts
@@ -37,9 +37,8 @@ export async function createChemAlchemy({
     // Get the current user with optimized getClaims()
     const user = await getAuthenticatedUser(supabase);
 
-    // The gang's edition comes with the credits read: an elixir is custom_equipment,
-    // which needs an edition_id, and the gang it is crafted for is the only context
-    // that has one. Gangs have no edition of their own — it derives via the gang type.
+    // Elixirs are custom_equipment and need an edition; the gang is the only context
+    // that has one, derived via its gang type.
     const { data: gangData, error: gangFetchError } = await supabase
       .from('gangs')
       .select(`
@@ -55,8 +54,6 @@ export async function createChemAlchemy({
       throw new Error('Failed to fetch gang information');
     }
 
-    // Gate on the capability, not on the caller: the button is also hidden client-side,
-    // but a gang type id alone does not say which edition it belongs to.
     const gangEdition = gangEditionJoin(gangData);
     if (!hasChemAlchemy(gangEdition?.slug)) {
       throw new Error('Chem-Alchemy is not available for this gang edition');
--- a/app/actions/chem-alchemy.ts
+++ b/app/actions/chem-alchemy.ts
@@ -4,6 +4,7 @@ import { createClient } from "@/utils/supabase/server";
 import { invalidateGangStash, invalidateGangCredits } from '@/utils/cache-tags';
 import { getAuthenticatedUser } from '@/utils/auth';
 import { logEquipmentAction } from '@/app/actions/logs/equipment-logs';
+import { gangEditionJoin, hasChemAlchemy } from '@/types/edition';
 
 interface ChemEffect {
   name: string;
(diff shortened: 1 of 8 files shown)


joeseos:
## Code Review

@joeseos Reviewed the diff (`ef4f0e6`) against the rest of the codebase, including the `custom_shared_set_edition` trigger and the sibling `getUserCustomSkills` helper for cross-referencing. Two real, reachable correctness bugs found (one already flagged inline by the Codex bot — confirmed and detailed below), plus a couple of smaller consistency issues.

### High

**1. Directly-collected items in `shareCollection` skip edition filtering entirely — confirms the existing Codex comment on `app/actions/customise/custom-share.ts:702`**
I independently verified this one: `equipmentIds`, `gangTypeIds`, `fighterTypeIds`, `skillIds`, and `tradingPostIds` are seeded straight from `collection.items` (`custom-share.ts:684-688`) with no call to `partitionByEdition` — only the *cascaded* additions (fighters found via gang types, skills found via skill types) get filtered. I checked `addCollectionItem` (`app/actions/customise/custom-collections.ts:187-197`): it only checks `.eq('user_id', user.id)`, never edition, so a collection containing a cross-edition item is reachable today. When that happens here, the bad row goes straight into the insert batch, the `custom_shared_set_edition` trigger rejects the whole batch — and by then this function has already deleted the collection's existing shares (`custom-share.ts:746-752`), so the collection ends up completely unshared instead of partially shared with the "skipped N assets" warning this PR is meant to provide. The direct sets need the same `partitionByEdition` treatment the cascaded ones get, before the delete/insert.

**2. `edition_slug` for custom skills omits the system-skill-type fallback — `app/api/users/[id]/route.ts:124-165`**
Only `custom_skill_types` is joined with `editions:edition_id (slug)`; the system `skill_types` table isn't, even though it also has `edition_id` in the schema. So every custom skill built on a *system* skill type (a legitimate, common case — see the `skill_type_name` fallback two lines below) gets `edition_slug: null`, which the file's own new comment says "reads as N23" downstream. This is a real divergence from established behavior: `getUserCustomSkills` (`app/lib/customise/custom-skills.ts:79-80`, unrelated to this PR) already does it correctly —
```ts
edition_slug: editionSlugFromJoin(skill.custom_skill_types?.editions)
  ?? editionSlugFromJoin(skill.skill_types?.editions),
```
— and the `custom_shared_set_edition` trigger's own SQL comment says *"Custom wins, matching getUserCustomSkills"*, confirming that's the intended rule this route should follow too. Fix: join `skill_types (name, editions:edition_id (slug))` here as well and fall back the same way.

### Medium

**3. `shareCustomFighter`'s warning can be silently overwritten — `custom-share.ts` ~150-215**
`shareCustomGangType` and `shareCollection` both accumulate into a `warnings: string[]` and `.join('. ')` at the end, but `shareCustomFighter` uses a single `let warning`. If cross-edition skill types get dropped (`warning = droppedWarning(...)`) *and* the subsequent skill-share insert also fails, the second assignment clobbers the first — the user never sees the "skipped N skill types from a different edition" message, only the generic share-failure one. Should use the same array pattern as the other two functions for consistency and to not lose information.

### Low

**4. Ownership-scoping fix in `copy_custom_collection.sql` may be incomplete.** The new `AND cft.user_id = cgt.user_id` guard is applied only to the fighter-types-via-gang-type closure. The structurally identical skill-types-via-`fighter_type_skill_access` closure (a few lines below) has no equivalent check. Worth confirming whether the same "another user's asset attached to your item" leak this PR fixes for fighter types is also reachable for skill types.
5. `droppedWarning`'s leading count (`dropped.length`) can be inconsistent with the shown name list when some dropped rows have a falsy `name` — cosmetic only.
6. The `.map(f => ({ ...f, name: (f as any).fighter_type }))` step (used twice, to satisfy `droppedWarning`'s `{name}` shape) could be dropped by aliasing directly in the select (`name:fighter_type`) instead.

### Verified, not flagged
- `friendlyShareError`'s substring matches (`'from a different edition'`, `'cannot resolve the share edition'`) checked against the actual `RAISE EXCEPTION` messages in `custom_shared_set_edition` — both match.
- `gangEditionJoin`'s alias list correctly covers both `gang_types!gang_type_id` and the `custom_gang_type:custom_gang_types!custom_gang_type_id` alias used in `chem-alchemy.ts`.
- `resolveFighterEditionId`'s `.eq('gang_type_id', ...)` against the `gang_types` table is correct (it has both a bigint `id` and a separate uuid `gang_type_id`, and this matches the convention used throughout the rest of the codebase).

#1 and #2 are reachable bugs that undercut this PR's own stated goal (graceful, non-destructive handling of cross-edition assets), not edge-case nitpicks — worth fixing before merge. #3 is a quick consistency fix. #4-6 are optional follow-ups.

---

PR #2032, opened 2026-08-19
Rename Downtime to Post-Cycle for N26

FlorianWeissDev:
## Why

N26 renames the between-battles phase from **Downtime** to **Post-Cycle**. The rig glitch **Clear all** modal, its success toast and the gang log all still said "Downtime" regardless of edition.

## What changed

Display wording only. The code path behind it is untouched and keeps the N23 naming across every edition — `clearMethod === 'downtime'`, `clearGlitchesDowntimeMutation`, the `clearRigGlitchesDowntime` action and the `rig_glitches_cleared_downtime` action type are all unchanged.

- **`types/edition.ts`** — new `downtimePhase` row in `EDITION_CAPABILITIES` with a `downtimePhaseName` accessor, following the existing `beastSubtype` / `beastSubtypeName` precedent for string-valued rows. No slug comparisons anywhere else.
- **`components/fighter/fighter-injury-list.tsx`** — the method tab and the success toast read the label from the accessor.
- **`app/actions/fighter-injury.ts`, `app/actions/logs/fighter-logs.ts`** — `logFighterAction` takes an optional `edition_slug` and words the description accordingly.
- **`utils/log-types.ts`, `components/log-modal.tsx`** (+ the gang, fighter and campaign call sites) — `getLogTypeLabel` takes an optional edition slug.

## Reviewer notes

- **The unresolved-edition fallback is its own constant**, `UNRESOLVED_EDITION_DOWNTIME_PHASE`, not a reach into `EDITION_CAPABILITIES.downtimePhase.n23`. A future edition missing from the registry must not silently inherit another edition's naming, same reason an unknown slug gets `NO_CAPABILITIES` rather than an alias. Note this differs from `beastSubtypeName` and `initiativeAndMentalCharacteristicSuffix`, which still use the `?? …n23` form; aligning those was left out of scope.
- **Log descriptions are frozen at write time**, so this affects new rows only — existing rows keep their stored wording. The Type column is rendered live and does follow the edition, so an old row will show the new type label next to an old description. That is intentional, not a bug.
- **The log's edition slug is resolved server-side** in `clearRigGlitchesDowntime`, via the gang's gang type, reusing the `credits` lookup it already runs rather than trusting a client-supplied value.
- **`LOG_TYPE_LABELS` is unchanged.** The few labels an edition rewords resolve through a small `EDITION_LABELS` map first; callers that pass no slug behave exactly as before.
- **`LogModal` applies the slug to both the Type column and the action-type filter**, so the two cannot disagree.
- **The commits are separable.** The three consumer commits touch disjoint file sets and each depend only on the first, so any of them can be dropped without breaking the rest.

## Testing

`tsc --noEmit` and `eslint` clean on the touched files. Label resolution asserted at code level across `n23`, `n26`, `null`/`undefined` and an unregistered slug. Smoke-tested locally against seeded Spyrer gangs on both editions: the tab, the toast, the log description and the log Type column all follow the edition.

> [!NOTE]
> Not exercised on staging yet.

### DIFF
--- a/app/actions/fighter-injury.ts
+++ b/app/actions/fighter-injury.ts
@@ -11,6 +11,7 @@ import { revalidateTag } from 'next/cache';
 import type { GangLogActionResult } from './logs/gang-logs';
 import { countsTowardRating, hasKilledStatusFlag } from '@/utils/fighter-status';
 import { getFighterTotalCost } from '@/app/lib/shared/fighter-data';
+import { gangEditionSlug } from '@/types/edition';
 
 export interface AddFighterInjuryParams {
   fighter_id: string;
@@ -577,7 +578,11 @@ export async function clearRigGlitchesDowntime(params: {
 
     const { data: gang, error: gangError } = await supabase
       .from('gangs')
-      .select('credits')
+      .select(`
+        credits,
+        gang_types ( editions:edition_id ( slug ) ),
+        custom_gang_types ( editions:edition_id ( slug ) )
+      `)
       .eq('id', fighter.gang_id)
       .single();
 
@@ -609,6 +614,7 @@ export async function clearRigGlitchesDowntime(params: {
       fighter_name: fighter.fighter_name,
       action_type: 'rig_glitches_cleared_downtime',
       old_value: deletedCount,
+      edition_slug: gangEditionSlug(gang),
       oldCredits: financialResult.oldValues?.credits,
       oldRating:  financialResult.oldValues?.rating,
       oldWealth:  financialResult.oldValues?.wealth,
--- a/app/actions/logs/fighter-logs.ts
+++ b/app/actions/logs/fighter-logs.ts
@@ -3,6 +3,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { createGangLog, GangLogActionResult } from "./gang-logs";
 import { formatFinancialChanges, formatXpBreakdown } from "./log-helpers";
+import { downtimePhaseName } from "@/types/edition";
 
 export interface FighterLogParams {
   gang_id: string;
@@ -26,6 +27,8 @@ export interface FighterLogParams {
   /** OOA/kills values for combined XP+OOA log line */
   old_ooa?: number;
   new_ooa?: number;
+  /** Gang's edition, for descriptions whose wording the edition renames. */
+  edition_slug?: string | null;
   oldCredits?: number;
   oldRating?: number;
   oldWealth?: number;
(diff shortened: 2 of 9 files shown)


joeseos:
## PR Review

@FlorianWeissDev Reviewed against the full codebase (no CLAUDE.md present in this repo). Nice, tightly-scoped PR — the writeup in the description is thorough and the design choices (own `UNRESOLVED_EDITION_DOWNTIME_PHASE` constant, server-resolved edition slug, log descriptions frozen at write time) all check out against the existing `types/edition.ts` conventions (`answerFor`/`can`, the `beastSubtypeName` precedent).

### Verified
- `downtimePhaseName` / `EDITION_CAPABILITIES.downtimePhase` follow the established pattern correctly, including the completeness check via `satisfies Record<...>`.
- All three `<LogModal>` call sites (`gang.tsx`, `fighter-details-card.tsx`, `campaign-page-content.tsx`) now pass `editionSlug`; no call site was missed.
- `getLogTypeLabel`'s new `editionSlug` param is optional and defaults safely; `useMemo` deps in `log-modal.tsx` correctly include `editionSlug`.
- `gangEditionSlug(gang)` in `clearRigGlitchesDowntime` matches the aliases (`gang_types`, `custom_gang_types`) selected in the query — resolves correctly.
- `campaignData.edition_slug` is a real, pre-existing field (confirmed other call sites already use it), so the new prop isn't reading from nothing.

### Findings

**Low — `utils/log-types.ts:59`**
`LOG_TYPE_LABELS['rig_glitches_cleared_downtime']: 'Downtime glitch removal'` is now dead code. `getLogTypeLabel` checks `EDITION_LABELS[actionType]` first, and `EDITION_LABELS` has an unconditional entry for this exact key, so the `LOG_TYPE_LABELS` entry can never be reached. Not a bug (behavior is correct either way), but it's a latent trap — a future editor could "fix" the `LOG_TYPE_LABELS` copy for this key and be confused when nothing changes. Suggest deleting that now-unreachable line, or adding a one-line comment noting `EDITION_LABELS` supersedes it.

### Verdict: **Approve**

No correctness, security, or edge-case issues found. The one finding above is cosmetic and non-blocking.

---

joeseos:
## Follow-up Review (new commit `768c2f9`)

@FlorianWeissDev Reviewed the one new commit since the last pass (`Drop the superseded cleared-glitches log label`). It's a single-line deletion in `utils/log-types.ts`.

### Previous findings

- **Low — dead `rig_glitches_cleared_downtime` entry in `LOG_TYPE_LABELS` (`utils/log-types.ts:59`)** → **Resolved.** The unreachable entry has been deleted. Verified this is safe: `getLogTypeLabel` checks `EDITION_LABELS[actionType]` first, and `EDITION_LABELS` carries an unconditional entry for `rig_glitches_cleared_downtime` (`utils/log-types.ts:121`), so lookups for that key still resolve — now solely through the edition-aware path, which is the intent. The three other references to the action type (`app/actions/fighter-injury.ts:615`, `app/actions/logs/fighter-logs.ts:15,149`) are write-side / switch cases and are unaffected by the label-map change.

### New findings

None. The commit removes only dead code; no bug, regression, edge-case, or convention issue introduced.

### Verdict: **Approve**

The single outstanding finding from the prior review is fixed and no new issues surfaced. Good to merge.

---

PR #2035, opened 2026-08-19
Feature/tactics cards

joeseos:
Adds a **Gang Tactics** section to the gang page's *Battles* tab, wiring up the `tactics_cards` / `gang_tactics_cards` tables that until now existed only in the database.

The section lists the cards a gang holds — D66 range, name, and a notepad tooltip for the user's own description — with the standard edit/delete row actions and an **Add** button opening a checkbox picker over the edition's catalogue. Cards already held show checked and disabled rather than hidden, so the full deck stays visible.

**N26-only.** The catalogue is edition-scoped and only N26 is seeded, so this adds a `gangTacticsCards` row to `EDITION_CAPABILITIES` rather than rendering an empty section for N23. N23 gangs skip the query entirely.

No migration — the tables and RLS already exist.

## Data flow

| | |
|---|---|
| Gang's cards | Server-loaded via `getGangTacticsCards` in `app/lib/shared/gang-data.ts`, cached under a new `BASE_GANG_TACTICS_CARDS` tag. Joins `tactics_cards` for name and D66, the same way `getGangStash` (two functions above it) joins `equipment` and `custom_equipment` for `equipment_name`. |
| Catalogue | Lazy — fetched only when the Add modal opens, via `useQuery` (`enabled: isAddModalOpen`) against `app/api/tactics-cards/route.ts`. |
| Mutations | Server actions in `app/actions/gang-tactics-cards.ts`, guarded by `checkPermissionCached` plus an edition check. The add path re-validates that every posted card id exists in the gang's edition. |

`gang_tactics_cards` stays a pure link table — `gang_id`, `tactics_cards_id`, `description`. Name and D66 are deliberately *not* denormalised onto it, matching `fighter_equipment` / `fighter_skills` / `gang_stash`, which store the FK and join for the name. (The app only snapshots a name where the entity might not exist in any catalogue — `campaign_territories.territory_name`, `campaign_battles.scenario`, `fighter_effects.effect_name` — which isn't the case here, since `tactics_cards_id` is `NOT NULL` with `ON DELETE RESTRICT`.)

List state lives in `gang-page-content.tsx` rather than the section component, because `Tabs` renders only `children[activeTab]` and so unmounts tab bodies on switch.

## Shared component touched: `components/ui/list.tsx`

Flagging this explicitly — it's a two-line change to a shared component inside a feature PR. On the root element of `List`:

```diff
-  className={`mt-6 ${className}`}
+  className={cn('mt-6', className)}
```

- **Why it's needed.** `List` hardcodes `mt-6`, and this section renders inside a `bg-card … p-4` wrapper, so the heading would sit 40px from the card top against 16px for the Battle Sessions card above it. Passing `className="mt-0"` doesn't fix that on its own: Tailwind emits `mt-0` before `mt-6`, and CSS resolves by stylesheet order rather than attribute order, so `mt-6` would win. `cn` uses tailwind-merge, which resolves the conflict properly.
- **Why it's inert for everyone else.** `gang-tactics-cards.tsx` is the only one of 13 `List` callers that passes `className`. For the other twelve it defaults to `""`, so `cn('mt-6', "")` → `"mt-6"` — identical output, minus a trailing space.
- **It also aligns the prop.** `list.tsx` was the only file in `components/ui/` that accepted a `className` prop and appended it rather than merging; the other ten `cn` users there all merge, so `className` now behaves as it does on every other primitive in that folder.

Happy to split this into its own PR if preferred — the only cost is the heading spacing until it lands.

## Verification

`tsc --noEmit` clean. `next build` compiles and passes its TypeScript pass. `eslint` reports 6 errors / 3 warnings, all pre-existing — the identical count appears on a clean tree, and the one in `app/gang/[id]/page.tsx` is the long-standing "JSX within try/catch" rule firing on the `GangPageContent` element, only shifted in line number.

**Not verified at runtime.** The environment this was built in has no Supabase credentials, so `next build` fails at static prerender of `/confirm-email` and `/_not-found` (neither touched here) and the app can't be started. The following still needs a real environment:

1. N26 gang → *Battles* tab → Gang Tactics card renders below Battle Sessions, **Add** on the right.
2. Add → picker lists the 18 seeded N26 tactics in D66 order (11-12 … 65-66); tick two, confirm, rows appear.
3. Reopen Add → those two show checked and disabled.
4. Edit → save a description → notepad icon appears in `Desc.`, hover shows name + text; over 1500 chars disables Save.
5. Delete → confirm → row disappears. Hard-reload → all state persists.
6. Switch tabs and back → list still correct.
7. N23 gang → no Gang Tactics card, and no `/api/tactics-cards` request.
8. Non-owner on an N26 gang → list read-only, Add/Edit/Delete disabled.

https://claude.ai/code/session_01Bgty9QD32KvwTXcRGGjNtV


joeseos:
## PR Review

@joeseos Reviewed against the full codebase (no CLAUDE.md present in this repo), including the `gang_tactics_cards`/`tactics_cards` migration, RLS policies, and the existing edition-capability / permission-check conventions.

### Verified
- **Auth/authorization**: `authoriseGangTactics` correctly chains `getAuthenticatedUser` → `checkPermissionCached(user.id, gangId, gang.user_id)` → `hasGangTacticsCards(gangEditionSlug(gang))`, matching the pattern used elsewhere (e.g. `fighter-skill-access.ts`). `updateGangTacticsCardDescription` / `deleteGangTacticsCard` both scope their mutation by `.eq('id', ...).eq('gang_id', ...)`, so even though RLS also enforces gang ownership/arbitrator status, there's no IDOR path via a guessed `gangTacticsCardId`.
- **Catalogue validation**: `addGangTacticsCards` doesn't trust client-supplied IDs — it re-validates against `tactics_cards` scoped to the gang's resolved `edition_id` before inserting. Since `hasGangTacticsCards` already gates on a resolved edition slug, `editionId` can't practically be null when this code path is reached.
- **Idempotency**: `upsert(..., { onConflict: 'gang_id,tactics_cards_id', ignoreDuplicates: true })` matches the DB unique constraint (`gang_tactics_cards_gang_id_tactics_cards_id_key`) exactly, and the client's Map-merge on the returned rows correctly handles the "already owned" re-add case.
- **Migration/RLS**: `20260819180600_create_tactics_cards.sql` policies mirror the existing owner-or-accepted-campaign-arbitrator pattern used by other gang-scoped tables.
- Char limit (`TACTICS_DESCRIPTION_CHAR_LIMIT`) is enforced both client- and server-side.
- `hasGangTacticsCards` gate correctly skips the extra query in `app/gang/[id]/page.tsx` for N23 gangs (`Promise.resolve([])`), and `compareTacticsCards`/`formatD66Range` handle null D66 values reasonably (unnumbered cards sort last).

### Findings

**Low — `components/gang/gang-tactics-cards.tsx` (edit modal, `isEditOverLimit`)**
`isEditOverLimit` is computed from the raw `editDescription.length`, but `handleSaveDescription` checks the length of the *normalised* (trimmed) description before saving. If a description sits right at the 1500-char boundary with leading/trailing whitespace, the counter/Save-button state can disagree slightly with what actually gets validated server-side. Cosmetic only — the server re-validates and will reject if truly over limit, so nothing incorrect gets persisted.

**Low — scope note, `components/ui/list.tsx`**
The `mt-6 ${className}` → `cn('mt-6', className)` fix is a genuine bug fix (previously a caller's own margin utility would lose the CSS specificity/ordering battle against `mt-6`), but it's a drive-by change to a shared component bundled into a feature PR. Worth calling out in review even though the fix itself is correct and necessary for `<GangTacticsCards className="mt-0">` to work.

**Note (not a finding)** — Tactics cards render inside the "Battles" tab (`gang-page-content.tsx`) rather than a dedicated tab or the main gang info area. Assuming that's an intentional product placement rather than a leftover from where it was easiest to slot in.

### Verdict: **Approve**

Core logic (authz, catalogue validation, cache invalidation, RLS) is sound and follows existing conventions closely. Findings above are cosmetic/non-blocking.

---

PR #2036, opened 2026-08-20
Add support for effects granting skills to fighters

joeseos:
## Summary
This PR adds functionality for fighter effects (such as bionic arms or equipment) to grant skills to fighters automatically. When an effect that grants a skill is applied, the fighter receives that skill at no XP or credit cost. The skill is removed automatically when the effect is removed through cascade delete.

## Key Changes

- **New `grantSkillsForEffects()` function** in `app/actions/equipment.ts`: Core logic that creates fighter_skills rows linked to effects via fighter_effect_skills. Uses service role client for the linking operation since it's derived from user-written data rather than user-directed.

- **Effect skill granting in equipment operations**: Integrated skill granting into:
  - `insertEffectWithModifiers()` - grants skills when effects are created
  - `deleteEquipmentFromFighter()`, `moveEquipmentToStash()`, `sellEquipmentFromFighter()` - revalidates BASE_FIGHTER_SKILLS cache tag when effects are removed
  - `applyEffectsForEquipmentOptimized()` in add-fighter.ts - grants skills during fighter creation

- **Admin UI enhancements** in `admin-fighter-effects.tsx`:
  - Added "Grants Skill (optional)" dropdown for non-skill effect types
  - Displays granted skill in effect badges
  - Fetches available skills from `/api/admin/skills` endpoint
  - Properly separates skill_id meaning: owning skill for skill-category effects vs. granted skill for equipment effects

- **New utility function** `grantedSkillFromEffect()` in `utils/effect-modifiers.ts`: Extracts the skill_id from effect type_specific_data when it represents a granted skill (not an owning skill).

- **Skill precedence logic** in `fighter-data.ts` and `gang-data.ts`: Granted skills never override purchased skills - if a fighter already has a skill they bought, the granted version is ignored.

- **UI label update** in `fighter-skills-list.tsx`: Changed "Lasting Injury" fallback to "Effect" to reflect that skills can now come from effects other than injuries.

## Implementation Details

- Removal requires no code - the cascade from fighter_equipment → fighter_effects → fighter_effect_skills → fighter_skills handles cleanup automatically
- Skill grants are fighter-scoped; vehicle-only effects grant nothing
- The service role client is used only for the fighter_effect_skills linking operation, keeping the table itself restricted
- Cache invalidation ensures skill lists are refreshed when effects are added/removed


joeseos:
## PR Review

@joeseos Reviewed against the full codebase (no CLAUDE.md present in this repo), including the `fighter_skills`/`fighter_effect_skills` schema, RLS policies, `add_fighter_injury.sql` (the pattern this PR says it mirrors), and the various consumers of `fighter_skills` (`get_available_skills`, `fighterAlreadyHasSkill`, the free-skill recount in `fighter-advancement.ts`).

### Verified
- **Service role usage is justified, not an escalation**: `fighter_effect_skills` has no INSERT/UPDATE/DELETE RLS policy for `authenticated` (only SELECT), so a regular client insert would be rejected outright. `createServiceRoleClient` is an established pattern (14+ other files already use it), and both rows it links were just written under the caller's own RLS-checked `supabase` client, so this isn't a new privilege-escalation surface.
- `fighter_skills` INSERT policy (`Users can create skills for their own fighters`) is satisfied by the normal fighter-ownership/arbitrator paths this function is called from — no RLS gap there.
- The `skillCategoryTypeIds` exclusion correctly prevents a `'skills'`-category effect (where `skill_id` means "the skill this effect belongs to") from being misread as a grant of that same skill.
- Cache invalidation additions in `deleteEquipmentFromFighter`/`moveEquipmentToStash`/`sellEquipmentFromFighter` correctly guard on `fighter_id` being non-null (vehicle-only N23 equipment has none; N26 vehicles arrive as a real `fighter_id`).
- `/api/admin/skills` already supports `edition_id` filtering and admin-gates via `checkAdmin`, so the new "Grants Skill" dropdown reuses an existing, already-authorized endpoint correctly.

### Findings

**Medium — `app/actions/equipment.ts`, `grantSkillsForEffects` (~line 118)**
The function always inserts a *new* `fighter_skills` row for a granted skill, with no check for whether the fighter already has that skill_id. This diverges from `add_fighter_injury.sql`, which the code comment explicitly says this mirrors — that SQL function does `INSERT ... WHERE NOT EXISTS (... fighter_id = ... AND skill_id = ...)`, and reuses the existing row's id if one is already there. As written, a fighter who already owns a skill (purchased, or granted by another effect/injury) and then picks up an equipment effect granting the same skill ends up with two `fighter_skills` rows for that `(fighter_id, skill_id)` pair.

The two display paths this PR patches (`getFighterSkills`, `getGangFightersList`) correctly resolve which row wins for display via the new precedence check, so this won't visibly break the Skills list. But other existing consumers of `fighter_skills` that count/check rows without that precedence logic will see the duplicate: `fighterAlreadyHasSkill` (fighter-advancement.ts) and the "remaining skill count" used to recompute a fighter type's `free_skill` eligibility after deleting a skill will both be inflated by the extra granted row. Suggest adding the same `NOT EXISTS`/find-or-reuse check `add_fighter_injury.sql` uses before inserting.

**Medium — `app/actions/equipment.ts`, `grantSkillsForEffects` (~line 160, the final `.update(...)` call)**
```ts
await supabase
  .from('fighter_skills')
  .update({ fighter_effect_skill_id: link.id })
  .eq('id', skill.id);
```
This call's result isn't checked. The two steps before it (`fighter_skills` insert, `fighter_effect_skills` link insert) both have explicit failure handling — including a compensating delete when the link insert fails, per the "without the link the skill has no provenance and no cascade, so it goes" comment right above it. If this final update fails, the same failure mode occurs (the row now has no `fighter_effect_skill_id`), but nothing rolls it back: the granted skill becomes a permanent, provenance-less row on the fighter — it won't cascade-delete when the equipment/effect is removed, and it'll display as an ordinary (indistinguishable from purchased) free skill forever. Suggest checking the error and deleting `skill.id`/the link on failure, same as the branch above it.

**Low — `app/actions/equipment.ts`, `grantSkillsForEffects` (~line 108, `fighter_effect_types` category lookup)**
```ts
const { data: types } = await supabase
  .from('fighter_effect_types')
  .select('id, fighter_effect_categories ( category_name )')
  .in('id', candidates.map(effect => effect.fighter_effect_type_id));
```
No `error` check. If this query fails, `types` is `undefined`, `skillCategoryTypeIds` ends up empty, and every candidate — including `'skills'`-category effects, which should be excluded — proceeds through the grant loop. Low likelihood (this table has an open `SELECT` policy for all authenticated users), but worth a bail-out on error given the categories check exists specifically to prevent that misinterpretation.

### Verdict: **Changes requested** (non-blocking, but worth addressing)

Auth/security model is sound. The two Medium findings are both about the same root cause — the new grant path doesn't fully match the existing injury-grant pattern's idempotency/cleanup guarantees — and are cheap to fix given the surrounding code already has the compensating-delete pattern in place for the other failure branch.

---

joeseos:
## Follow-up PR Review

@joeseos Reviewed the changes since the last review — commit `8e617d1` ("Address PR review on grantSkillsForEffects"), diffed against `6c52580` (the commit the previous review covered). Scope: `app/actions/equipment.ts` and `app/actions/fighter-advancement.ts`.

### Previous findings

**Medium — duplicate `fighter_skills` row when a fighter already has the granted skill** (`grantSkillsForEffects`, insert-always vs. `add_fighter_injury.sql`'s find-or-reuse)
**Partial / addressed by design decision.** The insert-always behavior itself is unchanged, but it's now a documented, deliberate choice: reusing an existing row would mean selling the equipment cascades away a skill the fighter bought with XP, which is worse than a duplicate row. The concrete fallout I'd flagged — `fighterAlreadyHasSkill` and the `free_skill` recount — turns out to only actually break the `free_skill` recount (`fighterAlreadyHasSkill` is a `.limit(1).maybeSingle()` existence check, so duplicate rows don't affect its correctness). That recount is now fixed by excluding effect-granted rows (`is('fighter_effect_skill_id', null)`) in `fighter-advancement.ts` line ~1504. I checked for other `count`-based consumers of `fighter_skills` and found none besides this one. Reasonable resolution.

**Medium — unchecked `.update({ fighter_effect_skill_id })` with no compensating cleanup on failure** (`grantSkillsForEffects`, ~line 160)
**Resolved.** Now checks `backfillError` and, on failure, deletes both the `fighter_skills` row and the `fighter_effect_skills` link, matching the cleanup pattern used in the branch above it.

**Low — unchecked error on the `fighter_effect_types` category lookup** (`grantSkillsForEffects`, ~line 108)
**Resolved.** Now checks `typesError`/`!types` and bails out (`return false`, granting nothing) instead of failing open with an empty exclusion set.

### New findings

None of note. The fix commit is narrowly scoped to the three items above and doesn't touch the admin UI or `add-fighter.ts` paths, which weren't implicated by the prior findings. One very minor, not-worth-blocking-on observation: in the new `backfillError` cleanup branch, deleting the `fighter_skills` row first will already cascade-delete the `fighter_effect_skills` link (per the PR's own description of the cascade), so the subsequent explicit `serviceClient...delete()` on the link is redundant — harmless, just an extra no-op round trip.

### Verdict: **Approve**

All three previously-raised findings are resolved or reasonably addressed with documented rationale. No new correctness, security, or null-handling issues in the delta.

---

joeseos:
## Second Follow-up PR Review

@joeseos Reviewed the changes since the last review — commits `729c462`, `0a66e27`, `604c546` (current head), diffed against `8e617d1` (the commit the previous review covered). Scope: loadout-aware display of equipment-granted skills, plus a comment-trimming pass.

### Previous findings
All three from the original review remain resolved/reasonably addressed (see prior comment) — nothing in this delta touches that logic except comment wording in `grantSkillsForEffects`, which is unchanged functionally.

### New findings

**Low — `components/fighter/fighter-equipment-list.tsx`, line ~26 — exported function declared mid-import-block (readability)**
The new `isEquipmentInActiveLoadout` export sits between two `import` statements:
```ts
import FighterLoadoutsModal from '@/components/fighter/fighter-loadouts-modal';

export function isEquipmentInActiveLoadout(...) { ... }

import { Badge } from '@/components/ui/badge';
```
This is valid ES module syntax (imports don't have to be contiguous) so it compiles fine, but splitting the import block with a function definition hurts scanability. Suggest moving the function below all imports.

### Verified logic
- The new gang-list skill filter (`gang-data.ts` ~line 1608) correctly mirrors the existing effects filter immediately below it: both key off `activeLoadoutEquipmentIds` and only apply when the skill/effect has a `fighter_equipment_id` parent (injuries/purchases, which have none, are always shown) — matches the commit's stated intent to hide on the gang card.
- The fighter-page mute path (`fighter-skills-list.tsx` `isOutOfLoadout` + `List`'s new `rowClassName`) uses the same `isEquipmentInActiveLoadout` helper now shared from `fighter-equipment-list.tsx`, so the two pages can't drift out of sync on what counts as "in loadout."
- `granted_by_equipment_id` is correctly threaded through both `getFighterSkills` and `getGangFightersList` (new `fighter_equipment_id` selected off `fighter_effects`), and both consuming components treat it as optional/undefined-safe (`!!skill.granted_by_equipment_id`, `activeLoadoutId ?? null` defaults).
- The "Trim comments" commit is comment-only — confirmed via diff, no logic touched — and it kept the decision-rationale comments call out in the commit message (service-role justification, backfill rollback, free_skill exclusion), not just line-restating ones.

### Verdict: **Approve**

Only a cosmetic nit on code organization; no functional, security, or null-handling issues in this delta.

---

PR #2040, opened 2026-08-20
Log tactics card rolls, adds and removals to the gang log

joeseos:
Three new action types go through the existing createGangLog, so the log
modal picks them up without any enum to extend: tactics_card_roll,
tactics_card_added and tactics_card_removed.

The roll happens in the browser, so it needs a server hop to record it.
That action takes only the dice and resolves the card itself, rather than
trusting a posted card id. Repeated Roll clicks are held behind the same
2s cooldown the injury roll uses.

Adds log one row per card actually inserted, read back from the upsert so
re-adding an existing card logs nothing. Delete reads the card name before
removing the row, since the client's copy isn't trusted.

### DIFF
--- a/app/actions/gang-tactics-cards.ts
+++ b/app/actions/gang-tactics-cards.ts
@@ -4,6 +4,12 @@ import { createClient } from '@/utils/supabase/server';
 import { getAuthenticatedUser } from '@/utils/auth';
 import { checkPermissionCached } from '@/utils/user-permissions';
 import { invalidateGangTacticsCards } from '@/utils/cache-tags';
+import type { GangLogActionResult } from './logs/gang-logs';
+import {
+  logRolledTacticsCard,
+  logTacticsCardAdded,
+  logTacticsCardRemoved
+} from './logs/gang-tactics-logs';
 import { gangEditionJoin, gangEditionSlug, hasGangTacticsCards } from '@/types/edition';
 import {
   GANG_TACTICS_CARD_SELECT,
@@ -101,7 +107,7 @@ export async function addGangTacticsCards(params: {
     // The browser can post any uuid, so don't trust what the picker sent.
     let catalogueQuery = supabase
       .from('tactics_cards')
-      .select('id')
+      .select('id, name')
       .in('id', tacticsCardIds);
 
     if (auth.context.editionId) {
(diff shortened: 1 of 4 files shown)


joeseos:
## PR Review

@joeseos Reviewed against the full codebase (no CLAUDE.md present in this repo), including `add_fighter_injury`'s cooldown precedent, the `DiceRoller`/`rollD66Outcome` mechanics, RLS on `gang_tactics_cards`, and `createGangLog`'s error-handling contract.

### Verified
- **Server never trusts client-posted card identity.** `verifyAndLogRolledTacticsCard` re-resolves the card from `total` alone (`d66_min <= total <= d66_max`, edition-scoped), and `deleteGangTacticsCard` reads the card name from the DB row before deleting rather than trusting whatever name the client happens to be holding. Both match the PR description exactly.
- **"Log only what was actually inserted" is correctly implemented.** Switching from a separate `.in('tactics_cards_id', ...)` select to `.upsert(...).select(...)` with `ignoreDuplicates: true` means the `RETURNING` clause only contains genuinely-new rows (Postgres `ON CONFLICT DO NOTHING` excludes untouched conflicting rows from `RETURNING`), so `logTacticsCardAdded` is only fired for cards that weren't already owned — matches "re-adding an existing card logs nothing." This also incidentally fixes the add-toast's count to reflect only real additions rather than the full requested selection.
- **Logging is correctly non-blocking without needing extra try/catch.** `createGangLog` swallows its own errors internally and always resolves `{success: false, error}` rather than throwing (confirmed in `gang-logs.ts`), so a log-write failure in `addGangTacticsCards`'s `Promise.all(...)` or the delete path can't turn an already-successful DB mutation into a false "failed to add/remove" error for the user.
- `rollD66Outcome()` produces `total = tens*10+ones` (proper D66 notation, e.g. `[5,2] → 52`), which lines up correctly with the catalogue's `d66_min`/`d66_max` range comparisons — no dice-sum-vs-D66-notation scale mismatch.
- Reuses `authoriseGangTactics` for the new roll-logging action rather than duplicating the permission/edition check — no new auth surface.
- `RollOutcome` type was already imported in `gang-tactics-cards.tsx` from a prior merged PR (the dice roller itself predates this diff), so no missing-import/compile concern.

### Findings

**Low — `app/actions/gang-tactics-cards.ts`, `verifyAndLogRolledTacticsCard` (~line 148)**
```ts
const { data: card, error } = await query.maybeSingle();
```
`.maybeSingle()` throws if more than one row matches. Nothing in the schema prevents two `tactics_cards` rows in the same edition from having overlapping `[d66_min, d66_max]` ranges (only `d66_pair_check`/`d66_order_check` exist — no exclusion constraint on the range itself). Today's seed data is non-overlapping, so this won't fire in practice, but a future admin data-entry mistake (via the catalogue, which has no UI-level overlap validation either) would turn every "Roll D66" click for that gang into a hard failure rather than a graceful "ambiguous card" error. Not this PR's bug to fix, but this is the first code path that turns a range overlap into a thrown error instead of a display-only quirk — worth a `.limit(1)` (or a future exclusion constraint) if that's a realistic data-entry risk.

### Verdict: **Approve**

Small, well-scoped PR. The security-relevant design choices (server-side card resolution, upsert-returning for accurate logging) are correct and consistent with the rest of the codebase's conventions. The one finding above is a minor robustness gap, not a functional bug.

---

PR #2043, opened 2026-08-22
Use fighter_variant alongside fighter_specialisation

joeseos:
PR #2042 added `fighter_variant` to `fighter_types`,
  `custom_fighter_types` and `fighters` and
  backfilled it. Nothing read it. This makes the
  application use it, so a fighter can hold a variant
  (Bonecrusher) and a specialisation (Heavy) at once
  instead of one overwriting the other.

  `fighter_variant` is an additional field, not a
  replacement. The ~80 catalogue rows that were only
  ever variant labels stop being referenced; the 8 real
  specialisations, and the 203 `fighter_types`
  rows that legitimately carry one, are untouched. It is
  generic to both editions — most of those
  variant labels are N23 (Goliath Natborn/Vatborn/Unborn,
  House Agent 434, Warrior (432)).

  ## Telling the two apart

  On `fighter_types` the 134 variant rows and the 203
  real-specialisation rows are perfectly
  disjoint, and where both exist the names always agree.
  So `fighter_variant IS NOT NULL` is an exact,
  edition-agnostic discriminator and no hardcoded id list
  is needed anywhere:

  - `get_fighter_types_with_cost` gates the
  `fighter_specialisations` join on `ft.fighter_variant IS
    NULL`, so `specialisation` now means only the 
  Specialist pick.
  - The write paths derive the same way: a type row's 
  `fighter_specialisation_id` is real exactly when
    the row has no variant.

  There is no equivalent rule for `fighters`, where 62 
  rows hold a stale id that *contradicts* their
  variant (specialisation `Natborn`, variant `Vatborn`), 
  so no name comparison works. Those are
  legacy dirt for the cleanup migration rather than 
  something to gate in code.

  ## Dropdowns re-keyed to fighter type ids

  The variant dropdowns used the catalogue uuid as the 
  option value and skipped any row without a
  `specialisation`. Both assumptions break the moment the 
  cleanup migration nulls
  `fighter_types.fighter_specialisation_id` on variant
  rows — the pickers would have silently gone
  empty. They now key on `fighter_type.id` throughout,
  which is what `FighterAddModal` already did.

  That collapses a lot: a byte-identical duplicate option
  builder in the edit modal, the `''` and
  `"default"` sentinels,
  `currentVariantOrSpecialisation()`, two O(n²) label-noun
  scans, and a
  ~70-line type-resolution block in the submit path.
  Opening a Haunt now preselects Bonecrusher by
  id rather than by matching a label string.

  The edit modal only submits the fighter type when it
  actually changed, so saving an unrelated field
  cannot make the server re-derive and clear a
  specialisation.

  ## Authoring

  Nothing could author a variant, so the next Haunt
  variant an admin created would have gone straight
  back into `fighter_specialisations`. Both admin forms
  now have a **Fighter Variant** text field
  writing `fighter_types.fighter_variant`, and a **Fighter
  Specialisation** picker over existing
  catalogue rows. The picker is a dropdown rather than
  free text, which removes the upsert that minted
  a catalogue row on any new name and the PUT that renamed
  a *shared* row for every fighter type
  pointing at it — including, potentially, one of the 8
  real specialisations. Adding a genuinely new
  specialisation now needs a separate route.

  `POST`/`PUT` on `/api/admin/fighter-types` persist both
  columns, `is_default` means neither variant
  nor specialisation, and the equipment and gang-lineage
  pickers label rows by both.

  ## Display

  `fighter-details-card`, `fighter-card` and `print-gang`
  each inlined the same
  `{type} ({subtypes}), {variant}, {specialisation}` chain
  of JSX ternaries mid-markup. Each now
  builds it in a local variable above the return.
  print-gang keeps its denser separator and gains the
  specialisation it previously dropped.

  ## Not in scope

  The variant is not relevant to promotion or
  advancements, and appears nowhere in those paths.

  Nothing authors `custom_fighter_types.fighter_variant`
  yet — no UI sets it and 0 of 2,553 rows have
  a value — so the code branches that could only ever have
  read `null` are gone, along with a round
  trip that fetched it.

  ## Cleanup migration

  `20260823120000_narrow_fighter_specialisation_id.sql` is
  included as a file and **not applied** —
  run it manually after merge. It re-runs the backfill
  first, because deployed `add-fighter` does not
  yet write `fighter_variant` and production is still
  creating rows that need it (the count grew from
  267 to 293 during this work); nulling the ids before
  repairing them would lose the label. It then
  drops the variant references from `fighters` and
  `fighter_types`.

  Dry-run against production, read-only: 52,925 stale ids
  nulled, 25,840 real specialisations kept,
  **0 fighters lose a variant label**.

  The variant rows in `fighter_specialisations` are left
  in place, along with the now-unused `POST`
  and `PUT` on `/api/admin/fighter-specialisations` — a
  small follow-up once this has settled.

  ## Verified

  `tsc --noEmit` and `npm run build` both clean. The
  `get_fighter_types_with_cost` join change was
  simulated against production data: 0 variant rows leak a
  specialisation, all 203 real ones are kept.
  `get_fighter_types_with_cost` only deploys on merge to
  main, so the dropdown keeps its old label in
  preview until then.


joeseos:
## Code Review

@joeseos Reviewed the full diff (25 files, ~1900 lines) against the surrounding write paths, the two SQL RPCs, and the included-but-not-applied cleanup migration.

### Summary
This makes `fighter_variant` (added inert in #2042) actually drive the app: the `get_fighter_types_with_cost` join now gates on `fighter_variant IS NULL` so `specialisation` means only the Specialist pick, the write paths (`add-fighter`, `edit-fighter`, `copy-fighter`) derive/propagate it consistently, the admin catalog forms get a real variant field plus a specialisation dropdown (replacing the old free-text-upsert flow), and the dropdown/display logic across several components is re-keyed to `fighter_type.id` and collapsed considerably. This is a large, well-reasoned refactor with a genuinely simpler end state — most of the diff is deletion of ad-hoc string-matching logic.

### Findings

**1. Medium — stale `fighter_specialisation_id` semantics left in `fighter-edit-modal.tsx`.**
`components/fighter/edit-fighter/fighter-edit-modal.tsx:590`:
```ts
setSelectedSpecialisationId((fighter.fighter_specialisation as any)?.fighter_specialisation_id || '');
```
Everywhere else in this file, `selectedSpecialisationId` was re-keyed to hold a **`fighter_type.id`** (see the `useState` init at line 212, and every other `setSelectedSpecialisationId` call in the diff). This one line, in the "re-init form when the fighter prop changes" block, still assigns the old specialisation-catalog-id value. In the common case this self-heals: the same render pass also flips `hasExplicitlySelectedType` to `false`, which causes the `fighterTypeInitData` memo/effect a few lines down to immediately recompute and overwrite `selectedSpecialisationId` with the correct `fighter_type.id` before anything paints — so I don't think this is currently visible to users. But it's dead/incorrect code relying on that self-correction happening every time, and it's inconsistent with the rest of this refactor. Worth updating to match (or just dropping the line and letting the effect below own it) so a future edit to the init-effect ordering doesn't resurrect a real bug.

**2. Low — cleanup migration drops the drift guard its prerequisite used.**
`supabase/migrations/20260823120000_narrow_fighter_specialisation_id.sql` reuses the same hardcoded 8-id "real specialisation" allowlist as `20260821100000_add_fighter_variant.sql`, but doesn't repeat that migration's `DO $$ ... RAISE EXCEPTION` guard, which explicitly existed to fail loudly rather than silently misclassify a real specialisation as a variant if the catalog drifted. `POST /api/admin/fighter-specialisations` (which can mint a new catalog row with a fresh id) is left in place per the PR description, and this migration is meant to be applied manually at some point after merge — so there's a window, however narrow, where a specialisation added after merge and before this runs would get its `fighter_specialisation_id` silently nulled by the final `UPDATE ... WHERE fighter_specialisation_id NOT IN (...)` instead of being caught. Given the prior migration treated this exact scenario as worth a hard failure, it'd be consistent to carry the same guard here, or at least note in the migration comment that it assumes no new specialisation was added since the last dry-run.

Nothing else rose to blocking. Some things I specifically checked and found correct:
- `add-fighter.ts` / `edit-fighter.ts` / `copy-fighter.ts`: `fighter_specialisation_id` is nulled exactly when the type row has a variant (`fighterTypeData.fighter_variant ? null : fighterTypeData.fighter_specialisation_id`), matching the "perfectly disjoint" invariant described in the PR body. `fighter_variant` is fetched via `select('*', ...)` in `add-fighter.ts`, so it isn't silently `undefined`.
- `get_fighter_types_with_cost.sql`: the new `LEFT JOIN fighter_specialisations fspec ON fspec.id = ft.fighter_specialisation_id AND ft.fighter_variant IS NULL` correctly suppresses `specialisation` for variant rows without needing a hardcoded id list at query time.
- The migration's two-step backfill-then-null ordering is correct: it fills `fighters.fighter_variant` from either the type row or a non-specialist `fighter_specialisation_id` *before* nulling anything, and only where `fighter_variant IS NULL`, so it won't clobber a value the deployed app already wrote correctly, and the 62 contradicting rows described in the PR body get their stale id dropped without touching their (already correct) variant.
- `fighter-edit-modal.tsx` submit path: `shouldUpdateFighterType` only fires when the type combobox was explicitly touched or the resolved type id differs from the fighter's current one, which does back up the PR's claim that saving an unrelated field can't cause the server to re-derive and clear a specialisation.
- Admin forms (`admin-create-fighter-type.tsx`, `admin-edit-fighter-type.tsx`): the specialisation picker is now a dropdown over existing catalog rows rather than free text, so the old failure mode (renaming a shared specialisation row out from under every other fighter type pointing at it) is gone.
- Display components (`fighter-details-card`, `fighter-card`, `print-gang`, `admin-gang-lineage`) consistently render `variant` and `specialisation` as separate, independently-optional pieces rather than one overwriting the other.

### Checklist
| Area | Status |
|---|---|
| Correctness | ✅ (see note 1 — currently self-healing, not a live bug) |
| Edge cases / null handling | ✅ |
| Security | ✅ (no new surface; admin routes unchanged in auth posture) |
| Performance | ✅ (removes a round trip for unused `custom_fighter_types.fighter_variant`) |
| Readability/maintainability | ✅ (net large simplification) |
| Project conventions | ✅ (matches the guard/comment style of the prerequisite migration — see note 2 for the one place it doesn't) |
| DRY/YAGNI | ✅ (collapses duplicated option-builder/type-resolution logic as described) |

### Verdict: **Approve** (non-blocking notes above worth a look before or shortly after merge, given note 2 is time-sensitive relative to when the migration actually gets run)


---

PR #2044, opened 2026-08-22
updated rpc to be edition aware

Djidiouf:


joeseos:
## Code Review

@Djidiouf Reviewed this against the full codebase (schema, related RPCs, and the canonical edition-resolution logic in `app/lib/shared/gang-data.ts`).

### Summary
Adds an `edition_slug` output column to `get_gang_details(uuid)`, resolved via `gang_types`/`custom_gang_types` → `editions`. Small, targeted, 8-line diff.

### Findings

**None blocking.** A couple of low-severity notes for awareness:

1. **Low — file header says "obsolete, do not update".** `get_gang_details.sql` opens with:
   > "This Code is obsolete. Do not use it or update it. We now use the library `app/lib/shared/gang-data.ts` to get the gang details. This code is kept here for compatibility with the Rule Snatcher Tool."

   Worth flagging since it reads as a hard "don't touch" — but prior migrations (`20260801120000_fighter_multiple_classes.sql`, `20260805120000_rename_fighter_sub_types_to_fighter_specialisations.sql`) explicitly keep this RPC in sync with schema/rename changes, so updating it here is consistent with established practice, not a violation. Might be worth updating that header comment in a follow-up so it doesn't keep looking stale/contradictory to future contributors.

2. **Low — `edition_id` nullability asymmetry (pre-existing, not introduced by this PR).** `gang_types.edition_id` is nullable while `custom_gang_types.edition_id` is `NOT NULL`. Existing rows were backfilled to the current edition by migration `20260630095837`, so `COALESCE(gt.edition_id, cgt.edition_id)` should resolve correctly for existing gangs today. Just noting that a future `gang_types` row inserted without `edition_id` would yield `edition_slug = NULL` here — expected per that migration's documented rollout plan, not a regression from this change.

### Correctness check
- `chk_gang_type_exclusive` guarantees a gang has exactly one of `gang_type_id`/`custom_gang_type_id` set, so the `COALESCE` in the new join is safe (no ambiguity between the two sources).
- Join pattern (`LEFT JOIN editions ed ON ed.id = ...`, trailing `edition_slug text` column) matches the existing precedent in `get_fighter_types_with_cost.sql`.
- Logic is semantically equivalent to the canonical `gangEditionJoin`/`gangEditionSlug` resolution in `types/edition.ts` used by `gang-data.ts`.
- No alias collisions (`ed`, `cgt` are unused elsewhere in the function).
- No security/permission changes; grants (`REVOKE`/`GRANT`) untouched.
- Performance impact negligible — two extra `LEFT JOIN`s on PK/indexed columns in a single-row-result function.

### Checklist
| Area | Status |
|---|---|
| Correctness | ✅ |
| Edge cases / null handling | ✅ (see note 2) |
| Security | ✅ |
| Performance | ✅ |
| Readability/maintainability | ✅ |
| Project conventions | ✅ (see note 1) |
| DRY/YAGNI | ✅ |

### Verdict: **Approve**


---

PR #2050, opened 2026-08-24
Add edition-scoped scenarios and D6 scenario rolling for N26

joeseos:
## Summary
This PR adds edition-specific scenario filtering and introduces a D6 dice roll feature for selecting scenarios in Necromunda 2026 (N26). Scenarios are now tracked with their associated edition, allowing campaigns to display only scenarios relevant to their edition.

## Key Changes

- **Edition-scoped scenarios**: Added `edition_slug` field to `Scenario` type and updated all scenario queries to include and filter by edition
  - Modified scenario API endpoints to scope duplicate checks per edition
  - Updated `getScenariosCached` to fetch and attach edition information via `withEditionSlug`
  - Added `sameEditionForDisplay` helper function to filter scenarios by edition

- **D6 scenario rolling for N26**: Implemented new `scenarioD6Roll` edition capability
  - Added `DiceRoller` component to battle creation modal for editions that support scenario rolling
  - Integrated `rollNd6Outcome` utility to roll a single D6 and match against numbered scenarios
  - Feature is only available in N26 (disabled for N23)

- **Scenario filtering across components**:
  - `CreateBattleModal`: Filters available scenarios by edition before display and rolling
  - `ActiveSession`: Uses `useMemo` to filter scenarios by edition when updating battle session
  - `CampaignBattleLogModal`: Filters scenarios by campaign edition when loading for battle editing
  - `CampaignBattleLogsList` and `CampaignBattleSessions`: Pass `editionSlug` prop through component hierarchy

- **API improvements**:
  - Scenario duplicate validation now scoped to edition (allows same scenario number in different editions)
  - Updated error messages to clarify edition-specific constraints
  - Changed `.single()` to `.maybeSingle()` for more robust query handling

## Implementation Details

- The `withEditionSlug` helper transforms scenario rows with nested edition data into a flat structure with `edition_slug` field
- Scenario filtering uses display-level logic rather than action guards, since all scenario rows are edition-backfilled
- D6 rolling is conditionally rendered based on `hasScenarioD6Roll` capability check
- Cache tag updated from `global-scenarios` to `global-scenarios-with-edition` to reflect new data structure

### DIFF
--- a/app/api/admin/scenarios/route.ts
+++ b/app/api/admin/scenarios/route.ts
@@ -66,16 +66,20 @@ async function _POST(request: Request) {
       );
     }
 
-    // Check for duplicate scenario number
-    const { data: existing } = await supabase
+    // Scenario numbers restart per edition, so the check is scoped to one.
+    let duplicateQuery = supabase
       .from('scenarios')
       .select('id')
-      .eq('scenario_number', numericScenarioNumber)
-      .single();
+      .eq('scenario_number', numericScenarioNumber);
+    duplicateQuery = edition_id
+      ? duplicateQuery.eq('edition_id', edition_id)
+      : duplicateQuery.is('edition_id', null);
+
+    const { data: existing } = await duplicateQuery.maybeSingle();
 
     if (existing) {
       return NextResponse.json(
-        { error: 'A scenario with this number already exists' },
+        { error: 'A scenario with this number already exists for this edition' },
         { status: 409 }
       );
     }
@@ -140,17 +144,21 @@ async function _PATCH(request: Request) {
       );
     }
 
-    // Check for duplicate scenario number (excluding current scenario)
-    const { data: existing } = await supabase
+    // Duplicate check, excluding the current scenario.
+    let duplicateQuery = supabase
       .from('scenarios')
       .select('id')
       .eq('scenario_number', numericScenarioNumber)
-      .neq('id', id)
-      .single();
+      .neq('id', id);
+    duplicateQuery = edition_id
+      ? duplicateQuery.eq('edition_id', edition_id)
+      : duplicateQuery.is('edition_id', null);
+
+    const { data: existing } = await duplicateQuery.maybeSingle();
 
     if (existing) {
       return NextResponse.json(
-        { error: 'A scenario with this number already exists' },
+        { error: 'A scenario with this number already exists for this edition' },
         { status: 409 }
       );
     }
--- a/app/api/campaigns/battles/route.ts
+++ b/app/api/campaigns/battles/route.ts
@@ -1,6 +1,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { NextResponse } from "next/server";
 import { normaliseParticipants, territoryClaimerFor } from "@/utils/battle-participants";
+import { withEditionSlug } from "@/types/edition";
 
 export async function GET(
   request: Request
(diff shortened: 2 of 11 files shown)


joeseos:
## Review

Reviewed the diff against the full call graph for `scenarios` and the `types/edition.ts` capability/display-filter conventions.

### Blocker

**`app/gang/[id]/battle-session/[sessionId]/page.tsx:50-53` — scenarios feeding `ActiveSession` are never edition-tagged (high)**

This is the server component behind `renderBattleSessionPage`, which both `/gang/[id]/battle-session/[sessionId]` *and* `/campaigns/[id]/battle-session/[sessionId]` route through (the campaign page is a one-line re-export of it). Its scenarios query wasn't touched by this PR:

```ts
supabase
  .from('scenarios')
  .select('id, scenario_name, scenario_number')   // no editions:edition_id (slug)
  .order('scenario_number'),
```

That result is passed straight into `<ActiveSession scenarios={scenarios || []} .../>` with no `edition_slug` on any row. `active-session.tsx`'s new filter is:
```ts
const editionScenarios = useMemo(
  () => scenarios.filter((s) => sameEditionForDisplay(s.edition_slug, session.edition_slug)),
  [scenarios, session.edition_slug]
);
```
`sameEditionForDisplay` treats a missing slug as `'n23'` (`(a ?? 'n23') === (b ?? 'n23')`). Every scenario here has `edition_slug === undefined`, so it reads as n23 regardless of the scenario's real edition. For an n23 session that happens to still compare equal, but for an **N26 session** (`session.edition_slug === 'n26'`, which `get-battle-session-data.ts:24` does populate correctly) every scenario now fails the comparison — `editionScenarios` is always `[]`. The Scenario `Combobox` in an active N26 battle session goes empty, and `editionScenarios.find((s) => s.scenario_name === session.scenario)` can no longer find the already-selected scenario either, so an in-progress N26 battle's scenario silently stops resolving/displaying.

This is the same pattern already applied correctly in `app/api/campaigns/battles/route.ts` and `app/lib/reference-data.ts` in this same PR — just needs the same treatment here:
```ts
supabase
  .from('scenarios')
  .select('id, scenario_name, scenario_number, editions:edition_id (slug)')
  .order('scenario_number'),
```
and `(scenarios || []).map(withEditionSlug)` before handing it to `ActiveSession` (import `withEditionSlug` from `@/types/edition`).

### Notes, not blocking

- **`app/actions/campaigns/[id]/battle-logs.ts:575` `getBattleData`** has the same un-tagged scenarios query, but it appears to have no callers left in the tree (grep finds only the definition) — looks like dead code rather than a live bug, worth confirming/removing separately.
- **`app/lib/campaigns/[id]/get-campaign-data.ts:446`** also selects scenarios without edition, but only to resolve names for already-recorded past battles by id (no `sameEditionForDisplay` filtering applied downstream), so it's not exposed to this bug.
- **DiceRoller / D6 matching**: `getRange` maps each scenario to `{min: scenario_number, max: scenario_number}` and `rollNd6Outcome(1)` only ever produces 1–6. That's fine as long as N26's numbered scenarios all fall in 1–6, but `app/api/admin/scenarios/route.ts` only validates `scenario_number >= 1` with no upper bound — if a 7+ numbered N26 scenario is ever added via the admin UI, it becomes silently unreachable by the roller (no error, just never selectable). Not asking for a fix here, just flagging it as a latent trap given `scenarioD6Roll` is edition-gated but the admin form doesn't know about that constraint.
- Everything else — the duplicate-check edition scoping in `app/api/admin/scenarios/route.ts`, the `.single()` → `.maybeSingle()` cleanup, the prop-threading of `editionSlug` through `CampaignBattleSessions`/`CampaignBattleLogsList`/`CampaignBattleLogModal`, and the `Scenario` type consolidation into `types/campaign.ts` — looks correct and consistent with the existing `withEditionSlug`/`sameEditionForDisplay` conventions.

**Verdict: Changes requested.** The `ActiveSession` scenario blank-out for N26 battles is a real regression and should be fixed before merge; everything else in the PR is sound.

cc @joeseos

---

PR #2051, opened 2026-08-24
Show N23/N26 breakdowns on admin stats

Djidiouf:
Wire by-edition gang and campaign counts into the admin stats API and modal so totals and activity windows show edition splits under each count.

### DIFF
--- a/app/api/admin/stats/route.ts
+++ b/app/api/admin/stats/route.ts
@@ -2,8 +2,7 @@ import { NextResponse } from 'next/server'
 import { createClient } from "@/utils/supabase/server";
 import { checkAdmin } from "@/utils/auth";
 import { getUserCount } from '@/app/lib/get-stats-user';
-import { getGangCount } from '@/app/lib/get-stats-gang';
-import { getCampaignCount } from '@/app/lib/get-stats-campaign';
+import { getCampaignCountsByEdition, getGangCountsByEdition } from '@/app/lib/get-stats-by-edition';
 import { getGangActivityStats } from '@/app/lib/get-stats-gang-activity';
 import { getCampaignActivityStats } from '@/app/lib/get-stats-campaign-activity';
 
@@ -21,8 +20,8 @@ export async function GET() {
     const [userCount, gangCount, campaignCount, gangActivity, campaignActivity] =
       await Promise.all([
         getUserCount(),
-        getGangCount(),
-        getCampaignCount(),
+        getGangCountsByEdition(),
+        getCampaignCountsByEdition(),
         getGangActivityStats(),
         getCampaignActivityStats(),
       ]);
@@ -42,4 +41,3 @@ export async function GET() {
     );
   }
 }
-
--- a/app/lib/get-stats-activity.ts
+++ b/app/lib/get-stats-activity.ts
@@ -1,0 +1,40 @@
+import { unstable_cache } from 'next/cache';
+
+import type { ActivityStats, EditionCounts } from '@/types/stats';
+
+const ACTIVITY_WINDOWS_DAYS = [14, 30, 90, 180] as const;
+
+/**
+ * Shared 24h-cached activity windows for admin stats.
+ * Each period is loaded via the provided by-edition fetcher.
+ */
+export function makeActivityStats(
+  fetcher: (since: string) => Promise<EditionCounts | null>,
+  cacheKey: string,
+  tag: string
+): () => Promise<ActivityStats | null> {
+  const getCached = unstable_cache(
+    async () => {
+      if (!process.env.SUPABASE_SERVICE_ROLE_KEY) {
+        return null;
+      }
+
+      const cutoffFor = (days: number) =>
+        new Date(Date.now() - days * 24 * 60 * 60 * 1000).toISOString();
+
+      const [last2Weeks, last1Month, last3Months, last6Months] =
+        await Promise.all(
+          ACTIVITY_WINDOWS_DAYS.map((days) => fetcher(cutoffFor(days)))
+        );
+
+      return { last2Weeks, last1Month, last3Months, last6Months };
+    },
+    [cacheKey],
+    {
+      tags: [tag],
+      revalidate: 86400,
+    }
+  );
+
+  return getCached;
+}
--- a/app/lib/get-stats-by-edition.ts
+++ b/app/lib/get-stats-by-edition.ts
@@ -1,0 +1,216 @@
+import { TAGS } from '@/utils/cache-tags';
+import { unstable_cache } from 'next/cache';
+
+import { createServiceRoleClient } from '@/utils/supabase/server';
+import { EDITION_N26 } from '@/types/edition';
+import type { StatWithEdition } from '@/types/stats';
+
+type ServiceRoleClient = ReturnType<typeof createServiceRoleClient>;
+
+async function getN26EditionId(
+  supabase: ServiceRoleClient
+): Promise<string | null> {
+  const { data, error } = await supabase
+    .from('editions')
+    .select('id')
+    .eq('slug', EDITION_N26)
+    .maybeSingle();
+
+  if (error) {
+    console.error('Error fetching n26 edition id:', error);
+    return null;
+  }
+
+  return data?.id ?? null;
+}
+
+function splitEditionCounts(
+  total: number | null,
+  n26: number | null
+): StatWithEdition {
+  if (total === null || n26 === null) {
+    return { total, n23: null, n26: null };
+  }
+
+  return {
+    total,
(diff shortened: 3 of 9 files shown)


joeseos:
## Review — Show N23/N26 breakdowns on admin stats

@Djidiouf Thanks — the modal wiring and the type split (`StatWithEdition` / `ActivityStatsWithEdition`) are clean, and using the `chk_gang_type_exclusive` invariant to sum official + custom gangs is correct. Two issues need fixing before merge, plus a few worth considering.

---

### High

**1. `app/lib/get-stats-by-edition.ts:10-25` — reimplements an existing cached helper**

`getN26EditionId` duplicates `getEditionIdBySlug` in `app/lib/editions.ts`, which already resolves slug → id off a 1-hour `unstable_cache` of the `editions` table. The version here is uncached and runs once per `fetch*CountsByEdition` call, so a cold `/api/admin/stats` request makes **8 identical round-trips** to `editions` (2 totals + 4 gang windows + ... ) that the shared helper would serve from cache.

```ts
import { getEditionIdBySlug } from '@/app/lib/editions';
import { EDITION_N26 } from '@/types/edition';

const n26EditionId = await getEditionIdBySlug(EDITION_N26);
```

That also drops the `ServiceRoleClient` type alias and the `supabase` parameter threading.

**2. `app/lib/get-stats-by-edition.ts:115-117` and `:149-151` — totals regress to N/A when the edition lookup fails**

When `getN26EditionId` returns `null` (row missing, or the query errored), the whole function returns `null`. The modal then shows `N/A` for the Gangs and Campaigns totals *and* all eight activity windows — even though the total-count query would have succeeded and was the only thing displayed before this PR. A failure to resolve one edition shouldn't take down the numbers that don't depend on it.

`splitEditionCounts` already models exactly this degradation, so let the total run first and pass `null` for the breakdown:

```ts
const n26EditionId = await getEditionIdBySlug(EDITION_N26);
// ... build totalQuery ...
const [{ count, error }, n26] = await Promise.all([
  totalQuery,
  n26EditionId ? countN26Gangs(n26EditionId, since) : Promise.resolve(null),
]);
```

---

### Medium

**3. Query amplification on `/api/admin/stats`**

A cold request goes from ~10 count queries to ~35, and the new ones are more expensive than the old ones: `count: 'exact'` over `gangs` joined `!inner` to `gang_types` / `custom_gang_types`, repeated for each of the four activity windows, none of them cached per-call. The activity results are cached for 24h so this is a cold-path cost, but it will grow linearly with the gangs table.

Worth considering a single grouped query per entity (a `count(*) ... group by edition slug` RPC) instead of one query per edition per window — that would collapse gang activity from 16 queries to 4 and give you every edition in one shot, which also addresses #4.

**4. `app/lib/get-stats-by-edition.ts:27-40` — `n23` derived by subtraction hardcodes a two-edition world**

`n23 = total - n26` means *everything that isn't N26* is reported as N23. That silently absorbs:
- gangs whose `gang_types.edition_id` is `NULL` (the column is nullable),
- campaigns with a `NULL` `campaign_type_id` or a `campaign_types.edition_id` of `NULL`,
- **any future edition** — an N27 would be counted as N23 with nothing failing.

`types/edition.ts` is emphatic on this point ("Adding a slug to `EditionSlug` is one compile error per row, so a new edition answers each question on its own"). Stats are display-only rather than behaviour, so it's a softer case than `EDITION_CAPABILITIES`, but the hardcoded `n23`/`n26` keys in `types/stats.ts:8-11` bake the same assumption into the API contract and the component. Prefer counting per edition and keying off `EditionSlug`:

```ts
export interface StatWithEdition {
  total: number | null;
  byEdition: Partial<Record<EditionSlug, number>> | null;
}
```

with rows that resolve to no edition in their own bucket rather than folded into N23.

**5. `app/lib/get-stats-by-edition.ts:37` — `Math.max(0, …)` hides an impossible state**

If `n26 > total` the data is inconsistent and the clamp turns it into a plausible-looking `0`. That state is reachable if the gang joins ever fan out — `gang_types`' primary key is the composite `(id, gang_type_id)`, and today only the separate `gang_types_gang_type_id_key` UNIQUE keeps one row per `gang_type_id`. Losing that constraint would inflate the N26 count and this code would report a silently wrong split instead of surfacing it. Suggest logging and returning a `null` breakdown when `n26 > total`.

---

### Low

**6. `types/stats.ts:1-6` — `ActivityStats` is now dead.** Nothing in the tree references it (`get-stats-gang-activity.ts` and `get-stats-campaign-activity.ts` were its only consumers). Delete it, and once it's gone the `WithEdition` suffixes are just noise — `ActivityStatsWithEdition` → `ActivityStats`, `StatWithEdition` → `EditionCounts` or similar.

**7. `components/admin/admin-stats-modal.tsx:24-31` — breakdown renders even when unavailable.** When only the split fails you get `N/A (N23)` / `N/A (N26)` stacked under a perfectly good total, which reads as broken. `if (edition?.n23 == null && edition?.n26 == null) return null;` keeps the tile clean. Also the `text-right` breakdown sits under a `text-center` value in the same tile — looks unintentional.

**8. Third copy of the total-count query.** `get-stats-gang.ts` and `get-stats-campaign.ts` still carry their own `count: 'exact'` queries for the public `/api/stats/*` endpoints. Those could now be `(await getGangCountsByEdition())?.total ?? null`, sharing one cache entry instead of two.

**9. `get-stats-gang-activity.ts` and `get-stats-campaign-activity.ts` are now identical** apart from the table/tag names — a small `makeActivityStats(fetcher, tag, key)` factory would remove ~35 duplicated lines. Optional; only worth it if a third entity is likely.

---

### Verdict: **changes requested**

Blocking on #1 and #2 — both are small edits. #3–#5 are worth a decision before this becomes the shape other stats copy; #6–#9 are cleanup.

---

joeseos:
@Djidiouf Reviewed the new commit (`9d6d0ed`, since the last review on `1dc58a6`).

### Previous findings

| # | Finding | Status |
|---|---|---|
| 1 | Reimplemented `getEditionIdBySlug` uncached | **Resolved** — now imports and calls the shared cached helper via `resolveEditionId`, wrapped in try/catch since `getEditionIdBySlug` throws on error. |
| 2 | Totals regress to N/A when edition lookup fails | **Resolved** — `fetchGangCountsByEdition`/`fetchCampaignCountsByEdition` now run the total query unconditionally; `buildEditionCounts` only suppresses the breakdown, not the total. |
| 3 | Query amplification (single grouped query per entity) | **Outstanding**, as expected — not blocking, flagged as "worth considering." Note the fix for #4 below roughly doubles the windowed query count (both editions are now queried explicitly instead of one + subtraction), which matches the "16 queries" estimate from the original comment. |
| 4 | `n23` derived by subtraction hardcodes a two-edition world | **Resolved** — counts are now gathered per-`EditionSlug` into `byEdition: Partial<Record<EditionSlug, number>>`, with `types/stats.ts` and the component updated to match. No more silent absorption of unrecognized/future editions into N23. |
| 5 | `Math.max(0, …)` hides an impossible state | **Resolved** — `buildEditionCounts` now compares `counted > total`, logs an error, and returns `byEdition: null` instead of clamping. |
| 6 | Dead `ActivityStats` type / noisy `WithEdition` suffixes | **Resolved** — old `ActivityStats`/`StatWithEdition` removed, new `EditionCounts`/`ActivityStats` naming in place. |
| 7 | Breakdown renders even when unavailable | **Resolved** — `EditionBreakdown` now returns `null` when both `byEdition[N23]` and `byEdition[N26]` are missing. The `text-right`/`text-center` alignment nit mentioned alongside it wasn't touched, but that was cosmetic and optional. |
| 8 | Third copy of the total-count query | **Resolved** — `getGangCount`/`getCampaignCount` now delegate to `getGangCountsByEdition()?.total`/`getCampaignCountsByEdition()?.total`, sharing the same cache entry and tag (`TAGS.globalGangCount()` / `TAGS.globalCampaignCount()` are still wired to the shared cache, so existing revalidation call sites keep working). |
| 9 | Duplicated activity-window logic | **Resolved** — extracted into `app/lib/get-stats-activity.ts`'s `makeActivityStats` factory, used by both `get-stats-gang-activity.ts` and `get-stats-campaign-activity.ts`. |

### New findings

None. Checked for stale references to the removed identifiers (`StatWithEdition`, `ActivityStatsWithEdition`, `splitEditionCounts`, `getN26EditionId`, `countN26*`) — none remain anywhere in the tree. The `/api/admin/stats` route and its consumers still line up with the new shapes. Edge cases (partial edition-lookup failure, total vs. breakdown independence, `counted > total`) are handled sensibly.

### Verdict: **approve**

Nice cleanup — this is a solid iteration on the by-edition modeling and it generalizes correctly beyond N23/N26.

---

PR #2053, opened 2026-08-24
fix(admin): use shared Modal component for equipment adjusted cost dialog

v0rp-z:
## Summary

The **Add Equipment Adjusted Cost** dialog in the admin *Edit Fighter Type* form is not using the shared [`components/ui/modal.tsx`](components/ui/modal.tsx) `Modal` component that the rest of the app uses. 

This PR swaps it for the shared component.

## Changes

- `components/admin/admin-edit-fighter-type.tsx`: replaced the dialog markup with `<Modal>`.
- State reset for the three form fields (`showAdjustedCostDialog`, `selectedAdjustedCostEquipment`, `adjustedCostAmount`) is now centralised in a single `onClose` handler

## Test plan

  - Save adds the chip, closes the modal, resets the form
  
## Screengrabs
**Old:** 
Screenshot From 2026-08-24 23-13-16

**New:** 
Screenshot From 2026-08-24 23-07-23 Screenshot From 2026-08-24 23-07-06 Screenshot From 2026-08-24 23-07-13

### DIFF
--- a/components/admin/admin-edit-fighter-type.tsx
+++ b/components/admin/admin-edit-fighter-type.tsx
@@ -2364,95 +2364,65 @@ export function AdminEditFighterTypeModal({ onClose, onSubmit }: AdminEditFighte
                 )}
 
                 {showAdjustedCostDialog && (
-                  <div
-                    className="fixed inset-0 bg-black/50 dark:bg-neutral-700/50 flex items-center justify-center z-50"
-                    onClick={(e) => {
-                      if (e.target === e.currentTarget) {
-                        setShowAdjustedCostDialog(false);
-                        setSelectedAdjustedCostEquipment("");
-                        setAdjustedCostAmount("");
-                      }
+                  <Modal
+                    title="Equipment Adjusted Cost Menu"
+                    helper="Select equipment and enter an adjusted cost"
+                    width="sm"
+                    onClose={() => {
+                      setShowAdjustedCostDialog(false);
+                      setSelectedAdjustedCostEquipment("");
+                      setAdjustedCostAmount("");
                     }}
+                    onConfirm={() => {
+                      setEquipmentDiscounts(prev => [
+                        ...prev,
+                        {
+                          equipment_id: selectedAdjustedCostEquipment,
+                          adjusted_cost: parseInt(adjustedCostAmount),
+                        },
+                      ]);
+                    }}
+                    confirmText="Save Adjusted Cost"
+                    confirmDisabled={!selectedAdjustedCostEquipment || !adjustedCostAmount || parseInt(adjustedCostAmount) < 0}
                   >
-                    <div className="bg-card p-6 rounded-lg shadow-lg w-[400px]">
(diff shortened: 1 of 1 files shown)


joeseos:
## Code Review

@v0rp-z Reviewed against `components/ui/modal.tsx` and the sibling "Gang-Specific Cost" dialog in the same file.

### Summary
Swaps a hand-rolled backdrop/dialog for the shared `Modal` component in the admin Edit Fighter Type form's "Equipment Adjusted Cost" dialog. Clean, well-scoped, single-file change.

### Findings

None blocking, one trivial observation:

- `onConfirm` no longer re-checks `adjusted_cost >= 0` before pushing to `equipmentDiscounts` (the old code did, in addition to disabling the button) — it now relies solely on `confirmDisabled` to gate the click. Since `Modal`'s `handleConfirm` only invokes `onConfirm` when the button isn't `disabled`, and the native `disabled` attribute blocks the click entirely, this is safe in practice. The only way to hit a `NaN`/negative value here would be forcing a value into the `type="number"` input past what a browser normally allows (which typically just resets to `""` and is already caught by `!adjustedCostAmount`), so this isn't a real regression — just noting it since it's a slight behavioral narrowing from "validate twice" to "validate once, trust the disabled state."

### Correctness check
- `Modal`'s `handleConfirm` calls `onClose()` automatically unless `onConfirm` explicitly returns `false`; since the new `onConfirm` is a plain void callback, closing + resetting via the centralized `onClose` handler happens exactly as described (confirmed by tracing `components/ui/modal.tsx`).
- `confirmDisabled={!selectedAdjustedCostEquipment || !adjustedCostAmount || parseInt(adjustedCostAmount) < 0}` is equivalent to the old inline `if (selectedAdjustedCostEquipment && adjustedCostAmount) { if (adjusted_cost >= 0) {...} }` guard, just moved from the click handler to the button's disabled state — same validation, matches how the adjacent Gang-Specific Cost dialog (`admin-edit-fighter-type.tsx:2515-2529`) already does it.
- Equipment `<select>`, cost `<Input>` (including the existing keydown guard against typing `-`), and the "already has an adjusted cost" filter on the equipment list are all preserved verbatim — no behavior lost in the port.
- This is a direct, near-identical match of an existing pattern already used elsewhere in the same file, so it's a net consistency win, not a new pattern introduced.
- Pure UI refactor — no server actions, RPCs, or data model touched. No security or performance surface.

### Checklist
| Area | Status |
|---|---|
| Correctness | ✅ |
| Edge cases / null handling | ✅ (see trivial note above) |
| Security | ✅ (not applicable — client-only dialog) |
| Performance | ✅ (not applicable) |
| Readability/maintainability | ✅ (net simplification, removes duplicated dialog chrome) |
| Project conventions | ✅ (matches the existing `Modal` usage pattern in the same file) |
| DRY/YAGNI | ✅ |

### Verdict: **Approve**


---

v0rp-z:
Superseded by https://github.com/joeseos/mundamanager/pull/2059 — reopened from the main repo now that I have push access.

PR #2054, opened 2026-08-24
fix(admin): use shared Modal component for equipment adjusted cost dialog (create)

v0rp-z:
## Summary

Follow-up to #2053. The **Add Equipment Adjusted Cost** dialog in the admin *Create Fighter Type* form had the same old modal that #2053 fixed for the edit form. This PR applies the same refactor here.

## Changes

- `components/admin/admin-create-fighter-type.tsx`: replaced the old dialog with `<Modal>`
- State reset for the three form fields (`showAdjustedCostDialog`, `selectedAdjustedCostEquipment`, `adjustedCostAmount`) is centralised in a single `onClose` handler

## Test plan

  - Save adds the chip, closes the modal, and resets the form

### DIFF
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -13,6 +13,7 @@ import { hasAlignment, hasSaveCharacteristic, allowsMultipleSubtypes, hasStartin
 import { toggleFighterSubtype } from '@/utils/fighter-subtype-picker';
 import { getSkillSetRank } from "@/utils/skillSetRank";
 import { compareEquipmentCategories } from "@/utils/getEquipmentCategoryRank";
+import Modal from '@/components/ui/modal';
 
 interface AdminCreateFighterTypeModalProps {
   onClose: () => void;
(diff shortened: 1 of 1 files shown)


joeseos:
## Review — use shared Modal for the Create Fighter Type adjusted-cost dialog

@v0rp-z Verified this against `components/ui/modal.tsx`: `Modal.handleConfirm` calls `onConfirm()` then `onClose()` unless the result is explicitly `false`, and your `onConfirm` returns `undefined`, so the "Save closes + resets the form" behaviour is preserved correctly. The nested-modal z-index (`z-[100]` inner vs `z-50` outer) and `useScrollLock`'s ref-counted lock/unlock both handle this being a modal-within-a-modal without issue. Diff is otherwise a faithful, minimal port of the old markup onto the shared component — no logic touched beyond the dialog chrome.

### Checklist

- **Correctness** — Behavior-equivalent to the old dialog (fields, filtering, disabled state, close-and-reset). No regression found.
- **Edge cases / null handling** — One small one, see low note below.
- **Security** — N/A, no new data flow.
- **Performance** — N/A.
- **Readability/maintainability** — Improves it; removes ~30 duplicated lines of modal chrome and centralizes reset into `onClose`.
- **Project conventions** — Matches the shared `Modal` usage pattern and mirrors sibling PR #2053's approach 1:1, so the two forms stay consistent with each other.
- **DRY/YAGNI** — Net reduction in duplicated markup. One minor internal duplication noted below.

### Notes (non-blocking, low severity)

1. **`parseInt(adjustedCostAmount) < 0` guard is now the only defense against a bad value.** The old code re-checked `adjusted_cost >= 0` inside the click handler before pushing to `equipmentDiscounts`, in addition to disabling the button. That inline recheck is gone — `onConfirm` now pushes whatever `parseInt(adjustedCostAmount)` produces unconditionally. In practice `confirmDisabled` already prevents the click for negative/empty values, so this isn't reachable through normal UI interaction, but it's slightly less defensive than before. Not blocking.
2. **`parseInt(adjustedCostAmount)` is computed twice** (once in `confirmDisabled`, once in `onConfirm`). Could hoist to a local `const adjustedCost = parseInt(adjustedCostAmount)` above the `Modal` and reuse it in both places. Purely cosmetic, and the same pattern already exists in #2053.
3. PR description says #2053 "fixed" the edit-form dialog — that PR is still open, not merged. Doesn't matter functionally since this PR only touches `admin-create-fighter-type.tsx` and doesn't depend on #2053 landing first, but worth updating the description language.

### Verdict: **approve**

---

v0rp-z:
Superseded by https://github.com/joeseos/mundamanager/pull/2060 — reopened from the main repo now that I have push access.

PR #2056, opened 2026-08-25
Show an advancement icon beside XP on N26 gang fighter cards

Djidiouf:
Surfaces open rank Advancements next to the XP value with TbArrowBigUpFilled and a react-tooltip, using the same openAdvancementsFor derivation as the fighter Advancements list.

### DIFF
--- a/app/lib/fighter-advancements.ts
+++ b/app/lib/fighter-advancements.ts
@@ -1,22 +1,56 @@
 import { getGangFightersBundle } from '@/app/lib/shared/gang-data';
+import { countAdvancementsTaken } from '@/utils/advancementRanks';
 
 /**
  * Get gang fighters (id/name/status columns) — selector over the shared
  * gang fighters bundle, so it reads the same cache entry as the gang page
  * instead of maintaining a duplicate copy of the gang's fighter list.
+ *
+ * Also carries starting_xp and advancements_taken so the fighter-page
+ * Combobox can derive open Advancements without a second fetch.
  */
 export const getGangFighters = async (gangId: string, supabase: any) => {
   const bundle = await getGangFightersBundle(gangId, supabase);
+
+  const advancementEffectsByFighter = new Map<string, unknown[]>();
+  for (const effect of bundle.effects) {
+    if (!effect.fighter_id) continue;
+    const category =
+      effect.fighter_effect_type?.fighter_effect_category?.category_name;
+    if (category !== 'advancements') continue;
+
+    const list = advancementEffectsByFighter.get(effect.fighter_id);
+    if (list) {
+      list.push(effect);
+    } else {
+      advancementEffectsByFighter.set(effect.fighter_id, [effect]);
+    }
+  }
+
+  const skillsByFighter = new Map<string, Record<string, { is_advance?: boolean }>>();
+  for (const skill of bundle.skills) {
+    if (!skill.fighter_id) continue;
+    const key = skill.id ?? `${skill.fighter_id}-${skill.created_at}`;
+    const existing = skillsByFighter.get(skill.fighter_id) ?? {};
+    existing[key] = { is_advance: skill.is_advance || false };
+    skillsByFighter.set(skill.fighter_id, existing);
+  }
+
   return bundle.fighters.map((f: any) => ({
     id: f.id,
     fighter_name: f.fighter_name,
     fighter_type: f.fighter_type,
     xp: f.xp,
+    starting_xp: f.starting_xp ?? null,
+    advancements_taken: countAdvancementsTaken(
+      { advancements: advancementEffectsByFighter.get(f.id) ?? [] },
+      skillsByFighter.get(f.id) ?? {},
+    ),
     killed: f.killed,
     retired: f.retired,
     enslaved: f.enslaved,
     starved: f.starved,
     recovery: f.recovery,
-    captured: f.captured
+    captured: f.captured,
   }));
 };
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -9,7 +9,7 @@ import { TypeSpecificData } from '@/types/fighter-effect';
 import { createClient } from '@/utils/supabase/client';
 import { getSkillSetRank } from "@/utils/skillSetRank";
 import { characteristicRank } from "@/utils/characteristicRank";
-import { openAdvancementsFor } from "@/utils/advancementRanks";
+import { countAdvancementsTaken, openAdvancementsFor } from "@/utils/advancementRanks";
 import { List } from "@/components/ui/list";
 import { UserPermissions } from '@/types/user-permissions';
 import { useMutation, useQuery } from '@tanstack/react-query';
@@ -3374,9 +3374,9 @@ export function AdvancementsList({
       });
   }, [advancements, advancementSkills]);
 
-  const advancementCount = advancements.length + advancementSkills.length;
+  // Taken count: characteristic effects plus is_advance skills (shared with gang cards).
+  const advancementCount = countAdvancementsTaken({ advancements }, skills);
 
-  // advancementCount is the taken count: effects plus is_advance skills.
   const openAdvancements = openAdvancementsFor(
     editionSlug,
     fighterStartingXp,
--- a/components/fighter/fighter-page.tsx
+++ b/components/fighter/fighter-page.tsx
@@ -691,6 +691,7 @@ export default function FighterPage({
               onValueChange={(value) => router.push(`/fighter/${value}`)}
               placeholder="Select a fighter..."
               className="w-full"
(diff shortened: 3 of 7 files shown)


joeseos:
@Djidiouf Reviewed this against the codebase. Nice, focused change — logic tracks with the existing `openAdvancementsFor` derivation and `react-tooltip` is already a project dependency, so no new dependency risk.

## Findings

**1. [Medium] Duplicate "advancements taken" counting logic**
`utils/advancementRanks.ts:158-166` (new `countAdvancementsTaken`) duplicates logic that already exists inline in `components/fighter/fighter-advancement-list.tsx:3307-3309,3377`:
```ts
const advancementCount = advancements.length + advancementSkills.length;
```
where `advancements` is `effects.advancements` (see `components/fighter/fighter-page.tsx:903`) and `advancementSkills` filters `skills` by `is_advance` — i.e. the exact same rule the new function encodes. Now there are two independent implementations of "how to count taken advancements" that can silently drift (e.g. if a future edition adds another advancement source). Since you already extracted this into a shared utility, consider having `fighter-advancement-list.tsx` call `countAdvancementsTaken(effects, skills)` too (or as quick follow-up) instead of keeping its own inline version.

**2. [Low] `openAdvancements` isn't memoized**
`components/gang/fighter-card.tsx:463-468` computes `openAdvancements` directly in the render body, while the sibling `stats` derivation just above it (line ~397) uses `useMemo`. `FighterCard` has local state (`isMultiline`, `compactNaturalHeight`) that triggers re-renders unrelated to fighter data, so this recomputes `countAdvancementsTaken`/`openAdvancementsFor` (including an `Object.values(skills).filter(...)` pass) on every such re-render. Cheap in practice, but worth wrapping in `useMemo` for consistency with the existing pattern in this file and to avoid the redundant work.

Nothing else stood out — null/undefined handling for `effects`/`skills`/`starting_xp` is correct (`countAdvancementsTaken` and `openAdvancementsFor` both degrade to 0 safely), the icon is correctly gated to editions with an XP ladder without a hardcoded edition check, and the `print:hidden` tooltip / visible icon split for print view looks intentional.

## Verdict
**Approve** — no correctness, security, or edge-case bugs found. The two notes above are non-blocking cleanup/consistency suggestions, not blockers.

---

joeseos:
@Djidiouf Reviewed the latest commit (`3e27cac` — the Combobox change) against the previous review.

## Previous findings
1. **[Medium] Duplicate "advancements taken" counting logic** — **Resolved.** `fighter-advancement-list.tsx` now calls `countAdvancementsTaken({ advancements }, skills)` instead of the inline `advancements.length + advancementSkills.length`.
2. **[Low] `openAdvancements` not memoized** — **Resolved.** `fighter-card.tsx` now wraps the derivation in `useMemo`.

## New findings

**1. [Medium] N26 fighter switcher loses the XP figure entirely when no Advancement is open**
`components/fighter/fighter-page.tsx:641-649`:
```ts
const xpSuffix =
  !isCumulativeXp && f.xp !== undefined && f.xp !== null
    ? ` (${f.xp} XP)`
    : '';
```
`xpSuffix` is suppressed for *every* cumulative-XP (N26) fighter, independent of `showAdvancementIcon`. So once a fighter has no open Advancement, the dropdown shows neither the icon nor any XP number — the XP figure just disappears, whereas before this PR it always showed `(X XP)`. That contradicts the commit message ("Replaces (XP) with `TbArrowBigUpFilled` **when a fighter has open rank Advancements**") and is inconsistent with the pattern this same PR establishes in `fighter-card-stats-table.tsx`, where the numeric XP value is always shown and the icon is only appended on top. Suggest keeping `(X XP)` for N26 fighters too, and only swapping/augmenting when `showAdvancementIcon` is true.

**2. [Medium] `combobox.tsx` closed-state rendering change has app-wide blast radius**
`components/ui/combobox.tsx:361` changed:
```diff
- {selectedOption.displayValue || selectedOption.label}
+ {selectedOption.label}
```
This is the shared `Combobox` used in ~20 other call sites with `ReactNode` labels (gang alliance selector, `FighterAddModal`, `stash-tab`, `vehicle-edit`, etc.), not just the fighter switcher this PR targets. Several of those intentionally keep decorations out of the closed-state summary — e.g. `utils/gang-combobox-option.tsx`'s `labelSuffix` (the "Already assigned" badge), whose own doc comment says it "does not affect displayValue" precisely so the badge doesn't show once selected. With this change, `label` (including any badge/suffix) is now always rendered when the combobox is closed everywhere in the app, not just here. This might be fine, but it's a scope-widening behavior change on a shared component that isn't mentioned in the PR description and wasn't visually verified across those other consumers — worth a deliberate pass (or scoping the fix to just the fighter Combobox) rather than a blanket component change.

Everything else — the `getGangFighters` enrichment (effect/skill shapes match existing `fighter_effect_type.fighter_effect_category.category_name` and `is_advance` usage elsewhere in the codebase), null handling in `openAdvancementsFor`/`countAdvancementsTaken`, and the `isCumulativeXp` edition gating — looks correct.

## Verdict
**Changes requested** — finding #1 is a real, user-visible regression for N26 fighters without open Advancements; finding #2 warrants at least a quick visual check of the other Combobox consumers (or narrowing the fix to the fighter switcher) before merge.

---

joeseos:
@Djidiouf Reviewed the latest commit (`007e489`) against the previous review.

## Previous findings
1. **[Medium] N26 fighter switcher loses the XP figure entirely when no Advancement is open** (`fighter-page.tsx:641-649`) — **Outstanding.** The `xpSuffix` logic is unchanged in this commit: it's still gated on `!isCumulativeXp`, so for N26 fighters `xpSuffix` is always `''` regardless of `showAdvancementIcon`. A cumulative-XP fighter with no open Advancement still shows neither the icon nor `(X XP)` in the switcher — the original regression described in the last review still applies.
2. **[Medium] `combobox.tsx` closed-state rendering change has app-wide blast radius** — **Resolved.** The unconditional `{selectedOption.label}` change has been replaced with an opt-in `showLabelWhenClosed` prop (`components/ui/combobox.tsx:52-58,91,369-371`), defaulting to `false`. That default reproduces `main`'s original `displayValue || label` behavior exactly, so all ~20 other `Combobox` consumers (gang alliance selector, `FighterAddModal`, `stash-tab`, `vehicle-edit`, etc.) are unaffected. Only the fighter switcher in `fighter-page.tsx:694` opts in via `showLabelWhenClosed`. Well-scoped fix, clearly documented with a doc comment explaining the tradeoff.

## New findings
None. The diff since the last review is minimal and self-contained (2 files, +13/-2): the new `showLabelWhenClosed` prop and its single call site. No new bugs, edge cases, or readability issues introduced.

## Verdict
**Changes requested** — finding #1 from the previous review (XP figure silently disappearing for N26 fighters without open Advancements) is still unresolved and is a user-visible regression versus current `main` behavior. Finding #2 is fully addressed.

---

joeseos:
@Djidiouf Reviewed the latest commit (`92cb280`) against the previous review.

## Previous findings
1. **[Medium] Duplicate "advancements taken" counting logic** — **Resolved** (unchanged since last review).
2. **[Low] `openAdvancements` not memoized** — **Resolved** (unchanged since last review).
3. **[Medium] N26 fighter switcher loses the XP figure entirely when no Advancement is open** (`fighter-page.tsx:641-649`) — **Resolved.** `xpSuffix` is no longer gated on `!isCumulativeXp`. For cumulative-XP (N26) fighters it now always renders `(currentXp/nextTierStart XP)` via the new `nextTierStartFor` call, independent of `showAdvancementIcon`; the icon is purely additive on top. This matches the exact `${currentXp}/${nextTierStart ?? '–'}` pattern already used in `fighter-details-card.tsx:407-415` and `fighter-xp-modal.tsx:405`, so it's consistent with the app's existing convention for cumulative-XP display rather than introducing a new format.
4. **[Medium] `combobox.tsx` closed-state rendering change has app-wide blast radius** — **Resolved** (unchanged since last review; `showLabelWhenClosed` opt-in prop, default `false`, only the fighter switcher opts in).

## New findings
None. The diff since the last review is a single, self-contained commit touching only `components/fighter/fighter-page.tsx` (+15/-8): it fixes finding #3 by reusing an existing utility (`nextTierStartFor`) and an existing formatting convention. Null/undefined handling is sound — `nextTierStart` degrades to `'–'` when there's no next tier (Legend of the Underhive, or an edition with no ladder), and the `f.xp === undefined || f.xp === null` guard is preserved unchanged for the non-cumulative branch, so there's no regression there.

Minor, non-blocking observation: `openAdvancementsFor` and `nextTierStartFor` each independently call the internal `tierStartsFor(editionSlug)` lookup per fighter in the map — trivial cost (a small array literal), not worth restructuring for.

## Verdict
**Approve** — all previously outstanding findings are now resolved, and no new correctness, security, or edge-case issues were introduced by this commit.

---

PR #2058, opened 2026-08-25
Add challenge system for gang battles in campaigns

joeseos:
## Summary
This PR implements a challenge-based battle system for campaigns, allowing gangs to issue and respond to challenges before filing battle reports. The system introduces a new battle lifecycle with multiple states (pending, issued, accepted/declined) and separates the challenge workflow from traditional battle logging.

### Battle Status Lifecycle
- Added `BattleStatus` type with five states: `challenge_pending`, `challenge_issued`, `challenge_accepted`, `challenge_declined`, and `played`
- Challenges are created in `challenge_pending` state and progress through issuing/responding before being converted to `played` when the battle report is filed
- Added helper functions `battleStatusOf()` and `isPlayedBattle()` to handle legacy battles without a status field (defaulting to `played`)
- Added visual status badges with color coding in the battle logs table

### Challenge Round Generation
- New `generateChallengeRound()` action creates one `challenge_pending` slot per accepted gang in a campaign
- Only campaign arbitrators can open challenge rounds
- Supports multiple rounds per cycle for flexible scheduling

### Challenge Workflow Actions
- `issueChallenge()`: Allows a gang owner to pick an opponent and territory at stake for their pending challenge
- `respondToChallenge()`: Allows the challenged gang to accept or decline the challenge
- Both actions include permission checks to ensure only the relevant gang owner (or arbitrator) can act

### UI Components
- New `CampaignChallengeModal` for issuing challenges with opponent selection, territory staking, scenario, and notes
- New `CampaignChallengeRoundModal` for arbitrators to open challenge rounds
- Added action buttons in battle logs table for pending/issued challenges
- Challenges are sorted above played battles in the logs list regardless of sort order

### Battle Log Updates
- Modified `updateBattleLog()` to support converting challenges to played battles via `status` parameter
- Territory release logic now only applies to `played` battles (challenges stake territory, not claim it)
- Updated `deleteBattleLog()` to preserve territory ownership when deleting challenges
- Battle statistics exclude challenges from win/loss records

### Data Model
- Added `status`, `challenger_gang_id`, and `challenged_gang_id` fields to Battle interface
- Added `current_cycle` field to campaign data for UI convenience
- Updated database migration to add `challenge_pending` to status constraint

### Permission Checks
- New `ownsGangOrArbitrates()` helper validates that user owns the gang or runs the campaign
- Challenge actions re-verify specific gang ownership beyond RLS policies


joeseos:
## Code review — @joeseos

Reviewed the six commits against the codebase. Four findings, one blocking.

Note on conventions: there is no `contribution.md` / `CONTRIBUTING.md` in the repo — the contributor conventions live in `README.md` ("How to Contribute as a Developper?"). Branch naming (`feature/…`) and the migration conventions in `supabase/README.md` are followed. Cache invalidation uses the post-`refac/cache-tags` API (`invalidateCampaign`) throughout, and the new actions return `{ success, error }`, matching the dominant convention rather than the local throw-style of `createBattleLog`/`updateBattleLog`.

---

### 🔴 HIGH — the staked territory cannot be set when issuing a challenge

**`components/campaigns/[id]/campaign-battle-log-modal.tsx:472` and `:1059`**

```ts
return hasAnyWinnerSelected ? territories : ([] as BattleLogTerritory[]);
```
```tsx
{hasAnyWinnerSelected && availableTerritories.length > 0 && ( … territory combobox … )}
```

The territory field is gated behind `hasAnyWinnerSelected`. While "Battle played" is unticked there are no winners, so the combobox is hidden and `availableTerritories` is empty — the challenger has no way to nominate the territory they are staking.

That is the point of a challenge, and it contradicts the design this PR is built on ("the challenger picks the stake along with the opponent"). As it stands a challenge can only ever record the two gangs, and the stake gets entered after the battle, which is the behaviour the feature set out to replace.

**Fix:** show the territory whenever there is a challenge, not only when a winner exists:

```ts
const territoryPickable = hasAnyWinnerSelected || isUnplayedChallenge;
```
and use that in both the memo and the JSX guard. Worth double-checking the label too — it reads "Claimed Territory" / "Contested Territory", and for a challenge it should say something like "Territory at stake".

---

### 🟡 MEDIUM — `respondToChallenge` reports success when nothing was updated

**`app/actions/campaigns/[id]/battle-logs.ts`, `respondToChallenge`**

```ts
const { error: updateError } = await supabase
  .from('campaign_battles')
  .update({ status: … })
  .eq('id', battleId)
  .eq('status', 'challenge_issued');

if (updateError) throw updateError;
return { success: true };
```

The `.eq('status', …)` guard is there for concurrency, but the result is never checked. Zero rows affected is not an error in PostgREST, so both a lost race (someone already answered) and an **RLS denial** return `{ success: true }` — the user sees "Challenge answered" and the toast, while the row is unchanged. The optimistic UI then disagrees with the server until the next refresh.

**Fix:** `.select('id')` on the update and treat an empty result as a failure:

```ts
const { data, error } = await supabase…select('id');
if (error) throw error;
if (!data?.length) return { success: false, error: 'This challenge has already been answered' };
```

---

### 🟡 MEDIUM — challenge-round count includes gangs that have not accepted

**`components/campaigns/[id]/campaign-battle-logs-list.tsx`** (`availableGangs` effect) → passed as `gangCount` to the round modal.

`availableGangs` is built from every `member.gangs` entry with no status filter, but the loader returns gangs with `status IN ('ACCEPTED','PENDING')` and `generateChallengeRound` inserts only for `status = 'ACCEPTED'`. So the confirm dialog can promise "This creates 7 challenges" and the server creates 5.

The convention already exists elsewhere — `campaign-page-content.tsx` filters `g.status === 'ACCEPTED'` when building `campaignGangs` for battle sessions.

**Fix:** filter to accepted gangs for the count (and consider it for the opponent combobox, so a pending member cannot be challenged — that part is pre-existing behaviour of the battle-log modal, not a regression from this PR).

---

### 🟢 LOW — `challenged_gang_id` recomputed on every edit

**`app/actions/campaigns/[id]/battle-logs.ts`, `updateBattleLog`**

```ts
if (existingBattle.challenger_gang_id) {
  updatePayload.challenged_gang_id =
    participants.find((p) => p.gang_id !== existingBattle.challenger_gang_id)?.gang_id ?? null;
}
```

Runs on every update of any row carrying a `challenger_gang_id`, including long after it is `played`, and with three or more participants `.find()` picks whichever happens to be first. Harmless today because the column is only read for accept/decline on an issued challenge, but it is a silent write with no guard.

**Fix:** only recompute while the row is still a challenge (`existingBattle.status !== 'played'`).

---

### What looks good

- The staked-vs-claimed guards in `updateBattleLog` and `deleteBattleLog` are the right call and are the subtle bug in this feature — deleting a pending slot no longer strips the stake's owner.
- Reusing `updateBattleLog` for the conversion rather than adding a `completeChallenge` action, and seeding `participants` so the existing RLS policy covers the new flows with no policy changes.
- Excluding non-`played` rows from the standings and the gang page — easy to miss, would have quietly skewed every record.
- Migration is guarded and re-appliable; the Discord triggers key on `'played'` so no webhook change was needed.

### Verdict

**Changes requested** — on the strength of the HIGH finding. The two mediums are worth folding into the same round. Everything else is sound.

---

Djidiouf (approved):

PR #2059, opened 2026-08-25
fix(admin): use shared Modal component for equipment adjusted cost dialog

v0rp-z:
## Summary

The **Add Equipment Adjusted Cost** dialog in the admin *Edit Fighter Type* form is not using the shared [`components/ui/modal.tsx`](components/ui/modal.tsx) `Modal` component that the rest of the app uses. 

This PR swaps it for the shared component.

## Changes

- `components/admin/admin-edit-fighter-type.tsx`: replaced the dialog markup with `<Modal>`.
- State reset for the three form fields (`showAdjustedCostDialog`, `selectedAdjustedCostEquipment`, `adjustedCostAmount`) is now centralised in a single `onClose` handler

## Test plan

  - Save adds the chip, closes the modal, resets the form
  
## Screengrabs
**Old:** 
Screenshot From 2026-08-24 23-13-16

**New:** 
Screenshot From 2026-08-24 23-07-23 Screenshot From 2026-08-24 23-07-06 Screenshot From 2026-08-24 23-07-13

### DIFF
--- a/components/admin/admin-edit-fighter-type.tsx
+++ b/components/admin/admin-edit-fighter-type.tsx
@@ -2364,95 +2364,65 @@ export function AdminEditFighterTypeModal({ onClose, onSubmit }: AdminEditFighte
                 )}
 
                 {showAdjustedCostDialog && (
-                  <div
-                    className="fixed inset-0 bg-black/50 dark:bg-neutral-700/50 flex items-center justify-center z-50"
-                    onClick={(e) => {
-                      if (e.target === e.currentTarget) {
-                        setShowAdjustedCostDialog(false);
-                        setSelectedAdjustedCostEquipment("");
-                        setAdjustedCostAmount("");
-                      }
+                  <Modal
+                    title="Equipment Adjusted Cost Menu"
+                    helper="Select equipment and enter an adjusted cost"
+                    width="sm"
+                    onClose={() => {
+                      setShowAdjustedCostDialog(false);
+                      setSelectedAdjustedCostEquipment("");
+                      setAdjustedCostAmount("");
                     }}
+                    onConfirm={() => {
+                      setEquipmentDiscounts(prev => [
+                        ...prev,
+                        {
+                          equipment_id: selectedAdjustedCostEquipment,
+                          adjusted_cost: parseInt(adjustedCostAmount),
+                        },
+                      ]);
+                    }}
+                    confirmText="Save Adjusted Cost"
+                    confirmDisabled={!selectedAdjustedCostEquipment || !adjustedCostAmount || parseInt(adjustedCostAmount) < 0}
                   >
-                    <div className="bg-card p-6 rounded-lg shadow-lg w-[400px]">
-                      <h3 className="text-xl font-bold mb-4">Equipment Adjusted Cost Menu</h3>
-                      <p className="text-sm text-muted-foreground mb-4">Select equipment and enter an adjusted cost</p>
+                    <div className="space-y-4">
+                      <div>
+                        <label className="block text-sm font-medium mb-1">Equipment</label>
+                        <select
+                          value={selectedAdjustedCostEquipment}
+                          onChange={(e) => setSelectedAdjustedCostEquipment(e.target.value)}
+                          className="w-full p-2 border rounded-md"
+                        >
+                          <option value="">Select equipment</option>
+                          {filteredEquipment
+                            .filter(item => !equipmentDiscounts.some(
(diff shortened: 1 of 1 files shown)


joeseos:
### Review — @v0rp-z

Nice cleanup — matches the existing `Modal` usage pattern already established for the Gang Cost dialog in the same file (`components/admin/admin-edit-fighter-type.tsx:2485`).

**Findings**

1. **[Medium] Missing revalidation inside `onConfirm`** — `components/admin/admin-edit-fighter-type.tsx:2379-2387`
   The new `onConfirm` unconditionally pushes to `equipmentDiscounts` with no guard:
   ```tsx
   onConfirm={() => {
     setEquipmentDiscounts(prev => [
       ...prev,
       { equipment_id: selectedAdjustedCostEquipment, adjusted_cost: parseInt(adjustedCostAmount) },
     ]);
   }}
   ```
   It relies solely on `confirmDisabled` to gate this. The sibling dialog in the same file, `handleAddGangCost` (line 1355), re-validates inputs *inside* the handler and returns `false`/toasts on invalid data — defense-in-depth in case the handler is ever invoked outside the disabled-button path (e.g. pasted negative values bypass the `onKeyDown` guard, though currently caught by `confirmDisabled`). Not a functional regression today, but worth aligning with the established pattern for consistency and robustness. Suggest extracting an `onConfirm` handler that re-checks `selectedAdjustedCostEquipment`/`parseInt(adjustedCostAmount) >= 0` before pushing (similar to `_handleAddAdjustedCost`, which already exists nearby but is currently unused/dead code — could be revived here instead of the new inline closure).

2. **[Low] Modal close-then-reset semantics verified correct** — no action needed. `Modal`'s `handleConfirm` calls `onConfirm()` then `onClose()` when the result isn't explicitly `false` (`components/ui/modal.tsx`), so the centralized `onClose` reset (clearing `showAdjustedCostDialog`/`selectedAdjustedCostEquipment`/`adjustedCostAmount`) correctly fires after a successful Save, matching the old inline behavior. Confirmed this is not a regression — just flagging that it was checked.

**Checklist**
- Correctness: ✅ (aside from finding #1, which is non-blocking)
- Edge cases / null handling: ✅ equivalent to prior behavior
- Security: ✅ no new surface (client-side form state only)
- Performance: ✅ no change
- Readability/maintainability: ✅ improved — removes ~80 lines of duplicated modal markup
- Project conventions: ✅ now consistent with the Gang Cost dialog's `Modal` usage
- DRY/YAGNI: ⚠️ see finding #1 — inline `onConfirm` duplicates logic that already exists (dead) as `_handleAddAdjustedCost`

**Verdict: Approve** (finding #1 is a nit, not a blocker)


---

joeseos:
### Review (updated commit) — @v0rp-z

Reviewed only the new commit (`20e5061`) since the last review.

**Previous findings**

1. **[Medium] Missing revalidation inside `onConfirm`** — **Resolved.** `onConfirm` is now `handleAddAdjustedCost`, a named handler that mirrors the sibling `handleAddGangCost` pattern: it re-checks `selectedAdjustedCostEquipment`/`adjustedCostAmount` are set, re-parses and validates the cost (`isNaN` / `< 0`), and toasts + `return false` on invalid input (which keeps the `Modal` open per its `handleConfirm` contract, confirmed in `components/ui/modal.tsx`). The previously-dead `_handleAddAdjustedCost` was removed rather than left as unused code.
2. **[Low] Modal close-then-reset semantics** — Still correct; unaffected by this commit (reset stays in the centralised `onClose`, which still fires since the success path implicitly returns `undefined`, and `undefined !== false`).

**New findings**

None. The new handler's success path returns `undefined` rather than `handleAddGangCost`'s explicit `true`/manual reset+close — a minor stylistic asymmetry between the two sibling handlers, but functionally correct (verified against `Modal`'s `result !== false` check) and arguably preferable since it avoids duplicating the centralised `onClose` reset logic. Not flagging as an issue.

**Checklist**
- Previous findings addressed: ✅ (finding #1 resolved, finding #2 unaffected)
- New bugs/regressions: none found
- Edge cases / null handling: ✅ empty fields, non-numeric input, and negative values are all caught before mutating state
- Readability / conventions: ✅ matches `handleAddGangCost` validation style

**Verdict: Approve**

---

v0rp-z:
Extracted handleAddAdjustedCost following the handleAddGangCost pattern — validates and returns false inside the handler, this has been wired into onConfirm, so even with a paste input it cannot be bypassed. See image below with negative value. 

image

PR #2060, opened 2026-08-25
fix(admin): use shared Modal component for equipment adjusted cost dialog (create)

v0rp-z:
## Summary

Follow-up to #2053. The **Add Equipment Adjusted Cost** dialog in the admin *Create Fighter Type* form had the same old modal that #2053 fixed for the edit form. This PR applies the same refactor here.

## Changes

- `components/admin/admin-create-fighter-type.tsx`: replaced the old dialog with `<Modal>`
- State reset for the three form fields (`showAdjustedCostDialog`, `selectedAdjustedCostEquipment`, `adjustedCostAmount`) is centralised in a single `onClose` handler

## Test plan

  - Save adds the chip, closes the modal, and resets the form

### DIFF
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -316,6 +316,24 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
     setSpecialRules(prev => prev.filter(rule => rule !== ruleToRemove));
   };
 
+  const handleAddAdjustedCost = () => {
+    if (!selectedAdjustedCostEquipment || !adjustedCostAmount) {
+      toast.error('Please select equipment and enter an adjusted cost');
+      return false;
+    }
+
+    const cost = parseInt(adjustedCostAmount);
+    if (isNaN(cost) || cost < 0) {
+      toast.error('Please enter a valid cost (must be 0 or greater)');
+      return false;
+    }
+
+    setEquipmentDiscounts(prev => [
+      ...prev,
+      { equipment_id: selectedAdjustedCostEquipment, adjusted_cost: cost },
+    ]);
+  };
+
   const handleSubmit = async () => {
     // Check if selected fighter subtype is Crew
     const isCrew = selectedFighterSubtypes.includes('Crew');
@@ -1155,15 +1173,7 @@ export function AdminCreateFighterTypeModal({ onClose, onSubmit }: AdminCreateFi
                     setSelectedAdjustedCostEquipment("");
                     setAdjustedCostAmount("");
                   }}
-                  onConfirm={() => {
-                    setEquipmentDiscounts(prev => [
-                      ...prev,
-                      {
-                        equipment_id: selectedAdjustedCostEquipment,
-                        adjusted_cost: parseInt(adjustedCostAmount),
-                      },
-                    ]);
-                  }}
+                  onConfirm={handleAddAdjustedCost}
                   confirmText="Save Adjusted Cost"
                   confirmDisabled={!selectedAdjustedCostEquipment || !adjustedCostAmount || parseInt(adjustedCostAmount) < 0}
                 >
--- a/components/admin/admin-create-fighter-type.tsx
+++ b/components/admin/admin-create-fighter-type.tsx
@@ -13,6 +13,7 @@ import { hasAlignment, hasSaveCharacteristic, allowsMultipleSubtypes, hasStartin
 import { toggleFighterSubtype } from '@/utils/fighter-subtype-picker';
 import { getSkillSetRank } from "@/utils/skillSetRank";
 import { compareEquipmentCategories } from "@/utils/getEquipmentCategoryRank";
+import Modal from '@/components/ui/modal';
 
 interface AdminCreateFighterTypeModalProps {
   onClose: () => void;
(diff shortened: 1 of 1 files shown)


joeseos:
## Review

Reviewed against `components/ui/modal.tsx` and the sibling refactor already in `admin-edit-fighter-type.tsx` (the `showGangCostDialog` → `Modal` migration, since the "Equipment Adjusted Cost" dialog in the edit form hasn't been migrated yet, despite the PR description).

### Correctness
Traced through `Modal`'s `handleConfirm`: it calls `onConfirm()`, and only if the result `!== false` does it call `onClose()`. Since this PR's `onConfirm` returns `undefined` (not `false`), `onClose` always fires right after — which is what resets `showAdjustedCostDialog`, `selectedAdjustedCostEquipment`, and `adjustedCostAmount`. So despite `onConfirm` here not doing the reset itself, the three-field reset described in the PR body is actually correct and behaviorally equivalent to the old inline dialog. Cancel / ×/ backdrop click all route through the same `onClose`, also correct.

### Findings

**1. (Low) `onConfirm` skips the old handler's inline re-validation, relying solely on `confirmDisabled`**
`components/admin/admin-create-fighter-type.tsx` ~L1170-1178

The old code re-checked `selectedAdjustedCostEquipment && adjustedCostAmount` and `adjusted_cost >= 0` *inside* the click handler before pushing to `equipmentDiscounts`, in addition to disabling the button. The new `onConfirm` just does:
```tsx
onConfirm={() => {
  setEquipmentDiscounts(prev => [...prev, {
    equipment_id: selectedAdjustedCostEquipment,
    adjusted_cost: parseInt(adjustedCostAmount),
  }]);
}}
```
with no validation, trusting `confirmDisabled` alone to gate it. That mirrors the disabled-button behavior fine for normal use, but it's inconsistent with the established pattern in the same file's `handleAddGangCost` (used by the neighboring Gang-Specific Cost modal), which validates inside the handler and returns `false` + a toast on bad input rather than trusting the disabled state alone. Not a functional bug (a disabled `<button>` won't fire `onClick`), but worth aligning for defense-in-depth/consistency if you're touching this again.

**2. (Info, pre-existing, not introduced by this PR) `confirmDisabled` doesn't catch `NaN`**
`parseInt(adjustedCostAmount) < 0` is `false` when `parseInt` returns `NaN`, so a non-numeric string wouldn't be blocked by `confirmDisabled`. This logic is copied verbatim from the pre-refactor code, so it's not a regression — just flagging since it's now visible in a smaller diff. Not blocking.

No other issues found — imports, JSX structure, `Button`/`Input` usage elsewhere in the file, and the `width="sm"` sizing all check out, and `equipmentDiscounts` dedup filtering is untouched.

### Checklist
- Correctness: ✅ (see note above on findings #1)
- Edge cases / null handling: ✅ minor pre-existing gap noted (#2)
- Security: ✅ no concerns
- Performance: ✅ no concerns (like-for-like state updates)
- Readability/maintainability: ✅ net reduction of ~30 lines, removes duplicated modal chrome
- Project conventions: ✅ matches the `Modal` migration pattern established by #2053 / `handleAddGangCost`
- DRY/YAGNI: ✅ this is exactly the point of the PR

### Verdict: **Approve**

Nice cleanup — the migration is faithful to the original behavior. Finding #1 is optional polish, not a blocker.

cc @v0rp-z


---

joeseos:
## Follow-up review (commit 3d77271)

Reviewed just the new commit since the last review — `fix(admin): validate inputs inside adjusted cost onConfirm handler`.

### Previous findings

**1. (Low) `onConfirm` skipped inline re-validation — RESOLVED.**
`handleAddAdjustedCost` now re-checks `selectedAdjustedCostEquipment`/`adjustedCostAmount` are present, parses the cost, and rejects `isNaN(cost) || cost < 0` with a toast + `return false` (which keeps the Modal open per its `handleConfirm` contract, since `result !== false` is what triggers `onClose`). This is a byte-for-byte match of `handleAddAdjustedCost` in `admin-edit-fighter-type.tsx:1057-1073`, so the two sibling forms are now consistent with each other and with `handleAddGangCost`'s validate-in-handler pattern.

**2. (Info, pre-existing) `confirmDisabled` doesn't catch `NaN` — effectively mitigated, not literally fixed.**
The `confirmDisabled` prop itself (`parseInt(adjustedCostAmount) < 0`) is untouched and still won't disable the button for a non-numeric string. But that no longer matters for data integrity: `handleAddAdjustedCost` now independently guards with `isNaN(cost)`, so a `NaN` input can no longer reach `setEquipmentDiscounts`. Only cosmetic residue is the Save button appearing enabled when it will immediately toast-reject. Not worth a follow-up on its own.

### New changes review

- The extraction is a pure refactor: same logic that was previously inline in `onConfirm`, now named and gated by the two validation checks. No new state, no new imports beyond what's already used (`toast` was already imported/used elsewhere in the file).
- Success path relies on implicit `return undefined` (not `return true`) to let `Modal` call `onClose`, matching the sibling file's exact style — consistent, not a bug.
- Failure paths correctly `return false` before touching `equipmentDiscounts`, so invalid data can't be appended.
- Edge cases: empty selection, empty amount, non-numeric amount, negative amount — all covered by the two checks. No behavior change for the valid-input path.
- No regressions spotted in the surrounding JSX/wiring; `onConfirm={handleAddAdjustedCost}` is a direct drop-in for the old closure.

### Verdict: **Approve**

Finding #1 is fully addressed and matches the established sibling pattern; #2's residual gap is cosmetic only. No new issues introduced.

cc @v0rp-z

---

v0rp-z:
Same as #2059 - Extracted handleAddAdjustedCost following the handleAddGangCost pattern — validates and returns false inside the handler, this has been wired into onConfirm, so even with a paste input it cannot be bypassed. See image below with negative value.

Screenshot From 2026-08-25 16-55-28

PR #2062, opened 2026-08-25
Fix Fancy card print missing the top-bar stroke on iOS

Djidiouf:
Keep fancy-print on the body while Fancy is selected, and render the stroke as an img so iOS Safari still prints it. Document LAN mobile testing and allowedDevOrigins in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,13 +66,18 @@ For questions about contributing, feel free to ask in our [Discord server](https
    ```
 
 4. **Mobile device testing**
-   On your desktop, make sure you have the NODE_ENV variable setup with the value 'development'. Then run the following:
-   ```bash
-   npm run dev --host
+
+   `npm run dev` already listens on the LAN. The startup log prints a **Network** URL (for example `http://192.168.1.115:3000`). Open that URL on a phone on the same Wi-Fi.
+
+   Next.js still blocks HMR and other `/_next` assets from that origin until you allow it in `next.config.js`:
+
+   ```js
+   allowedDevOrigins: ['192.168.1.115'],
    ```
-   
-   On your mobile device, when connected to your wifi, access the website using the IP of your desktop on your local network (e.g. 192.168.1.5):
-   http://192.168.1.5:3000
+
+   Use the same IP as the Network URL, then restart the dev server. If you skip this, the page may load but you will see `Blocked cross-origin request to Next.js dev resource` in the terminal.
+
+   Do not commit this `next.config.js` change when you push. `allowedDevOrigins` is your local IP and should stay off the repo.
 
 5. **Access to the DB Schema**
    The Supabase DB schema can be accessed through https://supabase-schema.vercel.app/
--- a/app/globals.css
+++ b/app/globals.css
@@ -215,6 +215,13 @@ Enforcing `cursor: pointer` has the default value. */
     background-image: none !important;
   }
 
+  /* CSS backgrounds on this bar are stripped above; iOS Safari also often skips
+     printing nested background-images. The stroke is an <img> (.fancy-print-top-bar-art)
+     and is shown only for Fancy print (body.fancy-print). */
+  .fancy-print-top-bar-art {
+    display: none !important;
+  }
+
   .print-fighters {
     zoom: 0.55;
   }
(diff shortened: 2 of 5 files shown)


joeseos (commented):
@Djidiouf Thanks for the fix — nice root-cause writeup on the iOS Safari `window.print()` non-blocking race. Reviewed the diff against the surrounding code.

## Summary

Solid fix. Switching the top-bar stroke from a CSS `background-image` swap to a persistent `` toggled by a `useEffect` (instead of add/remove-around-`window.print()`) correctly addresses the described race, and it's a more portable fix (not an iOS-only hack) since it no longer depends on `background-image` painting reliably during print on any browser.

I traced the print/eco toggle logic, the `body.fancy-print` class lifecycle, and the on-screen `FighterCard` usage (also rendered outside of printing, in the normal gang cards view) to check for regressions — didn't find any. The `` has no display rule outside `@media print`, so it renders unconditionally on screen just like the old unconditional inline `background-image` did — consistent with pre-PR behavior, not a regression.

## Findings

**1. (Low) Duplicated `` markup between `fighter-card.tsx` and `print-gang.tsx`**
`components/gang/print-gang.tsx` extracts a shared `FancyPrintTopBarArt` component (used 4x within that file), but `components/gang/fighter-card.tsx:607-614` inlines the identical `` markup (same src, classes, alt, style) instead of importing/reusing it. Minor DRY violation — consider moving `FancyPrintTopBarArt` to a shared location (or a small local component in `fighter-card.tsx`) both files can import.

**2. (Low) Vestigial CSS left behind**
`print-gang.tsx`'s `printStyle === 'eco'` branch still includes `[&_.fancy-print-top-bar]:bg-none!`, which no longer does anything since `.fancy-print-top-bar` never gets a `background-image` set anymore (moved to the ``). Not incorrect, just dead weight — could be removed alongside the new `[&_.fancy-print-top-bar-art]:hidden!` it sits next to.

**3. (Low) `print-color-adjust` on an ``**
`app/globals.css`: `body.fancy-print .fancy-print-top-bar-art { ... -webkit-print-color-adjust: exact; print-color-adjust: exact; }` — this property is meant for backgrounds/box-shadows; it's a no-op on a plain `` (carried over from when this was a `background-image` rule). Harmless, but can be dropped for clarity.

No correctness, security, or performance issues found. The `useEffect` dependency array (`[viewMode, printStyle]`) and cleanup are correct — cleanup and re-apply happen synchronously in the same commit, so there's no visible flicker, and unmount correctly clears the body class.

## Verdict: **Approve**

The three findings above are all low-severity cleanup items, not blockers.

---

joeseos:
@Djidiouf Reviewed the latest push (`1ed4ba5` → `f6a6cc2`).

**Previous findings**
None on file — this PR has no prior review threads (only the deploy-preview bot comment), so there's nothing to mark resolved/outstanding/partial.

**Review of the latest commit's changes**

This commit is a clean follow-up on top of the iOS fix already in the first commit (CSS `background-image` → real `` for the top-bar stroke). It:
- Extracts the duplicated `FancyPrintTopBarArt` component (previously defined separately in `fighter-card.tsx` and `print-gang.tsx`) into a shared `components/gang/fancy-print-top-bar-art.tsx`. Good DRY cleanup, both call sites now stay in sync.
- Removes `[&_.fancy-print-top-bar]:bg-none!` from the eco-print selector in `print-gang.tsx`. Verified `.fancy-print-top-bar` no longer carries any background image (that's now `.fancy-print-top-bar-art`, the ``), so this selector was dead — safe removal, no behavior change.
- Removes `-webkit-print-color-adjust`/`print-color-adjust` from `body.fancy-print .fancy-print-top-bar-art` in `globals.css`. These properties affect backgrounds/colors, not `` raster content, so they were no-ops on an `` element — harmless cleanup, not a functional regression.
- README rewrite for LAN mobile testing is accurate and adds a useful callout not to commit the local-IP `allowedDevOrigins` value; confirmed `next.config.js` wasn't touched in this PR, so no local IP leaked into the diff.

**New findings**
None blocking. One very minor nit, not worth a change: the new `` has both `alt=""` and `aria-hidden` — redundant (either alone marks it decorative) but harmless.

I didn't have a way to test the actual iOS print output in this environment, so I can't independently confirm the visual fix — the approach (raster `` instead of a CSS background-image, which iOS Safari is known to drop on print) is sound and matches the documented rationale in the CSS comment.

**Verdict: Approve**

---

PR #2064, opened 2026-08-25
fix(battles): allow picking a gang you own as the opponent

joeseos:
The New Battle modal excluded every gang belonging to the current user from the "Opponent Gang" picker. That filter only had an effect on the campaign page, which passes userId; the gang page passes none, so it listed own gangs and the two entry points disagreed. Owning both sides of a battle is legitimate, so the exclusion is gone.

The gang picked as "Your Gang" is still kept out of the opponent list, and own gangs already chosen as opponents are now kept out of the "Your Gang" dropdown, so a gang cannot end up fighting itself; the submitted gang ids are deduped as a final guard.

Also sorts the opponent list by name in the modal, so the campaign page matches the gang page's already-sorted list.

### DIFF
--- a/components/battle-session/create-battle-modal.tsx
+++ b/components/battle-session/create-battle-modal.tsx
@@ -187,6 +187,12 @@ export default function CreateBattleModal({
     setSelectedCampaignGangIds((prev) => prev.filter((id) => id !== campaignGangId));
   };
 
+  const selectMyGang = (campaignGangId: string) => {
+    setSelectedMyGangId(campaignGangId);
+    // A gang can't fight itself: claiming it as yours drops it as an opponent.
+    removeCampaignGang(campaignGangId);
+  };
+
   // Filter out gangs already added as opponents, already in the session, or from
   // another edition. Skirmish opponents are searched across all users, so unlike
   // campaign gangs nothing upstream has constrained them to one ruleset.
@@ -240,15 +246,13 @@ export default function CreateBattleModal({
       const scenarioName = selectedScenario === 'custom'
         ? customScenario.trim()
         : sortedScenarios.find((s) => s.id === selectedScenario)?.scenario_name;
-      const pickedGangIds = effectiveGangId ? [effectiveGangId] : [];
+      const allGangIds = effectiveGangId ? [effectiveGangId] : [];
 
       if (campaignId) {
-        pickedGangIds.push(...selectedCampaignGangIds);
+        allGangIds.push(...selectedCampaignGangIds);
       } else {
-        pickedGangIds.push(...opponents.map((o) => o.gangId));
+        allGangIds.push(...opponents.map((o) => o.gangId));
       }
-
-      const allGangIds = Array.from(new Set(pickedGangIds));
 
       if (allGangIds.length === 0) {
         return { success: false, error: 'At least one gang is required' };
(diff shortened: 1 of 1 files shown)


joeseos (commented):
@joeseos Reviewed against the full `CreateBattleModal` component and its callers (`battle-sessions-tab.tsx` from the gang page, `campaign-battle-sessions.tsx` from the campaign page, `active-session.tsx`).

## Summary

Confirmed the root cause described in the PR body: `myGangs` is only non-empty when both `campaignId` and `userId` are passed, and the gang page (`gang-page-content.tsx:749-756`) never passes `userId` to `GangBattleSessions`/`CreateBattleModal` — so `myGangs` was always `[]` there, making the old `!myGangs.some(...)` exclusion a no-op on the gang page while it was active on the campaign page. Removing it and aligning both entry points is correct.

The two new anti-self-battle guards are both necessary and sufficient given the modal's structure:
- `opponentCampaignGangs` still excludes `effectiveGangId`, so the currently-selected "Your Gang" can never also be picked as an opponent.
- The new filter on the "Your Gang" `Combobox` options (only reachable when `myGangs.length > 1`) excludes gangs already in `selectedCampaignGangIds`/`existingGangIds`, closing the reverse direction.
- When `myGangs.length <= 1` or `gangId` is fixed (gang-page entry), "Your Gang" is static text, not a combobox, so it's never at risk of colliding with the opponent list at all.

## Findings

**1. (Low) UX dead-end: "Your Gang" combobox can be starved to zero options**
`components/battle-session/create-battle-modal.tsx:110-111,146-153`. When a user has 2+ own gangs in a campaign (`myGangs.length > 1`) and hasn't picked "Your Gang" yet, `effectiveGangId` is `undefined`, so `opponentCampaignGangs` doesn't exclude any of the user's own gangs from the opponent list. If the user adds *all* of their own gangs as opponents before ever touching the "Your Gang" dropdown, that dropdown is left with zero options (every candidate is now in `selectedCampaignGangIds`), and `effectiveGangId` stays `undefined`. `canConfirm` doesn't require `effectiveGangId` to be set (it only checks `selectedScenario`/`hasOpponents`, unchanged by this PR), so the battle can still be created — functionally fine since the picked gangs are still all correctly included via `allGangIds`, but "Your Gang" is left visibly unset/on placeholder, which reads as broken. Only reachable now because own gangs are no longer excluded from the opponent list — worth a quick look, but not a data-integrity issue.

**2. (Low) Minor duplicated filter logic**
The `!selectedCampaignGangIds.includes(g.id) && !existingGangIds.includes(g.id)` predicate is now repeated between `opponentCampaignGangs` (line 150-151) and the "Your Gang" `Combobox` options filter (line 330-332). Could be pulled into a small shared helper (e.g. `isPickableCampaignGang(g)`), but it's only two call sites — not a strong DRY violation.

**3. (Nit) Client-side dedupe is likely redundant**
`Array.from(new Set(pickedGangIds))` (line 251) — `createBattleSession`'s server action fetches gangs via `.in('id', params.gang_ids)`, which Postgres already collapses to distinct matches, so this dedupe shouldn't be load-bearing server-side. Harmless to keep as a defensive guard given this PR is specifically about self-battle correctness, just noting it's belt-and-suspenders rather than fixing a real duplicate-insert risk.

No correctness bugs, security issues, or performance concerns found beyond the above. Sorting `opponentCampaignGangs` by name is a nice small consistency fix that matches the gang page's already-sorted list, as described.

## Verdict: **Approve**

All findings are low-severity/edge-case; nothing here should block merging.

---

PR #2065, opened 2026-08-26
Fixed/copy fighter default skills

joeseos:


joeseos (commented):
@joeseos Reviewed `app/actions/copy-fighter.ts` in full, plus cross-checked against `app/actions/add-fighter.ts` to verify the "default skill" semantics this PR relies on.

## Summary

Correct fix. Before this PR, a base ("fresh recruit") copy skipped `fighter_skills` entirely, so a fighter whose type grants a default skill (e.g. a Ganger with a default Combat skill) would lose that skill on a non-experienced copy — inconsistent with what `add-fighter.ts` gives a freshly recruited fighter of the same type. `getDefaultSkillIds`/`copySkills` close that gap by reading `fighter_defaults` for the fighter's (or beast's) type and only carrying over skills in that set when `copy_as_experienced` is falsy, inserted at zero cost.

I specifically checked whether this could under- or over-copy anything:
- **Skill-linked effects**: I initially suspected default skills with attached `fighter_effects` (via `fighter_skill_id`) would be dropped in a base copy since the effects filter (line ~544-548) excludes `category === 'skills'` / `effect.fighter_skill_id` for non-experienced copies. But `add-fighter.ts`'s own default-skill insert (line 759-780) never creates a matching `fighter_effects` row either — it only inserts the bare `fighter_skills` row. So this PR's behavior is faithful to what a real fresh recruit gets; no gap.
- **Cost accounting**: `calculateTotalCopyCost` for base copies doesn't add a line item for the newly-copied default skills, which is correct since they're inserted free (`credits_increase: 0`) and the type's base `credits` already prices them in.
- **Custom vs. stock skills**: `getDefaultSkillIds` only matches on `skill_id` (not `custom_skill_id`), documented in the function's comment as intentional since `fighter_defaults` has no `custom_skill_id` column today.
- **Beast fighters**: `beastFighter` is fetched with `*` plus `fighter_type_id`/`custom_fighter_type_id`, so the shared `copySkills` helper works identically for the beast-copy path. Good DRY win — this also de-duplicates what were previously two near-identical inline blocks (main fighter + beast fighter) into one helper.
- **Refactor correctness**: the experienced-copy path (`copyAsExperienced` true) still copies all `fighter_skills` rows unfiltered with original cost fields, byte-for-byte equivalent to the pre-PR behavior — the docstring's "always stock skills" caveat correctly bounds the current limitation.

## Findings

**1. (Low) Silent failure mode differs from every other copy step in this file**
`app/actions/copy-fighter.ts:99-102`. If the `fighter_defaults` read in `getDefaultSkillIds` errors, it `console.warn`s and returns an empty `Set`, which makes `copySkills` silently copy zero default skills — the overall `copyFighter` call still reports success. Every other failure in this function (equipment insert, skills insert, effects insert, beast fetch, etc.) triggers `rollbackFighter` and returns `success: false`. If this is a deliberate "degrade gracefully rather than block the whole copy over a metadata read" choice, worth a one-line comment saying so explicitly (the existing comment explains *what* happens, not *why* it's not treated as failure); otherwise consider surfacing it as a non-fatal warning in the result so the user knows their copy may be missing default skills.

No other correctness, security, or performance issues found. The UI copy change ("Base fighter with equipment and default skills (fresh recruit)") accurately reflects the new behavior.

## Verdict: **Approve**

The one finding above is low-severity and about failure-mode clarity, not a functional bug — doesn't block merging.

---

PR #2067, opened 2026-08-26
Fix/gang legacys in edit fighter

joeseos:


joeseos:
## Code Review

@joeseos Reviewed `fighter-edit-modal.tsx`, `FighterAddModal.tsx`, and `types/edition.ts` against the rest of the codebase, and traced the specialisation/variant/legacy field shapes through `app/lib/shared/gang-assembly.ts`, `types/fighter.ts`, and the other consumers (`fighter-page.tsx`, `fighter-card.tsx`). Found one bug that undermines the PR's own stated goal, plus a maintainability concern with the new specialisation list.

### Findings

**1. Wrong property name (`.id` instead of `.fighter_specialisation_id`) silently reintroduces the exact "retype wipes specialisation" bug this PR is fixing** — High
`components/fighter/edit-fighter/fighter-edit-modal.tsx:235, 626, 955`
All three reads of the fighter's current specialisation use `(fighter.fighter_specialisation as any)?.id`:
```ts
const [selectedFighterSpecialisationId, setSelectedFighterSpecialisationId] = useState<string>(
  (fighter.fighter_specialisation as any)?.id || ''
);
...
setSelectedFighterSpecialisationId((fighter.fighter_specialisation as any)?.id || '');
...
const currentSpecialisationId = (fighter.fighter_specialisation as any)?.id || '';
```
But the real shape of `fighter.fighter_specialisation` — per `types/fighter.ts:243-246` and how it's actually assembled in `app/lib/shared/gang-assembly.ts:680-683` and `:1325-1330` — is `{ fighter_specialisation: string; fighter_specialisation_id: string }`. There is no `.id` field anywhere a fighter is fetched from the server; every other consumer in the codebase (`fighter-page.tsx:869-870,950`, `fighter-card.tsx:545`, `print-gang.tsx:715`) correctly reads `.fighter_specialisation_id`.

Effect: for any fighter loaded fresh from the server (i.e. essentially all real fighters, since this data comes from the DB, not from this modal's own optimistic overlay), `.id` is always `undefined`, so:
- `selectedFighterSpecialisationId` initializes to `''` — the dropdown shows **"None" even when the fighter has a specialisation**, while the "Current: …" line right below it (which correctly reads `.fighter_specialisation`, the name field) shows the real value. Visibly inconsistent UI on every open.
- Worse, in `handleConfirm`, `currentSpecialisationId` is therefore also always `''`. If the user retypes the fighter (`submitData.fighter_type_id` becomes truthy) without touching the specialisation dropdown, the condition `selectedFighterSpecialisationId !== currentSpecialisationId || submitData.fighter_type_id` is true via the type-change branch, so the block runs with `selectedFighterSpecialisationId === ''`, which resolves to no matching entry in `N26_PROSPECT_SPECIALISATIONS` and submits `fighter_specialisation_id: null` — **wiping the fighter's real specialisation on any retype**, which is precisely the bug the PR's own comment says it's preventing ("a retype must not rewrite it").

This likely wasn't caught in manual testing because this PR's own optimistic-overlay code (line ~543) *does* add an `id` field to the in-memory fighter object after a save in the same session, masking the bug until a fresh page load.

Fix: read `.fighter_specialisation_id` in all three spots, not `.id`.

**2. `N26_PROSPECT_SPECIALISATIONS` (hardcoded prod UUIDs from a Prospect-promotion-specific module) is now the master option list for every N26 fighter's specialisation editor** — Medium
`components/fighter/edit-fighter/fighter-edit-modal.tsx:13, 957-961, 1137` / `utils/keepTypePromotionN26.ts:8-12`
That module's own docstring warns: *"Specialisation/skill UUIDs below are production catalog ids used as a preferred lookup cache... specialisation ids do not [fall back] — a wrong id fails the `fighters.fighter_specialisation_id` FK. Prefer resolving from the DB ... if these ever drift."* This PR reuses that list as the canonical, only source of options for a general-purpose editing UI covering every N26 fighter (not just the Prospect-promotion flow the module was written for). If these IDs ever diverge from a given environment's `fighter_specialisations` table (a reseeded/staging/self-hosted DB, a future catalog edit), saving a specialisation here will hard-fail on the FK, with no fallback — unlike every other dropdown in this same modal (`fighterTypes`, `availableLegacies`, subtypes), which are all fetched live from the DB. Consider fetching the specialisation catalog the same way, or at least hoisting this list out of the promotion-specific module into something scoped for general reuse.

### Checklist
- Correctness: ❌ — see #1, a real regression risk for the exact scenario this PR targets
- Edge cases / null handling: mostly ✅ elsewhere (custom fighter type nulls specialisation correctly, legacy-clear-on-incompatible-row fix in both modals looks correct and uses the right field name)
- Security: ✅ no concerns
- Performance: ✅ no concerns
- Readability/maintainability: see #2; the variant/specialisation decoupling itself is clearly explained via comments
- Project conventions: no `CLAUDE.md` in this repo to check against
- DRY/YAGNI: see #2

### Verdict: **Changes requested** (finding #1 should be fixed before merge — it's a silent data-loss bug on save, not just a display glitch)

---

joeseos:
Correcting my earlier response to review finding 1 — **the finding was right and my pushback was wrong.**

I argued that `fighter.fighter_specialisation` reaches this modal keyed `id`, because that is how `app/fighter/[id]/page.tsx:232-236` builds it. I missed `transformFighterData` in `components/fighter/fighter-page.tsx:334-351`, which rekeys it on the way into client state:

```js
fighter_specialisation: fighterData.fighter.fighter_specialisation ? {
  fighter_specialisation: fighterData.fighter.fighter_specialisation.fighter_specialisation,
  fighter_specialisation_id: fighterData.fighter.fighter_specialisation.id
} : undefined,
```

The modal's `fighter` prop comes from `convertToFighterProps({ ...fighterData.fighter })`, i.e. the transformed state — so `?.id` was always `undefined`, exactly as the review said, and the suggested fix (read `.fighter_specialisation_id`) was correct. My claim that it "would break the fresh-load case" was false.

Two further things I asserted and should retract:

- `fighter-page.tsx:869` / `:950` are **not** buggy. They read the converted object, so `.fighter_specialisation_id` is right there. There is no pre-existing bug at those lines, and the review was correct to cite them as the pattern.
- `:350` is not a bad read either — it is the conversion itself, consuming `.id` from the server object deliberately.

Fixed in `e536fa2`: all three sites read `.fighter_specialisation_id`, and the optimistic overlay in `onMutate` now emits the same `{ fighter_specialisation, fighter_specialisation_id }` pair rather than putting the server shape back into client state — which was the one place this PR actually introduced the inconsistency. The stopgap helper and both `as any` casts are gone; it typechecks without them, which is its own confirmation that the code and `types/fighter.ts:243-246` now agree.

On finding 2 — reusing `N26_PROSPECT_SPECIALISATIONS` as the option list — agreed in principle, deferred deliberately. Those 8 ids are currently exactly the in-use set, but `fighter_specialisations` holds 86 rows, the other 78 being variant labels orphaned by `20260823120000`. A live fetch is the right answer and becomes clean once those are deleted; adding an endpoint that needs client-side filtering today would be the worse of the two. Tracked for the catalog cleanup.

---

PR #2068, opened 2026-08-26
feat(cache): add conditional Redis cacheHandler for self-hosted deploys

joeseos:
Backs the singular Next.js `cacheHandler` with Redis when `USE_REDIS_CACHE=true`, so a self-hosted (Coolify) deploy shares its Data Cache between instances and keeps it across redeploys. Vercel sets neither variable, so no config keys are written at all and the built-in Data Cache is used exactly as before.

No application code changes: all 49 `unstable_cache` call sites and the tag helpers in `utils/cache-tags.ts` are untouched, and nothing migrates to `use cache`.

## How the conditional works

`next.config.js` spreads in `cacheHandler` + `cacheMaxMemorySize: 0` only when the flag is set, using `fileURLToPath(new URL(...))` since `require.resolve` does not exist under `"type": "module"`.

**Building without `REDIS_URL`** is guarded on both `NEXT_PHASE` and the absence of the URL; the build falls back to an in-process cache and never opens a socket. This matters because Coolify deliberately withholds `REDIS_URL` at build time.

## Storage

Entries are `v8.serialize`d, because page values carry `Buffer`s and a `Map<string, Buffer>` that JSON would destroy. `unstable_cache` entries survive a redeploy; rendered pages are namespaced per build id so a new build cannot serve the previous build's payloads.

```
<prefix>:v1:fetch:<key>          unstable_cache — survives deploys
<prefix>:v1:<buildId>:page:<key> APP_PAGE / APP_ROUTE / PAGES / IMAGE
<prefix>:v1:tag:<tag>            SET of entry keys
<prefix>:v1:rt:<tag>             invalidation marker, 30d TTL
```

## Tag invalidation

Tags are indexed as Redis sets, so `revalidateTag` deletes exactly the affected keys — no `KEYS` or `SCAN`. It also writes a per-tag invalidation marker, which is what covers implicit `_N_T_` soft tags (Next never stores those on `set`, so they can never appear in an index) and any index lost to `allkeys-lru` eviction.

## Redis instance requirements

The instance must be configured as a **cache**: `maxmemory` plus `maxmemory-policy allkeys-lru`. Entries carry no TTL by design, so under Redis's actual default (`noeviction`) the keyspace grows until writes fail with OOM and invalidations start being dropped — a worse failure mode than the eviction this design already handles. Documented in the README.

## One incompatibility found and fixed

`app/api/notifications/[id]/route.ts` uses `runtime = 'edge'`, and Next bundles the cache handler into the edge chunk, where a top-level `redis` import fails to resolve `node:crypto` and **breaks the build**. All Node-only imports are therefore lazy. The route itself is unchanged.

## Verification

- Default build (no env vars) registers no handler — `required-server-files.json` shows `cacheHandler: undefined`.
- Coolify build (`USE_REDIS_CACHE=true`, no `REDIS_URL`) succeeds. Measured against a live Redis with a reachable `REDIS_URL` present: **zero connections, zero keys written** during the build.
- Runtime against Redis 7.0: MISS → SET → HIT for both `FETCH` and `APP_PAGE`, page keys correctly namespaced under the real build id, entry TTL `-1` (no imposed expiry), one shared connection.
- 11/11 semantics checks pass, including Buffer/`Map` survival, precise tag deletion, array tags, `{ expire: 0 }`, soft-tag invalidation via marker, no false expiry on writes after invalidation, and graceful degradation on an unreachable Redis.
- Cross-process invalidation confirmed end-to-end: an external process invalidated a tag and the running server saw the miss and re-fetched.
- `eslint` clean for the changed files.

## Known cost

Warm renders are **~2 ms slower**, because `cacheMaxMemorySize: 0` makes every read a network hop where the default keeps a 50 MB in-process LRU. Five batches of 20 requests per configuration:

| | Default | Redis |
|---|---|---|
| `/contributors` | 2.9–3.8 ms | 4.6–6.3 ms |
| `/terms` | 2.7–3.4 ms | 4.6–5.9 ms |

(An earlier revision of this description said 4–6 ms. That came from single unreplicated samples and overstated the cost by roughly 2×; run-to-run variance on identical code is ~1.7 ms, so the figures above are medians over repeated batches.)

Measured on trivial static pages, so still worth re-checking on a real page in Coolify.

## Rollback

Unset `USE_REDIS_CACHE` and rebuild — returns to stock Next.js caching with no code change.

---

Note for deployment: the build requires `NEXT_PUBLIC_SUPABASE_URL` / `NEXT_PUBLIC_SUPABASE_ANON_KEY` to prerender `/confirm-email`. Pre-existing behaviour, but Coolify's build environment needs them present.

https://claude.ai/code/session_019oMe8t68sQ2TXDavdvJtgT

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,3 +13,26 @@ NEXT_PUBLIC_APP_URL=http://localhost:3000
 # HMAC key for one-click email unsubscribe tokens. MUST match the UNSUBSCRIBE_SECRET
 # set on the send-notification-email edge function (supabase secrets set).
 UNSUBSCRIBE_SECRET=a-long-random-string
+
+# --- Redis-backed Next.js cache (self-hosted / Coolify only) ---
+# Leave false on Vercel: with this unset the app uses Next's built-in Data Cache
+# exactly as before. Set to true only where a Redis instance is available.
+# Must be present at BUILD time, because next.config.js reads it to decide whether
+# to register the custom cacheHandler.
+USE_REDIS_CACHE=false
+
+# SECRET. Runtime only - it is intentionally NOT available during `next build`,
+# and the build must succeed without it. Only needed when USE_REDIS_CACHE=true;
+# if it is missing the handler falls back to a local in-memory cache.
+REDIS_URL=
+
+# Optional. Key namespace, default: munda-manager:next-cache
+# Only matters if two apps share one Redis instance.
+REDIS_CACHE_PREFIX=
+
+# Optional. Per-operation timeout in ms, default: 1000
+REDIS_CACHE_TIMEOUT_MS=
+
+# Optional. Set to 1 to log cache GET/SET/HIT/MISS and tag invalidations.
+# Logs keys and counts only, never cached values.
+NEXT_PRIVATE_DEBUG_CACHE=
--- a/README.md
+++ b/README.md
@@ -627,3 +627,42 @@ interface FighterProps {
 ```
 
 
+
+## Self-Hosted Caching (Coolify)
+
+On Vercel the app uses Next's built-in Data Cache and nothing here applies. When
+self-hosting, `cache-handler.js` can back the Next.js `cacheHandler` with Redis so
+cached data is shared between instances and survives a redeploy.
+
+It is off unless `USE_REDIS_CACHE=true` is set **at build time** — `next.config.js`
+reads it to decide whether to register the handler at all. `REDIS_URL` is a runtime
+secret and is deliberately not needed during `next build`; without it the handler
+falls back to an in-process cache, which is what makes the build work.
+
+| Variable | Required | Notes |
+|---|---|---|
+| `USE_REDIS_CACHE` | build | `true` enables the handler |
+| `REDIS_URL` | runtime | Secret. Absent ⇒ in-memory fallback |
+| `REDIS_CACHE_PREFIX` | no | Default `munda-manager:next-cache` |
+| `REDIS_CACHE_TIMEOUT_MS` | no | Default `1000` |
+| `NEXT_PRIVATE_DEBUG_CACHE` | no | `1` logs HIT/MISS/SET and invalidations |
+
+`unstable_cache` entries are stored under `…:v1:fetch:` and persist across deploys;
+rendered pages are stored per build id so a new build never serves the previous
+build's payloads. Tags from `utils/cache-tags.ts` are indexed as Redis sets, so
+`revalidateTag` deletes exactly the affected keys rather than scanning the keyspace.
+
+To confirm Redis is receiving entries:
+
+```bash
+redis-cli -u "$REDIS_URL" DBSIZE
+redis-cli -u "$REDIS_URL" --scan --pattern 'munda-manager:next-cache:*' | head -50
+redis-cli -u "$REDIS_URL" SMEMBERS 'munda-manager:next-cache:v1:tag:gang-<GANG_ID>'
+redis-cli -u "$REDIS_URL" MONITOR   # watch live traffic while browsing
+
+# flush the app's cache (e.g. after changing the shape of a cached value)
+redis-cli -u "$REDIS_URL" --scan --pattern 'munda-manager:next-cache:*' \
+  | xargs -r -n 500 redis-cli -u "$REDIS_URL" DEL
+```
+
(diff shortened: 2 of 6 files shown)


joeseos:
## Code Review

@joeseos Reviewed `cache-handler.js`, the `next.config.js` wiring, and the docs against the rest of the codebase. This is a well-isolated, off-by-default change — no `unstable_cache` call sites touched, correctly gated behind `USE_REDIS_CACHE`, and the edge-runtime / build-time degradation paths are handled thoughtfully (lazy Node imports, `NEXT_PHASE` + `REDIS_URL` gate, `error` listener registered so node-redis doesn't crash the process on a socket error). Traced through `get`/`set`/`revalidateTag` in detail; found no data-corruption or stale-serving bugs. A few things worth addressing before/after merge:

### Findings

**1. `BUILD_ID` is read from disk synchronously on every page/route cache request, not just once per process** — Medium
`cache-handler.js:311-325` (`#entryKey`)
The file's own comment says "Next builds a handler per request," and `this.buildId` is reset to `undefined` in the constructor every time. So for any `kind !== 'FETCH'` (i.e. every rendered-page cache get/set), `#entryKey` re-triggers `readFileSync(...BUILD_ID)` because the per-instance memoization never survives past a single request. `loadNode()`'s module-level cache only dedupes the dynamic import, not this read.
Impact: a blocking sync disk read on the hot path of every page cache hit/miss, instead of once per process — works against the async-everywhere design used elsewhere in this file, and adds up under load (this is likely also part of the measured 4–6 ms warm-render regression called out in the PR description).
Fix: hoist `buildId` resolution to a module-level `let cachedBuildId` (or an async memoized promise), computed once and reused by every handler instance.

**2. Design assumes Redis is configured with `maxmemory-policy allkeys-lru`, but this isn't called out as a requirement anywhere a deployer would see it** — Medium
`cache-handler.js` (no TTL is ever set on `SET`/tag-set writes — confirmed intentional, "entry TTL -1" per the PR description) + `README.md` new table
The whole "recovery via 30-day invalidation marker" design exists specifically to paper over entries or tag-indices lost to `allkeys-lru` eviction. If the Redis instance is left on Redis's actual default (`noeviction`), the keyspace grows unbounded until writes start failing with OOM instead of quietly evicting — a materially different failure mode than what's designed for. This is mentioned only in a source comment, not in the README variable table or setup instructions where an operator standing up the Coolify Redis instance would actually see it.
Fix: add a line to the README's Self-Hosted Caching section stating `maxmemory-policy allkeys-lru` (or similar) is required on the Redis instance.

**3. Small race in `revalidateTag` can drop a fresh write** — Low, self-healing
`cache-handler.js:230-256`
Tag members are read (`SMEMBERS`) in one `MULTI`, then deleted by key in a second `MULTI`. A `set()` for the same tag landing in the gap between those two round-trips gets its brand-new key blindly `DEL`'d by the second multi (it's already in the captured `keys` set). Doesn't cause stale-data serving — the marker already covers correctness — just an avoidable extra cache miss on next read. Not blocking, but worth a comment noting it's a known, accepted tradeoff if intentional.

**4. Consider a note on securing `REDIS_URL`** — Low (security, operational)
Cached entries can include arbitrary `unstable_cache`-wrapped fetch payloads. Worth a one-line README callout that the Redis instance should not be network-exposed / should require auth+TLS, same as any other datastore holding potentially sensitive response data — easy to overlook on a self-hosted box.

### Checklist
- Correctness: mostly ✅, see #1 (perf-adjacent correctness) and #3 (benign race)
- Edge cases / null handling: ✅ — edge runtime degradation, missing `REDIS_URL` at build, corrupted payload decode (`#decode` drops and treats as miss), Redis unreachable at runtime (falls back to in-memory LRU) all handled
- Security: ✅ code-wise; see #4 for an operational doc gap
- Performance: see #1; the measured warm-render slowdown is plausible and probably partly explained by it
- Readability/maintainability: good — comments explain *why*, not *what*; PR description matches implementation
- Project conventions: no `CLAUDE.md` in this repo to check against; style matches surrounding files
- DRY/YAGNI: ✅ — no scope creep, application code untouched as claimed

### Verdict: **Approve with suggestions** (non-blocking; #1 and #2 are worth doing before/soon after a production Coolify rollout, not blockers to merging behind the off-by-default flag)

---

PR #2071, opened 2026-08-27
feat(gang): add dedicated Resources modal with optional log reasons

Djidiouf:
Move credits/reputation/trade points/campaign resources out of Edit Gang into their own modal, and append optional per-resource reasons to gang logs.

### DIFF
--- a/app/actions/logs/gang-resource-logs.ts
+++ b/app/actions/logs/gang-resource-logs.ts
@@ -12,11 +12,46 @@ export interface LogGangResourceChangesParams {
   oldState: GangResourceState;
   newState: GangResourceState;
   user_id?: string;
+  /** Optional per-resource reasons keyed by the same names used in oldState/newState */
+  reasons?: Record<string, string>;
+}
+
+function capitalizeResourceName(resourceName: string): string {
+  return resourceName.charAt(0).toUpperCase() + resourceName.slice(1);
+}
+
+/** First line only: "Reputation increased by 5 (reason)" */
+function formatResourceDeltaSummary(
+  displayName: string,
+  oldValue: number,
+  newValue: number,
+  reason?: string
+): string {
+  const delta = Math.abs(newValue - oldValue);
+  const direction = newValue > oldValue ? 'increased' : 'decreased';
+  return reason
+    ? `${displayName} ${direction} by ${delta} (${reason})`
+    : `${displayName} ${direction} by ${delta}`;
+}
+
+/**
+ * Two-line resource change description, e.g.:
+ * Reputation increased by 5 (reason)
+ * Reputation: 10 → 15
+ */
+function formatResourceChangeDescription(
+  displayName: string,
+  oldValue: number,
+  newValue: number,
+  reason?: string
+): string {
+  const summary = formatResourceDeltaSummary(displayName, oldValue, newValue, reason);
+  return `${summary}\n${displayName}: ${oldValue} → ${newValue}`;
 }
 
 export async function logGangResourceChanges(params: LogGangResourceChangesParams): Promise<GangLogActionResult> {
   try {
-    const { gang_id, oldState, newState, user_id } = params;
+    const { gang_id, oldState, newState, user_id, reasons } = params;
     const logPromises: Promise<GangLogActionResult>[] = [];
 
     // Check if credits, rating, or wealth changed - use new format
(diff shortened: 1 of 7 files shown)


joeseos:
@Djidiouf Reviewed this PR against the codebase (extraction of resource editing into `GangResourcesModal`, plus optional per-resource reasons threaded into gang logs).

## Summary

This is a clean extraction: the credits/reputation/trade-points/campaign-resource UI moves out of `GangEditModal` into `GangResourcesModal` with no behavioral drift in the moved logic, and the new "reason" plumbing is threaded consistently end-to-end:

- `gang-resources-modal.tsx` → `credits_reason` / `reputation_reason` / `trade_points_reason` / `resources[].reason`
- `gang.tsx` → maps `resourceUpdates` into `resources` (looking up `resource_name` correctly) before calling `updateGang`
- `update-gang.ts` → builds a `reasons` record keyed to match the exact keys used in `oldState`/`newState` (including the `'trade points'` space-separated key), so `logGangResourceChanges` picks up reasons correctly
- `sanitizeResourceReason` caps length (100 chars, matches the `Input maxLength`) and collapses whitespace both client- and server-side
- The new two-line log description relies on `whitespace-pre-line` already applied to the description column in `log-modal.tsx`, so the `\n` renders as an actual line break — verified this isn't a display regression

I traced the reason keys and resource-name lookups all the way through (modal → `gang.tsx` → `update-gang.ts` → `gang-resource-logs.ts`) since that's the part most likely to silently break, and didn't find a mismatch.

## Findings

**Low — `components/gang/gang.tsx` (~line 608-613), toast copy inconsistency**
When a user opens the new "Resources" modal and only changes credits/reputation/trade points (no campaign resource), `resourceOnly` evaluates to `false` (since those keys aren't in the `['resources','campaign_gang_id']` allowlist), so the toast reads "Gang updated successfully" instead of something resource-flavored. Minor, but slightly confusing given the user clicked a button labeled "Resources". Not a regression (matches prior generic-update copy), just worth a look if you want the copy to feel intentional for the new modal.

**Low — `components/gang/gang-resources-modal.tsx`, `handleSave`**
When `Object.keys(updates).length === 0` (user opens the modal and clicks Save without changing anything), `handleSave` returns `false` silently — the modal stays open with no toast/explanation. Not harmful, just a small UX gap (pre-existing `GangEditModal` always closed regardless, so this is new but arguably more correct — just lacks feedback).

Nothing else stood out: no missing null checks I could break, no new negative-quantity validation gaps beyond the pre-existing credits-only guard (unchanged from before this PR), no XSS/injection concerns (reason text renders through plain JSX, not `dangerouslySetInnerHTML`), and the edition-gating for Trade Points (`hasTradePoints`) is preserved server-side.

## Verdict: **Approve**

No high/medium issues found. The two low-severity items are cosmetic/UX polish, not correctness bugs — happy to see this merged as-is or with either nit addressed.

---

joeseos:
@Djidiouf Reviewed the latest commit (`596a004`), diffed against the previously-reviewed commit (`f4a645f`).

## Previous findings

**Low — toast copy inconsistency (`gang.tsx` ~608-613): RESOLVED.**
`isResourceUpdate` now checks against an explicit `resourceUpdateKeys` set that includes `credits`/`credits_operation`/`credits_reason`/`reputation`/... /`trade_points_reason`, not just `resources`/`campaign_gang_id`. Saving only credits/reputation/trade points from the new Resources modal now correctly shows "Resources updated successfully". I traced this against `handleGangUpdate` — the `resourceUpdates` key is always deleted and replaced with `resources` before `mutateAsync` is called, so the key set matches what actually reaches `onSuccess`'s `variables`. No case where a resource-only save would still fall through to the generic message.

**Low — silent no-op on empty save (`gang-resources-modal.tsx`, `handleSave`): RESOLVED.**
`handleSave` now calls `toast.message('No changes to save')` before returning `false`. Confirmed in `components/ui/modal.tsx`'s `handleConfirm` that a `false` return only skips `onClose()` and otherwise no-ops (no competing toast), so this is the only feedback shown and there's no double-toast.

## New issues in this diff

None. The change is a tight, 19-line diff scoped exactly to the two previously flagged items — no new logic paths, no new null-handling surface, and `toast.message` is a valid Sonner v2 API (`sonner@^2.0.7` in `package.json`) even though it's the first use of it in this codebase.

## Verdict: **Approve**

---

PR #2072, opened 2026-08-28
Add rank-based Skill Access for N26 Venator gangs

v0rp-z:
## Summary

N26 Venator gangs get their own skill-access rule (gang picks 4 Skill Sets, ranks them 1-4, and each Hunt Leader / Hunt Champion / Specialist fighter gets Primary/Secondary access derived from those ranks). We have no logic for this currently, so Venator gang fighters get the "Skill Set not accessible" error and users have to work around it.

This PR adds a new "Skill Access" section to **Edit Gang** for Venator gangs on N26, wires it to a new `gang_skill_set_ranks` table, and auto-writes the derived per-fighter overrides into the existing `fighter_skill_access_override` table on gang rank save, fighter add, subtypes change, and copy.

## Testing

- Manual browser walkthrough on a locally-seeded Venator gang

#### Skill Access panel

1. **Panel**
   - Open Edit Gang on a non-Venator gang should be *no* Skill Access section.
   - Open Edit Gang on a Venator gang should show Skill Access section visible with 4 numbered dropdowns.

2. **Initial set** — pick 4 different skill sets and save. Expected: save succeeds, dropdowns retain the choices on reopen.

3. **Cross-exclusion** — pick skill X in rank 1, then rank 2/3/4 dropdowns must not offer X. Clear rank 1, and X reappears everywhere.

4. **Partial save rejected** — fill 1–3 slots (leave the rest blank), save, toast fires, modal stays open, DB unchanged.

5. **Unchanged re-save is a no-op** — reopen Edit Gang, don't touch ranks, save, *no* confirmation dialog.

6. **Rank change triggers warning** — change one rank's skill set,  save, confirm dialog appears.
   - Cancel = no DB change.
   - Accept = new ranks, fighter access recomputed.

#### Fighter skill picker (rulebook mapping)

7. **Hunt Leader** — add one to the Venator gang,  open skill picker, rank 1 & 2 skills are Primary, rank 3 & 4 are Secondary. All four ranks pickable.

8. **Hunt Champion** — rank 1 & 2 Primary, rank 3 Secondary. Rank 4 skill picker shows "not accessible" error.

9. **Hunt Specialist** — rank 1 Primary, rank 2 & 3 Secondary. Rank 4 shows "not accessible".

#### Message copy

10. **Empty-state (Venator, no ranks)** — new Venator gang, no ranks set, add a Leader,  try to pick a skill, error reads *"Your gang hasn't ranked its Skill Sets yet. Set them in Edit Gang - Skill Access."*

11. **Per-rank restriction (Venator, ranks set)** — Champion tries to pick a rank 4 skill, error reads the *original* copy: *"This Skill Set is not accessible to this fighter. Change their Skill Set access in: Edit Fighter,  Customise Skill Set Access."* (NOT the Venator message.)

#### Subtype lifecycle

12. **Promotion picks up new access** — add a Ganger (no auto access), then edit fighter, change subtype to Leader, save, fighter's skill picker now shows Primary/Secondary per the ranks.

13. **Demotion cleans up stale access** — add a Leader (auto access), then edit fighter, change subtype to Ganger,  save,  fighter's skill picker no longer offers the previously-Primary sets; picking one shows "not accessible".

14. **Fighter copy** — copy a Hunt Leader, the copy has the same Primary/Secondary access as the original.

#### Permissions

15. **Arbitrator can save ranks** — as a campaign arbitrator (not gang owner), open the Venator gang's Edit Gang, Skill Access, save 4 ranks, succeeds.

16. **Manual override still works** — on any fighter, Edit Fighter, Customise Skill Set Access, set a skill to Primary, save, fighter's picker reflects it.

#### Regression

17. **Non-Venator gang** — pick any non-Venator gang, add a fighter, browse the skill picker, no behavior change, no missing skills, no errors.

18. **Outcasts archetype preserved** — create an Underhive Outcasts gang with an archetype-backed fighter, verify the archetype's granted skill access is present. Add/edit/copy that fighter, access is still there afterwards (not wiped by the Venator sync).

## UI
**Desktop:** 
Screenshot From 2026-08-28 01-53-31Screenshot From 2026-08-28 01-53-42Screenshot From 2026-08-28 01-54-09

**Mobile:**
Screenshot From 2026-08-28 02-04-02


joeseos:
@v0rp-z Follow-up review on the latest sync (head `f07400b`).

## No new code changes since the last review

Head `f07400b` is the same commit that was already present when the previous review (comment above) was posted — every one of the 13 feature commits pre-dates that review's timestamp. The only thing that happened afterward was a `--force-with-lease` rebase that dropped three unrelated docs commits (spec + plan). So there's no new diff to review, and none of the previously flagged issues have been addressed yet.

## Previous findings — status

| Finding | Status |
|---|---|
| `syncFighter` wipes archetype-derived overrides for non-Venator gangs (`add-fighter.ts:1215`, `edit-fighter.ts:1745`) | **Outstanding.** Both call sites are still unconditional. Worth noting: the last commit (`f07400b`) extracted `isVenatorGang()` into `utils/venatorSkillAccess.ts` — exactly the helper the suggested fix needs — but it was never wired into these two call sites. The `add-fighter.ts` comment ("no-op for non-Venator gangs") is still inaccurate: for a non-Venator gang, `deriveOverrides` returns `[]`, and `rewriteFighterOverrides` still runs its unconditional `DELETE` before skipping the insert, so it actively wipes any existing `fighter_skill_access_override` rows (e.g. archetype-derived ones) rather than being a no-op. |
| Arbitrator permission gap on `gang_skill_set_ranks` RLS | **Outstanding.** `20260827120000_create_gang_skill_set_ranks.sql` INSERT/UPDATE/DELETE policies are unchanged — still owner/admin only, no `campaign_gangs`/`is_arb` clause. |
| Non-atomic delete-then-insert in `save-venator-skill-ranks.ts` | **Outstanding.** Still two separate round trips with no rollback path if the insert fails after the delete succeeds. |
| Misleading "hasn't ranked its Skill Sets yet" message | **Outstanding.** `fighter-skills-list.tsx:427-435` (and the advancement-list equivalent) still show this message for *any* denied access under `isVenatorGang`, including when ranks are set but the subtype simply has no access to that rank by rulebook design. |
| Minor: duplicated `VENATOR_SUBTYPES` list | **Outstanding.** Still separately defined in `syncVenatorSkillOverrides.ts:31` and `venatorSkillAccess.ts:52` — a missed opportunity given the same commit refactored `isVenatorGang`/`VENATOR_GANG_TYPE_NAME` into the shared module. |
| Minor: sequential per-fighter sync in `syncGang` | **Outstanding.** Still a `for` loop awaiting `rewriteFighterOverrides` one fighter at a time. |

## New findings

None — no code changed since the last review.

## Verdict

**Changes requested** (unchanged from the previous review). The most severe item — `syncFighter` silently erasing archetype-derived skill overrides on Underhive Outcasts gangs via `add-fighter`/`edit-fighter` — is a real data-loss bug on an existing feature and should be gated on `isVenatorGang(editionSlug, gangType)` (now available) before this merges.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `dccdd21e`), covering the work since the last review (which stopped at `f07400b`): commits `ada00a4`, `c285f4c`, `78a807e`, `055a4a1`, plus the merge-conflict resolution in `dccdd21e`.

## Previous findings — status

| Finding | Status |
|---|---|
| `syncFighter` wipes archetype-derived overrides for non-Venator gangs | **Resolved, with a new related regression** — see below. Instead of gating on `isVenatorGang()` as suggested, `ada00a4` makes `syncFighter`/`syncGang` return early when the gang has no `gang_skill_set_ranks` rows, and `rewriteFighterOverrides` now checks `overrides.length === 0` *before* deleting rather than after. Since `gang_skill_set_ranks` is only ever populated via the Venator save flow, this closes the original archetype-wipe scenario just as well as the suggested fix. |
| Arbitrator permission gap on `gang_skill_set_ranks` RLS | **Resolved.** `78a807e`'s new migration rewrites INSERT/UPDATE/DELETE policies to add the `campaign_gangs` / `private.is_arb` carve-out, matching the pattern used elsewhere (e.g. `fighter_ooa_records`, `battle_sessions`). |
| Non-atomic delete-then-insert in `save-venator-skill-ranks.ts` | **Resolved.** `78a807e` adds `public.save_gang_skill_set_ranks(p_gang_id, p_ranks)`, a `SECURITY INVOKER` plpgsql function doing the delete + insert in one transaction (still subject to the table's RLS as the invoker), called via `supabase.rpc(...)` instead of two round trips. |
| Misleading "hasn't ranked its Skill Sets yet" message | **Resolved.** `c285f4c` threads a `hasGangRanks` boolean (probed once server-side in `app/fighter/[id]/page.tsx`) through `fighter-page.tsx` → `SkillsList`/`AdvancementsList` → `SkillModal`/`AdvancementModal`, and the message now only fires when `isVenator && !hasGangRanks`, leaving the generic message for per-rank access restrictions. |
| Duplicated `VENATOR_SUBTYPES` list | **Resolved.** Single definition now lives in `utils/venatorSkillAccess.ts`; the copy in `syncVenatorSkillOverrides.ts` is gone, and `syncGang`'s fighter filter now goes through `deriveOverrides` itself instead of a separate subtype check. |
| Sequential per-fighter sync in `syncGang` | **Resolved.** `ada00a4` replaced the per-fighter loop with one batched `DELETE ... IN (fighter_ids)` and one multi-row `INSERT`. |

## New findings

**Medium — `utils/syncVenatorSkillOverrides.ts:41-61`: stale skill-access overrides aren't cleaned up when a fighter loses Venator-subtype eligibility.**

`syncFighter` now short-circuits before calling `rewriteFighterOverrides` in two places:
```ts
const ranks = await readGangRanks(fighter.gang_id, supabase);
if (ranks.length === 0) return;
...
const overrides = deriveOverrides(ranks, subtypes);
if (overrides.length === 0) return;   // <-- also skips syncGang's per-fighter equivalent
await rewriteFighterOverrides(fighterId, supabase, actingUserId, overrides);
```
and `rewriteFighterOverrides` itself now only deletes when `overrides.length > 0` (previously it always deleted, then conditionally inserted). That fixes the archetype-wipe bug, but it also means: for a Venator gang that **does** have ranks configured, if a fighter's `fighter_subtypes` are edited so they no longer include `Leader`/`Champion`/`Specialist` (e.g. a promotion/demotion, or `championLeaderTypePromotion` — which is enabled for N26, the same edition Venators use), `deriveOverrides` now returns `[]` for that fighter and `syncFighter` returns before ever deleting their previously-written rows. Those `fighter_skill_access_override` rows — which this feature itself wrote — are left in place indefinitely, so the fighter keeps whatever Primary/Secondary skill-set access they were granted under their old subtype. `edit-fighter.ts:1742-1748` calls `syncFighter` on every `fighter_subtypes` change, so this is reachable through the normal subtype-edit path, not just an edge case. `syncGang`'s equivalent batching (`targets = ... .filter(t => t.overrides.length > 0)`) has the same gap, though it's less likely to be hit in practice since it only runs right after a gang's ranks are (re)saved.

Suggested fix: only use `ranks.length === 0` as the "not a managed Venator gang, don't touch the table" guard. Once `ranks.length > 0`, still call `rewriteFighterOverrides` (i.e. still delete) even when the newly-derived `overrides` is empty for a given fighter — that fighter's own prior rows are legitimately ours to clear.

**Low — `app/fighter/[id]/page.tsx`: `gang_skill_set_ranks` probe runs for every gang, not just Venator ones.**

The `hasGangRanks` lookup added in `c285f4c` runs unconditionally in the `Promise.all` for every fighter page load, even for non-N26/non-Venator gangs where it will always be empty. Not a correctness issue, just an avoidable query — worth gating on `isVenatorGang(editionSlug, gang_type)` if that's cheap to check before the query, but low priority.

## Verdict

**Changes requested.** All six previously-flagged issues are functionally addressed and the fixes are solid (atomic RPC, correct arbitrator RLS pattern, correct message gating). The one open item is the medium-severity staleness bug above — it's a much smaller blast radius than the original archetype-wipe bug, but it's a real correctness gap in the same code path and worth closing before merge.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `db11036`), covering only the newest commit since the last review (which stopped at `dccdd21`): `db11036a` — `app/fighter/[id]/page.tsx` (15 lines) and `utils/syncVenatorSkillOverrides.ts` (35 lines).

## Previous findings — status

| Finding | Status |
|---|---|
| `syncFighter` wipes archetype-derived overrides for non-Venator gangs | Resolved (in `ada00a4`, confirmed still holding — `ranks.length === 0` guard is untouched by this commit). |
| Arbitrator permission gap on `gang_skill_set_ranks` RLS | Resolved (in `78a807e`, unaffected by this commit). |
| Non-atomic delete-then-insert in `save-venator-skill-ranks.ts` | Resolved (in `78a807e` via the `save_gang_skill_set_ranks` RPC, unaffected by this commit). |
| Misleading "hasn't ranked its Skill Sets" message | Resolved (in `c285f4c`, unaffected by this commit). |
| Duplicated `VENATOR_SUBTYPES` list | Resolved (in `ada00a4`, unaffected by this commit). |
| Sequential per-fighter sync in `syncGang` | Resolved (in `ada00a4`, batching preserved by this commit). |
| **Medium — stale overrides not cleaned up when a fighter loses Venator-subtype eligibility** | **Resolved.** `rewriteFighterOverrides` now always runs its `DELETE` before the `overrides.length === 0` check (previously an early return skipped the delete entirely), and `syncFighter` no longer short-circuits on empty derived overrides — only on `ranks.length === 0` (the "not a managed Venator gang" gate, still intact). `syncGang`'s batched path gets the same treatment: `targets` is no longer pre-filtered to `overrides.length > 0` before the batched `DELETE ... IN (fighter_ids)`, so every fighter in a ranked gang gets cleared; only the `INSERT` payload is filtered to non-empty overrides (and short-circuits via `rows.length === 0`). Traced through both a Leader→Ganger demotion and the N26 `championLeaderTypePromotion` path — `syncFighter`/`syncGang` correctly clear the stale rows in both. |
| **Low — `hasGangRanks` probe ran for every gang** | **Resolved.** The probe moved out of the parallel `Promise.all` into a separate block gated on `gangBasic && isVenatorGang(gangBasic.edition_slug, gangBasic.gang_type)`. Verified `isVenatorGang`'s signature/null-handling matches what's passed, and `gangBasic.edition_slug`/`gang_type` are the correct field names from `getGangCore`. Non-Venator gangs skip the query; Venator gangs take one extra sequential round trip after the batch, as intended. |

## New findings

None. Reviewed both changed files in full — the delete-then-conditionally-insert restoration in `rewriteFighterOverrides`/`syncGang` is correct and doesn't reintroduce the original archetype-wipe bug (still gated on `ranks.length === 0` before ever touching the table), and the `hasGangRanks` gating change is a straightforward, correctly-ordered refactor with no behavior change for non-Venator gangs.

## Verdict

**Approve.** All six findings from the original review and both findings from the second round are now resolved, and this commit doesn't introduce anything new.

---

joeseos:
@v0rp-z Reviewed head `db11036` specifically for **bloat / DRY / YAGNI** — not correctness. The six findings from the earlier rounds all still read as resolved; nothing below is a regression of those.

Every claim here was checked against the files at `db11036` rather than the diff. Two candidate findings were dropped during that check and I've noted them at the bottom so they don't get re-raised.

## 1. The skill-types fetch reimplements an existing API route

`components/gang/gang-edit-modal.tsx:136-145`:

```ts
const { data, error } = await supabase
  .from('skill_types')
  .select('id, name, editions:edition_id (slug)')
  .order('name');
if (error) throw error;
return (data ?? [])
  .filter((r: any) => (r.editions?.slug ?? null) === editionSlug)
```

`app/api/skill-types/route.ts` already does this, edition-scoped **in SQL** (`route.ts:69-73`), and six call sites already use it — including `components/customise/custom-fighters.tsx:601` and `custom-skills.tsx:69` with the identical `?edition_slug=${editionSlug}` shape this modal needs. This is also the only direct `from('skill_types')` client-side query in `components/`.

Beyond the duplication, the hand-rolled query drops two behaviours the route has:

- **Over-fetch.** It pulls every `skill_types` row across all editions and filters in JS.
- **No N26→N23 fallback.** `route.ts:78-91` falls back to the N23 catalog when an edition has no seeded skill types, with the comment *"N26 skill_types are not seeded yet"*. `supabase/seed.sql:111-119` seeds 8 skill types, all under edition `…023`, none for N26. Venators are N26-only, so **if that still holds in your target DB the four dropdowns come up empty**. Your screenshots show them populated, so N26 is presumably seeded in your environment now — but the raw query has no safety net if any edition isn't, and the route does. Worth a quick confirm either way.

(If you switch to the route, the only extra step is filtering `!t.is_custom`, unless custom skill sets should be rankable — which is its own question.)

## 2. `&gt;` now renders literally in the "not accessible" message — both files

This one is a real user-visible regression introduced by the refactor, so flagging it even though it's outside the bloat brief.

Before, the text was JSX text content, where the JSX transform decodes HTML entities:

```jsx
access in: Edit Fighter &gt; Customise Skill Set Access.   // renders: Edit Fighter > Customise
```

Now it's a string literal inside an expression container (`fighter-skills-list.tsx:430-437`, `fighter-advancement-list.tsx:2814-2821`). Entities are **not** decoded there — React sets it as text content verbatim, so users will see:

> This Skill Set is not accessible to this fighter. Change their Skill Set access in: Edit Fighter **&gt;** Customise Skill Set Access.

The adjacent Venator branch uses a raw `→` correctly, which suggests the `&gt;` was carried over unintentionally. Fix is `>` (or `→` for consistency).

## 3. The same block is duplicated verbatim in two files

`fighter-skills-list.tsx:430-437` and `fighter-advancement-list.tsx:2814-2821` are identical apart from indentation, both as an inline IIFE inside JSX:

```tsx
{(() => {
  const isVenator = isVenatorGang(editionSlug, gangType);
  return isVenator && !hasGangRanks
    ? "Your gang hasn't ranked its Skill Sets yet. Set them in Edit Gang → Skill Access."
    : "This Skill Set is not accessible to this fighter. Change their Skill Set access in: Edit Fighter &gt; Customise Skill Set Access.";
})()}
```

Both files already import from `@/utils/venatorSkillAccess`, so a `skillAccessDeniedMessage(editionSlug, gangType, hasGangRanks)` export there removes the duplication and the IIFE — and means finding #2 gets fixed in one place instead of two.

## 4. `gangType` is a new prop threaded four layers deep

`gang_type` (the display name) is added at `app/fighter/[id]/page.tsx:264`, then through `fighter-page.tsx` (the `Gang` interface, `transformFighterData`, two call sites), then `SkillsListProps` + `SkillModalProps`, then `AdvancementsListProps` + `AdvancementModalProps` — roughly 16 lines of pure plumbing.

Two things shrink this:

- **`gangType` and `hasGangRanks` are always consumed together**, only ever as `isVenator && !hasGangRanks`. That's one server-computed boolean, not two props on four interfaces.
- **`gang_type_id` was already on that payload** (`page.tsx:266`, pre-existing) and already reaches `fighter-page.tsx:355`/`367`. The repo's precedent for "is this gang type X" is ID-based — `archetypeEligibility.ts:6-7` uses `N23_UNDERHIVE_OUTCASTS_GANG_TYPE_ID` / `N26_UNDERHIVE_OUTCASTS_GANG_TYPE_ID` — whereas `isVenatorGang` does `(gangType ?? '').toLowerCase() === 'venators'`.

To be fair to the name-based choice: it also catches a **custom** gang type named "Venators", which an ID check wouldn't, since those carry `custom_gang_type_id` and a null `gang_type_id`. So it's a genuine tradeoff rather than a defect — but it's worth being deliberate about, because the display-name match is what forced the whole new prop path, and it breaks on a rename.

## 5. Two migrations where the second exists only to undo the first

`20260827120000_create_gang_skill_set_ranks.sql` creates three RLS policies; `20260828120000_add_arbitrator_to_gang_skill_set_ranks_policies.sql` drops all three and recreates them with the arbitrator clause. 60 lines whose entire purpose is reverting 25 lines added in the same unmerged PR, permanently in the migration history.

Worth squashing — **unless** `20260827120000` has already been applied to a shared preview/dev database, in which case squashing isn't free and this is fine as-is. Your call, you know the deploy state.

While in there: the **UPDATE** policy is written twice (once per migration) for an operation nothing performs. Confirmed by grepping every `gang_skill_set_ranks` reference at head — the only writes are the RPC's `DELETE` + `INSERT` (`20260828130000…sql:10-14`); the rest are `SELECT`s in `page.tsx:58`, `gang-edit-modal.tsx:156`, and `syncVenatorSkillOverrides.ts:7`.

## 6. Four parallel `useState`s instead of one array

`gang-edit-modal.tsx:165-168` declares `rank1..rank4` / `setRank1..setRank4`, which forces the 4-way expansion to be written out four separate times:

- the `useEffect` at `:170-176` (`byRank.get(1)` … `get(4)`)
- the unchanged-check at `:497-502` (`byRank.get(1) === rank1 && …`)
- the `ranks:` payload literal at `:507-512`
- the render array at `:625-630`

One `useState<string[]>(['', '', '', ''])` collapses all four into loops over the rank indices.

Minor related note: the `useEffect` deps on `existingRanks`, which is `data: existingRanks = []` — that default is a fresh array identity every render, so the effect re-runs on every render until the query resolves (and forever, for non-Venator gangs where the query is disabled). It settles harmlessly because React bails out when all four `setState` calls pass the same `''`, and your global `refetchOnWindowFocus: false` (`app/providers/query-client-provider.tsx:17`) rules out the mid-edit-reset scenario I checked for. So: wasted work, not a bug.

## Minor

- **`revalidateTag` instead of the house helper** — `save-venator-skill-ranks.ts:49` calls `revalidateTag(TAGS.gang(gangId), { expire: 0 })`. `invalidateGang` (`utils/cache-tags.ts:70`) is exactly that. 21 files under `app/actions/` use the helper; **zero** call `revalidateTag(TAGS.gang(...))` directly — this would be the first.
- **`syncFighter` on copied exotic beasts** — `copy-fighter.ts:798-802`. Beast subtypes come from the fighter type or default to `['Exotic Beast']` (`utils/exotic-beasts.ts:133`), so `deriveOverrides` returns `[]` in practice; it costs two SELECTs per beast per copy, plus a DELETE when the gang has ranks. Not wrong, just never doing anything.
- **Three validations for one invariant** — `save-venator-skill-ranks.ts:19-28`. The `'1,2,3,4'` check is unreachable from the only caller (which hardcodes ranks 1-4), and the duplicate-skill check restates the table's `UNIQUE (gang_id, skill_type_id)`. Defensible as server-action hardening since it's a public endpoint — just noting it.
- **`gang_skill_set_ranks_skill_type_id_idx`** — max 4 rows per gang and nothing filters on `skill_type_id`; its only real use is the `ON DELETE RESTRICT` FK check. The sibling `fighter_skill_access_override` table has no equivalent index, so this isn't house convention.
- **`await refetchRanks()`** at `gang-edit-modal.tsx:523`, immediately before `onClose()` at `:528` — a round trip whose only effect is warming the query cache for the next open. `setQueryData` with the values just saved does it for free.

## Checked and dropped

- **`updated_at` is never written** — true, but the sibling `fighter_skill_access_override` table has the same unused column (`20260203143002…sql:10`), so this matches existing convention. Not a finding.
- **`if (targets.length === 0) return;` in `syncGang`** — I initially read this as redundant against the later `rows.length === 0` guard. It isn't: it protects the `.in('fighter_id', [])` call for a gang with zero fighters, which is reachable if an owner sets ranks before adding anyone. Correct as written.

## Not bloat, for the record

The multi-subtype priority merge in `deriveOverrides` (`ACCESS_PRIORITY` plus the inner loop) looks like speculative generality but isn't — `types/edition.ts:108` sets `multipleFighterSubtypes: { n26: true }`, so an N26 fighter can genuinely hold more than one subtype and the merge is load-bearing.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `263edec5`), covering the three commits since the last review (which stopped at `db11036`): `d464631` (findings 2/3/4), `549c374` (finding 5 + minor 4), `263edec` (findings 1/6 + minors 1/2/5).

## Previous findings — status

All six correctness findings from the first two review rounds were already resolved before this sync and are untouched by it (confirmed no regression): archetype-wipe guard, arbitrator RLS, atomic RPC, message gating, `VENATOR_SUBTYPES` dedup, batched `syncGang`, and the stale-override cleanup on subtype loss.

From the third-round **bloat/DRY** review (head `db11036`):

| Finding | Status |
|---|---|
| 1. Hand-rolled `skill_types` query duplicates `/api/skill-types` | **Resolved.** `gang-edit-modal.tsx` now calls `fetch(\`/api/skill-types?edition_slug=${editionSlug}\`)` and filters `!r.is_custom`. Confirmed the route tags standard rows `is_custom: false` and custom rows `true`, so the filter does what's intended, and the route's N26→N23 fallback + server-side edition filter now apply here too. |
| 2. `&gt;` rendering literally in the "not accessible" message | **Resolved.** `skillAccessDeniedMessage()` uses a literal `>` character (not the HTML entity), so it now renders correctly as text content. |
| 3. Duplicated message IIFE in two files | **Resolved.** Both `fighter-skills-list.tsx` and `fighter-advancement-list.tsx` now call the shared `skillAccessDeniedMessage()` from `utils/venatorSkillAccess.ts`. |
| 4. `gangType` + `hasGangRanks` threaded as two props through four interfaces | **Resolved.** Collapsed into one server-computed `venatorNoRanks` boolean, threaded through `page.tsx` → `fighter-page.tsx` → `SkillsList`/`SkillModal` and `AdvancementsList`/`AdvancementModal`. Verified no stale `hasGangRanks`/`has_gang_ranks` references remain, and `SkillsList`'s now-unused `gangType` prop was correctly dropped (it's still passed to `AdvancementsList`, which needs it for unrelated N26 promotion logic — correct, not a leftover). |
| 5. Two migrations where the second only undoes the first | **Resolved.** Squashed into `20260827120000_create_gang_skill_set_ranks.sql`; the arbitrator carve-out ships in the original create migration. Also dropped the now-confirmed-unused UPDATE policy/grant and the `skill_type_id` index, consistent with the RPC (`20260828130000`) only ever doing DELETE + INSERT. |
| 6. `syncFighter` called on copied exotic beasts, always a no-op | **Resolved.** Call removed from `copy-fighter.ts`'s beast-copy path; the main fighter's `syncFighter` call (line 397) is untouched. |
| Minor: `revalidateTag` instead of `invalidateGang` | **Resolved.** `save-venator-skill-ranks.ts` now uses `invalidateGang(gangId)`. |
| Minor: `await refetchRanks()` round trip before close | **Resolved.** Replaced with `queryClient.setQueryData(['gang-skill-set-ranks', gangId], ...)` using the just-saved values. |
| Minor: four parallel `rank1..rank4` useStates | **Resolved.** Collapsed into `useState<string[]>(['', '', '', ''])` + a `setRank(i, value)` helper; the unchanged-check, payload, and render now all loop over the array instead of four hand-written copies. |
| Minor: three validations for one invariant | **Kept, as noted** (defensible as server-action hardening). Not re-flagging. |

## New findings

None. Reviewed all 10 changed files in full — the prop collapse is consistent end-to-end (no dangling `gangType`/`hasGangRanks` references), the `/api/skill-types` switch correctly filters `is_custom` and inherits the fallback behavior, the migration squash matches what the RPC actually needs (no UPDATE anywhere), and the rank-array refactor in `gang-edit-modal.tsx` preserves the exact save/validation/cross-exclusion semantics from before (the `taken` set now excludes by index instead of by value, which is strictly more correct for duplicate-value edge cases, not a behavior change in practice since ranks are cross-exclusive).

## Verdict

**Approve.** All findings from every review round (two correctness rounds + one bloat round) are resolved, and this sync doesn't introduce anything new.

---

joeseos:
@v0rp-z Convention check on the RLS policies in `supabase/migrations/20260827120000_create_gang_skill_set_ranks.sql` (head `263edec`). Two deviations from house convention, both in the same block.

## 1. Function calls aren't wrapped in `(SELECT …)`

`private.is_admin()` is wrapped, but `auth.uid()` and `private.is_arb(cg.campaign_id)` are not:

```sql
EXISTS (SELECT 1 FROM public.gangs g WHERE g.id = gang_id AND g.user_id = auth.uid())
...
  AND private.is_arb(cg.campaign_id)
```

Measured across the 344 policies in `supabase/schema/schema.public.sql`:

| Pattern | Follows | Deviates |
|---|---|---|
| `(SELECT auth.uid())` wrapped | 139 | 3 |
| `(SELECT private.is_admin())` wrapped | 239 | 6 |
| `(SELECT private.is_arb(...))` wrapped | 51 | 6 |
| `col IN (SELECT ...)` subquery | 78 | 6 use `EXISTS` |

This isn't cosmetic. `(SELECT fn())` lets Postgres hoist the call into an InitPlan and evaluate it once per statement instead of once per candidate row. `private.is_admin` / `private.is_arb` are SECURITY DEFINER helpers that each hit `profiles` / `campaign_members`, so unwrapped they re-run per row. `20260719120000_gate_arb_gang_access_on_accepted.sql` rewrote ~30 existing policies into exactly this wrapped shape, so it's a deliberate pattern rather than incidental style.

## 2. `EXISTS (SELECT 1 …)` instead of `col IN (SELECT …)`

Only 6 of 344 policies use `EXISTS`, and none of them are gang-scoped write policies. The nearest sibling to this table — `gang_tactics_cards`, added in `20260819180600_create_tactics_cards.sql`, a gang-scoped table with the identical owner/admin/arbitrator authority model — is the template this should follow.

Worth noting the `EXISTS` form also has a latent hazard: `WHERE g.id = gang_id` inside `EXISTS (… FROM public.gangs g …)` is an unqualified outer reference that only resolves correctly because `public.gangs` happens to have no `gang_id` column. The second `EXISTS` qualifies it (`gang_skill_set_ranks.gang_id`); the first doesn't. The `IN` form removes the asymmetry.

## Suggested replacement

Modelled on `gang_tactics_cards` — same branch order (admin → owner → arbitrator), same `cg.status = 'ACCEPTED'` gate. `CREATE TABLE`, `ENABLE ROW LEVEL SECURITY`, and the `GRANT` stay as they are; the SELECT policy is already correct and only gets renamed:

```sql
CREATE POLICY "Allow authenticated users to view gang skill set ranks"
    ON public.gang_skill_set_ranks
    FOR SELECT
    TO authenticated
    USING (true);

CREATE POLICY "Users can create skill set ranks for their gang"
    ON public.gang_skill_set_ranks
    FOR INSERT
    TO authenticated
    WITH CHECK (
        (SELECT private.is_admin())
        OR gang_id IN (
            SELECT g.id
            FROM public.gangs g
            WHERE g.user_id = (SELECT auth.uid())
        )
        OR gang_id IN (
            SELECT cg.gang_id
            FROM public.campaign_gangs cg
            WHERE cg.status = 'ACCEPTED'
              AND (SELECT private.is_arb(cg.campaign_id))
        )
    );

CREATE POLICY "Users can delete skill set ranks from their gang"
    ON public.gang_skill_set_ranks
    FOR DELETE
    TO authenticated
    USING (
        (SELECT private.is_admin())
        OR gang_id IN (
            SELECT g.id
            FROM public.gangs g
            WHERE g.user_id = (SELECT auth.uid())
        )
        OR gang_id IN (
            SELECT cg.gang_id
            FROM public.campaign_gangs cg
            WHERE cg.status = 'ACCEPTED'
              AND (SELECT private.is_arb(cg.campaign_id))
        )
    );
```

The authority granted is unchanged — owner, admin, or arbitrator of an `ACCEPTED` campaign gang — so QA item 15 (arbitrator saves ranks) still passes, and `save_gang_skill_set_ranks` being `SECURITY INVOKER` means it keeps running under these policies.

**This can be edited in place.** I checked the production project: `to_regclass('public.gang_skill_set_ranks')` is `NULL` with 0 policies, so the migration hasn't been applied anywhere and no `DROP POLICY` / follow-up migration is needed.

## Two adjacent notes

**Policy naming.** `"gang_skill_set_ranks insert: owner, admin, or arbitrator"` matches neither established style in the repo — quoted sentence names (`"Users can create tactic cards for their gang"`) or unquoted `<table>_<op>_policy` identifiers (`vehicles_user_insert_policy`). The names above use the sentence style its nearest sibling uses.

**Missing `supabase/functions/` file for the new RPC.** `save_gang_skill_set_ranks` is defined only in `20260828130000_add_save_gang_skill_set_ranks_rpc.sql`. Per `supabase/README.md`, each RPC also gets a source-of-truth copy in `supabase/functions/<name>.sql` plus a row in the README's *RPC Functions* table (Security → `Invoker`). Every comparable RPC-adding migration pairs with one — `accept_campaign_join_request`, `check_permission`, `copy_custom_collection`, `get_available_skills`. Non-blocking, just easy to miss.

joeseos:
@v0rp-z On `save_gang_skill_set_ranks` — this should be a plain server action, not an RPC.

## Why drop it

It's `SECURITY INVOKER` and its whole body is a `DELETE` + `INSERT`, so it grants nothing the client lacks — it only buys a transaction. That transaction is then undone two lines later: `saveVenatorSkillRanks` calls `syncGang`, which does its own unprotected `.delete().in('fighter_id', targetIds)` then `.insert(rows)` across every affected fighter. Bigger window, worse failure, no protection.

The repo's precedent for this exact shape is a plain server action — `saveFighterSkillAccessOverrides` (`app/actions/fighter-skill-access.ts:52-77`) does the same delete-then-insert on the same table, no RPC, in production. The four existing write RPCs are 77–360 lines of genuine multi-table or privileged work; this is 5 statements.

## What to do instead

1. Delete `supabase/migrations/20260828130000_add_save_gang_skill_set_ranks_rpc.sql`.
2. In `app/actions/gang/save-venator-skill-ranks.ts`, replace the `supabase.rpc(...)` block with:

```ts
const { error: deleteErr } = await supabase
  .from('gang_skill_set_ranks')
  .delete()
  .eq('gang_id', gangId);
if (deleteErr) return { ok: false, error: deleteErr.message };

const { error: insertErr } = await supabase
  .from('gang_skill_set_ranks')
  .insert(ranks.map(r => ({ gang_id: gangId, rank: r.rank, skill_type_id: r.skill_type_id })));
if (insertErr) return { ok: false, error: insertErr.message };
```

Everything else stays as-is. Permissions don't change — the RPC ran as the invoker, so the table's INSERT/DELETE policies were already doing the work. The validation above it still runs, and the success path is behaviourally identical, so QA items 2–6 are unaffected.

Worst case on a failed insert is the gang having zero ranks; `syncGang` and `syncFighter` both early-return on `ranks.length === 0`, so no fighter's overrides are touched and the user just re-picks the four dropdowns.

This also retires the "missing `supabase/functions/` file + README row" item from my earlier comment — that obligation only existed because of the RPC.

joeseos:
@v0rp-z Follow-up review on the latest sync (head `ac11746`), covering the one commit since the last review (which stopped at `263edec`): `ac11746` — `app/actions/gang/save-venator-skill-ranks.ts` and `supabase/migrations/20260827120000_create_gang_skill_set_ranks.sql` (with `20260828130000_add_save_gang_skill_set_ranks_rpc.sql` deleted).

## Previous findings — status

All correctness/bloat findings from earlier rounds remain resolved and untouched by this commit. The two items open at the last review:

| Finding | Status |
|---|---|
| RLS policy convention (unwrapped `auth.uid()`/`is_arb`, `EXISTS` instead of `IN (SELECT …)`, sentence-style naming) | **Resolved.** The INSERT/DELETE policies now match `gang_tactics_cards` exactly: `(SELECT auth.uid())`, `(SELECT private.is_admin())`, `(SELECT private.is_arb(cg.campaign_id))` all wrapped, `gang_id IN (SELECT …)` instead of `EXISTS`, and the sentence-style names ("Users can create/delete skill set ranks for/from their gang"). Edited in place on the same (unapplied) migration file, as suggested. |
| Drop the `save_gang_skill_set_ranks` RPC in favor of a plain server action | **Resolved.** `save-venator-skill-ranks.ts` now does the delete then insert directly via the Supabase client, mirroring `saveFighterSkillAccessOverrides`. The RPC migration file is deleted outright. Grepped the repo for `save_gang_skill_set_ranks` — no remaining references anywhere (no stale calls, no `supabase/functions/` copy, no README row, no generated types), so the RPC's removal is clean. |

Reintroducing the non-atomic delete-then-insert was itself the explicit ask in the prior comment, with the failure mode reasoned through there (gang ends up with zero ranks; `syncGang`/`syncFighter` both early-return on `ranks.length === 0`, so no fighter overrides get touched, and `saveVenatorSkillRanks` returns `ok:false` before calling `syncGang` on an insert failure) — not re-flagging it as this PR's own regression since it matches the codebase's existing precedent for this exact shape.

## New findings

None. Both changed files were read in full — the RLS rewrite preserves the same authority (owner, admin, or arbitrator of an ACCEPTED campaign gang) so QA item 15 is unaffected, the `GRANT` list correctly stays `SELECT, INSERT, DELETE` (no `UPDATE`, consistent with the already-dropped UPDATE policy), and the action's early-return-on-error ordering (delete → insert → sync → invalidate) means a delete or insert failure can't leave stale `fighter_skill_access_override` rows.

## Verdict

**Approve.** Every finding raised across all review rounds is now resolved, and this commit doesn't introduce anything new.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `4ca0ac9`), covering the one commit since the last review (which approved at `ac11746`): `4ca0ac9` — a merge of `main` into this branch.

## What changed

The merge brings in 17 files' worth of unrelated `main` commits (N26 named-type/specialisation handling in `keepTypePromotionN26.ts`, print/portrait fixes, gang-card cleanup, a CI workflow revert, schema snapshot update). None of it touches the Venator feature's own files. Diffed `ac11746` vs `4ca0ac9` against every Venator-owned path (`utils/syncVenatorSkillOverrides.ts`, `utils/venatorSkillAccess.ts`, `app/actions/gang/save-venator-skill-ranks.ts`, `components/gang/gang-edit-modal.tsx`, the skill/advancement list components, the `gang_skill_set_ranks` migration) — zero changes. The merge was clean (no conflict markers, no manual resolution needed).

Checked the one overlap risk: `add-fighter.ts`, `edit-fighter.ts`, and `copy-fighter.ts` are touched by both `main`'s new specialisation-clearing logic and this PR's `syncFighter` calls. Confirmed they don't interact — the `main` changes only read/write `fighter_specialisation`/`fighter_specialisation_id`, `deriveOverrides`/`syncFighter` only ever touch `fighter_skill_access_override`, and the `syncFighter` call sites themselves (`edit-fighter.ts:1782`, `add-fighter.ts:1219`, `copy-fighter.ts:410`) are byte-for-byte unchanged from `ac11746`.

## Previous findings — status

All findings from every review round (two correctness rounds, one bloat/DRY round, one RLS-convention round, and the RPC-removal round) were resolved as of `ac11746`, which was approved. This merge commit doesn't touch any of that code, so nothing regresses:

| Finding | Status |
|---|---|
| `syncFighter` wiping archetype-derived overrides on non-Venator gangs | Resolved, unaffected |
| Arbitrator RLS gap on `gang_skill_set_ranks` | Resolved, unaffected |
| Non-atomic delete-then-insert for gang ranks | Resolved (plain server action, matches codebase precedent), unaffected |
| Misleading "hasn't ranked Skill Sets" message | Resolved, unaffected |
| Duplicated `VENATOR_SUBTYPES` list | Resolved, unaffected |
| Sequential per-fighter `syncGang` | Resolved (batched), unaffected |
| Stale overrides not cleared on subtype loss | Resolved, unaffected |
| `hasGangRanks` probe running for every gang | Resolved (gated), unaffected |
| Bloat/DRY items (hand-rolled `skill_types` query, `&gt;` entity bug, duplicated message IIFE, prop threading, migration squash, parallel `useState`s, etc.) | All resolved, unaffected |
| RLS policy convention (unwrapped functions, `EXISTS` vs `IN`, naming) | Resolved, unaffected |
| `save_gang_skill_set_ranks` RPC removed in favor of plain server action | Resolved, unaffected |

## New findings

None. This commit is a no-op merge from the Venator feature's perspective.

## Verdict

**Approve.** No code changes to review beyond a clean merge from `main` that doesn't intersect this feature's logic. All prior findings remain resolved.

---

joeseos:
### `syncFighter` isn't gated by gang type — all gangs pay the extra roundtrips

`syncFighter` has no gang-type or edition check. Both queries run before it can bail:

```ts
const { data: fighter } = await supabase.from('fighters')
  .select('gang_id, fighter_subtypes').eq('id', fighterId).single();   // roundtrip 1
const ranks = await readGangRanks(fighter.gang_id, supabase);          // roundtrip 2
if (ranks.length === 0) return;                                        // only gate
```

The gate is *"does this gang have rank rows"*, not *"is this gang Venators"*, and it lands after both reads. So an N23 Escher gang pays 2 sequential roundtrips on every fighter add, every copy, and every edit that changes subtypes — it just exits afterwards.

Note this is inconsistent with the rest of the PR: `app/fighter/[id]/page.tsx:55` and `gang-edit-modal.tsx` both correctly gate their new query on `isVenatorGang(...)`. Only the server actions skipped it.

### Fix: gate at the three call sites, at zero cost

Each call site already has both values in scope, so no new queries are needed anywhere:

| File | Change | Edition already available |
|---|---|---|
| `add-fighter.ts` | add `gang_type` to the `gangs` select at L393 | `editionSlug` at L422 |
| `copy-fighter.ts` | add `gang_type` to the `gangs` select at L290 | `editionSlug` at L248 |
| `edit-fighter.ts` | add `gang_type` to the `gangs` select at L274 | `resolvedEditionSlug` at L1440 |

Then `if (!isVenatorGang(editionSlug, gangType)) return;` before calling `syncFighter`. `gangs.gang_type` is a plain `text` column (`schema.public.sql:5758`), so it rides along on selects that already run.

For `edit-fighter`, `resolvedEditionSlug` is set at L1440 inside `if (params.fighter_subtypes !== undefined)` — the same condition that gates `syncFighter` at L1782 — so it's always populated on that path.

Suggest also making `gangType` + `editionSlug` **required** params on `syncFighter` rather than adding an internal lookup: a future caller that forgets becomes a compile error instead of a silent roundtrip. The three `updateFighterDetails` callers in `fighter-advancement.ts` (L701, L861, L1437) are covered automatically, since they route through `edit-fighter`.

### `saveVenatorSkillRanks` has no server-side gang check

`app/actions/gang/save-venator-skill-ranks.ts` writes rank rows without verifying the gang is a Venators/N26 gang — the `isVenator` check exists only in `gang-edit-modal.tsx`. Worth a gang fetch + `isVenatorGang` guard before the delete/insert; that's what makes "has rank rows" trustworthy as a signal in the first place. Cold path, so the extra query costs nothing that matters.

### Separate bug: the override delete is unscoped and runs last

`rewriteFighterOverrides` deletes **all** `fighter_skill_access_override` rows for the fighter — no `access_level` or origin filter — and the delete runs *before* the `overrides.length === 0` early return. Two other places write that same table: `utils/archetypeEligibility.ts` and `app/actions/fighter-skill-access.ts` (the manual UI).

Ordering works against you in both actions — the archetype write happens first, then the sync wipes it:

- `add-fighter.ts`: archetype overrides at **L575**, `syncFighter` at **L1219**
- `edit-fighter.ts`: archetype overrides at **L1590 / L1613 / L1629**, `syncFighter` at **L1782**

So inside a Venator gang that has ranks, adding or copying a fighter with no Leader/Champion/Specialist subtype deletes that fighter's archetype and manual overrides and inserts nothing back. That's QA item #18, and it isn't only reachable via a mis-typed gang — it's the normal path for a Ganger in a ranked Venator gang. Scoping the delete to rows the sync owns would close it; the gang-type gate shrinks the blast radius but doesn't fix it.

---

joeseos:
@v0rp-z Reviewed the new commits since the last review comment (4ca0ac9 → 546e793, i.e. `52144f7` through the trailing merge — 13 commits). This was a single-pass review at high effort, done manually (no subagent fan-out/verify pass), so treat it as high-recall/needs-confirmation.

## Previous findings — status

All three findings from the last review comment are **resolved**:

1. **`syncFighter` not gated by gang type at call sites** — Resolved. `add-fighter.ts`, `copy-fighter.ts`, and `edit-fighter.ts` now fetch `gang_type`/`custom_gang_type_id` alongside data they already select, and wrap the `syncFighter` call in `isVenatorGang(...)`. Non-Venator gangs no longer pay the extra roundtrips. Verified all 8 `isVenatorGang(...)` call sites now pass the new required `isCustomGangType` argument consistently (`git grep` confirms no stale 2-arg or old-signature callers remain).

2. **`saveVenatorSkillRanks` had no server-side gang check** — Resolved. It now fetches `gang_type`/edition/`custom_gang_type_id` and rejects with `'This gang cannot rank Skill Sets.'` if `isVenatorGang` is false, before touching `gang_skill_set_ranks`.

3. **Unscoped override delete wiping archetype overrides (QA #18)** — Resolved correctly, and via a good approach: `rewriteFighterOverrides` and `syncGang` now scope their `DELETE` to `skill_type_id IN (ownedSkillTypeIds)` — only the gang's ranked skill sets — instead of wiping every override row for the fighter. `saveVenatorSkillRanks` now also captures `previouslyOwnedSkillTypeIds` before the rank delete/insert and threads it into `syncGang`, so a rank *change or clear* still cleans up stale rows for skill sets that fell out of the rank set. This closes the archetype-wipe bug while still handling rank-set churn. The `0a91ca9a` follow-up (broadening `edit-fighter`'s sync trigger to also fire on archetype-only edits) is a sensible companion fix, since the archetype block's own unscoped delete-then-insert (pre-existing, outside this PR's diff) would otherwise leave Venator rows gone until the next rank save.

## New findings

**Low — silent swallow of gang-fetch error in `edit-fighter.ts`** (`app/actions/edit-fighter.ts`, new block around the `syncFighter` call, ~L1786-1802)
```ts
const { data: syncGangRow } = await supabase
  .from('gangs')
  .select(`...`)
  .eq('id', fighter.gang_id)
  .single();
...
if (syncGangRow && isVenatorGang(...)) { ... }
```
The `error` from this select isn't destructured or logged. If the query fails transiently, `syncGangRow` is `null`, the `if` is skipped, and the Venator sync silently doesn't run — with no `console.error` trace, unlike every other `syncFighter`/`syncGang` call site in this PR which logs on failure. Worth a one-line `console.error` on the fetch error for parity and easier triage, but this is a corner case (query failing on an `.eq('id', ...)` lookup by PK is rare) so I wouldn't block on it.

No other new issues found — no regressions in the RLS migration (matches the `gang_tactics_cards`-style policy shape used elsewhere), the `gang-edit-modal.tsx` clear-ranks flow correctly falls into the `filled.length === 0 && hadPreviousRanks` branch (the earlier `filled.length !== 0 && filled.length !== 4` guard already rejects partial fills), and the `getGangCore` cache-key bump (`v3` → `v4`) correctly accompanies the new `venator_no_ranks` field being folded into that cached row.

## Verdict

**Approve.** All outstanding findings from prior rounds are addressed; the one new item is a minor logging gap, not a correctness bug.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `0319e65`), covering the one commit since the last review (which approved at `546e793`): `0319e65` — `app/api/skill-types/route.ts`, `components/gang/gang-edit-modal.tsx`, `app/actions/edit-fighter.ts`, plus doc-comment removal across `app/actions/fighter-skill-access.ts`, `app/lib/shared/gang-data.ts`, `20260831120000_allow_fighter_owner_to_edit_skill_overrides.sql`, `utils/cache-tags.ts`, `utils/syncVenatorSkillOverrides.ts`, `utils/venatorSkillAccess.ts`.

## Previous findings — status

All findings from every prior round (two correctness rounds, one bloat/DRY round, one RLS-convention round, the RPC-removal round, and the gang-type-gating round) remain **resolved** — this commit doesn't touch any of that logic.

One item is carried over, not addressed by this commit (it wasn't in scope, and the prior review explicitly didn't block on it):

| Finding | Status |
|---|---|
| Low — silent swallow of gang-fetch error in `edit-fighter.ts` (no `console.error` if the `syncGangRow` select fails, unlike every other sync call site) | **Outstanding, non-blocking.** `edit-fighter.ts:1784-1793` still doesn't destructure/log `error` from that select. Same corner case as before (PK lookup failing is rare) — worth a one-line fix for parity when convenient, not a merge blocker.

## New findings

**Medium — `strict_edition=1` removes the N23 fallback for the rank picker, but nothing in this repo actually seeds N26 `skill_types`**

`app/api/skill-types/route.ts:79` now skips the N23 fallback when `strict_edition=1`, and `gang-edit-modal.tsx:138` opts the Venator rank picker into it. That's the right rulebook behavior *if* N26's skill-type catalog is populated — but `supabase/seed.sql:111-120` only inserts skill types for edition `...023` (N23), and a repo-wide grep for `INSERT INTO ... skill_types` / the N26 edition UUID under `supabase/migrations/` turns up nothing. If the target database's N26 catalog is empty, the four rank dropdowns in Edit Gang → Skill Access now render with **zero options** instead of falling back to N23 — a silent break of the whole feature (QA items 1–6), with no empty-state message in `gang-edit-modal.tsx` (`skillTypes` just defaults to `[]` and the `options` filter at line 635 produces an empty `<select>`).

This exact risk was flagged in an earlier round ("your screenshots show them populated... presumably seeded in your environment now — worth a quick confirm either way"), so it's likely already handled outside the migration history. Given this commit deliberately trades the fallback safety net for correctness, worth an explicit confirm that N26 `skill_types` are seeded on whatever DB this deploys to before merge — or, cheaply, a distinct empty-state message in the modal ("no skill sets available for this edition yet") so a genuinely-empty catalog fails loud instead of silent.

No other issues in this commit — the rest is comment stripping (SQL policies, JSDoc, inline `//` notes) with no behavioral change, and the `['skill-types', editionSlug, 'strict']` query-key change correctly avoids colliding with the non-strict skill-types query used elsewhere (`skill-access-modal.tsx` already keys on `fighterId` + `editionSlug`, so no prior collision existed either, but the explicit `'strict'` tag makes the two variants unambiguous going forward).

## Verdict

**Approve**, contingent on confirming the N26 skill-types seeding question above before merge — everything else in this sync is either a resolved carry-over or a safe cleanup.

---

joeseos:
@v0rp-z Follow-up review on the latest sync (head `4eb5d41`), covering the one commit since the last review (which approved contingent on the N26 seeding question, at `0319e65`): `4eb5d41` — `components/gang/gang-edit-modal.tsx` only.

## Previous findings — status

| Finding | Status |
|---|---|
| Medium — `strict_edition=1` silently renders zero options if N26 `skill_types` are unseeded | **Resolved.** This commit adds the loud empty-state directly: `skillTypesLoaded && skillTypes.length === 0` now renders `"No Skill Sets are available for this edition yet. Please contact an admin."` in `text-destructive` (an established convention in this codebase — same pattern used in `skill-access-modal.tsx`, `form-message.tsx`, etc.) instead of four blank dropdowns. This doesn't answer whether N26 is actually seeded on the target DB, but it closes the "fails silent" risk the last review flagged — an empty catalog is now visibly reported rather than looking like a broken picker. |
| Low — silent swallow of gang-fetch error in `edit-fighter.ts` | **Outstanding, non-blocking** (unchanged, out of scope for this commit — same as noted in the last two reviews). |
| All other findings from every prior round (correctness, bloat/DRY, RLS-convention, RPC-removal, gang-type-gating) | Resolved, untouched by this commit. |

## New findings

None. The change is scoped and correct:

- `skillTypesLoaded` (`isSuccess` from the query) correctly gates the message so the pre-fetch/loading state still falls through to the existing "empty dropdowns" render rather than flashing the error message before the query resolves — no new loading-state regression.
- The query is `enabled: isVenator && !!editionSlug`, and this whole block only renders under `isVenator`, so `skillTypesLoaded` reaching `true` with zero rows can only mean a completed, successful, empty fetch — not a disabled or errored query masquerading as "empty."
- `ranks` state initialization (from `gang_skill_set_ranks`, independent of `skillTypes`) and the Save-button validation path are untouched by this diff. Traced the case where `skillTypes` is empty and the picker is hidden: `ranks` stays at whatever `existingRanks` initialized it to (all blank in the realistic unseeded-catalog case), so `filled.length === 0` and, with no previous ranks, Save is a no-op — no way to submit through a hidden picker.
- Only `gang-edit-modal.tsx` changed; no other Venator call sites (`syncFighter`, `syncGang`, `saveVenatorSkillRanks`, the RLS policies) are touched, so nothing from the correctness/RLS rounds is at risk of regressing here.

## Verdict

**Approve.** This closes out the one open item from the last review in the simplest available way (empty-state messaging rather than requiring a DB confirmation), and introduces nothing new.

---

joeseos (approved):

PR #2073, opened 2026-08-28
Remove the runaway campaign catalog cache purge

joeseos:
purgePreEditionCampaignCatalogCachesOnce was a one-time cleanup for the PR #1973 cache-key rename, but its guard flag is module scope, so "once" meant once per JS instance. On Vercel Fluid instances cycle constantly and tag expiry is persisted to a shared store, so it re-fired indefinitely and evicted two revalidate: false catalogs fleet-wide. The campaign page read both caches and then evicted them in after() during the same render, so they were never warm and every miss put a Supabase round trip on the render critical path.

Removes the purge and all five call sites, and reverts the four -with-edition keys in reference-data.ts now that nothing depends on the rename. Safe because those tags have been expired continuously since Aug 13; still worth running vercel cache purge --type data at deploy.

Also drops two redundancies found nearby: a duplicated invalidateCampaign call in the Discord callback, and the withReferenceInvalidation wrappers in the campaign-types and territories admin routes, whose handlers already call the broader invalidateCampaignCatalogLists on every ok-returning path. The same wrapper in five other admin routes is the only invalidation there and is left untouched.

### DIFF
--- a/app/actions/campaigns/[id]/campaign-territories.ts
+++ b/app/actions/campaigns/[id]/campaign-territories.ts
@@ -1,6 +1,6 @@
 'use server';
 
-import { invalidateCampaign, invalidateGangCampaignMembership, purgePreEditionCampaignCatalogCachesOnce } from '@/utils/cache-tags';
+import { invalidateCampaign, invalidateGangCampaignMembership } from '@/utils/cache-tags';
 import { createClient } from "@/utils/supabase/server";
 import { logTerritoryLost, logTerritoryClaimed } from "../../logs/gang-campaign-logs";
 import { getAuthenticatedUser } from '@/utils/auth';
@@ -310,7 +310,6 @@ export async function addTerritoryToCampaign(params: AddTerritoryParams) {
     if (error) throw error;
 
     invalidateCampaign(campaignId);
-    purgePreEditionCampaignCatalogCachesOnce();
 
     return { success: true };
   } catch (error) {
--- a/app/actions/create-campaign.ts
+++ b/app/actions/create-campaign.ts
@@ -1,6 +1,6 @@
 'use server'
 
-import { invalidateUser, invalidateCampaignCount, purgePreEditionCampaignCatalogCachesOnce } from '@/utils/cache-tags';
+import { invalidateUser, invalidateCampaignCount } from '@/utils/cache-tags';
 import { createClient } from "@/utils/supabase/server";
 import { getAuthenticatedUser } from "@/utils/auth";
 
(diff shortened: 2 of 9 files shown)


joeseos:
@joeseos Code review (high effort, single-pass — the Agent tool wasn't available in this session, so this is one careful pass through all the usual angles rather than the full multi-agent fan-out + verify).

**Summary: no correctness bugs found.** This is a clean, well-scoped revert/cleanup. I traced each removal against its stated justification:

- Confirmed the `withReferenceInvalidation` wrapper removal in `app/api/admin/campaign-types/route.ts` and `app/api/admin/territories/route.ts` is safe: read both full route files, and every `ok`-returning path in `POST`/`PATCH` already calls `invalidateCampaignCatalogLists()` directly before returning; all error returns are non-2xx, so the wrapper's `response.ok` branch was never doing anything extra on those files.
- Confirmed no dangling references to the removed `purgePreEditionCampaignCatalogCachesOnce` or the `-with-edition` cache keys anywhere in the tree, and no unused imports left behind (`TAGS`, `revalidateTag`, `after` are all fully removed from files that no longer use them).
- Confirmed the 5 other admin routes (`scenarios`, `gang-lineages`, `campaign-triumphs`, `equipment/trading-posts`, `gang-types`) each still define and use their own local `withReferenceInvalidation` copy as their only invalidation path, matching the PR description's claim that those were intentionally left alone.
- Checked for cache-key collisions between the reverted literal keys (`campaign-types`, `territories-list`, `global-scenarios`, `global-trading-post-types`) and any other `unstable_cache` call or string literal in the codebase — none found.
- Verified the discord-callback duplicate-call fix and the `after()` removal in the campaign page don't leave any other usage behind.

One low-confidence item I can't fully verify without visibility into the Vercel data-cache internals, flagging for your judgment rather than as a confirmed bug:

- `app/lib/reference-data.ts` (the 4 reverted keys) / `utils/cache-tags.ts`: the safety argument ("safe because those tags have been expired continuously since Aug 13") assumes the shared invalidation store's tag-bust affects *any* entry under a given tag regardless of key name, including a literal key like `campaign-types` that may have existed (dormant, unread) from *before* the original Aug 13 rename. If the cache backend's tag invalidation only marks entries it has already seen rather than retroactively covering an untouched key, a stale pre-rename entry lacking `edition_slug` could resurface once the key reverts. You already call this out and recommend `vercel cache purge --type data` at deploy as a belt-and-suspenders step — just flagging that this is a deploy-operational dependency rather than something enforced by the code itself, so it's worth double-checking that step actually runs.

No other findings across correctness, removed-behavior, cross-file callers, reuse/simplification, efficiency, altitude, or repo conventions (no applicable CLAUDE.md found).

---

PR #2075, opened 2026-08-28
fix(hooks): use useSyncExternalStore in useViewportWidth

joeseos:
The hook mirrored the module-level viewport store into useState via an effect, tripping react-hooks/set-state-in-effect. The module already was an external store (cached width + listener set + refcounted window listeners), so subscribe to it directly instead.

Behaviour is unchanged: getServerSnapshot returns DEFAULT_WIDTH for SSR/hydration, and subscribe() publishes the real innerWidth on mount so React picks it up on the snapshot re-read.

### DIFF
--- a/hooks/use-viewport-width.ts
+++ b/hooks/use-viewport-width.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useSyncExternalStore } from 'react';
 
 const DEFAULT_WIDTH = 800;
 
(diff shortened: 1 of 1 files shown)


joeseos:
@joeseos High-effort review of this PR.

**Note on methodology:** the Agent tool wasn't available in this session, so this is a single-pass review done in one context rather than the full multi-agent fan-out + subagent verify pass. Treat findings accordingly.

**Summary:** I traced the `useSyncExternalStore` conversion in detail, including how React's internal `subscribeToStore`/`updateStoreInstance` effects are ordered on mount. The pattern here — mutating the shared `viewportWidth` as a side effect of `subscribe()` and relying on React's built-in post-subscribe snapshot re-check to force a re-render — is exactly the officially-supported technique for this kind of external store, and I couldn't find a case where it produces a wrong or stale render (traced through initial mount, SSR/hydration via `getServerSnapshot`, batched multi-card mounts, and React StrictMode's double-invoke). The refcounted window-listener logic (`subscriberCount`/`listeners`) stays in lockstep before and after the change, and the single call site (`components/gang/fighter-card.tsx`) only consumes the returned number, so no call-site breakage.

Two minor, non-blocking notes:

1. **`hooks/use-viewport-width.ts` — new `subscribe()`, the `listeners.size === 0` gate**: this does change behavior versus the old code, not just the mechanism. Previously *every* mount called `setWidth(window.innerWidth)` unconditionally; now only the first concurrent subscriber refreshes the shared `viewportWidth` — later mounts while others are still up inherit the cached value. In the case that motivates `shouldStabiliseWidthOnResize` (a touch-device resize from browser-chrome show/hide), this actually looks like an improvement (a late-mounting card in a grid now matches its already-mounted siblings instead of picking up a "truer" but visually-inconsistent width). But the PR description says "Behaviour is unchanged," and this is a real, deliberate behavioral difference beyond the hook-signature swap — worth calling out explicitly in the description so it doesn't read as pure mechanical refactor if someone bisects a viewport-width bug later.

2. **`hooks/use-viewport-width.ts`, `subscribe()`**: the `typeof window !== 'undefined'` check guarding the `viewportWidth = window.innerWidth` line is unreachable in practice — `subscribe` is only ever invoked by React's client-side effect, never during SSR. Not wrong, just dead defensive code; could drop it for clarity since `subscribeToViewportResize()` right below already owns the SSR guard for this module.

No correctness bugs found beyond that. Nice use of the "publish on subscribe" pattern.

---

PR #2078, opened 2026-08-28
feat(fighters): keep N26 subtypes and specialisation independent of n…

Djidiouf:
…amed type

Named type and variant no longer rewrite subtypes; specialisation is Specialist-only and is stripped on add, copy, and edit when that subtype is missing.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -12,6 +12,7 @@ import { logFighterAction } from '@/app/actions/logs/fighter-logs';
 import { mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility';
 import { assertArchetypeAssignable } from '@/utils/assertArchetypeAssignable';
 import { editionSlugFromJoin, gangEditionSlug, type EditionJoin } from '@/types/edition';
+import { shouldClearSpecialisationForSubtypes } from '@/utils/keepTypePromotionN26';
 
 interface SelectedEquipment {
   equipment_id: string;
@@ -415,6 +416,10 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
       throw new Error('Gang not found');
     }
 
+    const fighterSource = effectiveFighterData as FighterTypeSource;
+    const editionSlug =
+      editionSlugFromJoin(fighterSource.editions) ?? gangEditionSlug(gangData);
+
     // Note: Authorization is enforced by RLS policies on fighters table
 
     // Check for adjusted cost based on gang type (only for regular fighters)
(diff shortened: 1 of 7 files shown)


joeseos:
@Djidiouf Reviewed this against `main` (head `0b813bc`). Solid, well-contained change — the edition-gated helpers (`namedTypeKeepsSubtypes`, `omitsNamedTypeSubtypeSuffix`, `shouldClearSpecialisationForSubtypes`, `hasN26SpecialistSubtype`) keep N23 behavior untouched while cleanly branching N26, and the client/server specialisation-clearing logic is duplicated for defense-in-depth rather than trusted single-sided, which is the right call for a DB-facing invariant.

**Checklist**

- **Correctness**: No bugs found in the main flow. Traced `add-fighter.ts`, `copy-fighter.ts` (including the beast-fighter branch), and `edit-fighter.ts`'s `updateFighterDetails` — the `resolvedEditionSlug` memoization guard (`=== undefined` check) correctly avoids redundant lookups across the three call sites that need it.
- **Bug fix bundled in**: `fighter-page.tsx`'s optimistic update switched from `optimistic?.fighter_specialisation ? … : prev…` to `'fighter_specialisation' in optimistic ? … : prev…` — this fixes a real bug where optimistically *clearing* a specialisation (`null`) would previously fall through to the falsy branch and keep showing the stale value. Good catch, matches the new `fighter_specialisation_id !== undefined` optimistic-overlay condition in `fighter-edit-modal.tsx`.
- **Edge cases**: Verified the "stale specialisation without Specialist subtype" case — the UI correctly hides the Specialisation combobox (`canHaveSpecialisation` gates on `hasN26SpecialistSubtype`) but still clears it server-side via `touchesSpecialisationScope` whenever type/subtypes/specialisation are actually touched, and client-side via `nextSpecialisationId = canHaveSpecialisation ? … : ''`. Rename-only saves correctly skip this (no spurious writes).
- **Security**: No issues — all queries go through Supabase's parameterized client, RLS still governs authorization as before.
- **Performance (low)**: `copyFighter` now does an unconditional extra round trip via `resolveFighterEditionSlug` (nested `fighters → fighter_types/custom_fighter_types/gangs → editions` select) even for gangs/editions where specialisations don't apply. Minor — one extra query per copy — but worth noting since `copyFighter` is already a hot, multi-query path (vehicle equipment/effects, beast fighters, etc.).
- **DRY (low)**: The `shouldClearSpecialisationForSubtypes(...) ? {null,null} : {source values}` ternary is repeated three times in `copy-fighter.ts` (main fighter + beast fighter branches). Could be a tiny local helper, but it's cheap enough that I wouldn't block on it.
- **Conventions**: Consistent with the existing edition-capability pattern in `types/edition.ts` and the "magic UUID with drift-risk comment" convention already established in `keepTypePromotionN26.ts`.

No `CLAUDE.md` exists in this repo, so no project-convention doc to check against beyond what's inferred from surrounding code, which this PR follows well.

**Severity summary**: 2 low-severity notes (perf round-trip, minor duplication), nothing blocking.

**Verdict: Approve**

---

PR #2079, opened 2026-08-28
feat(equipment): scope equipment lists by fighter specialisation

joeseos:
A fighter's specialisation follows the fighter; the equipment list follows the fighter type. The N26 promotion paths decouple them, so a Van Saar Tek Sniper promoted to Prime loses Long las and Refraction cloak -- and since neither has an equipment_availability nor a trading_post_equipment row, they become unreachable through the Equipment List and the Trading Post alike.

No fighter_type-based fix can express this: a promoted Prime is type=Prime, subtypes=[Leader, Specialist], specialisation=Sniper -- a combination no fighter_types row has or ever will.

fighter_type_equipment gains fighter_specialisation_id, joining the gang_variant_id and fighter_subtype scopes added by 20260822120000. A row either names a fighter type, or names none and relies on scope alone.

- Migration (DDL only): the column, the scope key rebuilt to include it, a CHECK that a row names at least one identity or scope column, and removal of one pre-existing all-NULL-scope row that would otherwise grant to every fighter once type-less rows are legal. The 14 live rows are re-authored by an admin.
- RPC: gang_data selects the fighter's subtypes and specialisation; the fte join admits scope-only rows and honours all three narrowing scopes. Both "did a row match?" proxies move from fte.fighter_type_id to fte.id -- a scope-only row matches with a NULL fighter_type_id, so the old test silently dropped it.
- include_legacy ($11) makes the legacy toggle explicit. $4 is the list-filter flag, not the toggle; callers gated legacy by withholding fighter_id, which stops working now the fighter row is always needed. NULL keeps the old meaning, so a stale bundle behaves as before during the deploy.
- Admin: the equipment PATCH delete is narrowed to the rows that modal owns -- unqualified by scope it also destroyed vehicle rows written by the vehicles route -- and a Specialisation Equipment section authors the new rows.

Also in the two files this already touches: filterEquipment wrapped in useCallback with its real deps, and six trivial edition filters in the admin modal unwrapped from useMemo. Repo lint 8 problems -> 6.


joeseos:
@joeseos Reviewed against `main` (head `5d4d832`). The overall design (specialisation as a type-less scope column on `fighter_type_equipment`, joined via the fighter's own row) is sound and well-documented — the migration comments and PR description make the reasoning easy to follow, and this PR also finishes wiring up `gang_variant_id`/`fighter_subtype` scoping into the RPC join, which the earlier `20260822120000` migration had added as columns but never actually consumed in the reader.

**Checklist**

- **Correctness (high)**: `supabase/functions/get_equipment_detailed_data.sql` lines 637-640 (the `REVOKE`/`GRANT` footer) still target the *old* 10-argument signature — `(UUID, TEXT, UUID, BOOLEAN, BOOLEAN, UUID, UUID, UUID, UUID[], UUID[])` — but the function definition above it now has 11 params (`include_legacy boolean` added). Line 8 of this same file (`DROP FUNCTION IF EXISTS get_equipment_detailed_data(uuid,text,uuid,boolean,boolean,uuid,uuid,uuid,uuid[],uuid[]);`, unchanged by this PR) already drops that exact 10-arg overload before the `CREATE OR REPLACE` runs. Since Postgres treats a changed parameter list as a distinct function (adding a parameter is not a "replace" of the old signature, confirmed by comparing against `main`'s copy of this file, which had the same 10-arg footer matching its then-current 10-arg definition), running this script top-to-bottom will: drop the old 10-arg function → create the new 11-arg one → then error on `REVOKE ALL ON FUNCTION ...(10 args)` because that signature no longer exists. Since `supabase/README.md` documents this file as what gets applied manually to deploy the function, this would break the deploy (and potentially leave `EXECUTE` ungranted to `authenticated`/`service_role` if the script aborts there). Fix: update all four `REVOKE`/`GRANT` lines to the new 11-arg signature (append `, BOOLEAN`).
- **Behavior change (medium)**: `admin-edit-equipment.tsx`'s save handler used to only include `fighter_types` in the PATCH body when `hasEditedFighterTypes` was true, and the route handler additionally gated the delete/insert on a `hasChanges` comparison. Both guards were dropped — the client now always sends `fighter_types: selectedFighterTypes` and `fighter_specialisation_equipment: ...`, and the route now unconditionally deletes-then-reinserts both sets of rows whenever the field is present (`app/api/admin/equipment/route.ts`, the `if (fighter_types !== undefined)` / `if (fighter_specialisation_equipment !== undefined)` blocks). So every equipment edit save — even one that only changes, say, the cost or description — now churns all plain-scope and specialisation-scope `fighter_type_equipment` rows for that equipment, widening the (pre-existing) delete→insert gap to every save instead of only saves that actually touch the fighter-type/specialisation associations. Worth confirming this is intentional; if not, restoring a "changed?" check before deleting would avoid the unnecessary writes and the wider race window.
- **Minor (low)**: `components/equipment/equipment.tsx` — `gangAffiliationId` is now only referenced in the prop type/destructure (the `hasGangAffiliation` variable that used it was removed). It's effectively dead going forward; fine to leave if other consumers still branch on it, but worth a quick check.
- **Edge cases**: The new `fte` join's type-less OR-branch (`fte.fighter_type_id IS NULL AND fte.vehicle_type_id IS NULL AND fte.custom_fighter_type_id IS NULL`) correctly falls through to NULL (non-matching) when `$6`/fighter_id isn't supplied and a row requires a specific subtype or specialisation, since `gd.fighter_subtypes`/`gd.fighter_specialisation_id` are NULL in that case and `NULL ? x` / `NULL = x` are both NULL — so scope-only rows can't over-grant when the fighter context is missing. Good.
- **Security**: No issues — RPC stays `SECURITY DEFINER` with the same grant model (once the footer signature is fixed), admin route still behind `checkAdmin`.
- **Migration**: Column add → dead-row delete → CHECK constraint → index rebuild ordering is correct (CHECK is added after the one violating row is removed, so it won't fail validation against existing data). `NULLS NOT DISTINCT` partial unique index correctly excludes vehicle rows, consistent with the prior migration's pattern.
- **Conventions**: Consistent with the existing scope-column pattern (`gang_variant_id`, `fighter_subtype`, `excluded`) from `20260822120000`, and the admin route's new DELETE properly narrows by every scope column so it no longer collides with vehicle-route-owned or subtype/variant-scoped rows — that fix (replacing the old unqualified delete) is a genuine improvement over the previous logic.

No `CLAUDE.md` in this repo to check against beyond established in-repo conventions, which this PR mostly follows well.

**Severity summary**: 1 high (deploy-breaking GRANT/REVOKE signature mismatch), 1 medium (unconditional churn on every equipment save), 1 low (dead prop).

**Verdict: Changes requested** — please fix the `REVOKE`/`GRANT` signatures before this ships, since that will break the manual RPC deploy step described in the PR itself.

---

PR #2080, opened 2026-08-28
fix(print): match gang-page portraits and pass starting XP

Djidiouf:
Print used the type catalog image and omitted starting_xp, so selected defaults were ignored and N26 cards showed N/A. Share image resolution across gang surfaces and replace the dead Cloudinary error fallback.


joeseos:
## Review

@Djidiouf Reviewed against the base branch and the wider codebase (`resolveGangImageUrl`, `FighterProps`, `gang-assembly.ts`, `gang-data.ts`). Summary below.

### What the PR does
- Centralizes gang-image resolution (`custom upload → selected type default → letter placeholder`) into `types/gang.ts#resolveGangImageUrl`, and reuses it in `gang.tsx`, `gang-card.tsx`, and `print-gang.tsx`. Verified `default_gang_image` / `default_image_urls` are populated correctly for both the standard and custom gang-type branches in `getGangType` (`app/lib/shared/gang-data.ts:418-465`) and threaded through in `app/gang/[id]/print/page.tsx:96-131`.
- Fixes the print `starting_xp`/N26 "N/A" bug by switching the `FighterCard` invocation in `print-gang.tsx` from an exhaustive prop list (which omitted `starting_xp`) to `{...fighter}` plus targeted overrides. Confirmed `starting_xp` is a real field on `FighterProps` (`types/fighter.ts:163`) so this now flows through. I checked the fields this spread relies on that aren't on the `FighterProps` type (`active_loadout_name`, `captured_by_gang_name`) — they're set at runtime in `gang-assembly.ts:731`, so they still pass through the spread correctly; the same file already relies on that same runtime shape elsewhere (`print-gang.tsx:739-740`).
- The removed `fighterBaseStats`/`fighterCurrentStats` construction was dead weight — `FighterCard` never actually consumes the `base_stats`/`current_stats` props (it recomputes its own `fighterData.base_stats`/`current_stats` internally from the individual characteristic props), so this is a safe cleanup, not a regression.
- Replaces the dead `res.cloudinary.com/dle0tkpbl/...` fallback with the shared `UNKNOWN_GANG_IMAGE_URL` constant consistently across `gang.tsx`, `gang-card.tsx`, `create-gang-modal.tsx`, and `campaign-page-content.tsx`. No leftover references to the old Cloudinary URL.
- `resolveGangImageUrl` correctly falls through to `undefined` on an empty/missing `.url` entry so the letter-placeholder path still works.

### Findings

**Low — `components/gang/print-gang.tsx:37-43` (`PrintGangPortrait`)**: the `` has no `onError` handler, unlike every other gang-portrait render in the app (`gang.tsx`, `gang-card.tsx`, `create-gang-modal.tsx`), which reset a broken `src` to `UNKNOWN_GANG_IMAGE_URL`. If a resolved image URL 404s at print time you'll get a broken-image icon instead of the fallback/placeholder. Low impact (print is a one-shot render), but worth matching the existing pattern for consistency. Not a blocker.

Nothing else stood out — no null-handling gaps, no security concerns (no new user input paths), and no performance concerns (this only touches render-time prop wiring, not query shape).

### Verdict: **Approve**

Solid, well-scoped fix with good DRY cleanup (shared `resolveGangImageUrl`, shared `UNKNOWN_GANG_IMAGE_URL`, removal of dead stat-fallback code). The one nit above is optional.

---

joeseos:
## Follow-up review (commit a4948f3)

@Djidiouf Reviewed just the new commit since the last review.

### Previous findings
- **Low — `print-gang.tsx` missing `onError` fallback**: **Resolved.** `PrintGangPortrait` now has an `onError` handler on the `` that resets `src` to `UNKNOWN_GANG_IMAGE_URL` on a load failure, matching the pattern used in `gang.tsx` (`handleImageError`) and `gang-card.tsx` (inline handler) — same `console.error` + fallback-URL reassignment, byte-for-byte consistent with `gang-card.tsx`'s version. Import of `UNKNOWN_GANG_IMAGE_URL` added correctly alongside the existing `resolveGangImageUrl` import.

No other findings carried over — everything else from the prior review was already approved.

### New findings
None. The change is a minimal, targeted 4-line addition scoped exactly to the flagged gap; no new logic paths, no null-handling concerns, and it doesn't touch anything else in the file. (Note: like the existing `gang.tsx`/`gang-card.tsx` handlers it copies, there's no guard against `UNKNOWN_GANG_IMAGE_URL` itself failing to load causing a repeat `onError` fire — but that's a pre-existing pattern in the codebase, not something this PR introduces, so not flagging it here.)

### Verdict: **Approve**

---

PR #2083, opened 2026-08-31
Improve Venator Skill Set ranking UX with drag-to-rank and partial ra…

Djidiouf:
…nks.

Allow 0–4 consecutive ranks, share Skill Set combobox grouping, and warn when ranking is incomplete in Skills/Advancement modals.

## Summary

- Replace the four fixed Venator Skill Set comboboxes in Edit Gang with an add-then-drag ranked list (0–4 Skill Sets), including mobile-friendly drag constraints.
- Allow saving partial rankings (consecutive ranks `1..n`) on both client and server; warn in Skills and Advancement modals when ranking is incomplete (`< 4`).
- Share Skill Set combobox grouping via `buildGroupedSkillSetComboboxOptions` / `getSkillSetGroupLabel`, and reserve Combobox clear-button space so empty/filled controls stay the same size.
- Keep the generic “This Skill Set is not accessible…” copy inlined in the Skills and Advancement modals (not in `venatorSkillAccess`). That message is not Venator-specific, so this restores the pre–Venator-ranking approach instead of housing it in a Venator util.

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] Edit a Venator gang → add/remove/reorder Skill Sets (desktop drag + mobile touch) → save with 2 ranks and confirm they persist as ranks 1 and 2
- [x] Save with 0 / 4 ranks; confirm clearing ranks works; confirm incomplete-rank warning appears under Skill Set comboboxes in Skills and Advancement modals until all four are set
- [x] With a Skill Set selected that the fighter cannot access, confirm the “not accessible…” message still shows (including when ranking is incomplete)

## Risk / rollout

- Low — no migration; behavior change is Venator-specific ranking UX plus shared Skill Set option grouping. Deploy includes new `@dnd-kit/modifiers` dependency. Existing gangs with 0 or 4 ranks behave as before; gangs can now also store 1–3 ranks for house ruling purposes.

### DIFF
--- a/app/actions/gang/save-venator-skill-ranks.ts
+++ b/app/actions/gang/save-venator-skill-ranks.ts
@@ -17,16 +17,17 @@ export async function saveVenatorSkillRanks(
 ): Promise<{ ok: true } | { ok: false; error: string }> {
   const { gangId, ranks } = params;
 
-  if (ranks.length !== 0 && ranks.length !== 4) {
-    return { ok: false, error: 'All four ranks must be set, or none.' };
+  if (ranks.length > 4) {
+    return { ok: false, error: 'At most four Skill Sets can be ranked.' };
   }
-  if (ranks.length === 4) {
+  if (ranks.length > 0) {
     const rankValues = ranks.map((r) => r.rank).sort((a, b) => a - b);
-    if (rankValues.join(',') !== '1,2,3,4') {
-      return { ok: false, error: 'Ranks must be exactly 1, 2, 3, 4.' };
+    const expected = Array.from({ length: ranks.length }, (_, i) => i + 1).join(',');
+    if (rankValues.join(',') !== expected) {
+      return { ok: false, error: 'Ranks must be consecutive starting from 1.' };
     }
     const skillIds = ranks.map((r) => r.skill_type_id);
-    if (new Set(skillIds).size !== 4) {
+    if (skillIds.some((id) => !id) || new Set(skillIds).size !== ranks.length) {
       return { ok: false, error: 'The same Skill Set cannot occupy two ranks.' };
     }
   }
--- a/app/fighter/[id]/page.tsx
+++ b/app/fighter/[id]/page.tsx
@@ -253,7 +253,7 @@ export default async function FighterPageServer({ params }: FighterPageProps) {
         gang_type: gangBasic.gang_type,
         gang_type_id: gangBasic.gang_type_id,
         custom_gang_type_id: gangBasic.custom_gang_type_id,
-        venator_no_ranks: gangBasic.venator_no_ranks ?? false,
+        venator_ranks_incomplete: gangBasic.venator_ranks_incomplete ?? false,
         gang_affiliation_id: gangBasic.gang_affiliation_id,
         gang_affiliation_name: gangBasic.gang_affiliation?.name,
         positioning: gangPositioning,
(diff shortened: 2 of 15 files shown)


joeseos (commented):
@Djidiouf Reviewed this against the codebase — nice cleanup of the shared combobox-grouping logic (`buildGroupedSkillSetComboboxOptions`) and the server-side validation generalizes cleanly to partial ranks (the `gang_skill_set_ranks` table already allows `rank BETWEEN 1 AND 4`, so no migration is needed). Renaming `venator_no_ranks` → `venator_ranks_incomplete` looks complete — no stale references left anywhere.

One behavioral regression worth confirming before merge, plus a minor nit.

### Findings

**1. [Medium] Removed confirmation before a destructive rank change**
`components/gang/gang-edit-modal.tsx` (`handleSave`, around the `isVenator` block, ~L624-660)

The old code warned the user with `window.confirm(...)` before saving changed/cleared ranks, because `saveVenatorSkillRanks` → `syncGang` → `rewriteFighterOverrides` deletes each fighter's `fighter_skill_access_override` rows for the affected Skill Sets and re-derives them from the new ranking. That side effect is unchanged in this PR (`utils/syncVenatorSkillOverrides.ts` still does the delete+reinsert), but the two `window.confirm` calls that warned about it are gone — saving now happens silently whenever `!ranksUnchanged`, whether ranks were edited or cleared entirely.

Failure scenario: a gang owner reorders/removes a Skill Set they'd previously ranked, not realizing it will wipe any manually-set "Customise Skill Set Access" overrides on their fighters. Previously they got a chance to cancel; now it just happens on Save.

This isn't mentioned in the PR description's summary or risk section, so it may be unintentional. Suggest restoring a confirm (or a less intrusive inline warning) gated on `hadPreviousRanks && !ranksUnchanged`, covering both the "ranks changed" and "ranks cleared" cases.

**2. [Low] Duplicated literal string across two components**
`components/fighter/fighter-advancement-list.tsx` and `components/fighter/fighter-skills-list.tsx`

Both now hardcode the identical "This Skill Set is not accessible to this fighter. Change their Skill Set access in: Edit Fighter > Customise Skill Set Access." string inline (previously centralized in `skillAccessDeniedMessage` in `venatorSkillAccess.ts`). The PR description explains the rationale for moving it out of the Venator util (it's not Venator-specific), which is reasonable — but leaving it as a shared generic constant (e.g. in a skills-ui util) rather than two copies would avoid the strings drifting apart later. Not blocking, just a maintainability nit.

### Other things checked, no issues found
- `deriveOverrides` in `venatorSkillAccess.ts` works correctly for partial (1-3) rank sets — no assumption of exactly 4 ranks.
- `getGangCore`'s cache key bump (`gang-core-v4` → `v5`) correctly invalidates cached data for the renamed field.
- New `@dnd-kit/modifiers` dependency and the touch/mouse sensor split (`useDndSensorsConfig('distance')`) look correctly scoped to this compact reorder list; scroll-lock on drag start/end/cancel is handled.
- Combobox padding change (`clearable ? "pr-[5.5rem]" : "pr-12"`) is a safe, purely visual sizing fix.
- No leftover references to `venator_no_ranks` / `venatorNoRanks` / `skillAccessDeniedMessage` anywhere in the codebase.

### Verdict: **Changes requested** (pending confirmation on Finding 1 — likely a quick fix or an intentional call either way)

---

joeseos (approved):

PR #2084, opened 2026-09-01
Make Venator Skill Set rank saves atomic and surface sync failures.

Djidiouf:
## Summary

- Make saving Venator Skill Set ranks atomic: one RPC replaces `gang_skill_set_ranks` and rewrites rank-derived `fighter_skill_access_override` rows in a single transaction, so a failed insert can no longer leave an empty rank table.
- Same for per-fighter sync on add/copy/edit (`replace_fighter_skill_access_overrides`). Add/edit now surface a skill-access warning toast; copy rolls back the new fighter if sync fails (and reports if that rollback fails).
- Restore a warning before changing or clearing existing ranks as an in-app modal (not `window.confirm`), because that rewrite resets custom Customise Skill Set Access on affected Skill Sets.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- [x] Edit a Venator gang with existing ranks → change or clear ranks → in-app confirm appears → Cancel leaves ranks unchanged; Continue saves and fighter access matches the new ranking
- [x] First-time ranking (no previous ranks) saves without the confirm
- [x] Failed rank save (e.g. before the migration is applied) shows an error and leaves Edit Gang open
- [x] Add / edit a Venator fighter when sync would fail: fighter save succeeds, skill-access warning toast is shown
- [x] Copy a Venator fighter: success copies access; if sync fails, copy is rolled back (or the error says the incomplete copy could not be removed)

## Database

  - `supabase/migrations/20260901072836_replace_gang_skill_set_ranks.sql` (defines both RPCs below)

**`replace_gang_skill_set_ranks`** (`supabase/functions/replace_gang_skill_set_ranks.sql`)  
Called when saving Edit Gang → Venator Skill Access. In one transaction it deletes the gang's rows in `gang_skill_set_ranks`, inserts the new 0–4 ranked Skill Sets, then deletes and re-inserts `fighter_skill_access_override` rows for those Skill Sets on every fighter in the gang (derived Primary/Secondary from rank + subtype).

**`replace_fighter_skill_access_overrides`** (`supabase/functions/replace_fighter_skill_access_overrides.sql`)  
Called when adding, copying, or editing a Venator fighter. In one transaction it deletes that fighter's override rows for the gang's ranked Skill Sets, then inserts the derived overrides for that fighter only.

Both are `SECURITY INVOKER` (table RLS still applies). Canonical copies live under `supabase/functions/`; the migration applies them to the database.

### Migration status

## Risk / rollout

Medium — **do not merge until the migration is applied.** After deploy, rank save calls `replace_gang_skill_set_ranks` and fighter add/copy/edit sync calls `replace_fighter_skill_access_overrides`. Without those functions, those paths will fail. No env vars or feature flags.

### DIFF
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -1231,7 +1231,6 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
               : [],
           },
           supabase,
-          user.id,
         );
       } catch (err) {
         console.error('syncFighter failed after add-fighter', err);
--- a/app/actions/add-fighter.ts
+++ b/app/actions/add-fighter.ts
@@ -1235,6 +1235,8 @@ export async function addFighterToGang(params: AddFighterParams): Promise<AddFig
         );
       } catch (err) {
         console.error('syncFighter failed after add-fighter', err);
+        archetypeSkillAccessWarning =
+          'Fighter added but skill access save failed. Please try again via Customise Skill Set Access.';
       }
     }
 
--- a/app/actions/copy-fighter.ts
+++ b/app/actions/copy-fighter.ts
@@ -431,6 +431,18 @@ export async function copyFighter(params: CopyFighterParams): Promise<CopyFighte
         );
       } catch (err) {
         console.error('syncFighter failed after copy-fighter', err);
+        const { error: deleteErr } = await supabase.from('fighters').delete().eq('id', newFighterId);
+        if (deleteErr) {
+          console.error('Failed to roll back copied fighter after syncFighter failure', deleteErr);
+          return {
+            success: false,
+            error: 'Failed to copy fighter skill access, and the incomplete copy could not be removed. Please delete it and try again.',
+          };
+        }
+        return {
+          success: false,
+          error: 'Failed to copy fighter skill access. Please try again.',
+        };
       }
     }
 
--- a/app/actions/copy-fighter.ts
+++ b/app/actions/copy-fighter.ts
@@ -427,7 +427,6 @@ export async function copyFighter(params: CopyFighterParams): Promise<CopyFighte
               : [],
           },
           supabase,
-          user.id,
         );
       } catch (err) {
         console.error('syncFighter failed after copy-fighter', err);
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -1817,6 +1817,7 @@ export async function updateFighterDetails(params: UpdateFighterDetailsParams): 
           );
         } catch (err) {
           console.error('syncFighter failed after edit-fighter', err);
+          archetypeSkillAccessWarning = ARCHETYPE_SKILL_ACCESS_WARNING;
         }
       }
     }
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -1813,7 +1813,6 @@ export async function updateFighterDetails(params: UpdateFighterDetailsParams): 
               subtypes: effectiveSubtypes,
             },
             supabase,
-            user.id,
           );
         } catch (err) {
           console.error('syncFighter failed after edit-fighter', err);
--- a/app/actions/gang/save-venator-skill-ranks.ts
+++ b/app/actions/gang/save-venator-skill-ranks.ts
@@ -2,7 +2,7 @@
 
 import { createClient } from '@/utils/supabase/server';
 import { getAuthenticatedUser } from '@/utils/auth';
-import { syncGang } from '@/utils/syncVenatorSkillOverrides';
+import { replaceGangRanks } from '@/utils/syncVenatorSkillOverrides';
 import { invalidateGang } from '@/utils/cache-tags';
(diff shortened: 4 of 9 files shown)


joeseos:
## Code Review

@Djidiouf Reviewed this against the current `main` and the surrounding call sites (add/copy/edit fighter, `gang-edit-modal.tsx`, RLS policies on `gang_skill_set_ranks` / `fighter_skill_access_override` / `fighters`).

**Overall**: solid fix. The atomicity problem is real (delete-then-insert with no transaction) and both new RPCs correctly wrap the rewrite in a single `plpgsql` function body. I traced through the specific scenarios called out in the PR description and they check out:

- `replace_gang_skill_set_ranks`: delete+insert of `gang_skill_set_ranks` and the rank-derived `fighter_skill_access_override` rows now happen in one transaction, so a failed insert can no longer leave an empty rank table. Clearing ranks (`p_ranks = []`) correctly deletes owned overrides and inserts nothing, since `deriveOverrides` returns `[]` for empty ranks.
- `replace_fighter_skill_access_overrides`: same shape for the single-fighter case; `p_fighter_id` is taken from the function argument rather than trusted from the payload, so a caller can't attach override rows to a different fighter — good.
- Both functions ignore the caller-supplied `user_id`/`o.user_id` and always stamp `auth.uid()` server-side, so a client can't forge the override owner.
- `SECURITY INVOKER` + `auth.uid()` guard preserves the exact same RLS-gated permission model the old app-level code relied on (verified against the `gang_skill_set_ranks` and `fighter_skill_access_override` RLS policies) — this isn't a privilege escalation, just an atomicity fix.
- `supabase/functions/*.sql` and the migration are byte-identical (diffed them, only comments differ) — good, avoids drift between the canonical copy and the applied migration.
- `copy-fighter.ts`: the rollback-on-sync-failure only needs to delete the just-inserted `fighters` row, since `syncFighter` is called before vehicles/equipment/skills are copied — no orphaned data risk there. The double-rollback-failure message path is a reasonable degrade.
- Confirm-modal flow in `gang-edit-modal.tsx`: `handleSave` returning `false` correctly stops `Modal`'s `onConfirm` from auto-closing (verified against `components/ui/modal.tsx`), so "Cancel leaves ranks unchanged" and "first-time ranking skips the confirm" both hold up.

### Minor / low severity

1. **Dead parameter left in place** (readability) — `utils/syncVenatorSkillOverrides.ts:26,54` (`syncFighter`, `replaceGangRanks`): `actingUserId` was renamed to `_actingUserId` instead of being removed, since the RPCs now derive the user from `auth.uid()` server-side and no longer use it. All four call sites (`save-venator-skill-ranks.ts`, `add-fighter.ts`, `edit-fighter.ts`, `copy-fighter.ts`) still pass `user.id`/`gang.user_id` into a parameter that's now unused. Suggest dropping the parameter entirely rather than prefixing with `_` — as-is it's a bit misleading (looks load-bearing, isn't).
2. **Nit**: `getVenatorRankSaveState()` (`gang-edit-modal.tsx:554`) is only called from one place, so the extraction doesn't buy much — not wrong, just an optional simplification.

Nothing here blocks merge; #1 is a one-line cleanup if you want to take it, otherwise fine to leave for a follow-up.

### Verdict: **Approve**

Correctness, edge cases, and the RLS/security model all hold up. Just flagging the reminder from the PR description itself: **the migration needs to be applied before this merges/deploys**, since the app code now unconditionally calls both new RPCs on the Venator add/copy/edit/rank-save paths.

---

joeseos (commented):
Don't think it needs to be atomic. We should be able to handle this just like everything else by adding the missing update RLS policy.

Should I merge and then later look into refactoring?

joeseos:
## Follow-up Review

@Djidiouf Reviewed just the new commit (`b570b12` → `9011cc3`) against the prior review's findings.

### Previous findings

1. **Dead `actingUserId` parameter** (`utils/syncVenatorSkillOverrides.ts`) — **Resolved.** The parameter is fully removed (not just renamed) from both `syncFighter` and `replaceGangRanks`, and all four call sites (`add-fighter.ts`, `copy-fighter.ts`, `edit-fighter.ts`, `save-venator-skill-ranks.ts`) were updated to stop passing `user.id`/`gang.user_id`. Confirmed `user` is still used for other purposes in each of those files (auth check, ownership checks, other `user_id` stamps), so this isn't an unused-variable regression — just the dead arg is gone. In `save-venator-skill-ranks.ts` the `getAuthenticatedUser(supabase)` call is kept (so the auth check itself is preserved) with its return value now correctly discarded.
2. **`getVenatorRankSaveState()` single-use extraction** (`gang-edit-modal.tsx`) — **Resolved.** Inlined back into `handleSave`; diffed the logic and it's an exact move, no behavior change.

### New findings

None. This commit is a pure mechanical cleanup addressing the two minor points from the last review — no new logic, no touched SQL/RPC files, no edge-case or null-handling changes to review.

### Verdict: **Approve**

---

joeseos (approved):

PR #2085, opened 2026-09-01
feat(battle-session): hide Blaze and Flesh Wounds in N26

joeseos:
The fighter action modal offered every condition token to every edition.
Blaze and Flesh Wounds are N23 rules that N26 does not have, so N26
players were offered conditions their edition doesn't use.

Add blazeCondition and fleshWounds to EDITION_CAPABILITIES with the
matching hasBlazeCondition / hasFleshWounds accessors, and filter both
pickers on the session's edition_slug, which FighterActionModal already
receives. The counters heading drops "Flesh Wounds" when it isn't shown.

CONDITION_BY_KEY keeps every definition so a value another edition
already recorded still renders its badge.

### DIFF
--- a/components/battle-session/participant-card.tsx
+++ b/components/battle-session/participant-card.tsx
@@ -52,7 +52,7 @@ import { createGangLog } from '@/app/actions/logs/gang-logs';
 import { updateFighterXp } from '@/app/actions/edit-fighter';
 import FighterCard from '@/components/gang/fighter-card';
 import type { BattleSessionFull, BattleSessionParticipant, BattleSessionFighter, SessionCondition, SessionInjuryRecord } from '@/types/battle-session';
-import { beastSubtypeName } from '@/types/edition';
+import { beastSubtypeName, hasBlazeCondition, hasFleshWounds } from '@/types/edition';
 
 interface ConditionDefinition {
   key: string;
@@ -187,6 +187,17 @@ function FighterActionModal({
     fighter.session_record?.conditions ?? []
   );
 
+  // Conditions the edition doesn't use aren't offered. CONDITION_BY_KEY keeps every
+  // definition, so a value another edition already recorded still renders its badge.
+  const sessionConditions = useMemo(
+    () => SESSION_CONDITIONS.filter((c) => c.key !== 'blaze' || hasBlazeCondition(editionSlug)),
+    [editionSlug]
+  );
+  const numericConditions = useMemo(
+    () => NUMERIC_CONDITIONS.filter((c) => c.key !== 'flesh_wound' || hasFleshWounds(editionSlug)),
+    [editionSlug]
+  );
+
   const openXpModal = async () => {
     setLoadingXp(true);
     try {
@@ -271,9 +282,11 @@ function FighterActionModal({
             </Button>
           </div>
           <div className="space-y-2 border-t pt-3 text-left">
-            <h4 className="text-sm font-medium text-neutral-500">Wounds & Flesh Wounds</h4>
+            <h4 className="text-sm font-medium text-neutral-500">
+              {hasFleshWounds(editionSlug) ? 'Wounds & Flesh Wounds' : 'Wounds'}
+            </h4>
             <div className="flex flex-col gap-2">
-              {NUMERIC_CONDITIONS.map((nc) => {
+              {numericConditions.map((nc) => {
                 const current = draftConditions.find((c) => c.key === nc.key)?.value ?? 0;
                 return (
                   <div key={nc.key} className="flex items-center justify-between">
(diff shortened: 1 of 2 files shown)


joeseos:
## Code review — @joeseos

Reviewed the 2-file / +27−4 diff against the full codebase. The approach is right: the gate lands in `EDITION_CAPABILITIES` rather than a slug comparison in the component, it reuses the `editionSlug` that `FighterActionModal` already receives (no new prop threading), and `CONDITION_BY_KEY` is deliberately left whole so already-stored values still render their badge. CI (`deploy-preview`) is green. No correctness, security, or performance blockers.

One scope gap is worth tracking, the rest are polish.

### Findings

**1. Flesh Wound boxes still print for N26 gangs — `components/gang/print-gang.tsx:436`, `:777`, `:800`, `:1379`, `:1402` — severity: medium**
This PR establishes `hasFleshWounds` as *the* edition capability for the rule, but the only other place Flesh Wounds surface to users is untouched: the gang print sheet offers a "Wounds/Flesh Wounds Boxes" option (`:436`) and renders four `FW` checkbox blocks — toughness-derived at `:777` / `:1379`, a fixed three boxes for crew at `:800` / `:1402`. An N26 gang still prints them. `print-gang.tsx` is already edition-aware (`hasTradePoints` at `:512`, `hasSaveCharacteristic` at `:724`, `hasAlignment` at `:1150`), so it is the same pattern.
*Fix:* gate each `FW` block on `hasFleshWounds(fighter.edition_slug ?? edition_slug)` and make the checkbox label conditional ("Wounds Boxes" when the edition has no flesh wounds). Note the `W` half must stay — N26 still has Wounds — so this is not a whole-block gate. Fine as a follow-up PR, but it shouldn't get lost: the capability now exists and the inconsistency is user-visible.

**2. `useMemo` is overhead here — `components/battle-session/participant-card.tsx:192-199` — severity: low**
The memos filter an 11-element and a 2-element array. The bookkeeping costs more than the filter, and nothing consumes referential stability — both results are `.map`ped inline. Every other edition gate in the codebase is a plain const (`components/equipment/purchase-modal.tsx:86`, `components/fighter/fighter-actions.tsx:175`, `components/gang/gang-edit-modal.tsx:202`).
*Fix:* drop both `useMemo` wrappers for plain `const`s.

**3. The same predicate is evaluated two independent ways — `participant-card.tsx:197` and `:286` — severity: low (DRY)**
`hasFleshWounds(editionSlug)` decides the heading text while the list is filtered by a separate call, so a reader has to confirm the two agree. Combined with finding 2:
```ts
const showBlaze = hasBlazeCondition(editionSlug);
const showFleshWounds = hasFleshWounds(editionSlug);
const sessionConditions = SESSION_CONDITIONS.filter((c) => showBlaze || c.key !== 'blaze');
const numericConditions = NUMERIC_CONDITIONS.filter((c) => showFleshWounds || c.key !== 'flesh_wound');
```
then reuse `showFleshWounds` for the heading at `:286`.

**4. Sibling capability rows are named inconsistently — `types/edition.ts:215` and `:217` — severity: low**
`blazeCondition` vs `fleshWounds`, added in the same commit for the same feature. More importantly, `hasFleshWounds(x)` reads like a per-fighter predicate — "does this fighter have flesh wounds?" — in a codebase where fighters genuinely accumulate them (`session_record.conditions`), and only the argument name disambiguates it.
*Fix:* rename to `fleshWoundCondition` / `hasFleshWoundCondition` to match its sibling and remove the ambiguity. Cheap now, three call sites.

**5. A null-edition session would silently lose both conditions — `types/edition.ts:366-370` — severity: low**
`can()` answers `false` for a null or unrecognised slug, so a session with no resolved edition hides Blaze and Flesh Wounds rather than falling back to N23. This matches every other `{ n23: true, n26: false }` accessor (`hasChemAlchemy`, `hasMasterCraftedWeapons`, `hasGuilderSales`), so it is the house convention and not a defect — and it is unreachable today: all 2,338 `battle_sessions` rows resolve to an edition (2,043 n23 / 295 n26, zero null). Flagging it only because `createBattleSession` explicitly tolerates a null `edition_id` (`app/actions/battle-sessions.ts:192-226`), so the path can appear later. No change requested; worth knowing.

### Noted, no action

- **Existing N26 values become unclearable.** 42 fighter records in N26 sessions already carry `blaze` (2) or `flesh_wound` (40). Their badges still render, but with the controls hidden they can no longer be decremented or toggled off from the modal. Confirmed as an accepted trade-off — recording it so it isn't rediscovered as a bug.
- **The gate is presentational only.** `updateSessionConditions` still accepts any key, so nothing stops a crafted request writing `blaze` into an N26 session. Consistent with how the rest of `session_record` is treated (free-form per-session scratch data, not rules-enforced), so no server-side validation is warranted here.

### Verdict

**Approve.** No blockers: behaviour is correct in both editions, conventions are followed, and CI is green. Findings 2–4 are small enough to fold into this PR if you want it tidy; finding 1 is the one worth opening a follow-up for, since the PR's rationale applies equally to the print sheet.

---

Djidiouf (approved):

PR #2089, opened 2026-09-02
Link alternate house gang lists to their root gang type so house iden…

Djidiouf:
…tity is a foreign key, not a name prefix.

## Summary

- Add ``parent_gang_type_id`` on ``gang_types`` so N26 gang lists (Wyld Hunt, Chymist Cult, Furnace Brutes, Unborn) reference their house.
- Keep the tree one level deep (parent must be a root) and backfill existing colon-named rows in the same edition.

**Why:** those gang lists are separate ``gang_types`` rows with their own fighter catalogs, but they are still House Escher / House Goliath for hatred, the list of gangs when creating a gang, and for potential other effects applying to a whole House. The only link today is the name (``House Escher: Wyld Hunt``). That is not a join key: ``Variant: Chaos Corrupted`` also contains a colon, and ``ILIKE 'House Escher%'`` is unsafe. Without a parent FK, unhiding the lists would treat them as extra houses, and Hatred (type of gang) on House Escher would not cover Wyld Hunt or Chymist Cult. This is also needed to add a secondary choice when creating a gang where a user chooses House Escher, for them to select if they want the standard gang, or one of the child gang (ie. Chymist cult, Wyld hunt)

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- [x] Migration not applied yet
- [ ] After apply: four children parented to House Escher / House Goliath; ``Variant: …`` rows stay roots

## Database

- [x] Migration added under ``supabase/migrations/``
- [x] RPC/SQL function changed: updated both ``supabase/migrations/`` and the matching file under ``supabase/functions/``

### Migration status

## Risk / rollout

- Low — additive nullable column, no app callers yet. Apply the migration before this PR is merged (or before the functions file is deployed from ``main``), or the ``UPDATE OF parent_gang_type_id`` trigger will fail.

### DIFF
--- a/supabase/README.md
+++ b/supabase/README.md
@@ -60,3 +60,4 @@ Syncing of those files is not automatic. If you update a function on Supabase, m
 | notify_gang_invite           | campaign_gangs     | Sends notification for gang invites       |
 | enqueue_notification_email   | notifications      | Queues an outbound email for a new notification |
 | handle_new_user              | auth.users         | Creates a profiles row on signup (`on_auth_user_created`) |
+| gang_types_parent_must_be_root | gang_types       | Keeps parent_gang_type_id one level deep (parent must be a root) |
--- a/supabase/functions/gang_types_parent_must_be_root.sql
+++ b/supabase/functions/gang_types_parent_must_be_root.sql
@@ -1,0 +1,44 @@
+-- Keep gang_types.parent_gang_type_id one level deep: the parent must itself be
+-- a root, and a type that already has children cannot become a child.
+--
+-- DEPLOY ORDER: apply migration 20260902172743_add_gang_types_parent_gang_type_id.sql
+-- (which creates the column) BEFORE this trigger goes live. UPDATE OF
+-- parent_gang_type_id will fail if the column does not exist.
+
+CREATE OR REPLACE FUNCTION public.gang_types_parent_must_be_root()
+RETURNS trigger
+LANGUAGE plpgsql
+SET search_path = public
+AS $$
+BEGIN
+  IF NEW.parent_gang_type_id IS NOT NULL THEN
+    IF EXISTS (
+      SELECT 1
+      FROM public.gang_types
+      WHERE gang_type_id = NEW.parent_gang_type_id
+        AND parent_gang_type_id IS NOT NULL
+    ) THEN
+      RAISE EXCEPTION 'parent_gang_type_id must reference a root gang type';
+    END IF;
+
+    IF EXISTS (
+      SELECT 1
+      FROM public.gang_types
+      WHERE parent_gang_type_id = NEW.gang_type_id
+    ) THEN
+      RAISE EXCEPTION 'cannot set parent_gang_type_id on a gang type that is already a parent of other gang lists';
+    END IF;
+  END IF;
+
+  RETURN NEW;
+END;
+$$;
+
+REVOKE ALL ON FUNCTION public.gang_types_parent_must_be_root() FROM PUBLIC;
+REVOKE EXECUTE ON FUNCTION public.gang_types_parent_must_be_root() FROM anon, authenticated;
+
+DROP TRIGGER IF EXISTS gang_types_parent_must_be_root ON public.gang_types;
+CREATE TRIGGER gang_types_parent_must_be_root
+  BEFORE INSERT OR UPDATE OF parent_gang_type_id ON public.gang_types
+  FOR EACH ROW
+  EXECUTE FUNCTION public.gang_types_parent_must_be_root();
--- a/supabase/migrations/20260902172743_add_gang_types_parent_gang_type_id.sql
+++ b/supabase/migrations/20260902172743_add_gang_types_parent_gang_type_id.sql
@@ -112,12 +112,25 @@ CREATE TRIGGER gang_types_parent_must_be_root
   EXECUTE FUNCTION public.gang_types_parent_must_be_root();
 
 -- One-time name heuristic: "{parent}: {suffix}" whose parent exists as an exact
--- gang_type in the same edition. Does not match Variant: … (no type named Variant).
+-- gang_type in the same edition. starts_with, not LIKE, so %/_ in a name are
+-- literal. A scalar subquery raises if a child matches more than one root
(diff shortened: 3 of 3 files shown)


joeseos:
@Djidiouf Reviewed this against the current schema (confirmed `gang_types` already has the `gang_types_gang_type_id_edition_id_key UNIQUE (gang_type_id, edition_id)` needed for the composite FK, and that the `deploy_supabase_functions.yml` workflow auto-runs any changed `supabase/functions/*.sql` on push to `main`). Overall the design (nullable self-FK + one-level-deep trigger + backfill) is sound. A few things worth addressing before/at merge:

## Findings

**1. (Medium) Backfill `UPDATE` has no dry-run/uniqueness safeguard** — `supabase/migrations/20260902172743_..._id.sql`, lines ~117-123
```sql
UPDATE public.gang_types AS child
SET parent_gang_type_id = parent.gang_type_id
FROM public.gang_types AS parent
WHERE ...
  AND child.gang_type LIKE (parent.gang_type || ': %');
```
If a child row's name ever happens to match more than one root's `"<root>: %"` prefix (ambiguous join), plain `UPDATE ... FROM` picks a match nondeterministically rather than erroring — unlike `MERGE`, there's no cardinality check. With only 4 known children today this is low-probability, but since the PR's own "How I tested" section has the post-apply verification unchecked, I'd suggest running the `WHERE` clause as a plain `SELECT child.gang_type, parent.gang_type` first to confirm exactly the 4 expected 1:1 matches before letting the `UPDATE` run against prod.

**2. (Low) `LIKE` pattern built from table data isn't escaped** — same file, backfill `WHERE` clause
`parent.gang_type || ': %'` is passed straight into `LIKE`, so if any `gang_type` value itself contains a literal `%` or `_`, it's interpreted as a wildcard rather than a literal character (e.g. an underscore in a name would match any single character there). Data is curated, not attacker-controlled, so this isn't a security issue, just a correctness edge case — `starts_with(child.gang_type, parent.gang_type || ': ')` would be safer than `LIKE` here and reads a bit clearer too.

**3. (Medium, process) Deploy-order hazard is real, not just theoretical** — I checked `.github/workflows/deploy_supabase_functions.yml`: it triggers on every push to `main` that touches `supabase/functions/*.sql` and runs `psql ... -f "$f"` with `ON_ERROR_STOP=1`. Since this PR adds `gang_types_parent_must_be_root.sql`, merging to `main` before the migration is manually applied will make CI immediately try `CREATE TRIGGER ... UPDATE OF parent_gang_type_id ON public.gang_types` against a column that doesn't exist yet, and the deploy job will hard-fail. You've called this out clearly in the file header and PR description, so this is really just a flag for whoever merges: apply the migration first, in the same sitting as the merge, not just "before merge" as a mental note — otherwise the very next `main` push (this one or an unrelated one) will fail CI.

**4. (Low, docs) `supabase/README.md` "Trigger Functions" table not updated** — the repo documents trigger functions there (`notify_campaign_member_added`, etc.) alongside their `supabase/functions/*.sql` file; `gang_types_parent_must_be_root` should probably get a row too, per that existing convention.

## Things that look correct and worth calling out
- The `edition_id IS NOT NULL WHEN parent set` CHECK correctly compensates for `MATCH SIMPLE` on the composite FK — good catch on that gap.
- The one-level-deep trigger's checks do see prior rows updated earlier in the *same* bulk statement (Postgres advances the command counter per row for BEFORE ROW triggers), so if the naming heuristic ever did produce a 3-level chain, the whole migration would abort loudly rather than silently corrupting the tree — safe failure mode.
- `REVOKE`s on the trigger function and duplicate migration/`functions/` trigger body match existing project conventions (e.g. `notify_campaign_member_added.sql`).

## Verdict
**Changes requested** — mainly item 1 (verify the backfill match set before running it against prod) and making sure item 3's deploy ordering is actually followed at merge time. Items 2 and 4 are minor and don't need to block.

---

joeseos (approved):

joeseos:
@Djidiouf Reviewed the two new commits since the last review (`02027b0` → `a505400`), which touch only `supabase/migrations/20260902172743_..._id.sql` and `supabase/README.md`.

## Previous findings

1. **(Medium) Backfill `UPDATE` had no cardinality/uniqueness safeguard** — **Resolved**, and via a stronger mechanism than what I suggested. The backfill now does `SET parent_gang_type_id = (SELECT parent.gang_type_id FROM ... WHERE ...)` — a scalar subquery — instead of `UPDATE ... FROM`. If a child ever matched more than one root, Postgres raises `more than one row returned by a subquery used as an expression` and the whole migration aborts, rather than silently picking a match at random. That's better than my original "run it as a SELECT first" suggestion, since it's enforced automatically rather than relying on the person applying it to remember to check. The outer `WHERE ... EXISTS (...)` uses the identical predicate, so it can't diverge from the scalar subquery's match set (can't be 0 rows when `EXISTS` is true, can't be >1 without erroring).

2. **(Low) `LIKE` pattern wasn't escaped** — **Resolved.** Swapped to `starts_with(child.gang_type, parent.gang_type || ': ')`, so a literal `%`/`_` in a `gang_type` name is no longer treated as a wildcard.

3. **(Medium, process) Deploy-order hazard** — **Outstanding**, as expected — this isn't something a code change fixes. The PR's own checklist still shows "Migration not applied yet," so the deploy-order requirement (apply the migration in the same sitting as merging to `main`, since CI auto-deploys `supabase/functions/*.sql` on every push) still applies at merge time. No action needed now, just don't lose track of it before merging.

4. **(Low, docs) `supabase/README.md` Trigger Functions table not updated** — **Resolved.** `gang_types_parent_must_be_root` now has a row, consistent with the existing convention.

## New findings

None. The rewritten backfill `UPDATE` is correct: `NOT DISTINCT FROM` on `edition_id` still lets two null-edition rows match each other same as before, and the `functions/gang_types_parent_must_be_root.sql` trigger body is untouched and still matches the migration's copy verbatim.

## Verdict

**Approve** — all addressable findings from the previous review are resolved. The one remaining item (deploy ordering) is a merge-time process step already flagged clearly in the PR description and migration comments, not a code defect.

---

PR #2090, opened 2026-09-03
refac: cut gang page CPU — weapon table derivation, plus cache invalidation gaps

joeseos:
## Summary

Started as "the weapon tables are slow". What survived investigation is smaller than that framing suggests, and the description below is deliberately conservative about it.

**1. Weapon table row derivation moved out of the render path.** `fighter-card-weapon-table.tsx` rebuilt its whole grouping/dedup/sort chain on every render, even though it depends only on the `weapons` prop. That chain now lives in a pure `utils/weapon-variants.ts` behind a single `useMemo`. `getVehicleWeapons` was duplicated — `fighter-card.tsx` and a second copy inlined in `print-gang.tsx`'s render loop, differing only in a fallback label — and is now one parameterised builder, which also let the print call site keep a stable array reference.

The one measured speedup inside that move: the two weapon-name sorts called `localeCompare(x, undefined, { sensitivity: 'base' })`, and passing options defeats V8's cached-collator fast path, so they built an `Intl.Collator` per comparison. A shared module-level collator makes that **2838 ms → 63 ms over 200k sorts (44.8x)**.

End to end, isolated benchmark of 76 weapon tables, four paired runs (this container has ~6% run-to-run variance, so single runs are not meaningful): **baseline median 80.36 ms → 68.34 ms, 1.18x** — 1.13x comparing minima. The rest of the value is structural: the pipeline runs once per weapon-set change rather than once per render, and the logic is testable in isolation.

The components sit inside the `gang-page-content.tsx` client boundary but are still SSR'd on every initial load, so this is billed compute as well as device time.

Honest bound: p50 gangs were never slow. The dominant large-gang cost is React building ~19k DOM nodes, which this PR does not touch.

**2. Cache invalidation.** Two changes, and the second is a consequence of the first rather than an independent fix:

- `invalidateGangFinancials` busted the whole fighters bundle on every credit change; it now targets `gangCore`/`gangOverview`. All 25 direct callers were audited to confirm the bundle is still busted wherever fighter rows actually change.
- `app/actions/update-vehicle.ts` now calls `invalidateGang` explicitly. On main, vehicle edits reached the bundle *via* `invalidateGangFinancials` busting `gang-{id}`; the narrowing above removes that, so this restores it. No bug existed on main — this is scaffolding for the change above, not a fix.

**Also:** `initializePositioningIfNeeded` wrote a default positioning map to the database during render. That write could never take effect — `revalidateTag` throws during render — so the cached `null` came back and the write repeated on every page load. Replaced with a pure `resolvePositioning`.

### Five things this PR tried and removed

All were mine. Recorded so they don't get re-proposed without the counter-argument.

1. **A `global-equipment` invalidation tag** (`24aed50`). Closed a real gap — admin catalogue edits don't reach the equipment copies PostgREST embeds per `fighter_equipment` row — but one tag carried the bundle and stash entries for *every* gang, so a single admin edit would drop them database-wide. Disproportionate to staleness that is cosmetic and self-healing on the next gang mutation. `app/api/admin/equipment/route.ts` is byte-identical to main.
2. **A React `cache()` layer** over the gang-data accessors plus a `getRequestClient` factory (`a390428`). Justified on a Redis round-trip that Vercel never pays: `next.config.js` puts `cacheHandler` and `cacheMaxMemorySize: 0` behind `USE_REDIS_CACHE`, which only Coolify sets. The real cost of a duplicate read is the `JSON.parse` — 0.216 ms at p50, 0.390 ms at p90, 3.071 ms for the largest gang — so the layer bought ~0.4 ms at p90, for a second caching mechanism in a codebase that had none.
3. **An eslint `no-restricted-imports` rule** (`b1cc3f1`). Enforced a constraint this branch invented, for the `cache()` layer above. Reworded to blame `unstable_cache`, it names one of nineteen equivalent modules — and server actions already import four of the others. `eslint.config.mjs` is byte-identical to main.
4. **A shared collator for the traits sort** (`cc8709f`). The 44.8x win comes from defeating per-comparison collator construction, which V8 only pays when `localeCompare` is given arguments. The traits sort was already bare `localeCompare()`, so routing it through a collator was **25% slower** (52.5 ms → 65.6 ms per 200k sorts). I had claimed three sites benefited; two do.
5. **Interning the variant map key** (`b1735a1`). Claimed as the second measured hot spot. Against the real distribution — 87.5% of owned weapons have exactly one profile — interning is **0.96x at one profile**, since the long suffix is built once per weapon either way and the map lookup is pure overhead. Net 1.03x–1.21x depending on ammo ownership: one to eight microseconds per gang, for ten lines and an invariant about uuids never containing a pipe.

**Not doing here**, deliberately: deduplicating the profile rows PostgREST embeds in the bundle (~4% of payload), the add-equipment modal RPC (different surface), and large-gang DOM weight (a UX change, not a refactor).

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] **Render parity, 460 cases, byte-identical** (6,380,568 bytes each side). A `renderToString` harness renders the real components against generated gangs at 7/12/20/76 fighters — vehicles, crew, master-crafted weapons, ammo profiles sharing a `weapon_group_id`, both editions, all view modes, plus the print roster — with baseline components pinned at the branch point. This is the gate for the grouping rewrite: the semantics are subtle (orphan-specials filter, duplicate counting via a `Set`, first-special-wins, integer-like key ordering) and a diff-read would not have caught a regression. Re-run after every revert above.
- [x] **Equipment-targeted effects on weapons**, added to the fixtures after review asked about them — the earlier run had none, so `applyWeaponModifiers` never fired and `effect_names` was always undefined. Fighter weapons now carry rig-glitch effects (strength/AP modifiers, plus a second lower-`sort_order` effect on some so the multi-name join and ordering fire), confirmed present in the output, parity still byte-identical.
- [x] Isolated benchmark of 76 weapon tables, four paired baseline/head runs, medians reported above rather than a single best-of.
- [x] Micro-benchmarks for each claimed optimisation individually — which is how items 4 and 5 above were caught.
- [x] `npx tsc --noEmit` clean; `npm run build` compiles and typechecks; `npx eslint` on the changed files clean apart from the pre-existing try/catch-around-JSX error in the gang page.
- [x] Invalidation audit: all 25 direct callers of `invalidateGangFinancials`.

## Risk / rollout

**Low–medium.** The refactor is gated on byte-identical output; the residual risk is the invalidation narrowing.

- No migration, no env vars, no feature flags.
- No cache key version bump and no new tags on existing entries. The only tag change is `invalidateGangFinancials` narrowing from `gang-{id}` to `gang-core-{id}` + `gang-overview-{id}`, with the `update-vehicle.ts` line above supporting it. That pair is what a second reader should look at.
- One guard added: `Object.freeze` on `NO_WEAPONS`, the empty array shared by every caller with no vehicle weapons, so a stray `push`/`sort` throws instead of corrupting the others.

### DIFF
--- a/app/@breadcrumb/fighter/[id]/page.tsx
+++ b/app/@breadcrumb/fighter/[id]/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { getRequestClient } from "@/utils/supabase/server"
+import { createClient } from "@/utils/supabase/server"
 import { getFighterBasic } from "@/app/lib/shared/fighter-data"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
@@ -18,7 +18,7 @@ export default async function FighterBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await getRequestClient()
+  const supabase = await createClient()
   
   // Reads the fighter/gang cached entries instead of an uncached query
   const fighterData = await getFighterBasic(id, supabase).catch(() => null)
--- a/app/@breadcrumb/fighter/[id]/page.tsx
+++ b/app/@breadcrumb/fighter/[id]/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { createClient } from "@/utils/supabase/server"
+import { getRequestClient } from "@/utils/supabase/server"
 import { getFighterBasic } from "@/app/lib/shared/fighter-data"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
@@ -18,7 +18,7 @@ export default async function FighterBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await createClient()
+  const supabase = await getRequestClient()
   
   // Reads the fighter/gang cached entries instead of an uncached query
   const fighterData = await getFighterBasic(id, supabase).catch(() => null)
--- a/app/@breadcrumb/gang/[id]/page.tsx
+++ b/app/@breadcrumb/gang/[id]/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { createClient } from "@/utils/supabase/server"
+import { getRequestClient } from "@/utils/supabase/server"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
   Breadcrumb,
@@ -17,7 +17,7 @@ export default async function GangBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await createClient()
+  const supabase = await getRequestClient()
 
   // Reads the gang page's cached core entry instead of an uncached query
   const gangData = await getGangCore(id, supabase).catch(() => null)
--- a/app/@breadcrumb/gang/[id]/page.tsx
+++ b/app/@breadcrumb/gang/[id]/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { getRequestClient } from "@/utils/supabase/server"
+import { createClient } from "@/utils/supabase/server"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
   Breadcrumb,
@@ -17,7 +17,7 @@ export default async function GangBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await getRequestClient()
+  const supabase = await createClient()
 
   // Reads the gang page's cached core entry instead of an uncached query
   const gangData = await getGangCore(id, supabase).catch(() => null)
--- a/app/@breadcrumb/gang/[id]/print/page.tsx
+++ b/app/@breadcrumb/gang/[id]/print/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { createClient } from "@/utils/supabase/server"
+import { getRequestClient } from "@/utils/supabase/server"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
   Breadcrumb,
@@ -17,7 +17,7 @@ export default async function GangPrintRosterBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await createClient()
+  const supabase = await getRequestClient()
 
   // Reads the gang page's cached core entry instead of an uncached query
   const gangData = await getGangCore(id, supabase).catch(() => null)
--- a/app/@breadcrumb/gang/[id]/print/page.tsx
+++ b/app/@breadcrumb/gang/[id]/print/page.tsx
@@ -1,5 +1,5 @@
 import { LuHouse } from "react-icons/lu";
-import { getRequestClient } from "@/utils/supabase/server"
+import { createClient } from "@/utils/supabase/server"
 import { getGangCore } from "@/app/lib/shared/gang-data"
 import {
   Breadcrumb,
@@ -17,7 +17,7 @@ export default async function GangPrintRosterBreadcrumb({
   params: Promise<{ id: string }>
 }) {
   const { id } = await params
-  const supabase = await getRequestClient()
+  const supabase = await createClient()
 
   // Reads the gang page's cached core entry instead of an uncached query
   const gangData = await getGangCore(id, supabase).catch(() => null)
--- a/app/@breadcrumb/user/[id]/page.tsx
+++ b/app/@breadcrumb/user/[id]/page.tsx
@@ -7,7 +7,7 @@ import {
   BreadcrumbPage,
   BreadcrumbSeparator,
 } from "@/components/ui/breadcrumb"
-import { createClient } from '@/utils/supabase/server'
+import { getRequestClient } from '@/utils/supabase/server'
 import { getUserProfile } from "@/app/lib/shared/gang-data"
 
 interface UserBreadcrumbProps {
@@ -18,7 +18,7 @@ export default async function UserBreadcrumb({ params }: UserBreadcrumbProps) {
   const { id } = await params;
   
   // Reads the cached profile entry instead of an uncached query
-  const supabase = await createClient();
+  const supabase = await getRequestClient();
   const profile = await getUserProfile(id, supabase).catch(() => null);
 
   const username = profile?.username || 'Unknown User';
--- a/app/@breadcrumb/user/[id]/page.tsx
+++ b/app/@breadcrumb/user/[id]/page.tsx
@@ -7,7 +7,7 @@ import {
   BreadcrumbPage,
   BreadcrumbSeparator,
 } from "@/components/ui/breadcrumb"
-import { getRequestClient } from '@/utils/supabase/server'
+import { createClient } from '@/utils/supabase/server'
 import { getUserProfile } from "@/app/lib/shared/gang-data"
 
 interface UserBreadcrumbProps {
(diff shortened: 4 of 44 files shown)


joeseos:
@joeseos Reviewed the full diff (20 files, 9 commits) rather than just the weapon-table piece the title/description describe, since the branch also rewrites cache-invalidation semantics, gang positioning, and adds per-request Supabase client memoization. Went deep on the caching change specifically since that's the part most likely to hide a silent, hard-to-notice bug (stale data with no error).

## What I verified carefully

**`invalidateGangFinancials` no longer busts `gang-{id}`** (`utils/cache-tags.ts`) — this is the riskiest change in the PR: previously every financial write also invalidated the fighters bundle as a side effect; now it only busts `gang-core-{id}` + `gang-overview-{id}`. I traced every one of the ~16 call sites of `invalidateGangFinancials` across `app/actions/**` and `app/api/**` to check whether the fighters bundle still gets busted when the surrounding write actually touches fighter/vehicle/equipment data:
- `utils/gang-rating-and-wealth.ts`, `update-gang.ts` (credits/reputation-only branches), `api/gangs/[id]/route.ts` (credits-only branch) — correctly gang-row-only, no fighter data touched, no bundle bust needed.
- `chem-alchemy.ts`, `edit-fighter.ts`, `sell-vehicle.ts`, `vehicle-hardpoints.ts`, `add-gang-vehicle.ts`, `move-to-stash.ts`, `move-from-stash.ts`, `fighter-advancement.ts`, `assign-vehicle-to-fighter.ts`, `equipment.ts`, `sell-equipment.ts`, `unassign-vehicle.ts`, `remove-vehicle-damage.ts`, `delete-vehicle.ts` — all have an unconditional `invalidateGang`/`invalidateFighter` call in the same action, so the bundle bust still happens regardless of the financials-tag change.
- `update-vehicle.ts` was the one place this actually broke (unassigned vehicle edits) — and it's fixed in the last commit (`b760879`). Good catch, and confirms the pattern above is the right one; I didn't find another instance of it.

This all checks out. Nice piece of work — genuinely hard to get this right across so many call sites.

**Weapon-variant grouping extraction** (`utils/weapon-variants.ts`) — read it line-by-line against the deleted inline version. The interned suffix-id map key, the `Intl.Collator` reuse (`.sort(traitCollator.compare)` is safe — `Intl.Collator.prototype.compare` is spec'd to return an already-bound function), and the deliberate choice to keep `baseGroups` a plain object rather than a `Map` (to preserve JS's integer-like-key reordering for numeric profile names) are all correct and match the deleted code's behavior exactly. Nothing to flag here.

**`resolvePositioning`** (`utils/fighter-positioning.ts`) — replacing the old write-during-render lazy-init (which could never actually invalidate its own cache tag, since `revalidateTag` throws during render, so the write silently repeated every load) with a pure derived default is a genuine bug fix, not just a refactor. Applying it to the battle-session page too is a correctness improvement (previously a never-reordered gang could show a different fighter order on the battle-session page vs. the gang page).

## Findings

**1. (Medium) PR scope vs. description** — Title/body describe only the weapon-table perf work, but 8 of 9 commits change cache-invalidation semantics app-wide, add request-level memoization (`getRequestClient`), and touch gang positioning. That's a much bigger, higher-blast-radius change than "lift weapon table row derivation out of the render path" suggests. Worth expanding the PR description so a future "why did gang X show stale data" investigation knows to look at this PR.

**2. (Low) Comment-only invariant, not enforced** — `app/lib/shared/gang-data.ts` new header comment: *"No server action may read these: it would see a pre-mutation value after its own write. Nothing under `app/actions/**` imports this module today."* I confirmed that's true right now, but nothing stops a future PR from importing `getGangCore` into a server action and silently reading stale (pre-write) data later in that same action. Consider an eslint `no-restricted-imports` rule scoped to `app/actions/**` against `@/app/lib/shared/gang-data` so this is enforced, not just documented.

**3. (Low) `NO_WEAPONS` is a shared mutable singleton** — `utils/weapon-variants.ts`, `export const NO_WEAPONS: Weapon[] = [];`. Nothing currently mutates it, but it's returned as a shared default from several call sites app-wide; a future accidental `.push()`/`.sort()` on a "vehicle weapons" array somewhere would silently corrupt it for every other fighter/vehicle in the app that hits the same default. `Object.freeze(NO_WEAPONS)` would turn that into a loud dev-time error instead of silent cross-render corruption.

**4. (Low, informational) No regression tests landed for the grouping rewrite** — the PR claims 458 rendered cases verified byte-identical, but the repo has no test framework at all (no jest/vitest, no test files anywhere), so this doesn't violate any existing convention — just flagging that this fairly intricate grouping/dedup/sort logic now has zero automated coverage against future regressions.

## Verdict
**Approve.** Nothing here blocks merge — findings are a documentation/description gap, a hardening suggestion, and two low-severity defensive nits. The core caching-semantics change is correct everywhere I checked, and the weapon-table extraction preserves behavior faithfully.

---

joeseos:
@joeseos Reviewed the 10 commits since the last review (`b760879` → `b1735a1`), diffed against just the PR's files to filter out the unrelated `main` merge (`e916e49`). Net effect of this batch: two guards were added then partly walked back, and three experimental changes flagged in the PR body's "Five things this PR tried and removed" were reverted — the `global-equipment` tag, the React `cache()`/`getRequestClient` layer, and the traits-sort collator/key-interning micro-opts (`24aed50`, `a390428`, `cc8709f`, `b1735a1`), plus their supporting commits (`6f7fe59`, `9fa5062`, `80a51e0`, `b1cc3f1`, `03dff21`).

I verified the reverts are clean: grepped the full tree at head for `invalidateEquipmentCatalog`, `globalEquipment`, `getRequestClient`, `traitCollator`, `suffixIds`, `withEquipmentInvalidation` — zero hits, no dangling imports or references. `app/api/admin/equipment/route.ts` is byte-identical to main again (confirmed via diff). The cache-tags.ts doc table and `TAGS` object are consistent post-revert.

## Previous findings

1. **(Medium) PR scope vs. description mismatch** — **Resolved.** The PR body now documents the caching/positioning changes in full and adds the "Five things this PR tried and removed" section, so the description matches what's actually in the diff.

2. **(Low) Comment-only invariant, not enforced (eslint rule suggestion)** — **Partial, in an interesting way.** The suggested rule was added (`6f7fe59`) and then deliberately reverted (`b1cc3f1`) along with the `cache()` layer that motivated it — reasoning given in the commit message (it enforced a constraint this branch invented, and would have wrongly implied only `unstable_cache` needs this treatment when 3 other equivalent modules are already imported by actions). That's a defensible call. But the revert also deleted the doc comment on `app/lib/shared/gang-data.ts` that named the hazard (a server action reading these accessors after writing in the same request gets the pre-write value) — and that hazard is real independent of `cache()`, since it's inherent to `unstable_cache` itself, not something the reverted layer introduced. Right now there are zero violations (confirmed: nothing under `app/actions/**` imports `gang-data.ts`), so this isn't live-broken, but the one-line warning comment is gone along with the lint rule. Low severity, but worth restoring just the comment even without enforcement.

3. **(Low) `NO_WEAPONS` shared mutable singleton** — **Resolved and stayed resolved.** `Object.freeze(NO_WEAPONS)` (`6f7fe59`) survived every later revert; still present at head.

4. **(Low, informational) No regression tests for the grouping rewrite** — **Still outstanding**, as expected — informational only, not a convention violation in a repo with no test framework, no action needed.

## New findings

None. This batch is almost entirely subtractive (reverting prior additions back toward `main`'s behavior), and I didn't find a revert that missed a call site or left an inconsistency. The one item worth a follow-up is #2 above (restore the comment, not the lint rule).

## Edge cases / null handling

Nothing new to check here — no new logic branches were added in this batch, only removals of previously-added logic (the interned-key map, the extra collator, the `cache()` wrapper, the equipment-invalidation wrapper). The surviving `resolvePositioning`/`buildWeaponVariantRows`/`buildVehicleWeapons` logic reviewed last time is untouched by these commits.

## Verdict

**Approve.** The reverts are clean and well-justified, the one still-open item (restore the stale-read comment on `gang-data.ts`) is a low-severity doc nit, not a blocker.

---

PR #2091, opened 2026-09-03
Feature/create gang archetype picker

Djidiouf:
…s with their parent on the user profile.

## Summary

- Hide child gang types (Chymist Cult, Wyld Hunt, Furnace Brutes, Unborn Gang) from the Create Gang combobox and show them as a Gang archetype radio group after the parent is chosen.

## Type of change

- [x] Feature

## How I tested

- [x] Create a House Escher gang with Standard selected — type and image behave as before
- [x] Create a House Goliath gang with Furnace Brutes / Unborn Gang — same as above
- [x] Create a gang with no archetypes (e.g. Cawdor) — no Gang archetype radios

## Database

No new migration in this PR. Relies on ``parent_gang_type_id`` from #2089.

## Risk / rollout

Low — create-gang UI plus profile label. Child types stay hidden until ``is_hidden`` is cleared.

### DIFF
--- a/app/actions/create-gang.ts
+++ b/app/actions/create-gang.ts
@@ -16,7 +16,7 @@ interface CreateGangParams {
   gangOriginId?: string | null;
   credits?: number;
   gangVariants?: string[];
-  defaultGangImage: number;
+  defaultGangImage: number | null;
 }
 
 export async function createGang({
--- a/app/api/gang-types/route.ts
+++ b/app/api/gang-types/route.ts
@@ -29,7 +29,7 @@ export async function GET(request: Request) {
     // Build query - include gang origin data
     let query = supabase
       .from('gang_types')
-      .select('gang_type_id, gang_type, alignment, image_url, default_image_urls, affiliation, gang_origin_category_id, editions:edition_id (slug)')
+      .select('gang_type_id, gang_type, alignment, image_url, default_image_urls, affiliation, gang_origin_category_id, parent_gang_type_id, editions:edition_id (slug)')
       .order('gang_type');
 
     // Only filter out hidden types if user is not admin
@@ -102,7 +102,10 @@ export async function GET(request: Request) {
       }
     }
 
-    // Add affiliations and origins to gang types that require them
+    // Add affiliations and origins to gang types that require them.
+    // Note: parent_gang_type_id (and all other selected columns) are carried
+    // through automatically by the `...gangTypeWithEdition` spread below.
+    // If the .select() query above is changed, verify this field is still included.
     const gangTypesWithAffiliationsAndOrigins = gangTypes.map((gangType) => {
       const availableOrigins = gangType.gang_origin_category_id
         ? (originsByCategory[gangType.gang_origin_category_id] || [])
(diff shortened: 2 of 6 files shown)


joeseos:
@Djidiouf Reviewed against the full codebase (`create-gang-modal.tsx`, `create-gang.ts`, `app/api/gang-types/route.ts`, `utils/gangTypeArchetypes.ts`), including how `parent_gang_type_id` was introduced in #2089 and how `default_gang_image` is used downstream (`update-gang-image.ts`, `schema.public.sql`).

## Checklist

- **Correctness**: `resolvedGangTypeId`/`resolvedGangType` correctly resolve to the root when `gangArchetypeId` is `null` and to the child otherwise, and all dependent reads (affiliation, origin, image gallery, create payload) were consistently switched over from the old `gangType`-based lookups. `gangArchetypeFor`/`hasParentGangType` are edition-scoped correctly via `editionGangTypes`, and custom gang types (explicit `parent_gang_type_id: null`) correctly never surface an archetype picker. No regressions found.
- **Edge cases**: Verified custom types, "no archetypes" gang types (e.g. Cawdor), and edition switching all reset `gangArchetypeId` correctly. The new gallery-index clamp effect for archetype switches only clamps down, never restores the preferred `DEFAULT_IMAGE_INDEX` when switching back to a type with more images — but that's the same pre-existing limitation the original `gangType`-switch clamp already had, not a regression from this PR.
- **Security**: No issues — same auth/select patterns as before, only an additional column added to an existing scoped query.
- **Performance**: Negligible — one extra selected column, no new queries.
- **Readability/maintainability**: Good — `gangTypeArchetypes.ts` is a clean, well-documented extraction, and the route.ts comment about `parent_gang_type_id` flowing through the object spread is a helpful guard against future select-list edits.
- **Conventions**: No `CLAUDE.md` in this repo; checked against `CONTRIBUTING.md`. No schema changes here (correctly relies on the migration from #2089), consistent with the "schema changes go in a migration" rule.
- **DRY/YAGNI**: No duplicated logic; the archetype-resolution logic is centralized well instead of being repeated across the affiliation/origin/image/submit call sites.

## Minor notes (non-blocking, low severity)

1. `components/create-gang-modal.tsx` (create payload) — `defaultGangImage` now sends `null` instead of `currentImageIndex` when the resolved type has no images, vs. always sending the constant index before. This is a genuine improvement (matches how `update-gang-image.ts` already treats `null` as "no default image"), just flagging it as slightly outside the archetype-picker scope described in the PR summary.
2. The archetype radio group (`Standard` + children) doesn't use a `<fieldset>`/`role="radiogroup"`, so it's not announced as a single group to screen readers — consistent with the existing checkbox groups elsewhere in this file, so not a new pattern, just worth knowing.
3. As called out in the PR description itself: since the child gang types are presumably still `is_hidden = true`, non-admin users won't see the archetype picker at all until that flag is cleared on the DB rows — worth double-checking that follow-up happens as part of rollout, since it's easy to merge this and forget the data-side toggle.

## Verdict: **Approve**

No correctness, security, or performance blockers found. The two "How I tested" cases plus the "no archetypes" case cover the meaningful paths well.

---

joeseos (approved):

PR #2092, opened 2026-09-04
Scope Equipment List grants by gang origin, gang variant and fighter subtype

joeseos:
## Summary

- `fighter_type_equipment` has carried `gang_origin_id`, `gang_variant_id` and `fighter_subtype` for a while, but the equipment admin only ever wrote bare `(equipment_id, fighter_type_id)` rows, so a grant could not be limited to gangs of a given origin or variant, or to fighters of a given subtype. Only the vehicle admin exposed the origin axis. No fighter-type row in the database carried any scope.
- "Fighter Types with this Equipment" keeps its one-click dropdown for unscoped grants and gains an **Add Scoped** dialog for the rest. Fighter Type is optional there, so a subtype rule can span every gang — this is what lets "Leaders and Champions of a Corrupted gang may purchase Psychic Familiars" be two rows rather than one per gang's Leader and Champion type.
- Grants now round-trip as `fighter_type_grants` objects rather than a flat id array.
- **Two data-loss bugs fixed in the same paths.** The equipment PATCH deleted on `equipment_id` alone, destroying the vehicle admin's rows on every save (1,023 rows exposed); the client gate meant to prevent it misfired precisely when such rows existed, because the GET returned vehicle rows as null `fighter_type_id`s. The fighter-type PATCH stripped scope columns and re-inserted them unscoped. Both routes now define the rows they own once and apply that to the read and the delete alike, so a screen can only delete what it can see and re-create.
- Also fixed: the "Availability per Gang Variant" dialog returned bare on a duplicate, and `Modal.handleConfirm` closes on anything but an explicit `false` — so it toasted *and* dismissed, discarding the entry.

`excluded` (deny rows) is deliberately out of scope and left for a follow-up.

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- [x] `npx tsc --noEmit` clean; `eslint` clean on all changed files apart from one pre-existing `filteredWeapons` React Compiler error that reproduces identically on `main`. `next build` compiles and typechecks.
- [x] Verified the new SQL predicates against live data rather than by inspection: a Leader fighter matches a `fighter_subtype = 'Leader'` rule and not a Champion one; a call with no fighter (`'[]'`) matches neither; all 17,468 existing rows still pass both new predicates, so no current grant is lost.
- [x] Verified the scope filters select exactly the 16,444 plain fighter grants, leaving the other 1,024 vehicle/orphan rows untouched.
- [x] Confirmed the deployed `get_equipment_detailed_data` did not read `gang_variant_id` or `fighter_subtype` on `fighter_type_equipment` before this change (dumped `pg_get_functiondef` from the live project).
- [ ] **Not run — needs a reviewer with the app running.** No app credentials in my environment, so the end-to-end checks are unverified: (a) the vehicle-row regression — pick equipment a vehicle grants, save it through the equipment admin, confirm the row count holds (on `main` it drops to zero); (b) a matching-origin gang sees the item on the fighter's Equipment List and a different-origin gang does not; (c) the add/save/reopen/remove round trip for all three scopes.

## Database

No migration is needed — the columns already exist from `20260822120000`, which added them and noted that the companion `get_equipment_detailed_data` change "is applied separately". That deploy never happened, so this PR supplies it. Following the convention of `20260723120000` and `20260822120000`, the function body lives in `supabase/functions/` only and is not duplicated into a migration.

The RPC needed three changes for subtype, not just a predicate:
1. A rule spanning every gang carries no `fighter_type_id`, and the join's identity branch required one, so such a row could never match. That branch now also accepts a row naming no fighter, vehicle or custom type — provided it names a subtype, without which the single all-NULL row in the table would match every fighter.
2. `gang_data` now carries the fighter's subtypes, coalesced to an empty array so a call with no fighter matches no subtype rule.
3. `is_fighter_list` keyed off `fighter_type_id`/`vehicle_type_id`, neither of which a gang-wide row has, so it now keys off `fte.id`.

### Migration status

**Apply `supabase/functions/get_equipment_detailed_data.sql` from the dashboard before merging.** Origin scoping works on merge because its predicate is already live; variant- and subtype-scoped grants will save correctly but have no effect until the function is applied.

## Risk / rollout

- **Medium.** The scope-ownership fix touches a path every equipment save goes through, and the RPC change touches the query behind every equipment list. Mitigating: the new predicates were verified against all 17,468 live rows without excluding any, `SELECT DISTINCT` absorbs the extra join fan-out from variant being multi-valued per gang, and no fighter-type row currently carries any scope, so nothing existing changes behaviour.
- Worth knowing for the `excluded` follow-up: dropping the broken `hasEditedFighterTypes` gate means the grant rewrite now runs on *every* save rather than almost never. That is why the read/delete/write scope had to be made symmetric here — the follow-up would otherwise have been the first thing to populate those columns and the first to silently destroy them.

### DIFF
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from 'next/server'
 import { createClient } from "@/utils/supabase/server";
 import { checkAdmin } from "@/utils/auth";
-import { WeaponProfileInput, EquipmentAvailability, EquipmentOriginAvailability, EquipmentVariantAvailability, GangAdjustedCost, GangOriginAdjustedCost } from "@/types/equipment";
+import { WeaponProfileInput, EquipmentAvailability, EquipmentOriginAvailability, EquipmentVariantAvailability, FighterTypeEquipmentGrant, GangAdjustedCost, GangOriginAdjustedCost } from "@/types/equipment";
 import {
   FighterEffectType,
   FighterEffectTypeModifier,
@@ -344,11 +344,15 @@ export async function GET(request: Request) {
           console.warn('Error fetching fighter types:', fighterTypesError);
         }
 
-        // Fetch fighter types that have this equipment
+        // Fetch fighter types that have this equipment. Vehicle rows belong to the
+        // vehicle admin and carry a null fighter_type_id, so they are filtered out
+        // rather than surfacing here as blank entries.
         const { data: equipmentFighterTypes, error: equipmentFighterTypesError } = await supabase
           .from('fighter_type_equipment')
-          .select('fighter_type_id')
-          .eq('equipment_id', id);
+          .select('fighter_type_id, gang_origin_id, gang_variant_id')
+          .eq('equipment_id', id)
+          .is('vehicle_type_id', null)
+          .not('fighter_type_id', 'is', null);
 
         if (equipmentFighterTypesError) {
           console.warn('Error fetching fighter types with equipment:', equipmentFighterTypesError);
@@ -617,7 +621,7 @@ export async function PATCH(request: Request) {
       is_editable,
       is_consumable,
       weapon_profiles,
-      fighter_types,
+      fighter_type_grants,
       gang_adjusted_costs,
       gang_origin_adjusted_costs,
       equipment_availabilities,
@@ -693,51 +697,42 @@ export async function PATCH(request: Request) {
       }
     }
 
-    // More robust fighter type association handling
-    if (fighter_types !== undefined) {
-      
-      // First, get current associations to ensure we don't lose data
-      const { data: currentAssociations, error: fetchError } = await supabase
+    // Handle fighter type associations
+    if (fighter_type_grants !== undefined) {
+      // Only the rows this screen owns: vehicle rows belong to the vehicle admin.
+      const { error: deleteError } = await supabase
         .from('fighter_type_equipment')
-        .select('fighter_type_id')
-        .eq('equipment_id', id);
-      
-      if (fetchError) {
-        console.error('Error fetching current fighter type associations:', fetchError);
-        // Continue with the operation even if this check fails
-      }
-      
-      // Only proceed with deleting & updating if:
-      // 1. We successfully fetched the current associations
-      // 2. The new list is different from the current list
-      if (currentAssociations) {
-        const currentIds = currentAssociations.map(a => a.fighter_type_id);
-        const hasChanges = JSON.stringify(currentIds.sort()) !== JSON.stringify([...fighter_types].sort());
-        
-        if (hasChanges) {
-          // Delete existing associations
-          const { error: deleteError } = await supabase
-            .from('fighter_type_equipment')
-            .delete()
-            .eq('equipment_id', id);
-          
-          if (deleteError) throw deleteError;
-          
-          // Insert new associations if there are any
-          if (fighter_types.length > 0) {
-            const { error: insertError } = await supabase
-              .from('fighter_type_equipment')
-              .insert(
-                fighter_types.map((fighter_type_id: string) => ({
-                  fighter_type_id,
-                  equipment_id: id,
-                  updated_at: new Date().toISOString()
-                }))
-              );
-            
-            if (insertError) throw insertError;
-          }
-        }
+        .delete()
+        .eq('equipment_id', id)
+        .is('vehicle_type_id', null)
+        .not('fighter_type_id', 'is', null);
+
+      if (deleteError) throw deleteError;
+
+      if (Array.isArray(fighter_type_grants) && fighter_type_grants.length > 0) {
+        // fighter_type_equipment_fighter_scope_uidx is NULLS NOT DISTINCT, so a
+        // repeated scope is a unique violation rather than a no-op.
+        const seen = new Set<string>();
+        const grantRecords = (fighter_type_grants as FighterTypeEquipmentGrant[])
+          .map(grant => ({
+            fighter_type_id: grant.fighter_type_id,
+            equipment_id: id,
+            gang_origin_id: grant.gang_origin_id ?? null,
+            gang_variant_id: grant.gang_variant_id ?? null,
+            updated_at: new Date().toISOString()
+          }))
+          .filter(record => {
+            const key = `${record.fighter_type_id}|${record.gang_origin_id ?? ''}|${record.gang_variant_id ?? ''}`;
+            if (seen.has(key)) return false;
+            seen.add(key);
+            return true;
+          });
+
+        const { error: insertError } = await supabase
+          .from('fighter_type_equipment')
+          .insert(grantRecords);
+
+        if (insertError) throw insertError;
       }
     }
 
--- a/app/api/admin/equipment/route.ts
+++ b/app/api/admin/equipment/route.ts
@@ -15,6 +15,20 @@ interface FighterTypeEquipment {
   equipment_id: string;
 }
 
+/**
+ * The fighter_type_equipment rows this screen owns: plain fighter-type grants,
+ * scoped at most by gang origin and variant. Vehicle rows belong to the vehicle
+ * admin, and gang-type / subtype / deny rows to screens that can express them —
+ * none may be read here, because the save deletes everything it reads.
+ * Applied to the read and the delete alike so the two cannot drift apart.
+ */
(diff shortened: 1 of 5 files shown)


joeseos:
@joeseos Reviewed the full diff (route changes, the RPC, the admin component, and the type). Traced the SQL predicate logic by hand against the schema and the referenced migrations (`20260822120000` for the new columns/unique index, `20260806120000` to confirm `fighter_subtypes` is a jsonb array of name strings so the `?` operator usage is correct) rather than just reading the diff in isolation.

## Findings

**1. (Medium) `core_equipment` gating in the RPC wasn't updated for the new gang-wide-subtype branch** — `supabase/functions/get_equipment_detailed_data.sql`, line 468:
```sql
OR (e.core_equipment = true AND (fte.fighter_type_id IS NOT NULL OR cftl.is_ftl IS NOT NULL OR $3 IS NULL))
```
This still checks `fte.fighter_type_id IS NOT NULL`. A new gang-wide-subtype grant (branch 5 of the `fte` join, `fighter_type_id`/`vehicle_type_id`/`custom_fighter_type_id` all NULL, only `fighter_subtype` set) has `fte.fighter_type_id` NULL even when it successfully matches — that's exactly why the `is_fighter_list` LATERAL a few lines below was changed from `fte.fighter_type_id IS NOT NULL OR fte.vehicle_type_id IS NOT NULL` to `fte.id IS NOT NULL` (per your own PR description, point 3). This second, near-identical check wasn't given the same fix, so if `core_equipment = true` is ever combined with a gang-wide-subtype-only grant, the item will pass the `fte` join and `is_fighter_list` would say `true`, but this WHERE-clause gate will still drop the row before it reaches the caller — the fighter's equipment list silently won't show it. It also contradicts the comment immediately above `ftl_flag` ("Single source of truth... so the predicate lives in exactly one place") — this is a second place. Real-world blast radius is probably narrow (`core_equipment` items are normally fixed starting gear tied to one fighter type directly, not the kind of thing this new broad-scoping feature targets), but it's a genuine latent bug in the feature's core logic, not a hypothetical. Suggest either reusing `ftl_flag.is_fighter_list` here too if that's semantically equivalent for this gate, or adding the same `fte.id IS NOT NULL AND fte.fighter_type_id IS NULL AND fte.vehicle_type_id IS NULL` branch explicitly.

**2. (Low, disclosed risk) Delete-then-insert isn't transactional, and now runs on every save** — `app/api/admin/equipment/route.ts`, the `fighter_type_grants` block (~714-750). This was already true before this PR (two sequential PostgREST calls, no explicit transaction), but removing the broken `hasEditedFighterTypes` gate means it now fires on *every* equipment save instead of almost never (as your own "Risk / rollout" section notes). If the `insert` fails after the `delete` succeeds — e.g. a scoped grant referencing a gang variant/origin that got deleted between opening the modal and saving, tripping the FK — the equipment item loses all its fighter-type grants with no rollback, on what could be a routine unrelated edit (e.g. just changing the cost). Given this PR's whole premise is fixing exactly this class of data-loss bug, it might be worth closing this window too — e.g. move the delete+insert into a small RPC/transaction — though I'd understand deferring it since it's a pre-existing pattern, not a new regression.

**3. (Low) `filteredFighterSubtypes` sorts with one shared `editionSlug` rather than per-item edition resolution** — `components/admin/admin-edit-equipment.tsx`, ~line 590. `filteredFighterTypes`'s sort resolves each item's own edition slug via `editionSlugOf(editions, a.edition_id)` for the case where `editionId` is unset and the list spans multiple editions; the new subtype sort just uses one fixed `editionSlug` for every row in that same fallback case. Cosmetic (dropdown ordering only, not a data bug) — not blocking.

## What I checked and found solid
- **Symmetric ownership fix**: traced all three routes (equipment admin, fighter-types admin, vehicle admin at `app/api/admin/vehicles/route.ts:391`, unchanged by this PR) to confirm their read/delete scopes are now mutually exclusive by construction — vehicle admin already deletes scoped to `vehicle_type_id`, so it can't step on the new fighter-type-scoped rows, and the new `scopeToFighterTypeGrants`/`scopeToUnscopedGrants` helpers correctly make each screen's delete match exactly what it read. No new cross-screen data-loss vector.
- **SQL join correctness**: parenthesization of the 5-branch identity OR is correctly grouped; the `gd.gang_variants ? fte.gang_variant_id::text` and `gd.fighter_subtypes ? fte.fighter_subtype` jsonb-array-containment checks match the exact pattern already used elsewhere in this same function for `equipment_availability`; traced through the "call with no fighter" case by hand and confirmed `COALESCE(f.fighter_subtypes, '[]'::jsonb)` correctly makes gang-wide subtype rows match nobody rather than everybody, as claimed.
- **`SELECT DISTINCT` fan-out**: confirmed no raw `fte.*` column is projected in the outer SELECT (only the derived `ftl_flag.is_fighter_list` boolean), so the extra join fan-out from a fighter potentially matching more than one `fte` row collapses correctly.
- **`Modal.handleConfirm` fix**: confirmed against the actual component (`if (result !== false) onClose()`) — the old bare `return;` on the duplicate-variant-availability path did fall through to close the modal despite the toast; `return false` fixes it, and the new "Add Scoped" dialog's own duplicate check correctly returns `false` too (doesn't reintroduce the same bug).
- **New dialog's `confirmDisabled` guard**: worked through the truth table — it correctly disables submitting a fighter-type-only selection (redundant with the plain dropdown) and, more importantly, disables the combination that would create an identity-less row that can never match any fighter (no fighter type, no subtype) — proactively preventing a second "all-NULL row" landmine like the one the SQL comment already warns about.

## Verdict
**Changes requested** — mainly #1, since it's a genuine correctness gap in the feature this PR adds, even though the practical combination that triggers it is narrow. #2 and #3 are worth a look but don't need to block.

---

joeseos:
Pushed `d6ff669`.

**#1 — fixed, and you're right that it's a real gap.** I'd actually considered the `core_equipment` gate while writing the subtype branch and left it alone on the reasoning that "exclusive to a single fighter" and a broad subtype rule are contradictory by intent. That was the wrong call: whatever the policy should be, having two near-identical checks disagree means the row passes the join, gets flagged `is_fighter_list = true`, and is then dropped by the `WHERE` gate — it disappears with no way to tell why from the admin side.

I took your second suggestion rather than the first, deliberately:

```sql
OR (e.core_equipment = true AND (
    fte.fighter_type_id IS NOT NULL
    OR (fte.id IS NOT NULL AND fte.fighter_type_id IS NULL AND fte.vehicle_type_id IS NULL)
    OR cftl.is_ftl IS NOT NULL
    OR $3 IS NULL
))
```

Reusing `ftl_flag.is_fighter_list` here isn't semantically equivalent — it also covers `ea_var` and `ea_origin`, so core equipment would newly pass for any gang holding a matching origin/variant availability row. A bare `fte.id IS NOT NULL` has the same problem in the other direction: the gate currently excludes `vehicle_type_id` matches, and that relaxation would be a behaviour change this PR has no reason to make. The explicit branch admits only the case the PR introduces.

Traced the four cases to confirm nothing else moves: `fighter_type_id` set → first branch, unchanged; vehicle row → still excluded, since the new branch requires `vehicle_type_id IS NULL`; the single all-NULL row → never matches the join at all, because that branch requires `fighter_subtype IS NOT NULL`; gang-wide subtype row → newly admitted, which is the fix.

Fair point that this contradicts the "single source of truth" comment above `ftl_flag`. It's still two places, because the two gates genuinely aren't the same predicate — worth a follow-up to reconcile, but not by widening one of them here.

**#3 — fixed** in the same commit. The subtype sort now resolves each row's own edition via `editionSlugOf(editions, a.edition_id)`, matching `filteredFighterTypes` directly above it.

**#2 — deferring, agreed as non-blocking.** You've read it correctly: pre-existing pattern, newly hot because dropping `hasEditedFighterTypes` means the rewrite runs on every save. Every sibling block in this handler has the same shape (availabilities, discounts, weapon profiles), so fixing it properly means moving the whole save into an RPC — a structural change to code this PR doesn't otherwise touch, and one that would be better done for all of them at once than for this block alone. Happy to raise it as its own issue.

Still outstanding, unchanged from the description: `supabase/functions/get_equipment_detailed_data.sql` needs applying from the dashboard before merge, and the three end-to-end checks in "How I tested" are unrun here for want of app credentials.

Djidiouf (approved):

PR #2093, opened 2026-09-04
fix(admin): edition-gate the Gang Origin and Gang Variant pickers

joeseos:
## Summary

- **The bug:** Edit Equipment already treats edition as its top-level filter, but the Gang Origin and Gang Variant selects rendered straight off the raw query results, so they listed every edition's rows at once. Not cosmetic — three variant names (Chaos Corrupted, Genestealer Infected, Malstrain Corrupted) exist in **both** n23 and n26 under different ids, and the duplicate entries are indistinguishable in the list, so an admin editing an n26 item could silently attach an n23 scope id and produce a grant that can never match.
- All four origin/variant dropdowns now offer only the selected edition's rows: the Scoped Equipment List Entry modal, plus Cost per Gang Origin and the two Availability dialogs.
- Both filters compare `edition_id`, matching the other 73 edition comparisons across `components/admin/`. `/api/admin/gang-origins` now returns `edition_id` (it exposed nothing edition-related before); `/api/gang-variant-types` returns it alongside the `edition_slug` it already flattened, so the three player-side callers keep the slug — they submit only `v.id`, never the rows, so the extra field is inert for them.
- The grant-chip labels stay **unfiltered on purpose**: selecting an equipment sets `editionId` directly with no prune, so an already-saved cross-edition scope is legitimately present on load and has to keep rendering its name to stay removable. Same selectable-vs-listed split as `admin-trading-post.tsx`.
- `handleEditionChange` now prunes what it was missing — origin- and variant-scoped adjusted costs and availabilities, and the origin/variant half of a fighter-type grant. A scope absent from the catalog is kept, never dropped: both queries are `enabled`-gated, so an empty list means "still fetching", not "foreign". A grant's `fighter_subtype` is deliberately still not judged — it is stored by name, and the same name is legitimately defined in more than one edition.
- `handleEditionChange` moved below the catalog queries it reads. Reading `gangOriginList` from above its declaration made React Compiler give up on preserving the manual memoization; the move clears that **and** the pre-existing bail-out on `filteredWeapons`, taking repo lint errors from 7 to 5.
- Under n26 the Gang Origin picker is legitimately empty — there are no n26 rows in `gang_origins` yet. Left as an empty picker rather than hidden, and no `EDITION_CAPABILITIES` row added: that is a data-authoring fact, not a rules statement, so it starts working by itself once n26 origins are authored.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] `npx tsc --noEmit` clean; `npm run lint` clean on all three changed files (repo total down from 7 errors to 5)
- [x] `npm run build` compiles and passes its TypeScript pass
- [x] Confirmed against the live DB that the duplicate-name variants are real (9 n23 + 3 n26 rows, three names shared) and that all 27 `gang_origins` rows are n23
- [x] Confirmed the two API changes are additive: re-read all four other `/api/admin/gang-origins` consumers and all three other `/api/gang-variant-types` consumers — each declares its own structural query generic and none re-posts the rows
- [ ] **Not done — no Supabase credentials in my environment, so I could not run the app.** Worth a manual pass before merge: switch edition n23→n26 in Edit Equipment and confirm Gang Variant lists only the 3 n26 rows; confirm a saved n23-scoped grant chip still shows its name and its ✕ still removes it; confirm switching edition prunes the origin/variant rows it should

## Risk / rollout

- **Low.** No env vars, no feature flags, no migration. Both API changes only add a field to a GET response.
- No behaviour change for player-side code: `/api/gang-variant-types` keeps returning `edition_slug`, and its three other callers are untouched.
- The one intentional behaviour change beyond the fix: switching edition now drops already-picked origin/variant rows that belong to another edition, consistent with what the handler already did for fighter types, trading posts and gang types.

### DIFF
--- a/app/api/admin/gang-origins/route.ts
+++ b/app/api/admin/gang-origins/route.ts
@@ -30,7 +30,6 @@ export async function GET(request: Request) {
     const transformedData = origins.map((origin: any) => ({
       id: origin.id,
       origin_name: origin.origin_name,
-      // Origins are edition-scoped; clients filter their pickers on this
       edition_id: origin.edition_id ?? null,
       category_id: origin.gang_origin_category_id,
       category_name: (origin.gang_origin_categories as any)?.category_name || 'Unknown'
--- a/app/api/admin/gang-origins/route.ts
+++ b/app/api/admin/gang-origins/route.ts
@@ -16,6 +16,7 @@ export async function GET(request: Request) {
       .select(`
         id,
         origin_name,
+        edition_id,
         gang_origin_category_id,
         gang_origin_categories!gang_origin_category_id (
           id,
@@ -29,6 +30,8 @@ export async function GET(request: Request) {
     const transformedData = origins.map((origin: any) => ({
       id: origin.id,
       origin_name: origin.origin_name,
+      // Origins are edition-scoped; clients filter their pickers on this
+      edition_id: origin.edition_id ?? null,
       category_id: origin.gang_origin_category_id,
       category_name: (origin.gang_origin_categories as any)?.category_name || 'Unknown'
     }));
--- a/app/api/gang-variant-types/route.ts
+++ b/app/api/gang-variant-types/route.ts
@@ -6,6 +6,7 @@ import { editionSlugFromJoin } from "@/types/edition";
 interface Variant {
     id: string;
     variant: string;
+    edition_id?: string | null;
     editions?: { slug: string } | { slug: string }[] | null;
 }
 
@@ -21,7 +22,7 @@ export async function GET(request: Request) {
 
         let query = supabase
             .from('gang_variant_types')
-            .select('id, variant, editions:edition_id (slug)')
+            .select('id, variant, edition_id, editions:edition_id (slug)')
             .order('variant')
 
         const { data, error } = await query;
@@ -31,6 +32,9 @@ export async function GET(request: Request) {
         const modelData = data.map((variant: Variant) => ({
             id: variant.id,
             variant: variant.variant,
+            // Slug for the player-side callers, which reason in slugs; id for the
+            // admin catalog editor, which filters and stores an edition_id
+            edition_id: variant.edition_id ?? null,
             edition_slug: editionSlugFromJoin(variant.editions),
         }));
 
--- a/app/api/gang-variant-types/route.ts
+++ b/app/api/gang-variant-types/route.ts
@@ -32,8 +32,7 @@ export async function GET(request: Request) {
         const modelData = data.map((variant: Variant) => ({
             id: variant.id,
             variant: variant.variant,
-            // Slug for the player-side callers, which reason in slugs; id for the
-            // admin catalog editor, which filters and stores an edition_id
+            // Slug for the player-side callers; id for the admin editor, which saves one
             edition_id: variant.edition_id ?? null,
             edition_slug: editionSlugFromJoin(variant.editions),
         }));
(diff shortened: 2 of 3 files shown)


joeseos:
@joeseos Reviewed against the current `main` (which already has #2092 merged, including the `core_equipment` gate fix — good, that landed). Traced the new `handleEditionChange` logic by hand and cross-checked both API changes against every other consumer, not just the ones this PR touches directly.

## What I verified

- **The three cross-edition-duplicate names are real and the fix is correctly scoped**: confirmed `origin.edition_id`/`variant.edition_id` flow straight through from Supabase as plain strings in both routes, and `filteredGangOrigins`/`filteredGangVariants` correctly fall back to the unfiltered list when `editionId` is empty (new-equipment / no-selection state) rather than showing nothing.
- **Additive API changes don't break other consumers**: read all 4 other `/api/admin/gang-origins` callers (`admin-create-skill`, `admin-edit-vehicle-type`, `admin-create-vehicle-type`, `admin-edit-skill`) and all 3 other `/api/gang-variant-types` callers (`create-gang-modal`, `gang.tsx`, `custom-trading-posts`) — none do strict/shape-validated parsing, and `custom-trading-posts.tsx` only reads `.edition_slug` off its own narrower cast, so the new `edition_id` field is genuinely inert for all of them, as claimed.
- **The pruning logic is conservative in the right direction**: `handleEditionChange`'s new origin/variant checks (`if (origin && origin.edition_id !== newEditionId) return false`) only drop a grant/cost/availability when the id was *found* and belongs to another edition — an id not yet resolvable (list still loading, since both queries are `enabled`-gated on `selectedEquipmentId`/dialog-open) or genuinely absent from the catalog is kept, matching the "still fetching, not foreign" reasoning in the PR body. I checked this against every state collection in the file that carries a `gang_origin_id`/`gang_variant_id` (`fighterTypeGrants`, `gangOriginAdjustedCosts`, `equipmentOriginAvailabilities`, `equipmentVariantAvailabilities`) and didn't find one that was missed.
- **Unfiltered chip labels are intentional and correctly limited to just the label lookups** (lines showing `gangOriginList.find(...)`/`gangVariantList.find(...)` for grant chips) — every other read site (the four dropdown renders) does go through `filteredGangOrigins`/`filteredGangVariants`. Confirmed by grepping every use of both lists in the file.
- All referenced state setters (`setGangOriginAdjustedCosts`, `setEquipmentOriginAvailabilities`, `setEquipmentVariantAvailabilities`, `setScopedGrant*`, etc.) exist with matching names — no stale/renamed-variable typos from the `handleEditionChange` move.

## Finding

**(Low) The new "nothing to add" disabled-guard was applied to one of three analogous buttons, not all three.** `components/admin/admin-edit-equipment.tsx` — the "Cost per Gang Origin" → "Add Origin" button gets:
```jsx
disabled={gangOriginList.length > 0 && filteredGangOrigins.length === 0}
```
but the structurally identical "Availability per Gang Origin" → "Add Origin" and "Availability per Gang Variant" → "Add Variant" buttons (a few dozen lines below/after) don't get the same guard. Practically: under n26 today (no n26 `gang_origins` rows), an admin can still open the Origin/Variant Availability dialogs to a dropdown with nothing but the placeholder option, where the Cost dialog correctly disables itself first. Not a correctness or data-integrity issue — worst case is a moment of "why is this empty" — just an inconsistency worth a one-line follow-up for the two other buttons. (The "Add Scoped" fighter-type-grant button is correctly left alone here — origin/variant are optional there, so an empty catalog doesn't make that dialog useless the way it does for the two Availability dialogs.)

## Verdict
**Approve.** The core fix is correct and more thoroughly cross-checked than the PR's own "not done" list suggests was possible without app access — I'd trust the reasoning here. The disabled-guard gap is minor enough not to block; happy to see it in this PR or a fast follow-up.

---

PR #2094, opened 2026-09-04
Promote Prospects to Gangers at 13XP+ should count advancements correctly

v0rp-z:
## Summary

The advancement count was over-counting by 1 for Prospects. 

The 13-XP tier is the (RAW: mandatory) Prospect to Ganger(+Specialist) promotion, not a normal advancement, but we want to make sure that users can add and advancement, or promote, if they choose to promote, it should consume an advancement.

**From ticket 1264:** A Prospect at 77 XP was showing 11 available advancements when they should only have 10.
## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] tested locally

https://github.com/user-attachments/assets/9af42348-cfec-4040-8411-7bd7c4621907

**Confirm Gangers are unaffected**

https://github.com/user-attachments/assets/78bcce76-9f71-405c-a333-f9e4af03211d

## Risk / rollout

- Low

### DIFF
--- a/app/lib/fighter-advancements.ts
+++ b/app/lib/fighter-advancements.ts
@@ -40,6 +40,7 @@ export const getGangFighters = async (gangId: string, supabase: any) => {
     id: f.id,
     fighter_name: f.fighter_name,
     fighter_type: f.fighter_type,
+    fighter_subtypes: Array.isArray(f.fighter_subtypes) ? f.fighter_subtypes : [],
     fighter_type_subtypes: Array.isArray(f.fighter_types?.fighter_subtypes)
       ? f.fighter_types.fighter_subtypes
       : [],
--- a/app/lib/fighter-advancements.ts
+++ b/app/lib/fighter-advancements.ts
@@ -40,6 +40,9 @@ export const getGangFighters = async (gangId: string, supabase: any) => {
     id: f.id,
     fighter_name: f.fighter_name,
     fighter_type: f.fighter_type,
+    fighter_type_subtypes: Array.isArray(f.fighter_types?.fighter_subtypes)
+      ? f.fighter_types.fighter_subtypes
+      : [],
     xp: f.xp,
     starting_xp: f.starting_xp ?? null,
     advancements_taken: countAdvancementsTaken(
--- a/app/lib/shared/gang-assembly.ts
+++ b/app/lib/shared/gang-assembly.ts
@@ -677,6 +677,9 @@ export function assembleGangFighters(
           label: fighter.label,
           fighter_type: fighter.fighter_type || fighterTypeInfo.fighter_type || 'Unknown',
           fighter_subtypes: fighter.fighter_subtypes || [],
+          fighter_type_subtypes: Array.isArray(fighterTypeInfo.fighter_subtypes)
+            ? fighterTypeInfo.fighter_subtypes
+            : [],
           fighter_specialisation: fighterSpecialisationInfo ? {
             fighter_specialisation: fighterSpecialisationInfo.specialisation_name,
             fighter_specialisation_id: fighterSpecialisationInfo.id
--- a/app/lib/shared/gang-data.ts
+++ b/app/lib/shared/gang-data.ts
@@ -137,6 +137,12 @@ export interface GangFighter {
   label?: string;
   fighter_type: string;
   fighter_subtypes: string[];
+  /**
+   * Catalog subtypes on the fighter_type row. Kept-type promotion
+   * (Prospect→Ganger/Specialist) leaves this on the origin type while live
+   * `fighter_subtypes` flip; it is the stable signal for "was recruited as X".
+   */
+  fighter_type_subtypes?: string[];
   fighter_specialisation?: {
     fighter_specialisation: string;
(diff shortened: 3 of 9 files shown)


joeseos:
Reviewed the diff against the full data-flow — traced `fighter_type_subtypes` from each of the three query/assembly functions (`getGangFighters`, `assembleGangFighters`, the pre-existing `fighterData.fighter.fighter_type.fighter_subtypes` path) through every consumer of `openAdvancementsFor` (`fighter-card.tsx`, `fighter-page.tsx`, `fighter-advancement-list.tsx` — confirmed these are the only 3 call sites, all updated) down to the actual render (`hasAvailableAdvancements={openAdvancements > 0}` in `fighter-card.tsx:706`).

## Correctness check
Verified the fix against the ticket's own repro numbers. N26 tier starts ≤77 XP: `4,7,10,13,19,25,31,37,49,61,73` = 11 tiers, so `advancementsEarnedFor` returns 11 for a 77XP Prospect with 0 starting XP. Before this PR, `openAdvancementsFor` = 11 − 0 = **11** (the reported bug). After: `hasN26ProspectPromotionOccurred` detects catalog-still-Prospect + live-subtypes-no-longer-Prospect, so `promotionCost` = 1, giving 11 − 0 − 1 = **10**. Matches the ticket exactly.

Edition gating (`hasProspectSpecialisationPromotion`: n26 `true`, n23 `false`) means N23 is untouched — matches the "Confirm Gangers are unaffected" test video. Demotion (`buildN26ProspectDemotionSubtypes` restores `'Prospect'` to live subtypes) correctly flips `hasN26ProspectPromotionOccurred` back to `false`, so undo isn't left double-counted.

## Findings

**(Low, informational — no change requested)** `components/fighter/fighter-page.tsx` now carries two different shapes for the same "catalog subtypes" concept side by side: the new flat `f.fighter_type_subtypes` (used for the `gangFighters` sibling-roster icons, line ~656) vs. the pre-existing nested `fighterData.fighter?.fighter_type?.fighter_subtypes` (used for the currently-viewed fighter's `fighterCatalogSubtypes` prop, line ~958). Both are correctly wired to real data — I checked each one back to its query (`fighter_types!fighter_type_id` join in `getGangFightersBundle`, and `fighterTypeData.fighter_subtypes` in `assembleFighterView`) — so this isn't a bug, just a naming divergence that could trip up whoever touches this file next.

**(Low, question, non-blocking)** `utils/keepTypePromotionN26.ts` documents Ganger→Champion and Champion→Leader as keep-type promotions too. Worth a quick confirm from the author: do those also RAW-trade an Advancement roll the way Prospect→Ganger does? If so it'd be a follow-up in the same spirit as this fix; if not (i.e. those are just rank-title thresholds with no advancement cost), no action needed. Not blocking since the PR is explicitly scoped to the Prospect case from ticket 1264.

## Other checks
- No security, null-handling, or performance concerns — all new lookups are `Array.isArray`/`?? []`-guarded and O(1)/O(n) over small arrays.
- No orphaned/duplicated logic; the new `AdvancementCountOptions` param is additive and optional, so every pre-existing caller still compiles/behaves the same by default.
- Docstrings on the new fields/functions are appropriately non-obvious-WHY (catalog-vs-live subtype distinction), not restating the what.

## Verdict: **Approve**

@v0rp-z

---

joeseos:
Amending my earlier "Low, informational" note about `fighter_type_subtypes` (flat) vs `fighter_type.fighter_subtypes` (nested) — on reflection this is worth a concrete follow-up rather than just an FYI.

**Why the split happened:** `FighterProps`/`GangFighter` (the types `FighterCard` consumes) already declare `fighter_type: string` — it's rendered directly as the type name on every card (`type={fighter.fighter_type}` in `sortable-fighter.tsx`). That key was already committed to being a plain string everywhere before this PR, so when this PR needed to expose "catalog subtypes" on that same shared type, reusing/nesting under `fighter_type` wasn't available without breaking every existing string consumer — hence the new, separately-named flat field `fighter_type_subtypes`. No new data is fetched (the underlying `fighter_types` join already carried this for every fighter); it's a naming split, not duplicated work.

**Suggested follow-up PR:** unify the shape so "fighter type" carries its subtypes with it everywhere, e.g.
```ts
fighter_type: { name: string; subtypes: string[] };
```
(or similar) across `FighterProps` / `GangFighter`, replacing the separate `fighter_type` (string) + `fighter_type_subtypes` (array) pair with one field — and reconciling that with the page-local nested shape `fighter-page.tsx` already builds in `transformFighterData` for the single viewed fighter.

Flagging as a separate PR rather than part of this one because the blast radius is real: `fighter_type` as a bare string is read directly (`fighter.fighter_type`, `type={fighter.fighter_type}`, etc.) across `fighter-card.tsx`, `sortable-fighter.tsx`, `print-gang.tsx`, `participant-card.tsx`, and likely more — every one of those call sites would need updating in the same change. Not something to bundle into a bug-fix PR, but worth tracking so the two representations don't keep drifting further apart as more consumers get added.

@v0rp-z

---

PR #2095, opened 2026-09-05
Feat/n26 chem alchemy equipment categories

Djidiouf:
Rank the N26 Chem-alchemy categories under their own super-category and strip the parenthetical suffix from display names. Align user-facing Chem-alchemy spelling.

## Summary
- Rank N26 Gaseous Ammo, Poisons, and Stimms under a Chem-alchemy super-category in the equipment modal
- Strip ``(Chem-alchemy)`` from category display names, matching the existing weapon-suffix behaviour
- Align user-facing Chem-alchemy spelling in the stash button, user guide, and related copy

## Test plan
- [x] On an N26 gang, open the trading post / equipment modal and confirm Chem-alchemy categories group under a Chem-alchemy header
- [x] Confirm those category labels show as Gaseous Ammo, Poisons, and Stimms (no parenthetical suffix)
- [x] Confirm N23 Chem-alchemy copy (stash button, user guide, landing page) uses Chem-alchemy

### DIFF
--- a/app/actions/chem-alchemy.ts
+++ b/app/actions/chem-alchemy.ts
@@ -57,7 +57,7 @@ export async function createChemAlchemy({
 
     const gangEdition = gangEditionJoin(gangData);
     if (!hasChemAlchemy(gangEdition?.slug)) {
-      throw new Error('Chem-Alchemy is not available for this gang edition');
+      throw new Error('Chem-alchemy is not available for this gang edition');
     }
 
     if (gangData.credits < totalCost) {
--- a/app/user-guide/page.tsx
+++ b/app/user-guide/page.tsx
@@ -8,7 +8,7 @@ const defaultUrl = process.env.NODE_ENV === 'development'
 
 // SEO constants - edit these to update all metadata
 const PAGE_TITLE = 'User Guide - How to Use Munda Manager';
-const PAGE_DESCRIPTION = 'Complete user guide for Munda Manager. Learn how to create gangs, manage fighters, run campaigns, use custom assets, and explore advanced features like Chem-Alchemy and Gene-Smithing for Necromunda.';
+const PAGE_DESCRIPTION = 'Complete user guide for Munda Manager. Learn how to create gangs, manage fighters, run campaigns, use custom assets, and explore advanced features like Chem-alchemy and Gene-Smithing for Necromunda.';
 const PAGE_DESCRIPTION_SHORT = 'Complete user guide for Munda Manager. Learn how to create gangs, manage fighters, run campaigns, and explore advanced features.';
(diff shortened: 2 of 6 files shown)


joeseos:
Reviewed the diff against the surrounding code (equipment category rendering in `components/equipment/equipment.tsx`, the N23 rank table, and everywhere `Chem-Alchemy`/`Chem-alchemy` strings live). Small, well-scoped PR — one consistency gap worth a look, nothing blocking.

## Findings

**1. (Medium) Spelling rename missed the `equipment_category` value written for custom chems, so newly-added stash items will still show the old casing**
- `app/actions/chem-alchemy.ts:74`
- `components/gang/stash-tab.tsx:159` and `:185`

These three sites still write `equipment_category: 'Chem-Alchemy'` (capital A). That value is rendered verbatim in the stash list's Category column (`components/gang/stash-tab.tsx:872`: `item.equipment_category || 'Equipment'`). After this PR, a user who creates a new custom chem will see the stash button read "Chem-alchemy" but the item's own Category cell in the same table read "Chem-Alchemy" — the exact inconsistency this PR sets out to fix, just one layer deeper (data instead of static copy). Test plan didn't cover this path since it only checks copy in the stash button/user guide/landing page, not the item list itself.

Suggested fix: update the three `equipment_category: 'Chem-Alchemy'` literals to `'Chem-alchemy'` for consistency. (No other code in the repo does an exact-string match against `'Chem-Alchemy'`, so this is safe to change — it's write-then-display-only, not used as a lookup key anywhere else.)

**2. (Low, non-blocking) Chained `.replace()` calls in `getEquipmentCategoryDisplayNameN26`**
`utils/equipmentCategoryRankN26.ts:124-129` — three sequential `.replace()` calls each stripping a different parenthetical suffix. Functionally fine since the suffixes are mutually exclusive, and it follows the pre-existing pattern for the two weapon-type suffixes, so not asking for a change here — just flagging in case a future 4th super-category suffix makes a single alternation regex (`/ \((Chem-alchemy|Close Combat Weapons|Ranged Weapons)\)$/`) worth consolidating to.

## Other checks
- New `Chem-alchemy` entries in `equipmentCategoriesN26` correctly append after `Wargear`, so `equipmentSuperCategoryRankN26`'s derived ordering and the display-name stripping both pick them up automatically — no other call site needed updating (confirmed via `getEquipmentCategoryRank.ts` / `components/equipment/equipment.tsx`).
- Gating (`hasChemAlchemy`, `hasEquipmentSuperCategories`) is untouched and orthogonal to this change — no edition-gating regression.
- No security, performance, or null-handling concerns; this is UI copy + a static lookup table.

## Verdict: **Approve** (non-blocking) — nice, tightly-scoped cleanup. Would recommend picking up finding #1 either in this PR or as an immediate fast-follow, since it's a one-line-per-site fix that completes the PR's own stated goal.

@Djidiouf

---

PR #2096, opened 2026-09-05
Leader advancement deducation fix

v0rp-z:
## Summary

Follow on from previous pr #2094 

**Tizdale reported:**
> Found an oddity on "fighter type change" 
> 
> Add a Prospect -> Add 40XP
> Promote to Ganger, Specialist -> 7 Advancements goes to 6.
> Promote to Champ, Advancements remain at 6. 
> Promote to Leader via Promote (Death of a Leader) 
> Type changes due to this, Promotions jump up to 7

this PR replaces `fighter_type.fighter_subtypes` with `fighters.fighter_specialisation_id` which is a much better signal for this work, as it persists all the way through to the Leader sub-type

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] local dev env

https://github.com/user-attachments/assets/1ce00cdd-d137-4041-8df6-40e1b03fcb9d

**Prospect manual edit to Champion** 

https://github.com/user-attachments/assets/93d7c078-a769-4791-9ff3-1f400af635a8

## Risk / rollout
- Low

### DIFF
--- a/app/lib/fighter-advancements.ts
+++ b/app/lib/fighter-advancements.ts
@@ -40,10 +40,7 @@ export const getGangFighters = async (gangId: string, supabase: any) => {
     id: f.id,
     fighter_name: f.fighter_name,
     fighter_type: f.fighter_type,
-    fighter_subtypes: Array.isArray(f.fighter_subtypes) ? f.fighter_subtypes : [],
-    fighter_type_subtypes: Array.isArray(f.fighter_types?.fighter_subtypes)
-      ? f.fighter_types.fighter_subtypes
-      : [],
+    fighter_specialisation_id: f.fighter_specialisation_id ?? null,
     xp: f.xp,
     starting_xp: f.starting_xp ?? null,
     advancements_taken: countAdvancementsTaken(
--- a/app/lib/shared/gang-assembly.ts
+++ b/app/lib/shared/gang-assembly.ts
@@ -677,9 +677,6 @@ export function assembleGangFighters(
           label: fighter.label,
           fighter_type: fighter.fighter_type || fighterTypeInfo.fighter_type || 'Unknown',
           fighter_subtypes: fighter.fighter_subtypes || [],
-          fighter_type_subtypes: Array.isArray(fighterTypeInfo.fighter_subtypes)
-            ? fighterTypeInfo.fighter_subtypes
-            : [],
           fighter_specialisation: fighterSpecialisationInfo ? {
             fighter_specialisation: fighterSpecialisationInfo.specialisation_name,
             fighter_specialisation_id: fighterSpecialisationInfo.id
--- a/app/lib/shared/gang-data.ts
+++ b/app/lib/shared/gang-data.ts
@@ -137,12 +137,6 @@ export interface GangFighter {
   label?: string;
   fighter_type: string;
   fighter_subtypes: string[];
-  /**
-   * Catalog subtypes on the fighter_type row. Kept-type promotion
-   * (Prospect→Ganger/Specialist) leaves this on the origin type while live
-   * `fighter_subtypes` flip; it is the stable signal for "was recruited as X".
-   */
-  fighter_type_subtypes?: string[];
   fighter_specialisation?: {
     fighter_specialisation: string;
     fighter_specialisation_id: string;
--- a/components/fighter/fighter-advancement-list.tsx
+++ b/components/fighter/fighter-advancement-list.tsx
@@ -189,13 +189,6 @@ interface AdvancementsListProps {
   fighterId: string;
   editionSlug?: string | null;
   fighterSubtypes: string[];
-  /**
-   * Catalog subtypes on the fighter's fighter_type row. Distinct from live
-   * `fighterSubtypes` because N26 keep-type promotion leaves the catalog on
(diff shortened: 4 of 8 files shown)


joeseos:
@v0rp-z Reviewed this against the surrounding fighter/advancement/promotion code. The Prospect→Ganger→Champion→Leader chain from the reported bug looks correctly fixed, but I found one correctness issue with the new signal that I think is worth addressing before merge.

## Findings

### 🔴 High — `hasN26ProspectPromotionOccurred` false-positives on natively-specialised fighters, wrongly deducting an Advancement they never traded away

**File:** `utils/keepTypePromotionN26.ts`, lines 63-70

```ts
export function hasN26ProspectPromotionOccurred(
  editionSlug?: string | null,
  fighterSpecialisationId?: string | null
): boolean {
  if (!hasProspectSpecialisationPromotion(editionSlug)) return false;
  if (!fighterSpecialisationId) return false;
  return N26_PROSPECT_SPECIALISATIONS.some((s) => s.id === fighterSpecialisationId);
}
```

This treats "current `fighter_specialisation_id` is one of the 8 catalog specialisation ids" as proof that the fighter traded in its 13-XP Advancement via the Prospect promotion. But that's not a reliable signal on its own — the same file already documents this exact ambiguity a few dozen lines below, for the *undo* check:

> ```
> * Not sufficient alone for undo — native specialised Gangers share these
> * specialisations.
> ```
> (`isN26ProspectPromotionUndoCandidate`, line 145)

Confirmed this is reachable, not just a hypothetical:
- `app/actions/add-fighter.ts:529` — `fighterInsertData.fighter_specialisation_id = fighterTypeData.fighter_specialisation_id;`. Any catalog fighter type recruited directly with a preset specialisation (a native "Ganger (Heavy)"/"(Gunner)"/etc., not a Prospect promotion) gets one of these same 8 ids written straight onto the fighter row.
- `app/actions/copy-fighter.ts:342` / `:352` and `:784`/`:796` copy `fighter_specialisation_id` verbatim regardless of how the source fighter acquired it.
- The migrations `20260821100000_add_fighter_variant.sql` and `20260823120000_narrow_fighter_specialisation_id.sql` both special-case exactly these 8 ids as "the eight real specialisations" that catalog `fighter_types` rows can carry natively — separate from the Prospect promotion flow.

**Failure scenario:** Recruit a fighter directly from a catalog type that has a built-in specialisation (e.g. a native Ganger (Gunner), never a Prospect), or copy such a fighter to another gang. `hasN26ProspectPromotionOccurred` returns `true` for it (assuming the edition has `prospectSpecialisationPromotion`), so `openAdvancementsFor` silently subtracts 1 from that fighter's available Advancements even though no promotion trade-in ever happened — the exact class of miscount this PR is trying to fix, just for a different fighter population.

**Suggested fix:** Don't rely on `fighter_specialisation_id` alone. Either:
- Compare against the fighter's *current* `fighter_type_id` catalog row's own default `fighter_specialisation_id` — if they match, the specialisation came from the type itself (native), not a promotion; a mismatch (e.g. current type is Champion/Leader, which has no native specialisation default, but the fighter still carries one) means it was carried over from a promotion. This needs the caller to pass the type's catalog default through, or
- Track provenance explicitly with a dedicated flag/column set only by `applyN26ProspectPromotion` (and cleared by copy/undo), instead of overloading `fighter_specialisation_id` for two different facts the way the pre-migration `fighter_specialisation_id` overloaded specialisation and variant (see the two migrations above — that overload was exactly the bug they were written to fix).

### 🟡 Low — no test coverage for the new counting logic

`hasN26ProspectPromotionOccurred` and the updated `openAdvancementsFor` have no unit tests. Given how easy it is to get this kind of provenance signal subtly wrong (as above), and that this is XP/Advancement accounting players will notice, a couple of table-driven cases (Prospect-promoted fighter vs. native specialised fighter vs. copied fighter) would have caught the issue above. Not blocking if the repo doesn't otherwise unit-test `utils/`, just flagging since none exist here.

## Everything else looks solid
- All three `openAdvancementsFor` call sites (`fighter-card.tsx`, `fighter-page.tsx` ×2, `fighter-advancement-list.tsx`) were updated consistently, including the `fighter-card.tsx` `useMemo` dependency array picking up `fighter_specialisation_id`.
- `fighter_specialisation_id` is correctly threaded through `getGangFighters`/`getGangFightersBundle` (it's a real selected column, not synthesized).
- Confirmed the Prospect-promotion skill grant is `is_advance: false` (`app/actions/fighter-advancement.ts:883`), so there's no double-deduction between `countAdvancementsTaken` and the new `prospectPromotionConsumed` flag — the design intent here is sound, just the detection signal is too broad.
- Verified the Champion→Leader recipe genuinely preserves the specialisation (`applyN26ChampionLeaderPromotion`, "Keep existing specialisation... do not clear or overwrite") and that promotion-undo only clears it for the Prospect case — both match the new docstring's claims.
- No `CLAUDE.md` or contributing conventions doc exists in this repo to check against.

The high-severity item is a real regression risk for a fighter population outside the one this PR was written to fix — worth resolving before merge, since it would ship a new (if narrower) advancement-miscount bug for native specialists/copied fighters.

---

joeseos (approved):

PR #2097, opened 2026-09-05
Added support for tactics cards packs

joeseos:
## Summary

Added tactics cards packs to enable expansion beyond the core tactics cards
-

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

Tested on local after adding the Goliath and Escher packs. 

- Further testing needed with "sub gang types" 
- No actual new tactics cards added

## Database

### Migration status

- [x] Migration already applied

## Risk / rollout

- Not yet tested with tactics cards as there's no data added yet
- Not yet tested with sub gang types/archetypes

### DIFF
--- a/app/actions/gang-tactics-cards.ts
+++ b/app/actions/gang-tactics-cards.ts
@@ -15,6 +15,7 @@ import {
   GANG_TACTICS_CARD_SELECT,
   normaliseTacticsDescription,
   TACTICS_DESCRIPTION_CHAR_LIMIT,
+  tacticsCardsPackFilter,
   toGangTacticsCard,
   type GangTacticsCard
 } from '@/types/tactics-card';
@@ -30,13 +31,17 @@ interface AddGangTacticsCardsResult extends GangTacticsResult {
 }
 
 interface GangTacticsContext {
-  editionId: string | null;
+  /** Every pack this gang may draw from, core included. */
+  availablePackIds: string[];
+  /** The pack the picker falls back to when none is chosen. */
+  corePackId: string | null;
 }
 
 /**
  * Authenticate, confirm the caller may edit this gang, and confirm the edition
  * has Gang Tactics. RLS enforces the first two again, but failing here returns
- * a usable message instead of a raw error.
+ * a usable message instead of a raw error. Resolves the gang's packs too, so
+ * neither caller has to repeat the scoping.
  */
 async function authoriseGangTactics(
   supabase: any,
@@ -49,7 +54,7 @@ async function authoriseGangTactics(
     .select(`
       id,
       user_id,
-      gang_types!gang_type_id ( editions:edition_id ( id, slug ) ),
+      gang_types!gang_type_id ( gang_type_id, parent_gang_type_id, editions:edition_id ( id, slug ) ),
       custom_gang_types!custom_gang_type_id ( editions:edition_id ( id, slug ) )
     `)
     .eq('id', gangId)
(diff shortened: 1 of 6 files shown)


joeseos:
@joeseos Reviewed this against the surrounding gang/edition code (`gangEditionJoin`, `checkPermissionCached`, the `gang_types.parent_gang_type_id` one-level-deep house model added in `20260902172743_add_gang_types_parent_gang_type_id.sql`) and the full picker/roll flow end-to-end. This is a clean, well-scoped feature — no correctness bugs found in the query logic, RLS, or React state handling. A few things worth confirming before merge.

## Findings

### 🟠 Medium — confirm the migration is applied before this merges

The PR body's Database section is unchecked (migration added ✓ implicitly via the new file, but "Needs maintainer to apply" / "Migration already applied" are both unticked). This isn't just paperwork here: `authoriseGangTactics`, the `/api/tactics-cards` route, and `verifyAndLogRolledTacticsCard` all query `tactics_cards_packs` and `tactics_cards.tactics_cards_pack_id` unconditionally — if this ships before `20260905120000_create_tactics_cards_packs.sql` is applied, every Gang Tactics read/write (not just the new pack picker) breaks immediately for every N26 gang, including the existing "view held cards" list. Worth an explicit confirmation in the PR before merging, per your own template's warning that migrations aren't auto-applied.

### 🟡 Low — `/api/tactics-cards` GET has no per-gang authorization check

`app/api/tactics-cards/route.ts` checks `getUserIdFromClaims` (authenticated) but, unlike `authoriseGangTactics` (which calls `checkPermissionCached(...).canEdit`), never checks that the caller can view/edit the specific `gang_id` passed in. Any authenticated user can pass an arbitrary gang id and learn its edition/gang-type-derived pack list. This matches the previous route's posture (it also had no gang-specific check, just needed an `edition_slug`), so it's not a regression, but the new version does resolve a real `gang_id` to its type/edition where before it only echoed back a slug the caller already knew — worth a quick check that this doesn't cross a privacy boundary for private/campaign-restricted gangs (RLS on `gangs` may already block this by returning null — if so this is a non-issue, just confirm).

### 🟢 Info — migration backfill assumes only N26 has `tactics_cards` rows today

```sql
INSERT INTO public.tactics_cards_packs (edition_id, gang_type_id, name)
SELECT e.id, NULL, 'Core Gang Tactics'
FROM public.editions e
WHERE e.slug = 'n26'
  ...
```
This is consistent with `gangTacticsCards: { n23: false, n26: true }` in `types/edition.ts`, so it should be a no-op assumption in practice. If it's ever wrong, the later `ALTER COLUMN tactics_cards_pack_id SET NOT NULL` will fail loudly and roll back the whole migration (it's wrapped in `BEGIN`/`COMMIT`), rather than silently leaving orphaned rows — so this is self-defending, just flagging the assumption for visibility.

## What I checked and confirmed correct
- The composite FK pattern (`tactics_cards_packs_gang_type_edition_fkey` on `(gang_type_id, edition_id)` → `gang_types(gang_type_id, edition_id)`) mirrors the existing `fighter_types_gang_type_edition_fkey` precedent exactly, including reusing the pre-existing `gang_types_gang_type_id_edition_id_key` unique constraint.
- `tacticsCardsPackFilter(gangTypeId, parentGangTypeId)` correctly covers exactly the one-level house/alternate-list tree enforced by `gang_types_parent_must_be_root` — no risk of missing a pack for a multi-level chain, because the schema doesn't allow one.
- Both mutation paths (`addGangTacticsCards`, `verifyAndLogRolledTacticsCard`) still validate server-side against `auth.context.availablePackIds` rather than trusting the client-posted pack/card ids — same "don't trust the picker" posture as before, correctly extended to packs.
- The D66 roll resolution is properly re-scoped to one pack (`'.eq('tactics_cards_pack_id', packId)`) so overlapping ranges across packs can't cross-match.
- Client state (`selectedPackId`/`selectedCardIds` reset together, `activePack` falling back to `corePack`, `ownedCardIds` computed globally rather than per-pack) all checked out — no stale-state or cross-pack contamination bugs.
- Dropping `tactics_cards_edition_id_name_key` in favor of `tactics_cards_tactics_cards_pack_id_name_key` is correct and necessary now that two packs in the same edition can have overlapping card names; verified the old constraint name matches what's actually in `schema.public.sql` before the `DROP CONSTRAINT`.
- No other call sites of `tactics_cards`/`TacticsCard` were missed (`gang-data.ts`'s `getGangTacticsCards` only touches `gang_tactics_cards`, unaffected).

## Verdict: **Approve**, pending confirmation that the migration will be applied at/before merge.

---

Djidiouf (approved):

PR #2099, opened 2026-09-06
Fixed so that the sign-up page is static

joeseos:
## Summary

- Fixed so that the sign-up page is static and that the email confirmation page now is public

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested the pages in browser

## Database

### Migration status

## Risk / rollout

- None

### DIFF
--- a/app/sign-up/page.tsx
+++ b/app/sign-up/page.tsx
@@ -1,12 +1,12 @@
 'use client';
 
 import { signUpAction } from "@/app/actions/auth";
-import { FormMessage, Message } from "@/components/form-message";
+import { FormMessage } from "@/components/form-message";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
 import Link from "next/link";
-import { useRouter } from "next/navigation";
-import { useState, use } from "react";
+import { useRouter, useSearchParams } from "next/navigation";
+import { Suspense, useState } from "react";
 import {
   EMPTY_PASSWORD_REQUIREMENTS,
   PASSWORD_ERROR_MESSAGE,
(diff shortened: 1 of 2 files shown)


joeseos:
@joeseos Small, well-targeted fix. Reviewed against the sibling auth pages and the proxy's routing logic.

## What this actually fixes
The root cause of "sign-up page not static": `Page(props: { searchParams: Promise<Message> })` consumes `searchParams` as a page prop, which forces Next.js to render the whole route dynamically per-request — even though the file is `'use client'` and almost everything in it is client-side state. Switching to `useSearchParams()` wrapped in `<Suspense>` is the correct, standard fix, and it's not a novel pattern here: `app/sign-in/page.tsx` already does exactly this (`Suspense` + `useSearchParams()` + per-key `.get()` calls, no `fallback` prop either), so this PR is just bringing `/sign-up` to parity with a fix that was already applied to its sibling page. Verified no other page in the app still has the broken `searchParams: Promise<...>` prop pattern.

## Checked and confirmed correct
- Dropping the `success` branch of `Message` when building `{ error: urlError }`: verified `signUpAction` (`app/actions/auth.ts`) only ever returns `{ error }` or `{ message }`, and the only two `router.push('/sign-up?...')` calls in this file use `message=` and `error=` — `success` was never populated for this page, so this isn't a behavioral regression, just dead-branch removal.
- `proxy.ts`'s `publicPaths` uses exact `pathname` matching (`publicPaths.includes(pathname)`), so adding `/confirm-email` is safe and doesn't need the prefix-match treatment `reset-password` gets.
- `/confirm-email` (`app/confirm-email/page.tsx`) is actually the **email-change** confirmation page (`EmailChangeConfirmPage`, `type: 'email_change'` OTP), not the new-signup verification flow — that one already goes through `/auth/callback`, which was already public. Making `/confirm-email` public is the right call regardless: a user changing their email typically opens the confirmation link from their email client in a browser session that isn't logged into the app, so gating it behind auth (redirecting to sign-in) would have made that flow unusable — but it's a separate concern from "the sign-up page," so much this PR's description ("email confirmation page is public") slightly undersells what's actually a second, independent bugfix bundled in here. Worth a one-line callout in the PR body for anyone reading it later, but not a reason to hold this up.
- No dangling references to the removed `Message` import or the old `props.searchParams` shape elsewhere in the file.

## Verdict: **Approve**

Nothing blocking. Only a documentation nit: the PR description could be a bit clearer that this bundles two independent fixes (static sign-up page + public email-change confirmation), since they're unrelated causes even though both touch "sign-up flow" tangentially.

---

Djidiouf (approved):

PR #2100, opened 2026-09-06
Rename Gang Variants to Gang Subtypes in the app.

Djidiouf:
Keep DB table/column names unchanged and map at the Supabase boundary; update UI copy and the user-guide image URL.

## Summary
- Renames user-facing Gang Variants to Gang Subtypes across create/edit gang, display, admin, trading posts, and docs.
- Renames app identifiers and moves ``/api/gang-variant-types`` to ``/api/gang-subtype-types`` with ``variant`` → ``subtype`` mapping.
- Leaves Postgres names (``gang_variant_types``, ``gangs.gang_variants``, ``gang_variant_id``) and catalog ``Variant: …`` strings unchanged for a follow-up migration.
- Renames Gang Archetype to Gang Variant in the UI only and left a TODO

## Test plan
- [x] Create a gang with subtypes; reload and confirm they persist
- [x] Edit gang subtypes; Outlaw sync with alignment still works
- [x] Home card / print show subtype names
- [x] Add Fighter unlocks for Secundan Incursion / Corrupted / Infested
- [x] Admin equipment: availability per gang subtype + scoped grants
- [x] Custom trading post rule scoped by gang subtype
- [x] User guide image loads from ``gang-subtypes.webp``

## Folow up PR

Will update the DB and do the code 2nd pass both for the renaming of gang variant to gang subtypes and then, the gang archetype to gang variant.


joeseos:
@Djidiouf Reviewed this against the full codebase (verified the actual merge-base diff — 481+/454- across 26 files — and cross-referenced every renamed identifier/route/shape repo-wide, not just within the diff hunks).

## Summary

This is a clean, thorough rename migration. I checked specifically for the classic incomplete-rename failure mode (a producer renamed but a consumer left reading the old field, or vice versa) across every touched shape (`gang_variants` id-array vs. `gang_subtypes` object-array, `{id, variant}` → `{id, subtype}`, `is_gang_variant`/`gang_variant_name` → `is_gang_subtype`/`gang_subtype_name`, the `/api/gang-variant-types` → `/api/gang-subtype-types` route + all its callers) and found no dangling references anywhere in the tree — every producer/consumer pair was updated together.

## Findings

**1. Medium — UI label collision between the two renames, in the same modal**
`components/create-gang-modal.tsx:530-537` and `:1320-1327`

The "Gang Archetype → Gang Variant" UI-only rename lands in the same component as the "Gang Variant → Gang Subtype" rename, producing two adjacent sections that now read:
```tsx
{/* TODO: UI label only for now (was "Gang archetype"). Concept is Gang Variant; additional variable renaming is left to do. */}
<span>Gang Variant</span>
...
<label>Gang Subtypes</label>
```
A user creating a gang sees a "Gang Variant" field (archetype choice) directly above a "Gang Subtypes" section — which is exactly what every existing player/doc has called "Gang Variants" up to now. This is already flagged with a TODO by the author as an acknowledged, temporary state pending the full archetype-rename pass, so I'm not blocking on it, but since both halves of the confusion ship in this same PR it's worth calling out — worth either finishing the archetype-side rename in the same PR, or holding the archetype label change for the follow-up so the two renames don't overlap on screen.

**2. Low (nit) — label capitalization inconsistency**
`components/admin/admin-edit-equipment.tsx` (scoped-grant summary line)

New label reads `Gang subtype: …` (lowercase "s") while the sibling labels built in the same joined string are `Origin: …` and `Subtype: …` (capitalized). Cosmetic only.

## Checklist

- **Correctness/regressions**: No bugs found. Rename is self-consistent end-to-end (verified `app/lib/get-user-gangs.ts`, `app/lib/shared/gang-data.ts`, `app/api/fighter-types/route.ts`, `app/api/gangs/[id]/route.ts`, `app/api/admin/equipment/route.ts`, `fighter-edit-modal.tsx`, `admin-edit-equipment.tsx`, `create-gang-modal.tsx`, `gang-edit-modal.tsx`, `gang.tsx`, `print-gang.tsx`, `gang-card.tsx`, `custom-trading-posts.tsx`).
- **Edge cases / null handling**: Preserved correctly through the DB-boundary mapping (`?? null`, `.filter(Boolean)`, `'Unknown'` fallbacks). `gang.tsx`'s `handleEditModalOpen` adds a defensive `item.subtype ?? item.variant ?? 'Unknown'` fallback — technically unnecessary now that the API always returns `subtype`, but harmless (minor YAGNI nit, not worth a change).
- **Security**: The dropped `checkAdmin` import in `app/api/gang-variant-types/route.ts` → `gang-subtype-types/route.ts` is *not* a regression — confirmed via the pre-PR file that `checkAdmin` was imported but never actually called; only `getUserIdFromClaims` gated the route, and that gate is preserved. Good catch removing the dead import (the route is correctly non-admin-gated since regular users need it for gang creation/editing).
- **Performance**: The batched `.in()` query pattern in `get-user-gangs.ts` is unchanged logic (pure rename), not a new perf change.
- **Readability/conventions**: Consistent with surrounding code style; comments correctly document the DB-column-vs-app-field mapping at each boundary (e.g. `// DB column remains gang_variants`), which will help whoever does the DB-side follow-up migration.
- **DRY/YAGNI**: No new duplication introduced.

## Verdict: **Approve**

No correctness or security issues. The one substantive item (#1) is UI copy, already self-flagged as a known interim state, and doesn't block merging a first-pass rename PR — but flagging it so it doesn't get lost before the archetype follow-up.

---

joeseos (approved):

PR #2101, opened 2026-09-06
Rename gang variants to gang subtypes in the database and app.

Djidiouf:
Add a migration for table/column/FK renames and Variant: → Subtype: catalog labels; update RPCs, seed, and drop the app mapping layer.

## Summary
- Adds migration ``20260906093510_rename_gang_variants_to_gang_subtypes`` (table/columns/indexes/FKs/policies + catalog ``Variant:`` → ``Subtype:``).
- Updates ``get_gang_details``, ``get_equipment_detailed_data``, and ``copy_custom_collection`` under ``supabase/functions/``.
- Drops the app mapping layer so code uses ``gang_subtype_types`` / ``gang_subtypes`` / ``gang_subtype_id`` directly.
- Updates ``seed.sql`` for the renamed catalog table.

## SQL to run (hard cutover — do this before/with app deploy)

Run in this order on the target database (staging then production). Do **not** rely on the app deploy alone; RPCs are not auto-applied by the migration.

### 1. Schema migration
- [ ] ``supabase/migrations/20260906093510_rename_gang_variants_to_gang_subtypes.sql``

### 2. RPC / function SQL (apply full file contents via SQL editor / dashboard)
Apply **after** the migration (they reference the new names):

- [ ] ``supabase/functions/get_equipment_detailed_data.sql`` — ``get_equipment_detailed_data``
- [ ] ``supabase/functions/copy_custom_collection.sql`` — ``copy_custom_collection``
- [ ] ``supabase/functions/get_gang_details.sql`` — ``get_gang_details`` (legacy / Rule Snatcher; still update for compatibility)

### 3. App deploy
- [ ] Deploy this branch so the app matches the new schema/RPC names

### Not a live-DB step
- ``supabase/seed.sql`` — local/dev seed only; not run against production

## Test plan
- [ ] Apply migration to staging/local
- [ ] Apply the three function SQL files above
- [ ] Create/edit gang subtypes; reload and confirm persistence
- [ ] Add Fighter unlocks for Secundan Incursion / Corrupted / Infested / Malstrain
- [ ] Admin equipment: availability + scoped grants by gang subtype
- [ ] Custom trading post rules scoped by gang subtype
- [ ] Copy gang preserves subtypes

### DIFF
--- a/app/actions/copy-gang.ts
+++ b/app/actions/copy-gang.ts
@@ -69,7 +69,7 @@ export async function copyGang(params: CopyGangInput): Promise<CopyGangResult> {
         credits: sourceGang.credits,
         reputation: sourceGang.reputation,
         trade_points: sourceGang.trade_points ?? 0,
-        gang_variants: sourceGang.gang_variants,
+        gang_subtypes: sourceGang.gang_subtypes,
         note: sourceGang.note,
         note_backstory: sourceGang.note_backstory,
         positioning: null, // Will be updated after fighters are copied
--- a/app/actions/create-gang.ts
+++ b/app/actions/create-gang.ts
@@ -55,7 +55,6 @@ export async function createGang({
     }
 
     // Insert the new gang (exclusive arc: gang_type_id or custom_gang_type_id, never both)
-    // DB column remains gang_variants
     const { data, error } = await supabase
       .from('gangs')
       .insert([{
@@ -71,7 +70,7 @@ export async function createGang({
         alignment,
         gang_affiliation_id: gangAffiliationId || null,
         gang_origin_id: gangOriginId || null,
-        gang_variants: gangSubtypes.length > 0 ? gangSubtypes : null,
+        gang_subtypes: gangSubtypes.length > 0 ? gangSubtypes : null,
         default_gang_image: defaultGangImage
       }])
       .select();
--- a/app/actions/customise/custom-trading-posts.ts
+++ b/app/actions/customise/custom-trading-posts.ts
@@ -254,7 +254,7 @@ export interface CustomTPAvailabilityRule {
   gang_type_id: string | null;
   custom_gang_type_id: string | null;
   gang_origin_id: string | null;
-  gang_variant_id: string | null;
+  gang_subtype_id: string | null;
   campaign_type_allegiance_id: string | null;
   alignment: string | null;
   availability: string | null;
@@ -464,13 +464,13 @@ export async function getAvailabilityRules(
         gang_type_id,
         custom_gang_type_id,
         gang_origin_id,
-        gang_variant_id,
+        gang_subtype_id,
         campaign_type_allegiance_id,
         alignment,
         availability,
         custom_gang_types (gang_type),
         gang_origins (origin_name),
-        gang_variant_types (variant),
+        gang_subtype_types (subtype),
         campaign_type_allegiances (allegiance_name)
       `)
       .eq('custom_trading_post_equipment_id', equipmentItemId)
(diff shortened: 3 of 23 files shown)


joeseos:
@Djidiouf Reviewed this PR against the full codebase (checked out the branch locally and grepped app + supabase code for stragglers).

## Summary
This is a clean, well-scoped rename. I traced every renamed identifier (`gang_variant_types`→`gang_subtype_types`, `variant`→`subtype`, `gang_variants`→`gang_subtypes`, `gang_variant_id`→`gang_subtype_id`, and the `Variant: `→`Subtype: ` catalog prefix) through:
- The migration (table/column/FK/index/policy renames + comments + catalog `UPDATE`)
- All three RPC files (`get_gang_details`, `get_equipment_detailed_data`, `copy_custom_collection`)
- Every app route/action/component touched (`app/actions/*`, `app/api/**`, `app/lib/shared/gang-data.ts`, `components/admin/admin-edit-equipment.tsx`, `components/customise/custom-trading-posts.tsx`, `components/gang/gang.tsx`)
- `seed.sql`

I found zero remaining references to the old names anywhere in app/TS/RPC code (confirmed via repo-wide grep for `gang_variant`, `.gang_variants`, `GangVariant`, `gangVariant`, and the `Variant:` prefix pattern). The `supabase/schema/schema.public.sql` dump still shows old names, but that file is only ever refreshed via separate `chore(db): update Supabase schema snapshot` commits per this repo's existing convention (confirmed via git log), so that's expected and not something this PR needs to touch.

The one place both the migration and `app/api/fighter-types/route.ts` need to agree — the catalog label going from `Variant: {subtype}` to `Subtype: {subtype}` (migration step 7's `UPDATE ... replace(gang_type, 'Variant: ', 'Subtype: ')` vs. the new `subtypeGangTypeName` helper) — is consistent both ways. Good catch keeping those two in lockstep since a mismatch there would silently make subtype-specific fighter unlocks disappear rather than error.

## Findings

**[Low] Deployment sequencing risk (inherent to hard-cutover renames, already flagged by the author)**
The migration renames columns/tables that `get_gang_details`, `get_equipment_detailed_data`, and `copy_custom_collection` depend on, but those RPC bodies live in `supabase/functions/*.sql` and are *not* auto-applied by the migration (per repo convention) — they need a separate dashboard/SQL-editor apply. If the migration lands before the RPC SQL is applied, gang details / equipment / copy-gang will throw against the renamed columns until the RPCs are redeployed. This is explicitly called out in the PR description's "SQL to run" checklist with the correct order, and mirrors the previously-shipped `20260806120000_rename_fighter_classes_to_fighter_subtypes` pattern, so this is a process reminder for whoever runs the deploy rather than a code change — just flagging it since there's no automated guard against applying them out of order.

**[Low / nit] Migration comment is confusing**
`supabase/migrations/20260906093510_rename_gang_variants_to_gang_subtypes.sql`, lines 68–74: the comment starts with "Unique scope index must be rebuilt" but then concludes no rebuild is needed (Postgres indexes track columns by attnum, so the rename propagates automatically). The reasoning is correct, just the opening clause reads as contradicting the actual conclusion — could reword the first line to something like "Confirm the unique scope index doesn't need a rebuild" to avoid confusing a future reader skimming the migration.

No correctness bugs, null-handling issues, security issues, or DRY/YAGNI concerns found. Didn't spot any adherence issues — this repo has no `CLAUDE.md`, so conventions were checked against the existing codebase patterns (e.g. the prior fighter_classes→fighter_subtypes rename), which this PR follows closely.

## Verdict
**Approve** — no blocking issues. The two notes above are informational/process, not required changes.

---

joeseos (approved):

PR #2103, opened 2026-09-06
Fixed invalidation duplication

joeseos:
## Summary

-   Both hot routes read the gang fighters bundle twice per render. On the fighter page, `getGangFighters` looked like a small helper for the navigation dropdown, but its first line fetched the whole gang bundle — the same entry the page had already requested. The gang page did the same via `getGangVehicles`, and `getGangFightersList` re-read `getGangCore` on top of that.

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [ ] 
- [ ] 

## Database

### Migration status

## Risk / rollout

  - No cache tags or invalidation logic changed. Nothing can go stale that wasn't already; every entry keeps its existing tag and key.
  - Rollback is clean. No cache poisoning in either direction, so a revert needs no purge.


joeseos:
## Review

@joeseos Checked out the branch and reviewed against the full codebase (not just the diff) — traced every relocated type/function to its consumers and diffed the removed fields/dead code against actual usages.

**Checklist**

- **Correctness**:
  - `unstable_rethrow(error)` added to all four page-level catch blocks (`campaigns/[id]`, `fighter/[id]`, `gang/[id]`, `gang/[id]/print`) that wrap a `notFound()`/`forbidden()`/`redirect()` call. This is a real bug fix, not just cleanup — before this, those Next.js control-flow throws were falling into the generic `catch` and getting `console.error`'d as failures even though they're expected navigation signals. Placement (before `console.error`) matches the documented Next.js pattern.
  - `selectGangFighterIndex` (replacing the deleted `app/lib/fighter-advancements.ts`) now excludes `effect.vehicle_id` rows when bucketing advancement effects (`utils/gang-assembly.ts:853`), matching `assembleGangFighters`' own `fighterEffectsRows` filter (line 101). The old `getGangFighters` didn't have this filter, so it's a genuine latent-bug fix (a vehicle-scoped effect row that also carries a `fighter_id` could have inflated `advancements_taken` in the fighter-page nav dropdown).
  - `getGangFightersList`'s new `gangEditionSlug`/`needsCore` branch is consistent with all three call sites: `gang/[id]/page.tsx` and `print/page.tsx` pass the slug they already hold (skipping a redundant `getGangCore`), and `battle-session/[id]/page.tsx` passes neither, so it still resolves core itself — matching the comment exactly.
  - `gang/[id]/page.tsx`'s `let pageProps` / JSX-outside-`try` restructuring is safe: every path through `catch` ends in a `throw`, so `pageProps` is always assigned by the time the `return` is reached.
- **Dead code removal (verified, not just trusted)**: repo-wide grep confirms `owned_beasts`/`beastFighters`/`ownedBeastsData` had zero consumers anywhere, and `gang_subtypes` was never read on the fighter page/component tree — both were pure wasted work (an extra `getGangSubtypes` call, a beast-remapping loop) being sent to the client for nothing. Same for the `alliance_type`/`gang_affiliation`/`gang_origin` raw objects and `gang_types` dropped from `gangDataForClient` — the `_name`/`_id` fields the client actually reads are untouched.
- **No stale references**: grepped for the old `app/lib/shared/gang-assembly`, `app/lib/fighter-advancements`, and old `GangFighter`/`GetGangFightersListOptions` import paths — nothing left pointing at the moved locations.
- **Edge cases / null handling**: `needsCore = callerSlug === undefined` correctly distinguishes "caller passed `null`" (edition genuinely unresolved, use it) from "caller passed nothing" (fetch core) — not conflated.
- **Security**: N/A, no auth/data-access changes; permission checks (`canViewHiddenGang`, `checkPermissionCached`) untouched.
- **Performance**: Net improvement — this removes duplicate bundle/core reads on the two hottest pages, exactly as described. The gang page's fighters+vehicles assembly is folded into a single `.then()` inside the existing `Promise.all` so the CPU pass overlaps other queries instead of serializing after them.
- **Readability/maintainability**: Good — comments explain *why* (e.g. the vehicle_id filter parity, the JSX-outside-try reasoning), not what. `types/gang.ts` becoming the shared home for `GangFighter`/`GangFightersBundle`/etc. cleans up what was previously a type owned by a data-loader module and imported sideways by a pure-transform module.
- **Conventions**: No CLAUDE.md in this repo; checked against CONTRIBUTING.md — no schema/migration involved (correctly unchecked in the PR template), nothing here conflicts with stated conventions.
- **DRY/YAGNI**: This is the point of the PR — collapsing duplicate bundle reads into shared selectors. No new duplication introduced.

**Findings**

1. **(Low) Empty test plan.** The "How I tested" checklist is unchecked/empty, but this PR touches control-flow/error-handling on four route pages (including `forbidden()`/`notFound()` paths) plus the gang and fighter page data assembly — a click-through of a hidden gang (403), a missing gang/fighter (404), and a normal gang+fighter page load would be worth doing/noting before merge, given the blast radius is larger than the title ("invalidation duplication") suggests.

**Severity**: Low (process note only — no code defect found)

**Verdict**: ✅ Approve

---

Djidiouf (approved):

PR #2104, opened 2026-09-07
Refactored exotic beast cache

joeseos:
## Summary

- A DRY consolidation (6 copies → 1, 'use server' removed, dead canEquipmentCreateBeasts deleted) and a latency fix (2 serial round trips → 0 on the buy path, 1 → 0 on four others).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Bought an Exotic Beast / Pet, checked the cost on the owner and the owner relation. Sold Exotic Beast equipment and cost on owner fighter is correct.

## Database

### Migration status

## Risk / rollout

- None

### DIFF
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -19,25 +19,7 @@ import { mapArchetypeSkillAccessToOverrides } from '@/utils/archetypeEligibility
 import { syncFighter } from '@/utils/syncVenatorSkillOverrides';
 import { isVenatorGang } from '@/utils/venatorSkillAccess';
 import { gangEditionSlug } from '@/types/edition';
-
-// Helper function to invalidate owner's cache when beast fighter is updated
-async function invalidateBeastOwnerCache(fighterId: string, gangId: string, supabase: any) {
-  // Check if this fighter is an exotic beast owned by another fighter
-  const { data: ownerData } = await supabase
-    .from('fighter_exotic_beasts')
-    .select('fighter_owner_id')
-    .eq('fighter_pet_id', fighterId)
-    .single();
-    
-  if (ownerData) {
-    // Invalidate the owner's cache since their total cost changed
-    invalidateFighter(ownerData.fighter_owner_id, gangId);
-
-    // Invalidate the owner's beast costs cache
-    // Without this, the owner's cost calculation uses stale beast data
-    revalidateTag(TAGS.fighter(ownerData.fighter_owner_id), { expire: 0 });
-  }
-}
+import { invalidateBeastOwnerCache } from '@/utils/exotic-beasts';
 
 async function getKilledStatusEffects(supabase: any, fighterId: string) {
   const { data: effects, error } = await supabase
--- a/app/actions/edit-fighter.ts
+++ b/app/actions/edit-fighter.ts
@@ -244,7 +244,8 @@ export async function editFighterStatus(params: EditFighterStatusParams): Promis
         starved,
         recovery,
         captured,
-        captured_by_gang_id
+        captured_by_gang_id,
+        fighter_pet_id
       `)
       .eq('id', params.fighter_id)
       .single();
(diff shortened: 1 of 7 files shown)


joeseos:
## Review

@joeseos Checked out the branch (both commits) and traced every call site of `invalidateBeastOwnerCache` across all 7 changed files plus `add-fighter.ts` (an importer of the same module not in the diff), and checked the DB schema for the `fighters.fighter_pet_id` FK to verify the new query logic.

**Did it break something? No — but one thing looked like a regression until I traced it through, worth recording:**

The five pre-existing copies of `invalidateBeastOwnerCache` weren't identical: the ones in `edit-fighter.ts` and `fighter-advancement.ts` additionally called `revalidateTag(TAGS.fighter(ownerId), { expire: 0 })` after `invalidateFighter(...)`, while the ones in `equipment.ts`, `move-to-stash.ts`, `move-from-stash.ts`, and `sell-equipment.ts` did not. The new unified helper (`utils/exotic-beasts.ts`) only calls `invalidateFighter`. I checked `utils/cache-tags.ts:75-78` — `invalidateFighter` already does `bust(TAGS.fighter(fighterId))` where `bust = (tag) => revalidateTag(tag, { expire: 0 })`, i.e. byte-for-byte the same call the two files had bolted on separately. So this consolidation removes a genuine duplicate, not a real second invalidation — no behavior change. (This redundancy is presumably the actual "duplication" the PR title refers to.)

**Checklist**

- **Correctness**:
  - `fighters.fighter_pet_id` is a FK to `fighter_exotic_beasts.id` (confirmed in `schema.public.sql:10317`), not the pet's own id — so `.eq('id', fighterPetId)` in the new helper is a valid direct PK lookup replacing the old `.eq('fighter_pet_id', fighterId)` scan. Both target the same row when the link is consistent.
  - The `fighterPetId === null` (skip) vs `=== undefined` (fall back to lookup) distinction is used correctly everywhere — `null` only ever means "row was read, confirmed not a pet," never "unknown."
  - `.single()` → `.maybeSingle()` fixes a real (if silent) issue: `.single()` returns a PostgREST error for every ordinary (non-beast) fighter — i.e. ~97% of calls — which was being swallowed by only destructuring `data`. `.maybeSingle()` returns `null` cleanly instead.
  - `deleteEquipmentFromFighter`'s guard (`equipmentBefore.fighter_id === params.fighter_id ? fighter.fighter_pet_id ?? null : undefined`) correctly avoids attaching a hint fetched for one fighter to the invalidation call for another.
  - `equipment.ts`'s new batched `fighterResult` query is gated on `params.fighter_id && !params.buy_for_gang_stash`, slightly broader than the old fighter-name-only fetch (`&& !params.custom_equipment_id && params.equipment_id`), so it now runs one extra `maybeSingle()` read for custom-equipment purchases where it's unused — harmless (parallelized with the gang query, not serial) but worth knowing it's not purely a subset of the old condition.
  - `'use server'` → `'server-only'` in `utils/exotic-beasts.ts` is safe: every importer is itself a `'use server'` action file (`app/actions/*.ts`), none are client components, so nothing was relying on these exports being individually-invocable Server Actions.
- **Dead code removal (verified)**: `canEquipmentCreateBeasts` has zero call sites repo-wide — confirmed dead before removal.
- **No stale references**: all 24 call sites of `invalidateBeastOwnerCache` pass a `fighterPetId` argument (or `undefined` implicitly where the row isn't at hand, e.g. none currently), matching the new signature.
- **Edge cases / null handling**: `ownerData?.fighter_owner_id` guard correctly handles a beast owned by no fighter (gang-stash beast) — no invalidation attempted, matches the added comment.
- **Security**: N/A — no auth/RLS changes, same tables/columns.
- **Performance**: Real win — skips a full-table-scan-by-`fighter_pet_id` lookup for ~97% of mutations (any fighter that isn't a beast), replacing it with a column already present on the row most callers now fetch anyway, plus the `equipment.ts` batching moves a serial fighter-name read into the opening parallel batch.
- **Readability**: Comments explain the non-obvious parts well (the FK direction, the `null` vs `undefined` contract, why `maybeSingle`).
- **Conventions**: No CLAUDE.md in this repo; nothing here conflicts with CONTRIBUTING.md. Note the PR template's "Type of change" / "How I tested" checklists are all unchecked/blank.
- **DRY/YAGNI**: This is the point of the PR — five divergent copies of the same helper collapsed into one in `utils/exotic-beasts.ts`, with the divergence (the redundant `revalidateTag`) resolved rather than papered over.

**Findings**

1. (Low) Empty test plan / unchecked "Type of change" boxes in the PR template — given this touches the invalidation path for every fighter/equipment mutation in the app (edit status, XP, advancements, equipment buy/sell/stash), a quick manual pass (buy equipment that grants a beast → confirm owner's total cost updates; edit a beast fighter's status → confirm owner's page reflects it) would be worth doing/noting before merge.

**Severity**: Low (process note only — no code defect found)

**Verdict**: ✅ Approve

---

Djidiouf (approved):

PR #2106, opened 2026-09-07
Add promoted_from_prospects tracker column

v0rp-z:
## Summary

Follow on PR from #2096 

Specialised fighters added to a gang, with a specialist UUID assigned nativity will bypass some of the new advancement deduction logic and cause them to not track their first advancement. To stop this from happening, we add a new `promoted_from_prospect` to track if the fighter has earned their UUID from a promotion or not, then we apply the appropriate logic.

Migration `20260907120000_add_promoted_from_prospect.sql` needs to be applied as part of the deploy. The column should exist in prod before the code that reads/writes it is live.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] local dev

https://github.com/user-attachments/assets/bd82484e-99df-4cb5-b187-79d425c78aeb

https://github.com/user-attachments/assets/160466cc-e499-4ab1-a81a-8db404a62ec7

## Database

### Migration status

## Risk / rollout

- medium, new db column, migration

### DIFF
--- a/app/actions/copy-fighter.ts
+++ b/app/actions/copy-fighter.ts
@@ -350,6 +350,7 @@ export async function copyFighter(params: CopyFighterParams): Promise<CopyFighte
       fighter_subtypes: sourceFighter.fighter_subtypes || [],
       fighter_specialisation: copySpecialisation.fighter_specialisation,
       fighter_specialisation_id: copySpecialisation.fighter_specialisation_id,
+      promoted_from_prospect: sourceFighter.promoted_from_prospect ?? false,
       fighter_variant: sourceFighter.fighter_variant,
       custom_fighter_type_id: sourceFighter.custom_fighter_type_id,
       fighter_gang_legacy_id: sourceFighter.fighter_gang_legacy_id,
--- a/app/actions/fighter-advancement.ts
+++ b/app/actions/fighter-advancement.ts
@@ -996,7 +996,7 @@ export async function applyN26ProspectPromotion(
       params.special_rules ??
       (Array.isArray(before.special_rules) ? before.special_rules : []);
 
-    return await applyKeepTypePromotionWithSkillGrant({
+    const promotionResult = await applyKeepTypePromotionWithSkillGrant({
       fighterId: params.fighter_id,
       before,
       newSubtypes,
@@ -1006,6 +1006,18 @@ export async function applyN26ProspectPromotion(
       skillId: skillResolved.skillId,
       creditsIncrease: N26_PROSPECT_PROMOTION_CREDITS,
     });
+
+    if (promotionResult.success) {
+      const { error: flagError } = await supabase
+        .from('fighters')
+        .update({ promoted_from_prospect: true })
+        .eq('id', params.fighter_id);
+      if (flagError) {
+        console.error('Failed to set promoted_from_prospect after Prospect promotion:', flagError);
+      }
+    }
+
+    return promotionResult;
   } catch (error) {
     console.error('Error applying N26 Prospect promotion:', error);
     return {
@@ -1455,6 +1467,15 @@ export async function deleteAdvancement(
                   ? 'Skill removed but Leader demotion failed — please restore Champion subtype in Edit Fighter.'
                   : 'Skill removed but Ganger demotion failed — please restore Ganger subtype in Edit Fighter.'),
           };
+        }
+        if (isProspectPromotionGrant) {
+          const { error: flagError } = await supabase
+            .from('fighters')
+            .update({ promoted_from_prospect: false })
+            .eq('id', params.fighter_id);
+          if (flagError) {
+            console.error('Failed to clear promoted_from_prospect on Prospect undo:', flagError);
+          }
         }
         demotedFighterDetails = {
           fighter_subtypes: demotedSubtypes,
--- a/app/actions/fighter-advancement.ts
+++ b/app/actions/fighter-advancement.ts
@@ -201,6 +201,7 @@ export interface AdvancementResult {
     fighter_specialisation: string | null;
     fighter_specialisation_id: string | null;
     special_rules: string[];
+    promoted_from_prospect?: boolean;
   };
   /** False when rating changed without a stash refund (e.g. Prospect promotion undo). */
   refund_stash?: boolean;
@@ -1233,7 +1234,8 @@ export async function deleteAdvancement(
       .select(`
         ${FIGHTER_WITH_EDITION_SELECT},
         fighter_subtypes, fighter_type, fighter_type_id,
-        fighter_specialisation, fighter_specialisation_id, special_rules
+        fighter_specialisation, fighter_specialisation_id, special_rules,
+        promoted_from_prospect
       `)
       .eq('id', params.fighter_id)
       .single();
@@ -1486,6 +1488,9 @@ export async function deleteAdvancement(
             ? null
             : (fighter.fighter_specialisation_id ?? null),
           special_rules: specialRules,
+          promoted_from_prospect: isProspectPromotionGrant
+            ? false
+            : (fighter.promoted_from_prospect ?? false),
         };
       }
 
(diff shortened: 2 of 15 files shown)


joeseos:
@v0rp-z Review of this PR.

## Findings

### 1. (High) Optimistic UI patch never sets `promoted_from_prospect`, so the fix doesn't take effect until page reload
`components/fighter/fighter-advancement-list.tsx:76-83` (patch type) and `components/fighter/fighter-page.tsx:941-972` (the `onFighterDetailsUpdate` handler that mutates local `fighterData.fighter` state).

`prospectPromotionConsumed` now derives from `promotedFromProspect`, which is only ever set from `fighterData.fighter?.promoted_from_prospect` fetched from the server (`fighter-page.tsx:940`). The `standalonePromotionMutation`'s `onMutate`/`onSuccess` handlers (`fighter-advancement-list.tsx:3010-3095`) call `onFighterDetailsUpdate?.(promotion)`, but `FighterPromotionResult`/the patch type carries `fighter_subtypes`, `fighter_specialisation(_id)`, `fighter_type(_id)`, `special_rules` — never `promoted_from_prospect`. And `fighter-page.tsx`'s patch handler doesn't touch that field either.

Net effect: right after a user runs the Prospect promotion, the client-side `promoted_from_prospect` stays `false` (its pre-promotion value), so `prospectPromotionConsumed` stays `false` and `openAdvancementsFor` doesn't deduct the slot in the UI. The user will see the stale "extra" open advancement until they reload the page — which is the exact symptom this PR sets out to fix. Previously (pre-PR) this worked immediately because the check used `fighter_specialisation_id`, which *was* included in the patch/optimistic update.

**Fix:** include `promoted_from_prospect: true` in the patch passed to `onFighterDetailsUpdate` in the `n26_prospect` promotion branch (and clear it symmetrically wherever the Prospect promotion is undone client-side), and thread it through `fighter-page.tsx`'s patch handler and the `onFighterDetailsUpdate` patch type.

### 2. (Medium) Migration backfill misses fighters already promoted past keep-type (Champion/Leader)
`supabase/migrations/20260907120000_add_promoted_from_prospect.sql:9-21`

The backfill only sets `promoted_from_prospect = TRUE` when the fighter's catalog `fighter_types.fighter_subtypes` still contains `"Prospect"`. But Champion→Leader promotion rewrites `fighter_type_id` to a Leader catalog row (per the removed comment in `keepTypePromotionN26.ts`: "Champion→Leader rewrites fighter_type_id"), and that Leader row won't carry the `Prospect` subtype — while `fighter_specialisation_id` is retained through that promotion. So any fighter who already went Prospect → Ganger+Specialist → Champion → Leader (or → Champion, if that also rewrites the row — worth double-checking) before this migration runs will *not* get backfilled to `true`.

This is a regression for exactly those rows: the old detection (`fighterSpecialisationId` matched against the eight Prospect specialisation ids, regardless of fighter type) correctly reported `prospectPromotionConsumed: true` for them; after this migration + deploy, they'll silently read as `false` and gain back an open advancement slot they shouldn't have.

**Fix:** broaden the backfill predicate to also catch fighters whose `fighter_specialisation_id` is one of the eight Prospect specialisation ids **and** current `fighters.fighter_subtypes` contains `Specialist` (the persistent marker of the original promotion, independent of any later type-change), not just fighters whose catalog type still shows `Prospect`.

### 3. (Low) New flag-setting updates swallow failures silently
`app/actions/fighter-advancement.ts:1010-1018` and `:1471-1479`

Both new `supabase.from('fighters').update({ promoted_from_prospect: ... })` calls only `console.error` on failure and continue — the overall promotion/undo is still reported as `success: true` to the caller even if the flag write fails. Since this flag is now the *sole* source of truth for `prospectPromotionConsumed` (the old multi-signal inference was deliberately removed), a failed write here silently reintroduces the exact bug this PR fixes, with no visible indication to the user or a way to detect/retry it. Given how central this column now is, consider at least a more prominent server-side alert (not just `console.error`), or making promotion fail loudly if the flag can't be persisted.

## Other notes
- `copy-fighter.ts:353` correctly preserves `promoted_from_prospect` on copy — good.
- The `fighter-basic-v4` → `v5` cache key bump in `fighter-data.ts:158` is appropriate for the new selected column.
- `keepTypePromotionN26.ts`'s simplified `hasN26ProspectPromotionOccurred` and updated docstring read well and clearly explain why the old inference broke.

## Verdict: Changes requested

The core migration/schema/type-plumbing work is solid, but finding #1 means the PR doesn't actually deliver the immediate-UX fix described (only fixes it after a reload), and finding #2 is a real data-correctness gap for any fighter that's already past keep-type promotion in prod. Both should be addressed before merge; #3 is worth a follow-up at minimum.


---

joeseos:
@v0rp-z Re-review of the latest commit (`eef155b1` "carry signal e2e", since the previous review).

## Previous findings

**#1 (High) — Optimistic UI patch never set `promoted_from_prospect` — Resolved.**
`fighter-promotion-modal.tsx:328` now sets `promoted_from_prospect: true` in the `n26_prospect` `onPromoted(...)` payload, which flows straight through `standalonePromotionMutation`'s `onMutate` → `onFighterDetailsUpdate?.(promotion)` (`fighter-advancement-list.tsx:3018`) → `fighter-page.tsx:952` merge (`patch.promoted_from_prospect ?? prev.fighter.promoted_from_prospect`). The undo path in `fighter-skills-list.tsx:538` symmetrically sets it to `false`. Verified the immediate-UX fix now works without a reload.

**#2 (Medium) — Migration backfill misses fighters already promoted past keep-type (Champion/Leader) — Outstanding.**
No changes to `supabase/migrations/20260907120000_add_promoted_from_prospect.sql` in this commit. The backfill predicate is unchanged and still only matches on catalog `fighter_subtypes @> '["Prospect"]'`, so fighters already rewritten to a Leader `fighter_type_id` still won't backfill to `true`. This is a one-time data-correctness gap that needs fixing before the migration runs in prod (or it needs to run once more, broadened).

**#3 (Low) — Flag-setting updates swallow failures silently — Outstanding.**
`app/actions/fighter-advancement.ts:1010-1018` and `:1471-1479` are untouched by this commit; still `console.error`-only on failure while returning `success: true`. Fine as a documented follow-up, as noted last time.

## New finding from this commit

**(Medium) Skill-delete rollback path loses `promoted_from_prospect` on a failed/false-positive Prospect undo**
`components/fighter/fighter-skills-list.tsx:525-530, 572-574, 606-608`

The optimistic delete handler sets `promoted_from_prospect: false` as soon as a Prospect-grant skill is deleted (line 538), but the `previousFighterDetails` snapshot used to roll back — both in the "server did not demote" branch (`onSuccess`, lines 565-578) and in `onError` (lines 604-609) — only restores `fighter_subtypes`, `fighter_specialisation`, and `fighter_specialisation_id`. It never restores `promoted_from_prospect`.

Concretely: if a user deletes the Prospect-promotion skill grant and the server rejects the demotion (a false positive, or a network/DB error), the subtypes/specialisation snap back to promoted correctly, but `promoted_from_prospect` stays stuck at the optimistic `false` — even though the fighter is still actually promoted in the DB. The fighter will show a spurious extra open Advancement slot until the page is reloaded, which is exactly the class of bug this PR exists to fix.

**Fix:** capture `promoted_from_prospect: promotedFromProspect` (or the resolved current value) in `previousFighterDetails` alongside the other three fields, and include it in both rollback call sites.

## Edge cases / conventions
No other issues spotted in the new lines — the type threading (`AdvancementResult['fighter_details']`, patch prop types across `fighter-advancement-list.tsx`/`fighter-skills-list.tsx`/`fighter-page.tsx`) is consistent, and non-Prospect promotion patches correctly omit the field so the `??` fallback preserves the existing value.

## Verdict: Changes requested

Finding #1 from the previous review is fixed. #2 and #3 remain outstanding (as before). The new rollback gap above should be fixed before merge since it reintroduces a stale-flag bug in a code path this PR specifically targets.

---

joeseos (approved):

PR #2108, opened 2026-09-07
Updated seed.sql with n26 data

joeseos:
## Summary

- Updated the seed.sql with n26 data

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- Haven't ran the seed on local, but should be good

## Database

### Migration status

## Risk / rollout

- Does not effect production and can be rolled back if needed

### DIFF
--- a/supabase/seed.sql
+++ b/supabase/seed.sql
@@ -23,17 +23,26 @@ ON CONFLICT (slug) DO NOTHING;
 -- ============================================================================
 -- 1. EQUIPMENT CATEGORIES
 -- ============================================================================
-INSERT INTO public.equipment_categories (id, category_name, created_at) VALUES
-('0d689aac-81cf-4a7d-81a0-953f2f5d6e47', 'Ammunition', now()),
-('9a3f7568-2e70-4d67-97cc-f5c9f21ac753', 'Armour', now()),
-('eafb69c2-3d86-44da-95a3-e815cdc63b3c', 'Basic Weapons', now()),
-('aceb626f-259d-45b2-8a36-0d9c7369969f', 'Close Combat Weapons', now()),
-('931fe7d2-4913-4cbc-b5d0-d34ef0865815', 'Grenades', now()),
-('2f58f3cb-d5c5-4620-86e6-a91e6428ca34', 'Heavy Weapons', now()),
-('18a867b0-42cb-42bb-b3a6-330fa3e65700', 'Personal Equipment', now()),
-('8e6fe32b-d70d-48a5-95c0-00441b502ae5', 'Pistols', now()),
-('1e3528d0-2064-4766-a23b-62b39ead07f4', 'Special Weapons', now()),
-('ad9d7b9b-7cea-48dc-9278-45e3e47a1aad', 'Weapon Accessories', now())
+-- Categories are edition-scoped and the same name may repeat across editions, so
+-- clients key on id rather than category_name.
+INSERT INTO public.equipment_categories (id, category_name, edition_id, created_at) VALUES
+('0d689aac-81cf-4a7d-81a0-953f2f5d6e47', 'Ammunition', '00000000-0000-0000-0000-000000000023', now()),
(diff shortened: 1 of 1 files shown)


joeseos:
## Review

The PR description flags the one open risk itself: "Haven't ran the seed on local, but should be good." I ran it, for real, against the actual current schema — not just read through it.

**Method:** spun up a throwaway local Postgres 16, loaded `supabase/schema/schema.public.sql` from current `main` (stubbing `auth.uid()`/`auth.role()`, `private.is_admin()`/`is_arb()`, `extensions.uuid_generate_v4()`, and an `auth.users` table — same setup this repo's own `docs/refactor-fighter-type-parents.md` describes as the way to rehearse a migration for free), then ran this PR's `supabase/seed.sql` head-to-tail.

**Result: it runs clean.** `BEGIN` → 40 `INSERT`s → `COMMIT`, zero errors, zero warnings.

I then queried the loaded data to check the specific correctness claims the PR's comments make, rather than taking them on faith:

- **No cross-edition equipment leak.** This is the important one. Before this PR, the four N23 `fighter_type_equipment` rules matched on subtype/name alone with no edition filter — and since `'Leader'`/`'Autogun'` etc. exist in both editions, they'd have silently also matched the new N26 House Escher fighter types, granting them N23-catalog equipment IDs alongside the N26-catalog ones from the new blocks. The PR's `AND ft.edition_id = ... AND e.edition_id = ...` additions on all four N23 rules are the right fix. Verified by joining the actual result: **139 rows are n23-fighter × n23-equipment, 26 are n26 × n26, zero rows cross editions, zero duplicate `(fighter_type_id, equipment_id)` pairs.**
- **`fighter_type_equipment_fighter_scope_uidx` (`NULLS NOT DISTINCT`) is real** — added in `20260822120000_add_variant_and_subtype_scope_to_fighter_type_equipment.sql` and still present (as `gang_subtype_id` post-rename) on current `main`. The PR's comment citing it is accurate, and the new N26 blocks don't collide with it (verified above — no duplicates).
- **`weapon_profiles` NOT NULL columns**: `acc_short`, `acc_long`, `damage`, `ammo` are `text NOT NULL` with no default on the real table. The four N26 profiles do supply `'-'` for all of them, confirmed by querying the inserted rows back.
- **N26 characteristic range**: `Ld/Cl/Wil/Int` are `[4, 10]` in `N26_FIGHTER_LIMITS` (`utils/characteristicLimits.ts` on `main`) — the seeded House Escher values (5–8) all fall inside that range.
- **`fighter_specialisation_id` FK**: `Sister (Specialist)` → `Sniper` resolves correctly against the production `fighter_specialisations` id used.
- **`fighter_effect_type_modifiers` signs**: the five N26 rows came back exactly as the inline comment describes (Leadership/Strength/Impressive-Scars-leadership `+1`, Weapon Skill/Save `-1`).
- **`equipment_categories` name reuse** ('Personal Equipment' in both editions) is safe — confirmed there's only a plain (non-unique) index on `category_name`, not a constraint, on current `main`.
- **`gang_types` `ON CONFLICT (gang_type_id)`** correctly targets the real `gang_types_gang_type_id_key` unique constraint (separate from the bigint `id` PK).

Nothing here needed my own judgment call — I checked it against the schema and got back actual query results, not inference. I have no blocking findings.

Two very minor, non-blocking observations:
- `equipment_categories` gains an `edition_id` column in this PR's first hunk but the comment explaining why ("clients key on id rather than category_name") is a little terse — fine as-is, just flagging in case a future reader wants more.
- The N26 `fighter_type_equipment` grants for House Escher are hand-picked illustrative examples (as the PR's own framing implies throughout — "illustrative, the rulebook numbers are not in this repo"), so this is dev/test fixture data, not a claim of rulebook accuracy. Worth keeping in mind if anyone later mistakes these for verified stats.

**Verdict: Approve.** Ran and verified end-to-end against the current schema; no correctness, FK, or constraint issues found.

cc @joeseos

---

v0rp-z (approved):

PR #2111, opened 2026-09-08
Refac/added fighter type availability

joeseos:
## Summary

- Added table fighter_type_availability to enable granting end excluding specific fighter types based on gang subtype (gang origin etc added as well for future support). No admin support atm, will be a follow up PR. Also backfills needed data as logic moved to db instead of in the api route.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- Will be tested in preview

## Database

### Migration status

## Risk / rollout

- None. New table, so production will not be affected by the migration or the RPC

### DIFF
--- a/app/api/fighter-types/route.ts
+++ b/app/api/fighter-types/route.ts
@@ -6,10 +6,14 @@ import { withEditionSlug } from '@/types/edition';
 
 type SupabaseServerClient = Awaited<ReturnType<typeof createClient>>;
 
-const subtypeGangTypeName = (name: string) => `Subtype: ${name}`;
+type GangSubtype = { id: string; subtype: string; edition_id: string | null };
 
-// Keyed by name so a later edition's re-issue of the subtype inherits the rule.
-const subtypesWithoutLeaders = new Set(['secundan incursion']);
+// The gang properties fighter_type_availability scopes against.
+type GangScope = {
+  gangTypeId: string | null;
+  gangOriginId: string | null;
+  gangSubtypes: GangSubtype[];
+};
 
 // Fetch the user's own custom fighters plus any shared with them through campaigns,
 // de-duplicated by id.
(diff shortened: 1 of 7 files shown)


joeseos:
@joeseos Review of this PR.

## Findings

### 1. (Low/verify) Deny-seed matches `gang_subtype_types.subtype` by exact-case literal string
`supabase/migrations/20260908120000_create_fighter_type_availability.sql` (the deny `INSERT ... VALUES (('Secundan Incursion', 'n23', 'Leader'), ('Malstrain Corrupted', 'n26', 'Brute'))` block, joined via `gst.subtype = v.subtype`)

This join is case- and spelling-exact against a hardcoded literal, with `ON CONFLICT DO NOTHING` and no row-count assertion — if the stored catalog value doesn't match byte-for-byte, the two deny rows silently insert 0 rows (no error). Notably, this very codebase already hit this exact class of bug once: the last commit's message describes 'Genestealer Infected' vs 'Genestealer Corrupted' being spelled differently between the gang-subtype catalog row and its fighter pool, silently breaking a `'Subtype: ' || subtype` join. Given that history, and that the commit messages show you validated this against real data (nice — the exact row/fighter counts in the commit messages are reassuring), it'd still be worth adding a cheap post-insert sanity check (e.g. `GET DIAGNOSTICS` row count + `RAISE EXCEPTION` if either literal matches zero rows) so a future rename of either subtype fails the migration loudly instead of quietly reverting this rule to "no deny."

### 2. (Info) Confirm the N26 "Malstrain Corrupted → deny Brute" rule is an intentional new rule, not just a port
Same migration file, and `app/api/fighter-types/route.ts` diff (removed `subtypesWithoutLeaders` set).

The pre-PR code only special-cased one deny: Secundan Incursion → remove `Leader`-subtype fighters. The Malstrain Corrupted → deny `Brute` rule is new — there's no equivalent in the code being replaced. Your commit messages make clear this is deliberate ("a gang holding Malstrain Corrupted may not add its own gang type's Brutes") and tested, so this isn't a bug — just flagging that it's a live gameplay-affecting behavior change for existing N26 Malstrain Corrupted gangs that isn't mentioned in the PR summary, worth a line there for anyone else reviewing/testing in preview.

### 3. (Nit) `is_gang_subtype` only recognizes subtype-scoped grants
`supabase/functions/get_fighter_types_with_cost.sql` (`granted` CTE / `is_gang_subtype` output column) and `app/api/fighter-types/route.ts`'s `hiddenGangTypes` filter (`fighter.is_gang_subtype || !hiddenIds.has(fighter.gang_type_id)`).

`is_gang_subtype` is only set when `gst.subtype_name` is non-null, i.e. only for grants scoped by `gang_subtype_id`. The schema explicitly supports `gang_origin_id`/`gang_type_id`-scoped grants too ("future support" per the table comment). If one of those is added later and happens to point at a fighter whose `gang_type_id` is itself a hidden pool type, it would get filtered out by the `hiddenGangTypes` check in the route despite being granted. Not an issue today since only subtype grants are seeded, but worth a `// TODO` or note for whoever adds origin/type-scoped grants.

## What looks solid
- The `rules`/`granted` CTE split in `get_fighter_types_with_cost.sql`, the `NOT EXISTS` deny anchored to `ft.gang_type_id = v_gang_type_id` (correctly scoping denies to the gang's own type even in the include-all call), and the `OR g.fighter_type_id IS NOT NULL` grant-survives-deny logic are all well-reasoned and clearly commented.
- Confirmed the `isGangAddition` branch in `route.ts` does **not** pass `p_gang_id`, so gang-addition (hired guns/hangers-on) fetches correctly stay unaffected by the new availability table — matches prior behavior of never mixing in subtype-pool fighters there.
- `fighter_type_availability_target_chk`/`_scope_chk` constraints correctly enforce the grant-must-name-a-fighter-type and scope-must-be-non-empty invariants described in the comments.
- Function arity cleanup (`DROP FUNCTION IF EXISTS ... (uuid, uuid, boolean, uuid)`) correctly avoids a PostgREST ambiguous-overload situation.
- The edition-aware `gangSubtypeRank` split (and its accompanying bug fix for the Genestealer Corrupted disappearing-from-picker regression) is a good, well-explained fix riding along in this PR.
- RLS policies and grants on the new table follow the existing `private.is_admin()` admin-write / authenticated-read pattern used elsewhere.

## Verdict: Approve

No blocking issues found — the migration and RPC logic are careful and the commit history shows real verification against data. Item #1 is worth a small follow-up for defense-in-depth, and #2 is just a suggestion to note the behavior change in the PR description before merge/preview testing.


---

v0rp-z (approved):

PR #2112, opened 2026-09-09
fix(equipment): stop writing null fighter_owner_id to fighter_exotic_beasts

joeseos:
## Summary

- Buying beast-granting wargear into the gang stash inserted a fighter_exotic_beasts row with a null fighter_owner_id, which RLS rejects: NULL IN (...) is NULL, not TRUE. Since fb9c4db9 that 403 threw before updateGangFinancials and logEquipmentAction could run, and the block's own rollback deleted from the legacy gang_stash table instead of fighter_equipment, so it matched nothing. The result was unpaid, unlogged stash rows that broke the rating + credits + stash = wealth invariant.

move-to-stash cleared the owner the same way. That write was rejected too, and its result went unchecked, so it was a silent no-op for every non-admin. Nothing depended on it: beast visibility comes from fighter_equipment.gang_stash via beast_equipment_stashed.

Restores pre-ab6cdb94 behaviour, in which stash-bought beast wargear grants no beast.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Bought a pet from the stash, moved it to a fighter, moved it back to the stash, then sold it. Also bought a pet to a fighter, pet fighter card shows and then sold the pet from the fighter. Gang rating and wealth checks out.

## Risk / rollout

- None

### DIFF
--- a/app/actions/equipment.ts
+++ b/app/actions/equipment.ts
@@ -809,29 +809,6 @@ export async function buyEquipmentForFighter(params: BuyEquipmentParams): Promis
       }
     }
 
-    // Handle beast creation for STASH equipment purchases
-    if (params.buy_for_gang_stash && !params.custom_equipment_id && params.equipment_id) {
-      try {
-        const beastResult = await createExoticBeastsForEquipment({
-          equipmentId: params.equipment_id,
-          ownerFighterId: null,  // No owner for stash
-          ownerFighterName: null,
-          gangId: params.gang_id,
-          userId: gang.user_id,
-          fighterEquipmentId: newEquipmentId
-        });
-
-        if (beastResult.success && beastResult.createdBeasts.length > 0) {
-          createdBeasts = beastResult.createdBeasts;
-          // Note: Don't add to rating since equipment is in stash
-        }
-      } catch (error) {
-        // Clean up the stash equipment that was just inserted since the beast it grants failed to create
-        await supabase.from('gang_stash').delete().eq('id', newEquipmentId);
-        throw error;
-      }
-    }
-
     // Calculate deltas for gang updates
     const totalRatingDelta = ratingDelta + createdBeastsRatingDelta + grantsRatingDelta;
     const stashValueDelta = params.buy_for_gang_stash ? ratingCost : 0;
--- a/app/actions/move-to-stash.ts
+++ b/app/actions/move-to-stash.ts
@@ -233,13 +233,6 @@ export async function moveEquipmentToStash(params: MoveToStashParams): Promise<M
       }
     }
 
-    // Clear beast owner AFTER successful stash update (prevents data loss if stash update fails)
-    // Keep the fighter_exotic_beasts row (tracks beast_equipment_stashed via fighter_equipment.gang_stash)
-    await supabase
-      .from('fighter_exotic_beasts')
-      .update({ fighter_owner_id: null })
-      .eq('fighter_equipment_id', params.fighter_equipment_id);
-
     // Rating delta: subtract equipment purchase_cost, effects, and beast equipment cost
     // BUT only if the fighter is active (not killed, retired, enslaved, or captured)
     // Inactive fighters are already excluded from rating calculations
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -220,7 +220,7 @@ export const getFighterTotalCost = async (fighterId: string, supabase: any): Pro
       .eq('fighter_id', fighterId),
     supabase
       .from('fighter_exotic_beasts')
-      .select('fighter_pet_id')
+      .select('fighter_pet_id, fighter_equipment!fighter_equipment_id (gang_stash)')
       .eq('fighter_owner_id', fighterId)
   ]);
 
@@ -244,7 +244,10 @@ export const getFighterTotalCost = async (fighterId: string, supabase: any): Pro
 
   // Owned exotic beasts roll their cost into the owner
   let beastsCost = 0;
-  const beastIds = (beastLinksRes.data || []).map((b: any) => b.fighter_pet_id);
+  // A stashed beast's cost sits in the gang stash, not on its owner.
+  const beastIds = (beastLinksRes.data || [])
+    .filter((b: any) => !b.fighter_equipment?.gang_stash)
+    .map((b: any) => b.fighter_pet_id);
   if (beastIds.length > 0) {
     const { data: beastData } = await supabase
       .from('fighters')
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -244,7 +244,7 @@ export const getFighterTotalCost = async (fighterId: string, supabase: any): Pro
 
   // Owned exotic beasts roll their cost into the owner
   let beastsCost = 0;
-  // A stashed beast's cost sits in the gang stash, not on its owner.
+  // Skip beasts created by equipment that is in the stash.
   const beastIds = (beastLinksRes.data || [])
     .filter((b: any) => !b.fighter_equipment?.gang_stash)
     .map((b: any) => b.fighter_pet_id);
--- a/app/lib/shared/fighter-data.ts
+++ b/app/lib/shared/fighter-data.ts
@@ -244,10 +244,11 @@ export const getFighterTotalCost = async (fighterId: string, supabase: any): Pro
 
   // Owned exotic beasts roll their cost into the owner
   let beastsCost = 0;
-  // Skip beasts created by equipment that is in the stash.
-  const beastIds = (beastLinksRes.data || [])
-    .filter((b: any) => !b.fighter_equipment?.gang_stash)
-    .map((b: any) => b.fighter_pet_id);
+  const beastLinks = beastLinksRes.data || [];
+  const stashedBeastIds = new Set(
+    beastLinks.filter((b: any) => b.fighter_equipment?.gang_stash).map((b: any) => b.fighter_pet_id)
+  );
+  const beastIds = beastLinks.map((b: any) => b.fighter_pet_id);
   if (beastIds.length > 0) {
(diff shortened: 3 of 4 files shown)


joeseos:
@joeseos Review of this PR.

## Verified root cause
Checked both claims against the actual RLS policies in `schema.public.sql`:
- `fighter_exotic_beasts` INSERT policy: `WITH CHECK (is_admin() OR fighter_owner_id IN (SELECT f.id FROM fighters WHERE f.user_id = auth.uid()))`. With `fighter_owner_id: null`, `NULL IN (...)` evaluates to `NULL`, not `TRUE`, so this correctly 403s for non-admins on the stash-purchase path — confirmed.
- `fighter_exotic_beasts` UPDATE policy also carries a `WITH CHECK`, not just `USING`, on the same predicate — so `move-to-stash`'s `.update({ fighter_owner_id: null })` would fail that check too for non-admins, and since the result wasn't inspected, it was a swallowed error, not a thrown one. Confirmed.
- The stash-purchase rollback deleting from `gang_stash` (a genuinely separate legacy table, its own `id` sequence, unrelated to `fighter_equipment.id`/`newEquipmentId`) instead of `fighter_equipment` — confirmed by checking `gang_stash`'s schema; that `.delete().eq('id', newEquipmentId)` really would match nothing.

Also confirmed the untouched fighter-purchase beast-creation path (`equipment.ts:785`, gated on `params.fighter_id && !params.buy_for_gang_stash`) is a separate branch from the one removed, so this doesn't touch beast creation for equipment bought directly onto a fighter — only the stash-purchase and move-to-stash paths are affected, as intended.

And confirmed the persisted rating/wealth math in `move-to-stash.ts` (`ratingDelta -= beastEquipmentCost`, `equipmentValue += beastEquipmentCost`) is computed independently of the `fighter_owner_id` clear that's being removed — so gang rating/wealth stay correct either way, matching your test notes.

## Finding: fighter-card total cost can still double-count a stashed beast (Medium)
`utils/gang-assembly.ts:445-466` (`beastCosts` calc) and `app/lib/shared/gang-data.ts:1080-1098` (`beastsOwned` query)

`beastCosts` sums every row in `ownedBeasts` (= `fighter_exotic_beasts` rows grouped by `fighter_owner_id`) whose beast fighter isn't killed/retired/enslaved/captured — it never checks whether the beast's granting equipment is stashed. The `beastsOwned` query already joins `fighter_equipment!fighter_equipment_id (gang_stash, ...)`, so the data needed is right there, just unused in the cost sum.

Before this PR, that didn't matter for non-admins (the `fighter_owner_id` clear silently failed anyway, so the row was already left dangling), but it did matter for **admins**: `is_admin()` bypasses the `WITH CHECK`, so admin-driven moves-to-stash previously *did* null `fighter_owner_id` successfully, which dropped the beast out of `beastsByOwner[fighterId]` and correctly removed its cost from the original owner's displayed total. After this PR, `fighter_owner_id` is never cleared for anyone (admin included), so an admin who moves beast-granting wargear to stash will now see the original fighter's card keep showing the beast's cost, even though the equipment (and implicitly the beast) is sitting in the gang stash.

This is a real, reproducible display regression for the admin path specifically (not a rating/wealth correctness issue — that math is independently correct per above). Since the fix data is already fetched, it seems low-cost to close this properly rather than trade one silent gap for a slightly different one: filter `ownedBeasts` (or the `beastCosts` reduce) on `!beastRel.fighter_equipment?.gang_stash` instead of relying on `fighter_owner_id` being nulled. Might be worth a fast follow-up rather than blocking this PR, since it's strictly an improvement over current prod behavior for everyone except admins.

## Everything else
The rest of the diff is a clean, well-justified deletion — no other issues found. The commit message's root-cause narrative checks out against the schema, and scope is appropriately narrow (doesn't touch the fighter-purchase beast path, doesn't touch `beast_equipment_stashed`'s actual source of truth for visibility).

## Verdict: Approve

The fix itself is correct and narrowly scoped to the reported bug. The fighter-card cost display gap above is worth a follow-up (or could be folded into this PR if you'd rather land it together), but it doesn't make anything worse for the vast majority of users and isn't a blocker.


---

joeseos:
@joeseos Re-review of the 3 commits pushed since the last review (`c7557ff`, `eea4238`, `1dc2a57`).

## Previous finding

**(Medium) Fighter-card cost double-counts a stashed beast for admins — Resolved, and generalized.**
`c7557ff` filters `ownedBeasts`/`myBeastLinks` on `!fighter_equipment?.gang_stash` in `utils/gang-assembly.ts`, exactly as suggested. `eea4238` goes further and fixes the same gap in `getFighterTotalCost` (`app/lib/shared/fighter-data.ts`), which turned out to be pre-existing since `ab6cdb94` and *not* admin-specific — the query there always matched stashed beasts for regular users too, so `killedRatingDelta`/`edit-fighter`'s active/inactive deltas were double-subtracting the beast's equipment cost whenever an owner with a stashed beast was killed or retired. Checked `move-to-stash.ts:236` — the null-owner write is confirmed gone from the current head, and `beastEquipmentCost` there sums the beast's own `fighter_equipment.purchase_cost`, matching what these two fixes now exclude. Good catch beyond the original ask.

## New finding: filtering drops the whole beast, not just its equipment cost (Medium)

`utils/gang-assembly.ts:446-465` (`beastCosts` reduce, fed by the `ownedBeasts` filter at line 205) and `app/lib/shared/fighter-data.ts:248-269` (`beastIds` filter feeding `beastsCost`)

Both fixes exclude a beast **entirely** from the owner's cost once its granting equipment is stashed — base `fighter_types.cost`, its own equipment, skills, effects, *and* `cost_adjustment` all drop out together, because the filter happens before the beast's data is even fetched (`beastIds` in `fighter-data.ts`; `ownedBeasts` in `gang-assembly.ts`).

But `move-to-stash.ts:236-260` only ever moves the beast's **equipment** cost into stash value (`beastEquipmentCost`) — it never touches skills, effects, or `cost_adjustment` on the beast. And beasts can carry all three: `edit-fighter.ts:1043-1044` has an explicit comment — *"Exotic Beast Advancements (skills/effects credits_increase + cost_adjustment) are intentionally NOT refunded"* — confirming beast advancements/cost_adjustment are a real, distinct, already-handled case elsewhere in the codebase, not merely hypothetical.

Concretely: a beast with a nonzero `cost_adjustment` (or a skill, or an injury effect) whose granting wargear is moved to stash will have that value vanish from the owner's rating with nothing added to stash — the exact `rating + credits + stash = wealth` invariant this whole PR exists to protect, broken again, just via a narrower door.

**Fix:** don't drop the beast row from the sum when stashed — still fetch/iterate it, but zero out only its own equipment-cost component (mirroring `move-to-stash.ts`'s `beastEquipmentCost`), keeping base cost, skills cost, effects cost, and `cost_adjustment` in the total. In `gang-assembly.ts` this means checking `beastRel.fighter_equipment?.gang_stash` per-beast inside the `beastCosts` reduce (zeroing `equipmentCost` there) rather than filtering `ownedBeasts` upstream; in `fighter-data.ts`, don't exclude stashed beasts from `beastIds` — instead zero their `fighter_equipment` contribution inside the `beastsCost` reduce.

Given the previous fix already shipped a real correctness improvement for the common case, and this gap is a narrower edge case (a beast with cost_adjustment/skills/effects specifically while its wargear sits in stash), it's reasonable to land as a fast follow-up rather than block here — your call given how the last "follow-up" note played out.

## Comment nit (non-blocking)
`1dc2a57` reworded the three new comments from "A stashed beast's cost sits in the gang stash, not on its owner's card" to "Skip beasts created by equipment that is in the stash." The new wording describes *what* the line does (redundant with the code) rather than *why* — the old phrasing was more useful. Not blocking.

## Verdict: Changes requested

Good, verified fix for the original finding — and a bonus fix for a pre-existing non-admin bug. The new gap above is real but narrow; flagging as changes-requested since it's the same invariant this PR is about, but it's reasonable to fold into a follow-up if you'd rather ship the bigger fix now.

---

joeseos:
@joeseos Re-review of `8ebedeb` ("Fixed exotic beasts costs"), the one commit pushed since the last review.

## Previous finding

**(Medium) Filtering dropped the whole beast, not just its equipment cost — Resolved.**

This commit implements exactly the suggested fix, consistently across all three sites:

- `utils/gang-assembly.ts:204` — `ownedBeasts` is back to unfiltered; the `beastCosts` reduce (`:454-457`) now zeroes only `equipmentCost` when `beastRel.fighter_equipment?.gang_stash` is true, leaving `baseBeastCost`, `skillsCost`, `effectsCost`, and `cost_adjustment` in the sum unconditionally.
- `utils/gang-assembly.ts:1275` (`assembleFighterView`) — same pattern: `myBeastLinks` unfiltered, `equipmentCost` conditionally zeroed per-beast (`:1289-1291`) rather than dropping the beast from `byEquipmentId`/`beastTotal` entirely.
- `app/lib/shared/fighter-data.ts:246-278` (`getFighterTotalCost`) — reworked to fetch all `beastIds` (not just non-stashed ones) and track `stashedBeastIds` as a `Set`, then zero only the `beastEquipment` component in the reduce (`:270-272`) while `beastSkills`, `beastEffects`, `baseBeastCost`, and `beast.cost_adjustment` still contribute.

Verified this now actually matches what `move-to-stash.ts` moves into stash value (equipment cost only), so a beast with `cost_adjustment`/skills/effects keeps contributing to the owner's rating while its granting wargear sits in stash — the wealth invariant holds for that edge case too. Also verified the `killed`/`retired`/`enslaved`/`captured` short-circuit for the beast fighter itself is untouched and orthogonal to this change.

The new comments ("Stashed equipment already moved its cost to stash value; the beast's advancements did not.") also address the earlier wording nit — they explain the *why* now.

No new issues spotted in this diff.

## Verdict: Approve

All findings from both prior rounds are resolved. This looks correct and complete.

---

v0rp-z (approved):

PR #2113, opened 2026-09-09
Fixed patreon webhook

joeseos:
## Summary

- Fix to get the patreon webhook working again.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Added preview env varible for the patreon webhook that's needed to test
- Update: tested in preview and now getting 200

## Risk / rollout

- No risk as it's currently not working

### DIFF
--- a/app/api/patreon/sync/route.ts
+++ b/app/api/patreon/sync/route.ts
@@ -1,7 +1,7 @@
 import { invalidateUser, invalidatePatreonSupporters } from '@/utils/cache-tags';
 import { NextRequest, NextResponse } from 'next/server';
 import { createClient } from '@supabase/supabase-js';
-import { createClient as createAuthClient } from "@/utils/supabase/server";
+import { createClient as createAuthClient, createServiceRoleClient, findAuthUserIdByEmail } from "@/utils/supabase/server";
 import { checkAdmin } from "@/utils/auth";
 
 /**
@@ -53,22 +53,6 @@ interface DatabaseUserData {
   tierTitle: string | null;
   tierId: string | null;
   discordRoles: string[] | null;
-}
-
-/**
- * Create service role Supabase client for admin operations
- */
-function createServiceRoleClient() {
-  return createClient(
-    process.env.NEXT_PUBLIC_SUPABASE_URL!,
-    process.env.SUPABASE_SERVICE_ROLE_KEY!,
-    {
-      auth: {
-        autoRefreshToken: false,
-        persistSession: false
-      }
-    }
-  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


joeseos:
@joeseos Review of this PR.

## Root cause check
The old `matchPatreonToUser` paginated through **every** Supabase auth user (`listUsers({ page, perPage: 1000 })` in a `while` loop) on **every single webhook call**, just to find one email match. That's a very plausible cause of the webhook silently failing/timing out as the user base grew — good catch, and the fix (a single GoTrue admin REST call with server-side `filter=<email>`) is the right direction.

Also verified:
- `profiles.patreon_user_id` does have a `UNIQUE` constraint (`profiles_patreon_user_id_key` in `schema.public.sql`) — so the docstring's "unique index" claim and the `.maybeSingle()` usage are both accurate/safe.
- Checking `patreon_user_id` before email is a genuine correctness improvement, not just perf: the old code tried email first, so a patron who changed their Patreon-linked email would misbehave (fail to match, or in a multi-tenant edge case even risk matching a different `profiles` row) instead of resolving via the stable, already-linked id.
- `processMemberDeletion` reuses `matchPatreonToUser('', patreonUserId)` — passing `''` for email correctly short-circuits the email path (`if (!patreonEmail) return null`), preserving the original patreon_user_id-only lookup semantics for delete events. Good reuse, not just a copy-paste.
- The `filter` param is a substring match per your comment, and the code correctly re-verifies the exact lowercased email against the returned candidates before accepting a match — right call, since a naive "take the first result" would risk mismatches.

## Finding: non-2xx GoTrue response is treated the same as "no user found" (Medium)
`app/api/patreon/webhook/route.ts:129-132`

```ts
if (!response.ok) {
  console.error(`Webhook user lookup failed: ${response.status} ${response.statusText}`);
  return null;
}
```

A `fetch` that *throws* (DNS failure, network error, timeout) correctly propagates up through `processPatreonWebhook`/`processMemberDeletion` to the top-level `try/catch` in `POST`, which returns a 500 — good, Patreon will retry. But a `fetch` that *completes* with a non-2xx status (e.g. the service key is rejected, GoTrue 5xx, rate limiting) returns `null` here, which the caller treats identically to "no matching user" — logging "No matching user found" and returning `success: true` (200). Since Patreon only retries on non-2xx, a transient auth/GoTrue failure here would be silently swallowed and never retried, and that patron's status update would just be lost. Given this route's whole purpose is landing dependably now, it'd be worth throwing (or otherwise signaling failure distinctly from "not found") on `!response.ok`, symmetric with how thrown exceptions are already handled correctly.

## Minor note (non-blocking)
`app/api/patreon/sync/route.ts` has its own separate, still-unfixed `matchPatreonToUser`/`getAllSupabaseUsers` that also does the full paginated `listUsers` walk. It's architecturally less bad there (cached once per sync run and reused across all patrons, rather than re-paginated per webhook call), so it's probably not urgent, but since it's the same underlying pattern that just bit you in the webhook, it might be worth the same `filter`-based lookup treatment at some point, and/or consolidating both into one shared helper so this fix doesn't have to be re-learned there too.

## Verdict: Approve

The core fix is sound, verified against the schema, and you've already confirmed 200s in preview. The `!response.ok` handling above is worth tightening, but I wouldn't block merge on it given the current state is "completely broken" and this is a strict improvement.


---

PR #2114, opened 2026-09-09
Add admin management for official gang types.

Djidiouf:
## Summary
- Add a Gangs tile on the Admin Dashboard to create and edit official ``gang_types`` rows (name, edition, alignment, hidden, affiliations, trading post, origin, parent/variant, default images).
- Replace the unused fighter-type POST on ``/api/admin/gang-types`` with validated create/update, lean GET for other admin screens, and ``?detailed=1`` for this modal.
- Make default-image credits optional and share credit attribution between create-gang and the image editor.

## Type of change
- [x] Feature

## How I tested
- [x] Opened Admin Dashboard → Gangs, created a gang type, then edited it
- [x] Confirmed default-image previews, optional credit fields, and that blanking a middle image URL is refused
- [x] Confirmed create-gang still lists types and shows credit lines

## Database
Not applicable — no migration.

## Risk / rollout
Low — admin-only catalog UI. Deletes are not exposed. ``global-gang-types`` cache is revalidated on create/update.

### DIFF
--- a/app/admin/page.tsx
+++ b/app/admin/page.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { LuUsers, LuSword, LuCar, LuBookOpen, LuScrollText, LuBookUser, LuHeartCrack, LuSearch, LuBell, LuHandshake } from "react-icons/lu";
+import { LuUsers, LuSword, LuCar, LuBookOpen, LuScrollText, LuBookUser, LuHeartCrack, LuSearch, LuBell, LuHandshake, LuHouse } from "react-icons/lu";
 import { LuChartColumn } from "react-icons/lu";
 import { PiFlagBannerFoldBold } from "react-icons/pi";
 import { LuSquarePen } from 'react-icons/lu';
@@ -18,6 +18,7 @@ import { AdminStatsModal } from "@/components/admin/admin-stats-modal";
 import { AdminScenariosModal } from "@/components/admin/admin-scenarios-modal";
 import { AdminInjuriesGlitchesModal } from "@/components/admin/admin-injuries";
 import { AdminAlliancesModal } from "@/components/admin/admin-alliances";
+import { AdminGangTypesModal } from "@/components/admin/admin-gang-types";
 import { AdminCampaignManagementModal } from "@/components/admin/admin-campaign-management";
 import { AdminSupportToolsModal } from "@/components/admin/admin-support-tools";
 import { AdminNotificationsModal } from "@/components/admin/admin-notifications-modal";
@@ -36,6 +37,7 @@ export default function AdminPage() {
   const [showScenarios, setShowScenarios] = useState(false);
   const [showInjuriesGlitches, setShowInjuriesGlitches] = useState(false);
   const [showAlliances, setShowAlliances] = useState(false);
+  const [showGangTypes, setShowGangTypes] = useState(false);
   const [showCampaignManagement, setShowCampaignManagement] = useState(false);
   const [showSupportTools, setShowSupportTools] = useState(false);
   const [showNotifications, setShowNotifications] = useState(false);
@@ -118,6 +120,12 @@ export default function AdminPage() {
       description: "Manage campaign types, territories, and triumphs",
       action: () => setShowCampaignManagement(true),
       icon: PiFlagBannerFoldBold
+    },
+    {
+      title: "Gangs",
+      description: "Manage gang types",
+      action: () => setShowGangTypes(true),
+      icon: LuHouse
     }
   ];
 
@@ -286,6 +294,12 @@ export default function AdminPage() {
         {showAlliances && (
           <AdminAlliancesModal
             onClose={() => setShowAlliances(false)}
+          />
+        )}
+
+        {showGangTypes && (
+          <AdminGangTypesModal
+            onClose={() => setShowGangTypes(false)}
           />
         )}
 
--- a/app/api/admin/gang-origin-categories/route.ts
+++ b/app/api/admin/gang-origin-categories/route.ts
@@ -1,0 +1,28 @@
+import { NextResponse } from 'next/server'
+import { createClient } from "@/utils/supabase/server";
+import { checkAdmin } from "@/utils/auth";
+
+export async function GET() {
+  const supabase = await createClient();
+
+  try {
+    const isAdmin = await checkAdmin(supabase);
+    if (!isAdmin) {
+      return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+    }
+
+    const { data, error } = await supabase
+      .from('gang_origin_categories')
+      .select('id, category_name')
+      .order('category_name');
+
+    if (error) throw error;
+    return NextResponse.json(data);
+  } catch (error) {
+    console.error('Error fetching gang origin categories:', error);
+    return NextResponse.json(
+      { error: 'Failed to fetch gang origin categories' },
+      { status: 500 }
+    );
+  }
+}
--- a/app/api/admin/gang-types/route.ts
+++ b/app/api/admin/gang-types/route.ts
@@ -17,6 +17,13 @@ const ALLOWED_ALIGNMENTS = [
   'Unaligned',
 ] as const;
 
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+type DefaultImagePayload = {
+  url: string;
+  credit?: { name?: string; url?: string; suffix?: string };
+};
+
 type GangTypePayload = {
   gang_type: string;
   edition_id: string;
@@ -26,10 +33,32 @@ type GangTypePayload = {
   trading_post_type_id: string | null;
   gang_origin_category_id: string | null;
   parent_gang_type_id: string | null;
-  default_image_urls: Array<{
-    url: string;
-    credit?: { name?: string; url?: string; suffix?: string };
-  }> | null;
+  default_image_urls: DefaultImagePayload[] | null;
+};
+
+type GangTypePatch = {
+  gang_type?: string;
+  edition_id?: string;
+  alignment?: string | null;
+  is_hidden?: boolean;
+  affiliation?: boolean;
+  trading_post_type_id?: string | null;
+  gang_origin_category_id?: string | null;
+  parent_gang_type_id?: string | null;
+  default_image_urls?: DefaultImagePayload[] | null;
+};
+
+type GangTypeRow = {
+  gang_type_id: string;
+  gang_type: string | null;
+  alignment: string | null;
+  is_hidden: boolean | null;
+  affiliation: boolean | null;
+  trading_post_type_id: string | null;
+  gang_origin_category_id: string | null;
+  parent_gang_type_id: string | null;
+  default_image_urls: unknown;
+  edition_id: string | null;
 };
 
 function emptyToNull(value: unknown): string | null {
@@ -37,6 +66,33 @@ function emptyToNull(value: unknown): string | null {
   if (typeof value !== 'string') return null;
   const trimmed = value.trim();
   return trimmed === '' ? null : trimmed;
+}
+
+function postgresCode(error: unknown): string | undefined {
+  if (!error || typeof error !== 'object' || !('code' in error)) return undefined;
+  const code = (error as { code?: unknown }).code;
+  return typeof code === 'string' ? code : undefined;
+}
+
+function mutationErrorResponse(error: unknown, fallback: string): NextResponse {
+  const code = postgresCode(error);
+  if (code === '22P02') {
+    return NextResponse.json({ error: 'Invalid id format' }, { status: 400 });
+  }
+  console.error(fallback, error);
+  return NextResponse.json({ error: fallback }, { status: 500 });
+}
+
+function parseUuid(
+  value: unknown,
+  field: string
+): { error: string } | { data: string | null } {
+  const trimmed = emptyToNull(value);
+  if (trimmed === null) return { data: null };
+  if (!UUID_RE.test(trimmed)) {
+    return { error: `${field} must be a valid UUID` };
+  }
+  return { data: trimmed };
 }
 
 function parseOptionalBoolean(
@@ -53,9 +109,42 @@ function parseOptionalBoolean(
   return { error: `${field} must be a boolean` };
 }
 
+function parseBooleanField(
+  value: unknown,
+  field: string
+): { error: string } | { data: boolean } {
+  if (typeof value === 'boolean') {
+    return { data: value };
+  }
+  return { error: `${field} must be a boolean` };
+}
+
+function parseAlignment(
+  value: unknown
+): { error: string } | { data: string | null } {
+  const alignment = emptyToNull(value);
+  if (alignment && !(ALLOWED_ALIGNMENTS as readonly string[]).includes(alignment)) {
+    return { error: 'Invalid alignment' };
+  }
+  return { data: alignment };
+}
+
+function parseGangTypeName(
+  value: unknown
+): { error: string } | { data: string } {
+  const gangType = emptyToNull(value);
+  if (!gangType) {
+    return { error: 'gang_type is required' };
+  }
+  if (gangType.length > 200) {
+    return { error: 'gang_type must be 200 characters or less' };
+  }
+  return { data: gangType };
+}
+
 function parseDefaultImageUrls(value: unknown):
   | { error: string }
-  | { data: GangTypePayload['default_image_urls'] } {
+  | { data: DefaultImagePayload[] | null } {
   if (value === undefined || value === null || value === '') {
     return { data: null };
   }
@@ -64,12 +153,14 @@ function parseDefaultImageUrls(value: unknown):
     return { error: 'default_image_urls must be an array or null' };
   }
 
-  const parsed: NonNullable<GangTypePayload['default_image_urls']> = [];
+  const parsed: DefaultImagePayload[] = [];
 
   for (const entry of value) {
     if (typeof entry === 'string') {
       const url = entry.trim();
-      if (!url) continue;
+      if (!url) {
+        return { error: 'default_image_urls entries require a valid http(s) image url' };
+      }
       if (!isValidHttpUrl(url)) {
         return { error: 'default_image_urls entries must be valid http(s) URLs' };
       }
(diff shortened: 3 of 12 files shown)


joeseos:
@Djidiouf Reviewed this PR against the current `main`. Overall this is a solid, well-scoped admin feature — input validation on the new route is thorough (URL validation, alignment enum, edition-scoped reference checks for trading post / parent gang type run in parallel), the fighter-type dead code was cleanly removed, and the credit-line de-duplication across `create-gang-modal.tsx` / `image-edit-modal.tsx` is a nice cleanup.

## Findings

**1. (Medium) No server-side guard against reassigning `edition_id` on an existing, in-use gang type**
`app/api/admin/gang-types/route.ts` — `_PATCH`

A gang's effective edition is derived entirely from its `gang_type_id` → `gang_types.edition_id` (see `gangEditionJoin`/`gangEditionSlug` in `types/edition.ts`), and `gangs` has no edition column of its own. `_PATCH` validates `edition_id` exists and that `trading_post_type_id`/`parent_gang_type_id` are consistent with the *new* edition, but nothing stops an admin request from changing `edition_id` on a gang type that already has live `gangs`/`fighter_types`/`tactics_cards_packs` rows — that would silently reclassify every existing gang of that type into a different edition (different fighter costs, equipment lists, tactics decks, etc.).

In practice the modal's `handleEditionChange` mitigates this in the intended UI flow — switching the edition dropdown while an existing type is selected just deselects it and resets the form rather than sending a changed `edition_id` — but that's a client-side behavior, not an API guarantee. A hand-crafted PATCH (or a future UI tweak) can still do it.

Suggested fix: in `_PATCH`, reject (or require explicit confirmation for) an `edition_id` change on a gang type that already has associated `gangs`/`fighter_types` rows, or simply make `edition_id` immutable after creation.

**2. (Low) "Origin Category" options are derived from existing origins, not all categories**
`components/admin/admin-gang-types.tsx` — `originCategories` (around line 223)

The dropdown is built by de-duplicating `category_id`/`category_name` out of the `gang_origins` list fetched from `/api/admin/gang-origins`, since there's no dedicated categories endpoint. A `gang_origin_categories` row with zero origins currently assigned to it can never show up here, so such a category is effectively unselectable for a gang type until at least one origin exists under it. Minor completeness gap, not a regression.

**3. (Low) Duplicated URL-validation logic between client and server**
`isPreviewableUrl` in `components/admin/admin-gang-types.tsx` vs `isValidHttpUrl` in `app/api/admin/gang-types/route.ts` implement the same http(s)-URL check independently. Not urgent, but could drift over time — worth sharing if you touch this area again.

**4. (Low) `parseBoolean` silently defaults invalid input to `false`**
`app/api/admin/gang-types/route.ts` — `parseBoolean`

Any non-boolean value for `is_hidden`/`affiliation` (e.g. a stray string) is coerced to `false` instead of raising a 400. Only reachable from this PR's own trusted frontend today, so low risk, but it can mask a bug quietly if that ever changes.

## Checklist
- Correctness: no crashing bugs found; edition-scoped validation (trading post / origin / parent gang type) looks correct against the schema's FKs and check constraints.
- Edge cases / null handling: good — `emptyToNull`, `normaliseDefaultImageUrls`, and the optional `DefaultImageCredit` fields are handled consistently across the touched files.
- Security: admin check (`checkAdmin`) is applied consistently on GET/POST/PATCH; no injection surface (Supabase query builder throughout).
- Performance: reference checks run via `Promise.all`, no N+1 loops.
- Readability/maintainability: good — validation is centralized in named helper functions.
- Conventions (CONTRIBUTING.md): no schema changes needed here since only existing columns are used; consistent with the rest of `app/api/admin/*`.
- DRY/YAGNI: only the minor URL-validation duplication noted above (#3).

## Verdict
**Approve** — the medium finding (#1) is worth a follow-up given it's a data-integrity foot-gun rather than a functional bug in the current UI flow, but it doesn't block merging this admin-only feature.

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`3fc353e`) against the previous review's findings. All four items from the last review are addressed, and I didn't find any new issues in the changes.

## Previous findings

1. **(Medium) No server-side guard against reassigning `edition_id` on an in-use gang type** — **Resolved.** `_PATCH` in `app/api/admin/gang-types/route.ts` now fetches the existing `edition_id` and calls a new `assertEditionChangeAllowed`, which blocks the edition change (409) whenever the gang type still has associated `gangs`, `fighter_types`, `tactics_cards_packs`, or child `gang_types` (via `parent_gang_type_id`) rows. Checked the underlying schema — `gangs.gang_type_id`, `fighter_types.gang_type_id`, and `tactics_cards_packs.gang_type_id` all exist and are queried correctly with `count: 'exact', head: true`.

2. **(Low) "Origin Category" options derived from existing origins, not all categories** — **Resolved.** New `GET /api/admin/gang-origin-categories` route added, reading directly from `gang_origin_categories` (`id, category_name`), and `admin-gang-types.tsx` now sources `originCategories` from this endpoint instead of de-duplicating from the origins list. The old `/api/admin/gang-origins` route is left in place and is still used by other admin components (vehicle type, skill, equipment editors), so no orphaned code.

3. **(Low) Duplicated URL-validation logic between client and server** — **Resolved.** Both sides now import a single `isValidHttpUrl` from the new `utils/http-url.ts`; the old server-only `isValidHttpUrl` and client-only `isPreviewableUrl` were removed.

4. **(Low) `parseBoolean` silently defaults invalid input to `false`** — **Resolved.** Replaced with `parseOptionalBoolean`, which returns `{ error }` for any non-boolean, non-nullish value (400 via the existing validation path) and only applies the default when the field is `undefined`/`null`. Consistent with the `{error} | {data}` pattern already used by the other validators in this file.

## New findings

None. The new `gang-origin-categories` route follows the same `checkAdmin` + try/catch + `NextResponse.json` conventions as the rest of `app/api/admin/*`. The edition-change guard runs before `assertEditionScopedRefs`, uses `Promise.all` (no N+1), and correctly treats "no change" (`currentEditionId === nextEditionId`) as a no-op without extra queries. No null-handling or readability concerns in the diff.

## Verdict

**Approve.**

---

joeseos (changes requested):
High

  - Removing a default image repoints ~102k gang portraits —
    components/admin/admin-gang-types.tsx:576. gangs.default_gang_image
    is a positional index into default_image_urls; Remove shifts every
    later entry down. 105,557 of 113,455 gangs are pinned to an index,
    102,015 above 0. No remap, no warning, no undo.
  - Rename leaves stale denormalized copies, and one gates behavior —
    app/api/admin/gang-types/route.ts:479. Nothing propagates a rename to
    gangs.gang_type / fighter_types.gang_type. isVenatorGang compares
    that copy by string, so renaming "Venators" keeps skill access for
    existing gangs and silently denies it to new ones. 19 gangs and 192
    fighter types already drift.
  - Dashboard cache never invalidated by gang-type edits — route.ts:360.
    getUserGangs caches gang-type images under user/gang tags with
    revalidate: false; only globalGangTypes() is revalidated, so home
    cards stay stale indefinitely.

Medium

  - PATCH is a full-row replace — route.ts:446. Omitted fields are
    written as defaults, not left alone. Latent: the modal always sends
    everything.
  - No unique index on (gang_type, edition_id) — route.ts:424. Duplicate
    names in one edition are creatable, and name-based lookups exist.
    Real constraint failures also collapse into a generic 500.
  - Blank string entry silently reindexes the image array — route.ts:72.
    Same positional corruption as above, but silent, while the object
    path returns a 400 for identical input. Only reachable by a
    hand-crafted request.

Low

  - Malformed UUID returns an opaque 500 — route.ts:200. No format check;
    PostgREST's 22P02 escapes to the catch-all. Admin-only.
  - credit.url reaches href without the PR's own validator —
    components/ui/default-image-credit-line.tsx:24. Inert, not XSS (React
    19 blocks javascript:), but isValidHttpUrl is imported two files
    over.
  - Edition-change guard is broader than needed — route.ts:305. Three of
    the four counted references already ON UPDATE CASCADE; only the gangs
    check is load-bearing.
  - assertParentGangType duplicates DB trigger/CHECK/FK rules —
    route.ts:243. Error strings are byte-identical to the trigger's;
    costs two round-trips and can drift. Does buy a clean 400 over a 500.
  - Redundant rawCredit branch — route.ts:104. Behaviorally equivalent to
    the creditFields computation above it.
  - Latent: NULL-edition gang types would be unreachable —
    admin-gang-types.tsx:203. Total lockout if such a row exists, but 0
    of 57 do today.
  - Mutations in API routes — route.ts:430. Diverges from CLAUDE.md, but
    pre-existing; this PR extends the pattern.

joeseos:
@Djidiouf Re-reviewed the latest commit (`c6b88a2`) against the previous review (`3fc353e`). This round adds partial-PATCH support, "no edition" gang types, denormalized name syncing to `gangs`/`fighter_types`, and remapping of gangs' pinned `default_gang_image` index when `default_image_urls` changes.

## Previous findings

1. **(Medium) No server-side guard against reassigning `edition_id` on an in-use gang type** — **Regressed (partial).** See new finding #1 below: the guard still exists but was narrowed and now misses two of the three cases it used to catch.
2. **(Low) "Origin Category" options derived from existing origins** — still **Resolved**, untouched by this commit.
3. **(Low) Duplicated URL-validation logic** — still **Resolved**, untouched by this commit.
4. **(Low) `parseBoolean` silently defaults invalid input to `false`** — still **Resolved** (`parseOptionalBoolean`/`parseBooleanField` unchanged in this commit).

## New findings

**1. (Medium) Regression — `assertEditionChangeAllowed` no longer checks `fighter_types`, `tactics_cards_packs`, or child `gang_types`**
`app/api/admin/gang-types/route.ts:502-522`

The previous commit's fix for finding #1 checked `gangs`, `fighter_types`, `tactics_cards_packs`, and child `gang_types` (via `parent_gang_type_id`) before allowing an `edition_id` change. This commit rewrote the function to query only `gangs`:

```ts
const { count: gangCount, error: gangError } = await supabase
  .from('gangs')
  .select('id', { count: 'exact', head: true })
  .eq('gang_type_id', gangTypeId);
...
if ((gangCount ?? 0) > 0) {
  return 'Cannot change edition_id while gangs still reference this gang type';
}
```
`fighter_types.gang_type_id` and `tactics_cards_packs.gang_type_id` are still live FK columns (confirmed in `schema.public.sql` / recent migrations), and both still tie game rules to a specific edition. As written, an admin can reassign `edition_id` on a gang type that has zero `gangs` rows but existing `fighter_types` or `tactics_cards_packs` rows (e.g. a type that was created, given fighter types/tactics packs, but never used to make a gang yet), silently reclassifying that catalog data into a different edition. The child-`gang_types` (variant) check is also gone, so a parent type with existing variants can have its edition changed without any guard, even though `assertParentGangType` requires parent/child to share an edition. This looks like an unintentional narrowing rather than a deliberate scope reduction — worth restoring the other three checks (or explaining why they're no longer needed, e.g. if `fighter_types`/`tactics_cards_packs` rows can no longer exist without a `gangs` row of that type).

**2. (Medium) `buildDefaultImageIndexMap`'s same-length shortcut mis-remaps pinned portraits on a remove+add edit**
`app/api/admin/gang-types/route.ts:227-251` (used by `remapPinnedDefaultImages`, `app/api/admin/gang-types/route.ts:546-601`)

```ts
if (oldUrls.length === newUrls.length) {
  oldUrls.forEach((_, index) => map.set(index, index));
  return map;
}
```
This assumes that whenever the array length is unchanged, positions are unchanged too. But the admin UI lets you both remove an image (`Remove` button, splices at index) and add one (`Add Image`, appends) in the same edit before saving. E.g. old `[A, B, C]`, remove `B` → `[A, C]`, then add `D` → final `[A, C, D]` — same length (3) as the original, but index 1 now holds `C` instead of `B`, and index 2 holds `D` instead of `C`. Because the lengths match, the function takes the identity-mapping fast path and skips content-based matching entirely, so a gang with `default_gang_image = 1` (pinned to the old `B`) silently ends up pinned to `C` instead, with no error and no indication anything changed. This directly contradicts the new UI copy ("Removing an image remaps later portraits and cannot be undone") — the remap doesn't actually happen correctly in this case. Given the function already has content-based matching logic for the length-changed case, the same matching (by URL, left-to-right, ignoring the length short-circuit) would give correct results here too — recommend removing the `oldUrls.length === newUrls.length` fast path, or otherwise flagging this known limitation in the code comment.

**3. (Low) Editing a "no edition" gang type silently claims it for the currently-filtered edition**
`components/admin/admin-gang-types.tsx:243`

```ts
setEditionId(gangType.edition_id || editionId);
```
When an admin selects a gang type with no `edition_id` for editing, the edition field is pre-filled with whatever edition happens to be selected in the page's filter dropdown (not left blank), and since `edition_id` is a required field on submit, saving *any* edit to that gang type (even an unrelated field, like fixing a typo in the name) will assign it to that edition. This is probably intentional given `edition_id` must be set to save at all, but it happens without an explicit "assign edition" action, and if the admin has the wrong edition selected in the filter, an edit could file the type under the wrong edition. Worth confirming this is the intended flow, or adding a small UI affordance (e.g. highlighting/requiring the edition field explicitly) when editing a previously edition-less type.

**4. (Low) `remapPinnedDefaultImages` error rollback isn't fully atomic**
`app/api/admin/gang-types/route.ts:588-598`

The catch block resets every entry in `changes` from its negative sentinel back to `from`, but if the failure happened partway through phase 2, some entries have already been successfully moved to `to` (and no longer match `-(from+1)`), so the "rollback" is a no-op for those and they stay migrated while others revert — leaving a mixed state rather than either fully applied or fully reverted. Low likelihood (requires a mid-loop DB failure) and not a data-loss risk beyond what a retry would need to sort out, but worth a comment noting this is best-effort, not transactional, if that's accepted.

## Checklist
- Edge cases / null handling: mostly solid (empty `default_image_urls` array clearing all pins, brand-new gang type with no prior images, "no edition" trading-post/parent-type reference checks) — verified these behave correctly; the two gaps above are the exceptions.
- Readability: good, the new helpers are small and the non-obvious `buildDefaultImageIndexMap` invariant is documented in a comment.
- Conventions: consistent with the rest of `app/api/admin/*` (checkAdmin, try/catch, `{error}|{data}` validators).

## Verdict

**Changes requested** — primarily for #1, since it's a regression of a previously-fixed data-integrity guard rather than a new limitation, and #2, since it can silently corrupt a gang's stored portrait selection with no error surfaced.

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`bbda3a0`) against the previous review's findings (`c6b88a2`).

## Previous findings (from the `c6b88a2` review)

1. **(Medium) Regression — `assertEditionChangeAllowed` only checked `gangs`** — **Resolved.** The function now runs `gangs`, `fighter_types`, `tactics_cards_packs`, and child `gang_types` (`parent_gang_type_id`) checks in `Promise.all`, matching the originally-intended guard. Verified the four queries against the schema's FK columns — all correct, no N+1.
2. **(Medium) `buildDefaultImageIndexMap`'s same-length shortcut mis-remapped pinned portraits on a remove+add edit** — **Resolved.** The `oldUrls.length === newUrls.length` fast path is gone; the function now always does left-to-right URL matching. Traced the `[A,B,C] → [A,C,D]` case by hand: old index 1 (`B`, removed) now correctly maps to `null` and old index 2 (`C`) maps to new index 1, instead of the previous identity-mapping bug that silently repinned `B`'s slot to `C`.
3. **(Low) Editing a "no edition" gang type silently claimed it for the filtered edition** — **Resolved.** `admin-gang-types.tsx` now shows an explicit amber warning ("This gang type has no edition. Saving will assign it...") when `willAssignEdition` is true, computed correctly from `!isCreateMode && selectedGangType && !selectedGangType.edition_id && editionId`.
4. **(Low) `remapPinnedDefaultImages` rollback isn't fully atomic** — **Resolved (documented).** A comment now explains the best-effort, non-transactional rollback behavior at the catch block, as suggested.

## New findings

None. Both fixes are minimal and correctly scoped — no new bugs, no null-handling gaps, no convention drift from the rest of `app/api/admin/*`.

## Still outstanding (from the earlier full review, unrelated to this commit's diff)

Not blocking this round, but worth tracking since they haven't been touched since the first full review:
- **(High)** Dashboard cache is still not invalidated on gang-type edits — only `TAGS.globalGangTypes()` is revalidated in `route.ts`; `getUserGangs`'s user/gang-tagged cache (used for home cards) is untouched.
- **(Medium)** No unique index/constraint on `(gang_type, edition_id)` — duplicate names in one edition remain creatable.

## Verdict

**Approve** — the two regressions flagged last round are correctly fixed and no new issues were introduced by this commit. The two items above are pre-existing gaps, not regressions from this change, and can be follow-ups.

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`695df45`) against the previous review's findings (`bbda3a0`). This commit only touches cache invalidation: `app/api/admin/gang-types/route.ts`, `app/lib/get-user-gangs.ts`, `utils/cache-tags.ts`.

## Previous findings

1. **(High) Dashboard cache never invalidated by gang-type edits** — **Resolved, but already fixed before this commit.** `TAGS.globalGangTypes()` was added to `getUserGangs`'s tag list back in `c6b88a2`, and `route.ts` already called `revalidateTag(TAGS.globalGangTypes(), ...)` on every successful POST/PATCH — so the dashboard cache was in fact wired correctly as of `bbda3a0`, despite the last review comment listing it as still outstanding. This commit doesn't fix a live bug; see new finding below.
2. **(High) Rename leaves stale denormalized copies** — still **Resolved** (`syncDenormalizedGangTypeName` in `c6b88a2`), untouched by this commit.
3. **(High) Removing a default image repoints gang portraits** — still **Resolved** (`remapPinnedDefaultImages` in `bbda3a0`), untouched by this commit.
4. **(Medium) PATCH is a full-row replace** — still **Resolved** (partial-PATCH support in `c6b88a2`), untouched by this commit.
5. **(Medium) No unique index on `(gang_type, edition_id)`** — still **Outstanding**, untouched by this commit.

Other previously-noted Low items (blank-string image reindex, malformed UUID 500, `credit.url` validator import, edition-change guard scope, `assertParentGangType` duplicating DB checks, redundant `rawCredit` branch, NULL-edition unreachable, mutations in API routes) are all untouched by this diff and remain as before.

## New findings

**1. (Low) The new `gang-type-{id}` tag is currently a no-op** — `utils/cache-tags.ts:185-188`, `app/lib/get-user-gangs.ts:187`

```ts
export function invalidateGangTypesCatalog(gangTypeId?: string) {
  bust(TAGS.globalGangTypes());
  if (gangTypeId) bust(TAGS.gangType(gangTypeId));
}
```
Every call site (`withReferenceInvalidation` in `route.ts`) busts `globalGangTypes()` unconditionally in the same call, so the new per-type tag never gets to matter — any write that busts `gang-type-{id}` also busts the global tag at the same instant, for every cache entry that would care. The comment in `get-user-gangs.ts` ("scoped to a type so an Escher edit does not wait on the global tag alone") describes a scenario that can't currently happen. Not a bug, but as written it's dead weight — either make `invalidateGangTypesCatalog` actually skip the global bust when a `gangTypeId` is known (the real point of scoping), or drop the per-type tag/plumbing until that's needed.

No other issues: the `response.clone().json().catch(() => null)` guard is safe, `body.gang_type_id` is reliably present on both POST and PATCH success responses (`GANG_TYPE_COLUMNS` includes it), cache keys were correctly bumped (`v2→v3`, `v3→v4`) for the changed `getUserGangIdsCached`/`getUserGangs` return/tag shapes, and the old `TAGS`/`revalidateTag` imports were cleanly removed from `route.ts` with no leftover usages.

## Verdict

**Approve** — no bugs or regressions in this commit. Finding #1 above is a minor, non-blocking cleanup item (the invalidation is still correct today, just wired more elaborately than it currently needs to be).

---

joeseos:
@Djidiouf Re-reviewed the latest commit (`9060356`) against the previous review's findings (`695df45`). This commit only touches cache-tag plumbing: `app/api/admin/gang-types/route.ts`, `app/lib/get-user-gangs.ts`, `utils/cache-tags.ts`.

## Previous findings

1. **(High) Dashboard cache never invalidated by gang-type edits** — still **Resolved**, untouched by this commit.
2. **(High) Rename leaves stale denormalized copies** — still **Resolved**, untouched by this commit.
3. **(High) Removing a default image repoints gang portraits** — still **Resolved**, untouched by this commit.
4. **(Medium) PATCH is a full-row replace** — still **Resolved**, untouched by this commit.
5. **(Medium) No unique index on `(gang_type, edition_id)`** — still **Outstanding**, untouched by this commit.
6. **(Low) The new `gang-type-{id}` tag was a no-op** — **Resolved.** This was the exact issue flagged last round: every write busted `globalGangTypes()` unconditionally, so the per-type tag never got a chance to matter. This commit removes it entirely rather than wiring it up — `TAGS.gangType`, `invalidateGangTypesCatalog(gangTypeId?)`'s optional param, and the `gangTypeIdsForTags` computation in `getUserGangs` are all gone, matching the "drop the per-type tag/plumbing until it's needed" option from the earlier review. `withReferenceInvalidation` no longer even needs to parse the response body to get `gang_type_id`, which is a nice simplification. Confirmed no other call site still references `TAGS.gangType(...)` or `invalidateGangTypesCatalog(id)`.

Other previously-noted Low items (blank-string image reindex, malformed UUID 500, `credit.url` validator import, edition-change guard scope, `assertParentGangType` duplicating DB checks, redundant `rawCredit` branch, NULL-edition unreachable, mutations in API routes) are all untouched by this diff and remain as before.

## New findings

None. Traced the two cache-key bumps and confirmed both are correct and necessary:
- `getUserGangIdsCached`'s return type narrowed from `{id, gang_type_id}[]` to `string[]` and its Supabase `select` narrowed from `'id, gang_type_id'` to `'id'` — cache key bumped `v3→v4` to avoid deserializing a stale shape; slightly cheaper query as a bonus.
- `getUserGangs`'s tag list dropped the per-gang-type tags — cache key bumped `v4→v5` for the same reason.

Verified `getUserGangIdsCached` and `getUserGangs` have no other call sites (`getUserGangs` is only used in `app/page.tsx`, whose usage is unaffected since the exported `Promise<Gang[]>` signature didn't change). The net cache-invalidation *behavior* is unchanged from `695df45` — `invalidateGangTypesCatalog()` still unconditionally busts `globalGangTypes()` on every successful admin write — this commit only removes dead plumbing that was never actually gated on the per-type tag.

## Verdict

**Approve** — this is a clean, low-risk cleanup with no functional changes and no new issues. The one remaining outstanding item (#5, unique index on `(gang_type, edition_id)`) is a pre-existing gap, not a regression, and can be a follow-up.

---

joeseos (approved):

PR #2115, opened 2026-09-09
Removed redundant API call for exotic beast

joeseos:
## Summary

-`getGangFightersBundle` queried `fighter_exotic_beasts` twice — once filtered by owner, once by pet. A beast's owner is always in the beast's own gang, so the owner-direction query already returns every row the pet-direction one did. Verified across all 23,331 links in production: zero are reachable only via the pet direction. Also drops two `equipment_name` embeds and the `id` column, none of which have a consumer since the `owned_beasts` payload was removed.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Data loads as it should

## Risk / rollout

- None

### DIFF
--- a/app/lib/shared/gang-data.ts
+++ b/app/lib/shared/gang-data.ts
@@ -789,8 +789,8 @@ export const getGangCampaigns = async (gangId: string, supabase: any): Promise<G
  *
  * Stage 1 (parallel): fighters, ALL gang vehicles (both keyed by gang_id)
  * Stage 2 (parallel): equipment (fighter + vehicle rows in one query),
- *   skills, effects (fighter + vehicle scopes in one query), exotic beasts
- *   (both directions), loadouts (+assignments embedded), captured-by names
+ *   skills, effects (fighter + vehicle scopes in one query), exotic beasts,
+ *   loadouts (+assignments embedded), captured-by names
  */
 export const getGangFightersBundle = async (gangId: string, supabase: any): Promise<GangFightersBundle> => {
   return unstable_cache(
--- a/app/lib/shared/gang-data.ts
+++ b/app/lib/shared/gang-data.ts
@@ -909,8 +909,7 @@ export const getGangFightersBundle = async (gangId: string, supabase: any): Prom
         equipment: [],
         skills: [],
         effects: [],
-        beastsOwned: [],
-        beastsPetOf: [],
+        beastLinks: [],
         loadouts: [],
         capturedByGangs: []
       };
@@ -1042,8 +1041,7 @@ export const getGangFightersBundle = async (gangId: string, supabase: any): Prom
         equipmentRes,
         skillsRes,
         effectsRes,
-        beastsOwnedRes,
-        beastsPetOfRes,
+        beastLinksRes,
         loadoutsRes,
         capturedByGangsRes
       ] = await Promise.all([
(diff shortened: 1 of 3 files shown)


joeseos:
Reviewed against `main`. The dedup logic itself checks out — I verified the invariant independently rather than just trusting the PR description:

- `fighters` for a `getGangFightersBundle(gangId)` call are always scoped by `gang_id` (capture only sets `captured`/`captured_by_gang_id` flags, it never moves a fighter's `gang_id`), so an owner and its pet are structurally guaranteed to be fetched via the same `fighterIds` set. Filtering `fighter_exotic_beasts` by `fighter_owner_id IN fighterIds` alone can't miss a pet-direction row.
- Confirmed the dropped `id` column and the two `equipment_name` embeds (`equipment!equipment_id`, `custom_equipment!custom_equipment_id`) have no reader anywhere in `utils/gang-assembly.ts` — only `beastRel.fighter_equipment?.gang_stash` is read off a beast link.
- The `beastsOwned`/`beastsPetOf` → `beastLinks` rename is complete; grepped the whole repo and only the 3 changed files reference those field names.

## Findings

**1. (High) Cache key not bumped for a payload-shape change on a `revalidate: false` cache**
`app/lib/shared/gang-data.ts` — `getGangFightersBundle`, cache key `` `gang-fighters-bundle-v4-${gangId}` ``

This function is wrapped in `unstable_cache(..., [`gang-fighters-bundle-v4-${gangId}`], { tags: [TAGS.gang(gangId)], revalidate: false })`. That cache never expires on its own — it's only rebuilt when `TAGS.gang(gangId)` is explicitly busted by a mutation on that gang. This PR changes the cached payload's shape (`beastsOwned`/`beastsPetOf` → single `beastLinks` field), but every gang whose cache entry was populated *before* this deploys and hasn't had a mutation since will keep serving the old-shaped JSON (`{ beastsOwned: [...], beastsPetOf: [...] }`, no `beastLinks` key) to the new code.

That's not a graceful-degradation case — `groupBy(bundle.beastLinks, 'fighter_owner_id')` and `bundle.beastLinks.forEach(...)` in `utils/gang-assembly.ts` (`assembleGangFighters`), plus `bundle.beastLinks.filter(...)` / `.find(...)` in `assembleFighterView`, all call array methods directly on `bundle.beastLinks` with no null guard. On `undefined` that's an immediate `TypeError`, which breaks the gang page and fighter page entirely for that gang until an unrelated mutation happens to bust its tag.

This is exactly the scenario the project already has a convention for — see `b8059cf` ("perf(venator): fold venator_no_ranks into getGangCore bundle"): *"Cache key bumped to v4 so old entries repopulate with the new field."* And `1bb2cf6` explicitly documents the exception: a version bump is safe to skip only when a stale entry degrades harmlessly (there, a missing status reads as an existing safe default and any mutation already rebuilds it). Here there's no such fallback — the field is unconditionally dereferenced.

Fix: bump the key to `` `gang-fighters-bundle-v5-${gangId}` ``.

## Checklist
- Correctness: solid once the cache key is bumped — see finding #1, which I'd treat as blocking given it's a guaranteed crash on stale-but-common cache entries, not a rare edge case.
- Edge cases / null handling: the `fighterIds.length > 0 ? ... : Promise.resolve({ data: [] })` guard is preserved correctly; the removed query's `fighter_pet_id` filter had no other special-casing to carry over.
- Security: no change to auth/query construction beyond column selection.
- Performance: genuine win — one round trip instead of two for every gang load.
- Readability/maintainability: cleaner; single source of truth for beast links matches the pattern already used in `app/lib/shared/fighter-data.ts`.
- Conventions: violates the project's own documented cache-versioning convention (see #1) — otherwise consistent with surrounding code.
- DRY/YAGNI: good — this removes the actual duplication rather than adding an abstraction over it.

## Verdict
**Changes requested** — please bump the `gang-fighters-bundle` cache key version before merging; otherwise this looks correct and is a good simplification.

---

joeseos:
The key already rotated. unstable_cache builds fixedKey as `${cb.toString()}-${keyParts.join(',')}` (unstablecache.js:55), and this PR changes the callback body — a removed query and a changed return literal — so every gang gets a fresh entry on first read after deploy. Old v4-shaped payloads are unreachable, not served. A v5 suffix would bump a key that has already changed. The failure mode the review routine describes is real for a shape change made outside the cached callback (in a helper it calls), where cb.toString() is unchanged — worth keeping in mind, but not this diff.

Djidiouf (approved):

PR #2117, opened 2026-09-11
Replace the N26 Advancements Combobox with a visible table.

Djidiouf:
## Summary

- Replace the N26 Advancements Combobox with a visible 2D6 table. Rolling in-app is optional; a roll is an upper bound, and the player can claim any result they rolled high enough for.
- Skill follow-on is chosen with in-row radios (row 12 is a single “any Skill Set” option). Primary/Secondary results still nudge matching Skill Sets, but Confirm stays available for house rules.
- N23 Champion and Ganger Specialist promotions keep the Primary copy and grey/italic non-Primary sets, while still allowing those sets to be confirmed. Skill-access no longer 400s on a non-Primary `skillTypeId`, so Roll/Select still appear.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] N26: open Advancements, skip Roll 2D6, claim a row, pick a characteristic or skill, Confirm; credits follow the claimed row
- [x] N26: Roll 2D6, exact band preselected, higher rows grey/italic but still clickable; gang log says you may take any result of N or less
- [x] N26: Primary/Secondary radios — non-matching Skill Sets grey/italic with a warning; Confirm still works; row 12 shows one “any Skill Set” radio
- [x] N23 Ganger: pair characteristic and Specialist promotion; non-Primary Skill Set greys, warning shows, Confirm still works
- [x] N23 Champion promotion: custom Skill Sets still list skills; Roll for Skill log uses Champion access; non-Primary Confirm works
- [x] Promotion modal labels still use muted foreground

## Risk / rollout

- Low — Advancements modal UI and gang-log wording only; no schema or env changes

### DIFF
--- a/app/actions/logs/gang-fighter-logs.ts
+++ b/app/actions/logs/gang-fighter-logs.ts
@@ -3,7 +3,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { createGangLog, GangLogActionResult } from "./gang-logs";
 
-import { formatRollOutcomeLine } from "@/utils/dice";
+import { formatN26AdvancementThresholdLabel, formatRollOutcomeLine } from "@/utils/dice";
 
 // Advancement Logging Functions
 
@@ -220,6 +220,14 @@ export async function logSkillAdvancementDeletion(params: AdvancementDeletionLog
   }
 }
 
+function mayTakeLowerAdvancementResults(diceData: unknown): boolean {
+  return (
+    !!diceData &&
+    typeof diceData === 'object' &&
+    (diceData as { may_take_lower_results?: unknown }).may_take_lower_results === true
+  );
+}
+
 function gangerAdvancementRollDetailLine(diceData: unknown): string | null {
   if (!diceData || typeof diceData !== 'object') return null;
   const d = diceData as { result?: unknown; dice?: unknown };
@@ -228,7 +236,10 @@ function gangerAdvancementRollDetailLine(diceData: unknown): string | null {
     Array.isArray(d.dice) && d.dice.length > 0 && d.dice.every((x) => typeof x === 'number')
       ? (d.dice as number[])
       : [d.result];
-  return formatRollOutcomeLine(d.result, dice);
+  const label = mayTakeLowerAdvancementResults(diceData)
+    ? formatN26AdvancementThresholdLabel(d.result)
+    : undefined;
+  return formatRollOutcomeLine(d.result, dice, label);
 }
 
 export async function logRolledGangerAdvancement(params: GangerAdvancementRollLogParams): Promise<GangLogActionResult> {
@@ -241,7 +252,10 @@ export async function logRolledGangerAdvancement(params: GangerAdvancementRollLo
         : '?';
 
     const detail = gangerAdvancementRollDetailLine(params.dice_data);
-    const firstLine = `Fighter "${params.fighter_name}" rolled ${resultNum} on the ${params.advancement_table} table, resulting in: "${params.outcome_label}"`;
+    const firstLine =
+      mayTakeLowerAdvancementResults(params.dice_data) && typeof resultNum === 'number'
+        ? `Fighter "${params.fighter_name}" rolled ${resultNum} on the ${params.advancement_table} table and may take any result of ${resultNum} or less`
+        : `Fighter "${params.fighter_name}" rolled ${resultNum} on the ${params.advancement_table} table, resulting in: "${params.outcome_label}"`;
     const description = detail ? `${firstLine}\n${detail}` : firstLine;
 
     return await createGangLog({
--- a/app/actions/logs/gang-fighter-logs.ts
+++ b/app/actions/logs/gang-fighter-logs.ts
@@ -3,7 +3,7 @@
 import { createClient } from "@/utils/supabase/server";
 import { createGangLog, GangLogActionResult } from "./gang-logs";
 
-import { formatN26AdvancementThresholdLabel, formatRollOutcomeLine } from "@/utils/dice";
+import { formatRollOutcomeLine } from "@/utils/dice";
 
 // Advancement Logging Functions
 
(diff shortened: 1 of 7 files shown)


joeseos (commented):
@Djidiouf Thanks for the PR — reviewed against the full diff and the surrounding flows. Overall the approach (visible table, upper-bound roll, in-row radios) is solid and well-documented. A few issues worth addressing before merge:

## Findings

### 1. Champion promotion silently lost its "requires a Primary skill set" warning (Medium-High)
`components/fighter/fighter-advancement-list.tsx`

The old `championPromotionRequiresPrimarySet` variable (previously computed from `advancementType === 'promotion_to_champion' && selectedSkillSetAccess !== 'primary'`) was repurposed/renamed into the new `n26SkillRequiresMatchingSet` (~line 1717), which only fires for `advancementType === 'skill'` under `isCumulativeXp`. The JSX block that rendered the amber warning under the Champion "Primary Skill Set" combobox (~line 2673-2689) was deleted along with it, and nothing replaced it.

Compare with the Ganger Specialist flow, which *did* keep its equivalent warning (`gangerSpecialistRequiresPrimarySet`, still rendered at line 2549). So now:
- Ganger Specialist promotion with a non-Primary set selected → shows "This advancement requires a Primary skill set."
- Champion promotion with a non-Primary set selected → grey/italic label only, **no warning text at all**.

This contradicts the PR description itself ("N23 Champion and Ganger Specialist promotions keep the Primary copy... while still allowing those sets to be confirmed" implies parity between the two flows). It looks like an unintentional side effect of reusing the variable name rather than a deliberate change.

**Fix:** Reintroduce a `championPromotionRequiresPrimarySet`-equivalent (e.g. `advancementType === 'promotion_to_champion' && !!selectedCategory && !championPreviewSkillAccessLoading && selectedSkillSetAccess !== 'primary'`) and render the warning under the Champion Skill Set combobox, matching the Ganger Specialist treatment.

### 2. N26 roll gang-log message is duplicated and drops the actual result name (Medium)
`app/actions/logs/gang-fighter-logs.ts` (`logRolledGangerAdvancement`, ~line 245-266; `gangerAdvancementRollDetailLine`, ~line 231-243)

For an N26 roll (`may_take_lower_results: true`), the generated description is now:
```
Fighter "X" rolled 7 on the Advancement table and may take any result of 7 or less
Roll 7 (4, 3): any result of 7 or less
```
Both lines say "may take any result of 7 or less" — the detail line's `formatN26AdvancementThresholdLabel` label duplicates what the first line already states, and `params.outcome_label` (the actual exact-band result name, e.g. "+1 Cool or +1 Willpower") is no longer logged anywhere in this message.

**Fix:** Either drop the label from `gangerAdvancementRollDetailLine` when `mayTakeLowerAdvancementResults` is true (just show the dice breakdown), or keep `outcome_label` in the first line and only use the threshold phrasing for the second line — but don't say the same thing twice.

### 3. Skill-access route now permits fetching skills for a fully denied skill type, not just non-Primary (Low-Medium)
`app/api/fighters/skill-access/route.ts` (~line 228-234)

The removed check previously gated `skillTypeId` on being Primary. The new comment says "Callers ... may list a non-Primary set and warn; do not 400 here" — but the replacement has no access-level check at all, so a `skillTypeId` for a skill set with **no defined access (effectively denied)** for the fighter type now also returns its skills, not just Secondary/Allowed sets. This is broader than "non-Primary" — worth confirming this is intended, since the corresponding `get_available_skills` RPC (used by the ordinary skill flow) still excludes `denied` skill types entirely (`supabase/functions/get_available_skills.sql` line 94/150), so there's now an inconsistency between the two access paths for "denied" sets specifically. Not a security hole (the `checkPermissionCached(...).canEdit` gate above it is untouched and still required), just a rules-accuracy question.

## Other notes
- `components/fighter/fighter-promotion-modal.tsx` and the `dice-roller.tsx` / `n26-advancement-result-table.tsx` additions look correct and are consistent with the codebase's conventions (render-phase state sync pattern, edition-scoped tables, etc.).
- The auto-select-when-only-one-option logic in `selectN26Row` and the credit/xp wiring for `isCumulativeXp` look correct on read-through.

Item 1 is the one I'd block on — it's a real regression in a shipped validation warning, not a style nit. Items 2-3 are worth fixing but not blocking if you'd rather follow up separately.

---

Djidiouf:
Findings 1 and 3 are intentional.

That matches “don’t restrict; warn, still allow.” Ganger already greys those sets and can Confirm. Tightening the route to Secondary/Allowed-only would break house-rule picks of a denied set. Same for the Promotion to Champion. Changes here are intentional.

joeseos:
@Djidiouf Re-reviewed the latest commit (955554a) against the prior review's findings.

## Previous findings

1. **Champion promotion missing "requires a Primary skill set" warning** — **Outstanding**, left as-is intentionally (per your comment). Confirmed no code change: `fighter-advancement-list.tsx` still only renders `gangerSpecialistRequiresPrimarySet` (line 2549) and `n26SkillRequiresMatchingSet` (line 2814); there's still no equivalent warning rendered under the Champion "Primary Skill Set" combobox. Noting for the record since it's a real asymmetry with the Ganger Specialist flow, but I won't re-block on it given you've called it intentional — worth a one-line note in the PR description so future readers don't read it as an oversight.

2. **N26 roll gang-log duplicated/dropped result name** — **Resolved.** The new commit removes the `formatN26AdvancementThresholdLabel` call from `gangerAdvancementRollDetailLine`, so the detail line is now dice-only (`Roll 7 (4, 3)`) and the threshold wording lives solely on the first line. No more duplication, and the unused import was cleanly dropped (the helper is still used elsewhere for UI labels, so no dead code). Verified via diff of `app/actions/logs/gang-fighter-logs.ts`.

3. **Skill-access route permits fetching skills for a fully denied skill type** — **Outstanding**, left as-is intentionally (per your comment). Confirmed no code change in `app/api/fighters/skill-access/route.ts` — the `skillTypeId` branch still has no access-level gate. Fine given the `canEdit` permission check above it is unaffected and this is a rules-accuracy question rather than a security issue, but the inconsistency with `get_available_skills.sql` (which still excludes `denied` sets) remains — worth a follow-up ticket if house rules ever need denied sets to be excluded from `/skill-access` too.

Only `app/actions/logs/gang-fighter-logs.ts` changed (3 lines added, 5 removed) — the targeted fix for finding #2 above. No new issues introduced: the branch logic in `logRolledGangerAdvancement` still correctly falls back to `outcome_label` for the non-N26 case and only drops it for the "may take lower results" case, which is the desired behavior.

## Verdict: **Approve**

Blocking finding is fixed; the two remaining items are explicitly acknowledged design decisions rather than bugs.

---

joeseos (approved):

PR #2118, opened 2026-09-11
Split the gang stash to not over invalidate

joeseos:
## Summary

- Split of the gang id cache so that common mutations that does not effect fighters does not bust the complete gang cache.

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Performed some actions to check that data is not stale

## Risk / rollout

- Low - Risk is that this change may result in stale cache, but will in those cases be followed up with fixes.

### DIFF
--- a/app/actions/equipment.ts
+++ b/app/actions/equipment.ts
@@ -1,6 +1,6 @@
 'use server'
 
-import { invalidateGang, invalidateFighter, invalidateGangCampaignMembership, invalidateGangStash, invalidateGangFinancials } from '@/utils/cache-tags';
+import { invalidateGang, invalidateGangRoster, invalidateFighter, invalidateGangCampaignMembership, invalidateGangStash, invalidateGangFinancials } from '@/utils/cache-tags';
 import { createClient, createServiceRoleClient } from "@/utils/supabase/server";
 
 import { updateGangFinancials, updateGangRatingSimple } from '@/utils/gang-rating-and-wealth';
@@ -909,6 +909,7 @@ export async function buyEquipmentForFighter(params: BuyEquipmentParams): Promis
       // If this fighter is a beast, invalidate the owner's cache
       await invalidateBeastOwnerCache(params.fighter_id, params.gang_id, supabase, fighterPetId);
     } else if (params.vehicle_id) {
+      invalidateGangRoster(params.gang_id);
       if (vehicleAssignedFighterId) {
         invalidateFighter(vehicleAssignedFighterId, params.gang_id); invalidateGangFinancials(params.gang_id);
       }
--- a/app/actions/sell-equipment.ts
+++ b/app/actions/sell-equipment.ts
@@ -1,6 +1,6 @@
 'use server'
 
-import { TAGS, invalidateGang, invalidateFighter, invalidateGangCampaignMembership, invalidateGangStash, invalidateGangFinancials } from '@/utils/cache-tags';
+import { TAGS, invalidateGang, invalidateGangRoster, invalidateFighter, invalidateGangCampaignMembership, invalidateGangStash, invalidateGangFinancials } from '@/utils/cache-tags';
 import { createClient } from "@/utils/supabase/server";
 import { getAuthenticatedUser } from "@/utils/auth";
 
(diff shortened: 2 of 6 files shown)


joeseos:
@joeseos Reviewed against the codebase, including tracing every call site that touches `invalidateGangFinancials`/`invalidateGang`/`invalidateFighter` to check for stale-cache regressions.

## Findings

**1. `utils/cache-tags.ts` — the split is correctly wired.** No severity (informational)
`invalidateGang` and `invalidateFighter` both still bust `gangRoster` unconditionally, and every existing call site that mutates fighter/vehicle data already goes through one of those two (or has an unconditional `invalidateGang(...)` at the end of the function, e.g. `sell-vehicle.ts`, `delete-vehicle.ts`, `add-gang-vehicle.ts`, `move-to-stash.ts`, `move-from-stash.ts`). I audited all 18 call sites of `updateGangFinancials` in `app/actions` and didn't find any that rely solely on `invalidateGangFinancials` (which deliberately skips `gangRoster`) for a mutation that actually changes roster contents. So the core risk called out in the PR description — a write that changes the roster silently not busting it — doesn't appear to materialize anywhere in the current codebase.

**2. The 3 added `invalidateGangRoster(...)` calls actually fix pre-existing bugs, not just preserve behavior.** Severity: informational (nice catch)
In `equipment.ts` (buy, vehicle branch), `sell-equipment.ts`, and `update-vehicle.ts`, I diffed against `main`: before this PR, the vehicle branch in all three only invalidated anything `if (vehicleAssignedFighterId)` / `if (params.assignedFighterId)`. Buying/selling equipment on an **unassigned** vehicle, or editing an unassigned vehicle's name/stats, busted *nothing* before this PR — a real latent staleness bug. The new unconditional `invalidateGangRoster(...)` calls fix that as a side effect of the split. Worth calling out in the PR description since it's more than a pure refactor.

**3. Minor: redundant invalidation in the "assigned" branch.** Severity: low
In all three touched files, the new `invalidateGangRoster(gangId)` fires unconditionally, and then `invalidateFighter(...)` (which also busts `gangRoster`) fires again when the vehicle *is* assigned to a fighter — e.g. `equipment.ts`:
```ts
invalidateGangRoster(params.gang_id);
if (vehicleAssignedFighterId) {
  invalidateFighter(vehicleAssignedFighterId, params.gang_id); invalidateGangFinancials(params.gang_id);
}
```
Harmless (just one extra `revalidateTag` call), and it matches the existing style elsewhere in the codebase (`sell-vehicle.ts` and `add-gang-vehicle.ts` already call `invalidateGang` twice), so not blocking.

**4. Doc comment table formatting.** Severity: low/nit
The new rows in the `utils/cache-tags.ts` header table wrap a cell across two comment lines (`gang-{id} gang core, campaign` / `captives`), which reads a little awkwardly compared to the original single-line-per-tag layout. Purely cosmetic.

**5. PR template.** Severity: low
"How I tested" only says "Performed some actions to check that data is not stale" — given this is a cache-invalidation change where the whole risk is silent staleness, a short note on which specific actions were exercised (e.g. "bought equipment for an unassigned vehicle, checked roster refreshed without reload") would make the claim easier to trust.

## Checklist
- **Correctness**: Verified — see findings 1–2. The tag split is sound and the three added calls plug real gaps rather than introducing regressions.
- **Edge cases**: Unassigned vehicles were the main edge case at risk here; this PR improves that case rather than regressing it.
- **Security**: N/A, no user input / auth surface touched.
- **Performance**: This is the point of the PR — fewer roster-bundle rebuilds on pure financial mutations. Net positive.
- **Readability/maintainability**: Good — the new doc comments in `cache-tags.ts` (e.g. on `invalidateGangFinancials`) clearly state the invariant ("deliberately does NOT bust gangRoster... a write that also changes roster contents must say so itself"), which is exactly the kind of thing that prevents this pattern from rotting later.
- **Project conventions**: Consistent with existing invalidation style (small helper functions per concern, doc comment above the tag table).
- **DRY/YAGNI**: No unnecessary abstraction; `invalidateGangRoster` is a minimal one-line helper matching the existing pattern of the other `invalidate*` functions.

## Verdict
**Approve.** The refactor is correctly scoped, verified against every fighter/vehicle mutation call site in `app/actions`, and incidentally fixes two pre-existing staleness bugs for unassigned-vehicle mutations. Only nits above, nothing blocking.

---

Djidiouf (approved):

PR #2119, opened 2026-09-11
Fixed genestealer corruption since the name changed from infected to …

joeseos:
…corrupted in n26

## Summary

- Fixed so that genestealer corruption work in n26 after name change from infected (still called infected in n23)
- Replaced the ref-plus-effect that seeded Venator skill-set ranks with the file's existing render-phase state sync, clearing the react-hooks/set-state-in-effect error while keeping the scroll-lock cleanup in an effect.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Checked in the edit gang modal and all options are now displayed
- Tested in create gang and the option Genestealer Corrupted now shows up

## Risk / rollout

- None

### DIFF
--- a/components/admin/admin-edit-equipment.tsx
+++ b/components/admin/admin-edit-equipment.tsx
@@ -12,7 +12,7 @@ import { WeaponProfileInput, emptyWeaponProfile, EquipmentGrants, EquipmentAvail
 import { HiX } from "react-icons/hi";
 import { getFighterSubtypeSortRank } from "@/utils/fighterSubtypeRank";
 import { gangOriginRank } from "@/utils/gangOriginRank";
-import { gangSubtypeRank } from "@/utils/gangSubtypeRank";
+import { getGangSubtypeRank } from "@/utils/gangSubtypeRank";
 import { AdminFighterEffects } from "./admin-fighter-effects";
 import { EditionSelect, useEditions, editionSlugOf } from '@/components/edition-select';
 import { hasLethalityStatline, hasTradePoints } from '@/types/edition';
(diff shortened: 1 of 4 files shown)


joeseos (commented):
@joeseos Reviewed against the full diff and the surrounding code.

## Root cause check
Confirmed the bug and that the fix addresses it correctly. `components/gang/gang-edit-modal.tsx` and `components/create-gang-modal.tsx` both filter the "Outlaw / Corrupted" subtype column with `.filter(v => (gangSubtypeRank[v.subtype.toLowerCase()] ?? -1) >= 10)`. With the old single `gangSubtypeRank` map keyed only on `"genestealer infected"`, an N26 gang subtype named `"Genestealer Corrupted"` fell through to `-1` and was silently filtered out of the picker entirely — matching the reported symptom exactly. Splitting into `gangSubtypeRankN23` / `gangSubtypeRankN26` behind `getGangSubtypeRank(editionSlug)` fixes this correctly.

I checked every remaining reference to the old `gangSubtypeRank` export across the repo (`git grep`) — all three consumers (`admin-edit-equipment.tsx`, `create-gang-modal.tsx`, `gang-edit-modal.tsx`) were migrated consistently, `editionSlug` is correctly in scope at every call site, and nothing references the removed named export anymore. No build breakage expected.

## Findings

**1. (Low) `gangSubtypeRankN26` only covers the "Outlaw / Corrupted" column (ranks 12-14), nothing in 1-9.**
`utils/gangSubtypeRank.ts`

If N26 has any subtype that belongs in the "Unaffiliated" (1-9) column, it would hit the exact same class of bug you just fixed — silently filtered out of `gang-edit-modal.tsx` / `create-gang-modal.tsx` (though still visible, just unsorted, in `admin-edit-equipment.tsx`, since that one only sorts and doesn't filter). I can't verify this from the repo alone since `gang_subtype_types` rows look like runtime/DB data rather than something seeded in migrations — worth a quick sanity check against the actual N26 catalog if you haven't already confirmed the full list renders.

**2. (Low-Medium) Unrelated refactor bundled into this fix**
`components/gang/gang-edit-modal.tsx` (~line 295-320)

The skill-type rank-seeding logic was refactored from a `useRef` + `useEffect` combo into a `ranksSeeded` state variable updated during render (the same render-phase-sync pattern used elsewhere in the codebase). On inspection this looks correct — the guards prevent any infinite-loop risk, and splitting the scroll-unlock into its own effect is a reasonable cleanup — but it's unrelated to the genestealer/gang-subtype fix and isn't mentioned in the PR's "How I tested" section. It touches the drag-and-drop skill rank editor in the same modal, so a reviewer skimming the title wouldn't expect that code path to have changed. Worth calling out explicitly (or splitting into its own PR) so it gets its own test pass rather than riding along.

## Verdict: **Approve**

The core fix is correct, minimal, and well-explained (good comments in `gangSubtypeRank.ts` on why N23/N26 can't share one map). Neither finding above is blocking — #1 is just worth a quick manual check, #2 is a process note rather than a bug.

---

Djidiouf (approved):

PR #2122, opened 2026-09-11
Fix missing space after "private" on the restricted gang record page

prvlaar:
Compiler was collapsing the whitespace between </strong> and "by"; use an explicit {' '} to guarantee it survives JSX compilation.

## Summary

-

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [ ] 
- [ ] 

## Database

### Migration status

## Risk / rollout

-

### DIFF
--- a/app/gang/[id]/forbidden.tsx
+++ b/app/gang/[id]/forbidden.tsx
@@ -11,8 +11,8 @@ export default function GangForbidden() {
         stranger — there is nothing for you here.&rdquo;
       </p>
       <p className="text-white/70 text-sm leading-relaxed">
-        This gang has been set to <strong>private</strong> by its owner. Only
-        the gang&apos;s owner and campaign arbitrators can view it.
+        This gang has been set to <strong>private</strong>{' '}
+        by its owner. Only the gang&apos;s owner and campaign arbitrators can view it.
       </p>
       <Button asChild>
         <Link href="/">Return to the Homepage</Link>


joeseos:
@prvlaar Thanks for the PR — reviewed against the codebase.

## Findings

**1. The original code did not actually have the reported bug (verified by compiling both versions)** — Severity: **low** (informational, not a regression)

`app/gang/[id]/forbidden.tsx` line 14-15 (before this change): `</strong> by its owner. Only` all sit on the *same source line*, separated by a literal space character. JSX's whitespace-trimming algorithm only strips/collapses whitespace that spans a line break (or is leading/trailing on the whole text node) — it does not touch a space that already exists mid-line. I compiled both the before and after versions with `@swc/core` (the compiler Next.js uses here, no custom Babel config in this repo) to confirm:

- Before: `"private"` + `" by its owner. Only the gang's owner and campaign arbitrators can view it."` — the space is already present in the compiled output.
- After: `"private"` + `' '` (from the added `{' '}`) + `"by its owner. Only ..."` — also correct, but only because the two adjacent string children get concatenated in the DOM.

So both versions render identically ("private by its owner."); this change doesn't appear to fix a real rendering bug. It's harmless either way, but since the PR's stated purpose is "compiler was collapsing the whitespace," it's worth double-checking what was actually observed (screenshot/browser?) before merging — if there's a real visual bug, this fix wouldn't be the cause, since the underlying compiled output was already correct.

**2. PR template not filled in** — Severity: **low**
The Summary, "How I tested," and Risk sections are still empty/placeholder. Given this is a text-only tweak, it's low-risk, but worth a one-line confirmation that the page was visually checked in a browser (per repo convention for UI changes) rather than just reasoned about.

## Checklist
- Correctness: see finding #1 — the fix doesn't change rendered output either way, so no regression, but the diagnosis in the PR description appears incorrect.
- Edge cases / null handling: N/A, static text only.
- Security: none applicable.
- Performance: negligible (one extra text child).
- Readability/maintainability: fine, though the second line now reads a bit awkwardly formatted (67-char line wrapped for a change that doesn't need it).
- Project conventions: no CLAUDE.md/Prettier config found in repo to check against; nothing violated.
- DRY/YAGNI: N/A, no logic changes.

## Verdict
**Approve** (low risk, no functional regression) — but please double check whether there was an actual observed rendering issue, since the compiled output shows the original text already rendered a space between "private" and "by".

---

prvlaar:
Withdrawing — turns out this isn't a real bug in the code (confirmed the compiled output already had the space in both versions). Apparently a Node.js version quirk on our end that'll go away with the upcoming upgrade, not something this fix addresses. Closing rather than merging a no-op change.

PR #2125, opened 2026-09-12
Fixed so that users can save a challenge without selecting a winner

joeseos:
## Summary

- Fixed so that users can save their campaign challenge without selecting a winner.
- Edited the "Battle played" to fit existing UI pattern better

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested creating a new challenge round end editing a challenge and saving without selecting a winner.

## Risk / rollout

- None

### DIFF
--- a/components/campaigns/[id]/campaign-battle-log-modal.tsx
+++ b/components/campaigns/[id]/campaign-battle-log-modal.tsx
@@ -534,7 +534,7 @@ const CampaignBattleLogModal = ({
     if (isSubmitting) return false;
 
     // Validate required fields
-    if (selectedScenario === '') {
+    if (showResultFields && selectedScenario === '') {
       toast.error("Please select a scenario");
       return false;
     }
@@ -564,7 +564,7 @@ const CampaignBattleLogModal = ({
       return false;
     }
 
-    if (!hasAnyWinnerSelected) {
+    if (showResultFields && !hasAnyWinnerSelected) {
       toast.error("Please select a winner");
       return false;
     }
@@ -765,22 +765,22 @@ const CampaignBattleLogModal = ({
       content={
         <div className="space-y-4">
           {isUnplayedChallenge && (
-            <div className="flex items-start gap-2 rounded-md border border-border bg-muted p-3">
-              <Checkbox
-                id="battle-played"
-                checked={battlePlayed}
-                onCheckedChange={(checked) => setBattlePlayed(checked === true)}
-                disabled={isLoadingBattleData}
-              />
-              <div>
+            <div>
+              <div className="flex items-center gap-2">
+                <Checkbox
+                  id="battle-played"
+                  checked={battlePlayed}
+                  onCheckedChange={(checked) => setBattlePlayed(checked === true)}
+                  disabled={isLoadingBattleData}
+                />
                 <Label htmlFor="battle-played" className="text-sm font-medium">
                   Battle played
                 </Label>
-                <p className="text-xs text-muted-foreground">
-                  Tick this once the battle has been fought to record the result and
-                  file the report. Leave it unticked to save the challenge.
-                </p>
               </div>
+              <p className="ml-6 text-xs text-muted-foreground">
+                Tick this once the battle has been fought to record the result and
+                file the report. Leave it unticked to save the challenge.
+              </p>
             </div>
           )}
 
--- a/components/campaigns/[id]/campaign-battle-log-modal.tsx
+++ b/components/campaigns/[id]/campaign-battle-log-modal.tsx
@@ -382,11 +382,17 @@ const CampaignBattleLogModal = ({
       s.scenario_name === battleToEdit.scenario
     );
 
+    // A challenge slot is created without a scenario, so treat "none" as its own
+    // state rather than falling through to an empty custom name.
+    const existingScenario = battleToEdit.scenario || battleToEdit.scenario_name || '';
     if (matchingScenario) {
       setSelectedScenario(matchingScenario.id);
-    } else {
+    } else if (existingScenario) {
       setSelectedScenario('custom');
-      setCustomScenario(battleToEdit.scenario || battleToEdit.scenario_name || '');
+      setCustomScenario(existingScenario);
+    } else {
+      setSelectedScenario('');
+      setCustomScenario('');
     }
 
     if (battleToEdit.created_at) {
@@ -837,6 +843,7 @@ const CampaignBattleLogModal = ({
             </label>
             <Combobox
               options={[
+                { value: '', label: 'No scenario selected' },
                 { value: 'custom', label: 'Custom' },
                 ...scenarios.map(scenario => ({
                   value: scenario.id,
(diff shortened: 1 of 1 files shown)


joeseos:
@joeseos Reviewed against the full component (`components/campaigns/[id]/campaign-battle-log-modal.tsx`), tracing how `selectedScenario`/`customScenario` get populated for a fresh, unplayed challenge.

## Findings

**1. The fix is incomplete — a third, identical validation check was left ungated and will still block saving.** Severity: **high**

This PR gates the "select a scenario" and "select a winner" checks with `showResultFields` in two places (line 537, 567), but there's a third check with the exact same shape that wasn't touched, in both places it's evaluated:

- `handleSaveBattle`, line 542:
  ```ts
  if (selectedScenario === 'custom' && !customScenario.trim()) {
    toast.error("Please enter a custom scenario name");
    return false;
  }
  ```
- The `formValid` memo, lines 717–718 (this one is worse — it disables the Save button entirely, not just a toast on submit):
  ```ts
  const customScenarioValid = selectedScenario !== 'custom' ||
    (selectedScenario === 'custom' && customScenario.trim() !== '');
  ```

The problem: when the modal opens for editing an **unplayed challenge that has no scenario yet** (the normal state right after `generateChallengeRound` creates a slot — see `app/actions/campaigns/[id]/battle-logs.ts` lines 690–697, which inserts rows with no `scenario` field at all), the populate effect around line 380–390 runs:
```ts
const matchingScenario = scenarios.find(s =>
  s.scenario_name === battleToEdit.scenario_name ||
  s.scenario_name === battleToEdit.scenario
);
if (matchingScenario) {
  setSelectedScenario(matchingScenario.id);
} else {
  setSelectedScenario('custom');
  setCustomScenario(battleToEdit.scenario || battleToEdit.scenario_name || '');
}
```
Since a freshly created challenge has neither `scenario` nor `scenario_name` set, `matchingScenario` is never found, so `selectedScenario` gets defaulted to `'custom'` with `customScenario` set to `''`. That makes `customScenarioValid` false and `formValid` false — **the Save button stays disabled** even with `showResultFields` correctly false, regardless of the winner-selection fix in this PR. If the button were ever enabled (e.g. stale state), `handleSaveBattle`'s line 542 would also block it with a toast.

This is exactly the scenario the PR title describes ("save a challenge without selecting a winner"), so I'd expect this to reproduce on the most common path: run a challenge round → open one of the generated challenge slots → try to save without ticking "Battle played". Worth double-checking in a browser before merging — my read of the state logic says it still blocks, which would mean the PR doesn't fully fix the reported bug for new challenges (it may still appear fixed for challenges that happen to already have scenario data from an older/different creation path, which could explain why manual testing didn't catch it).

**Suggested fix:** gate both checks the same way the other two were gated, e.g.
```ts
const customScenarioValid = !showResultFields || selectedScenario !== 'custom' ||
  customScenario.trim() !== '';
```
and
```ts
if (showResultFields && selectedScenario === 'custom' && !customScenario.trim()) { ... }
```

**2. "Battle played" checkbox/label/description restructure.** Severity: none (looks good)
Checked other checkbox+label usages in the codebase (e.g. `components/create-gang-modal.tsx`) — the plain `flex items-center gap-2` row without a bordered/muted container is indeed the established pattern elsewhere, and `ml-6` correctly matches the checkbox width + gap to align the description under the label. This part of the PR does what it says.

## Checklist
- **Correctness**: See finding #1 — high severity, likely leaves the reported bug only partially fixed.
- **Edge cases**: The specific edge case that matters most here (challenge with no scenario assigned yet) is the one left unhandled.
- **Security**: N/A.
- **Performance**: N/A, no change.
- **Readability/maintainability**: The two fixed checks read clearly; the third one being left inconsistent is the main issue.
- **Project conventions**: UI restructure matches existing patterns (see finding #2).
- **DRY/YAGNI**: N/A.

## Verdict
**Changes requested.** Please gate the remaining `customScenarioValid` check (both the `formValid` memo and the `handleSaveBattle` toast) the same way the scenario/winner checks were gated, and confirm in the browser that saving an unplayed challenge (fresh from a challenge round, no scenario set) works end-to-end without ticking "Battle played".

---

joeseos:
**Separate bug, not an incomplete fix.** The two can't co-occur: if `customScenarioValid` is false the button is disabled and `handleSaveBattle` never runs; if it's enabled, line 542 can't fire. Only the `formValid` half was live, and the originally reported bug is fixed by 4807a984.

**Fixed at the root instead of gating the check**, since gating would still open the modal on "Custom" with an empty box. The populate block assumed every battle has a scenario; "no scenario" is now its own state, and the combobox gained `{ value: '', label: 'No scenario selected' }` — the only one in the modal that lacked an empty option. No validation change needed: `scenarioValid` is already gated on `showResultFields`, so the scenario becomes required once "Battle played" is ticked.

joeseos:
Re-reviewed the latest commit (9e6d273) against the prior finding, tracing the new state logic in `components/campaigns/[id]/campaign-battle-log-modal.tsx`.

## Previous finding

**#1 — Third validation check left ungated (High)** — **Resolved, via root-cause fix rather than the gating I suggested**, and I think it's the better fix. Verified:

- The populate effect (~line 385-390) now has a third branch: when a `battleToEdit` has neither `scenario` nor `scenario_name` (the fresh-challenge-slot case), it sets `selectedScenario` to `''` instead of defaulting to `'custom'` with an empty `customScenario`. That was the actual root cause — the old code conflated "no scenario yet" with "custom scenario, not yet typed."
- The `Scenario` combobox gained an explicit `{ value: '', label: 'No scenario selected' }` option, so `''` is now a first-class, selectable state instead of an accidental fallthrough.
- Downstream, both places I flagged are neutralized without needing their own gate: `customScenarioValid` (line 723-724) and the `handleSaveBattle` toast (line 548) only trigger when `selectedScenario === 'custom'`, which no longer happens by default — so a fresh challenge slot now has `selectedScenario === ''`, `customScenarioValid` is `true`, and the toast never fires.
- The existing `scenarioValid` gate (`showResultFields ? selectedScenario !== '' : true`, line 720) still correctly makes scenario selection required once "Battle played" is ticked, so this doesn't weaken validation for played battles — it only unblocks the unplayed-challenge case.
- Traced the "already has a real/custom scenario" and "matches an existing scenario" branches — both unchanged and unaffected.

**#2 — "Battle played" checkbox restructure** — unaffected by this commit, still fine.

Only the populate effect and the scenario `Combobox` in this one file changed (per the diff between 4807a98 and 9e6d273). No new issues spotted:
- No duplicate option values (`''`, `'custom'`, and scenario UUIDs can't collide).
- `resetForm` already defaults `selectedScenario` to `''`, consistent with the new "no scenario" state.
- Scenario name resolution on save (line 630-635) correctly falls through to `''` for the new empty state, matching the pre-existing "no scenario" payload shape from `generateChallengeRound`.

## Verdict: **Approve**

The originally reported bug and the follow-up scenario-gating issue are both fixed at the source. Nice catch turning "no scenario" into an explicit state instead of overloading "custom."

---

Djidiouf (approved):

PR #2126, opened 2026-09-13
Added sort order tie break on gang rating on challenges

joeseos:
## Summary

- Added sort order tie break on gang rating on challenges on the campaign page

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested with "Challenge Round" on the campaign page to see if the fallback sort order is gang rating

## Risk / rollout

- None


joeseos:
@joeseos Reviewed against the full component (`components/campaigns/[id]/campaign-battle-logs-list.tsx`).

## Findings

**1. Refactor of the comparison block is behavior-preserving.** No severity (verified)
Traced both the old and new `sort()` callback by hand: for any pair where `aValue`/`bValue` differ, the new unified `comparison` computation produces the exact same result as the old separate string/number branches, and the `sortDirection` flip is applied identically. The only behavior change is what happens on a full tie, which is the point of the PR.

**2. `gangRatings` correctly reuses existing data, no new fetch/duplication.** No severity
`members[].gangs[].rating` is already the field used elsewhere in this file (e.g. gang extraction around line 456, standings around line 548), so this doesn't introduce a parallel/duplicate source of truth — good.

**3. Tie-break is correctly scoped to pending challenges only.** No severity
`aPending === 0 && bPending === 0` (i.e. neither side is `isPlayedBattle`) correctly limits the new rating tie-break to unplayed challenge slots — played battles that tie on every column still return `0` (unchanged, stable-sort behavior), matching the PR's stated scope ("tie break... on challenges").

**4. Minor: the rating tie-break ignores `sortDirection`.** Severity: low
```ts
if (aPending === 0 && bPending === 0) {
  return (gangRatings.get(a.challenger_gang_id ?? '') ?? 0)
    - (gangRatings.get(b.challenger_gang_id ?? '') ?? 0);
}
```
This always orders tied challenge slots ascending by challenger rating (weakest first), regardless of whether the user has the column sorted ascending or descending — unlike the primary comparison a few lines above, which is flipped by `sortDirection === 'asc' ? comparison : -comparison`. If that's deliberate (e.g. "weakest gang's challenge always surfaces first as a tiebreak, independent of the user's chosen direction"), worth a one-line comment saying so explicitly, similar to the existing "Challenges outrank played battles regardless of the chosen sort" comment above it. If it's not deliberate, flipping it too would be a one-line change:
```ts
const ratingComparison = (gangRatings.get(a.challenger_gang_id ?? '') ?? 0)
  - (gangRatings.get(b.challenger_gang_id ?? '') ?? 0);
return sortDirection === 'asc' ? ratingComparison : -ratingComparison;
```

**5. PR template.** Severity: low/nit
"Type of change" has no box checked — this reads as a small feature (new sort behavior), might be worth checking "Feature" for changelog/history clarity. Not blocking.

## Checklist
- **Correctness**: Verified — no regression in existing sort behavior, new tie-break correctly scoped to unplayed challenges only.
- **Edge cases**: Missing/legacy `challenger_gang_id` falls back to `0` via `?? ''` → map miss → `?? 0`, which is a sane default. Old rows with no `status` column count as played (per `battleStatusOf`), so they never hit this new code path — correct.
- **Security**: N/A, client-side sort only.
- **Performance**: `gangRatings` is a small `useMemo`'d Map keyed off `members`, negligible cost, correctly added to the `sortedAndFilteredBattles` dependency array.
- **Readability/maintainability**: Good — the refactor of the comparison branch is actually clearer than the original duplicated string/number logic, and the new tie-break has an explanatory comment.
- **Project conventions**: Consistent with existing code style in this file.
- **DRY/YAGNI**: No duplication introduced; minimal, targeted change.

## Verdict
**Approve.** Solid, well-scoped fix with a correct, verified refactor. Only a low-severity question about whether the rating tie-break should respect `sortDirection` — worth a quick confirmation or a clarifying comment, not blocking.

---

Djidiouf (approved):

PR #2127, opened 2026-09-13
Refactored campaign cache

joeseos:
## Summary

- **Fix: campaigns permanently 404 after an image or map edit.** `invalidateCampaign` emptied a shared cache entry, and the refill could land on a sessionless request whose RLS-filtered read returned zero rows with no error — caching `null` forever for every viewer.
- Campaign cached readers now fill from a service-role client created inside the `unstable_cache` callback and take no client from callers, so the fill can't depend on the caller's session. `/campaigns/[id]` redirects to sign-in instead of swallowing auth failures.
- Deleted `GET /api/campaigns/[campaignId]` — it had no auth check, sat outside the `proxy.ts` matcher, and busted the cache it was about to read.
- Removed `refreshData()` and its 8 call sites; campaign mutations now follow the project convention (optimistic patch → server action -> reconcile from returned `data` → rollback on failure). Map data and battle logs lifted into page state; settings save moved into the modal that owns it. Also fixes `refreshData()` silently wiping triumphs, battle sessions, cycle, Discord config and custom trading posts on every call.
- Split the single `campaign-{id}` tag into nine subject tags (core, members, territories, battles, allegiances, resources, map, captives, trading posts), so a battle log save no longer rebuilds the whole page's data. Each mutation busts only the subject it actually wrote.
- Dropped the blanket loop in `update-gang.ts` that busted every campaign entry of every campaign a gang belonged to on *any* gang update — replaced by per-gang `gang-overview-{id}` subscriptions.
- Removed ~290 lines of dead code found along the way (two uncalled server actions, an uncalled cached reader, unused imports).

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tried a variety of actions to check so that cache does not go stale, both on the campaign page and on the gang page (to check gang rating on the campaign page) 

## Risk / rollout

- Stale cache that will be picket up by trouble tickets


joeseos:
@joeseos This is a large, well-executed refactor. I reviewed it in depth: personally traced the security-critical architecture change, and ran two focused audits — one walking every mutation site against the new 9-tag cache scheme, one checking every `refreshData()` call site got a real replacement. Full breakdown below.

## Findings

**1. Wrong cache tag busted after feeding a starved fighter.** Severity: **high**
`app/actions/edit-fighter.ts`, the `'starve'` case (~line 754). This decrements a `campaign_gang_resources` row (Meat), which is squarely `campaignResources` territory ("resource catalog + per-gang resource quantities" per your own doc comment in `cache-tags.ts`). The diff shows:
```diff
-invalidateCampaign(campaignId);
+invalidateCampaignCaptives(campaignId);
```
That busts captives instead of resources — looks like a copy/paste from the capture/rescue path elsewhere in the same file, which correctly uses captives. On `main` the blanket `invalidateCampaign` happened to cover this by accident; the split exposes it. User-visible effect: after feeding a starved fighter, the campaign resources view keeps showing the pre-decrement Meat count forever (`revalidate: false`), while captives gets pointlessly busted. Fix is a one-line swap to `invalidateCampaignResources(campaignId)`.

I had an agent audit all ~15 other mutated action files against the new tag scheme (battle-logs, campaign-allegiances/gangs/join-requests/map/members/resources/settings/territories, update-campaign-image, custom-share, custom-trading-posts, update-gang, battle-sessions) plus a repo-wide grep for leftover `invalidateCampaign(`/`TAGS.campaign(` — this was the only one found. Everything else, including the trickier fan-in cases (`campaign-map.ts` correctly also busting `campaignTerritories` when a map object a territory references gets deleted; `custom-share.ts` correctly busting both `campaignTradingPosts` and `campaignCore` since `campaigns.custom_trading_posts` is a column read by `getCampaignBasic`; `update-gang.ts`'s replacement of the old blanket per-gang campaign loop with `gang-overview-{id}` subscriptions on the read side) checks out correctly.

**2. Service-role client for cache fills — verified as a deliberate, sound tradeoff, not a regression.** No severity (verified)
This is the biggest architectural change here and deserves the scrutiny: every cached campaign reader now fills via `createServiceRoleClient()` instead of the caller's session-scoped client, bypassing RLS entirely at fill time. I checked whether this could leak data to unauthorized viewers:
- `supabase/migrations/20260719140000_add_campaign_join_requests.sql` documents in a comment that `campaigns` and `campaign_members` both already have `USING (true) SELECT` policies `TO authenticated` — i.e. RLS was never doing per-membership filtering on reads, only gating "signed in or not." Write-side RLS (arb/admin/owner checks) is untouched by this PR.
- Your own new comment in `app/campaigns/[id]/page.tsx` states this explicitly and correctly: *"This is the authorisation gate for the whole page: the cached readers below fill from a service-role client and so do not enforce RLS themselves."* I confirmed `proxy.ts`'s matcher excludes all `api/` routes (`'/((?!api/|...).*)'`) but does cover `/campaigns/[id]`, and the page now also does its own `getAuthenticatedUser` + `redirect(signInPath(...))` rather than swallowing the auth failure — so unauthenticated users are stopped twice (middleware + page), independent of each other, before this data can reach them.
- The two still-public API routes (`.../data/route.ts`, `.../allegiances/route.ts`) already had no auth of their own; `data/route.ts` was already explicitly commented "intentionally public" and used a service-role client pre-PR. `allegiances/route.ts` had no explicit auth either, though previously an anonymous caller's request-scoped client would likely have hit the `TO authenticated` RLS wall and gotten nothing — post-PR it'll now return real data to anonymous callers too. Given allegiance catalogs are low-sensitivity and the sibling `data` endpoint already establishes "any campaign ID is publicly readable via this API," this is a minor consistency widening rather than a new class of leak, but worth a one-line callout in the PR description since it's a behavior change nobody asked for explicitly.

This part is correct and well-reasoned — flagging it mainly so it's clear it was actually checked, given how easy it'd be to wave through "swapped to service-role client" without checking what that implies.

**3. Deleted `GET /api/campaigns/[campaignId]/route.ts` — verified safe.** No severity (verified)
Confirmed it had no auth check on `main`, confirmed via the `proxy.ts` matcher that it was already outside middleware coverage (all API routes are), and grepped the whole PR branch for any remaining reference to the bare path — every surviving `/api/campaigns/${id}...` call carries a suffix (`/data`, `/allegiances`, `/resources`, `/logs`) that still exists as its own route. No dangling caller.

**4. `refreshData()` removal (8 call sites in `campaign-page-content.tsx`) — verified, no regressions, some real improvements.** No severity (verified)
Traced all 8 old call sites against their replacements. All have a working optimistic-patch → reconcile-from-response → rollback-on-error triple. Two are actually better than before: the battle-log delete path now correctly releases a claimed territory on delete (previously masked by the old blind full-page refetch), and territory-update failures now do a precise snapshot-restore instead of a full refetch.

**5. "~290 lines of dead code" claim — verified accurate.** No severity (verified)
`addGangToCampaignDirect`/`removeGangFromCampaignDirect` (`campaign-gangs.ts`, ~220 lines) and `getCampaignGangsForModal` (`get-campaign-data.ts`) were confirmed uncalled anywhere both before and after this PR — genuinely dead, not something this PR broke.

**6. Settings save moved into `campaign-edit-modal.tsx` — verified correct.** No severity (verified)
The modal's own `saveSettings()` now calls the action directly and hands the parent everything the old manual local-state patch set by hand (including `updated_at`); rename/status/image changes propagate correctly to the rest of the page via `onSaved`.

## Checklist
- **Correctness**: One real bug (finding #1), otherwise the extensive cache-tag split and RLS-bypass architecture change both check out under close inspection.
- **Edge cases**: Map-object deletion cascading to territory cache invalidation, and campaign-column writes via `custom-share.ts` needing to bust `campaignCore` too, were both handled correctly — these are exactly the kind of fan-in cases that are easy to miss in this kind of split.
- **Security**: The service-role RLS bypass is deliberate and matches existing `USING (true) TO authenticated` policy; page-level auth is preserved and improved (redirect instead of swallowing failure). Minor note on `allegiances/route.ts` becoming reachable by anonymous callers (finding #2).
- **Performance**: This is the point of the PR — a battle log save no longer rebuilds the whole campaign page's cache. Clear win.
- **Readability/maintainability**: Excellent doc comments throughout `cache-tags.ts` and `get-campaign-data.ts` explaining *why* each tag list is shaped the way it is (fan-in subscriptions, choke points) — this will make the next person's job much easier.
- **Project conventions**: The optimistic-patch → server action → reconcile → rollback pattern is applied consistently across all 8 migrated call sites.
- **DRY/YAGNI**: Net removal of ~290 dead lines plus a shared `shapeCampaignTerritory`/`CAMPAIGN_TERRITORY_COLUMNS` helper (verified used in `campaign-territories.ts`) to keep the optimistic-insert shape in sync with the cached-read shape — good instinct, prevents exactly the kind of drift bug this size of refactor usually introduces.

## Verdict
**Changes requested** — purely for the one-line fix in finding #1 (`invalidateCampaignCaptives` → `invalidateCampaignResources` in `edit-fighter.ts`'s starve case). Everything else in this PR, including the two riskiest pieces (the cache tag split and the service-role RLS bypass), held up under a full audit. Once that one line is fixed, this looks good to merge.

---

joeseos:
Re-reviewed the latest commit (72fee06) against the prior review's sole blocking finding.

## Previous finding

**#1 — Wrong cache tag busted after feeding a starved fighter (High)** — **Resolved.** Verified the fix directly:
- `app/actions/edit-fighter.ts`'s `'starve'` case now calls `invalidateCampaignResources(campaignId)` instead of `invalidateCampaignCaptives(campaignId)`.
- Confirmed `invalidateCampaignResources` in `utils/cache-tags.ts` (line 168-170) busts `TAGS.campaignResources(id)` → `campaign-resources-{id}`, documented at line 27-28 as covering exactly "resource catalog + per-gang resource rows" — the right tag for a `campaign_gang_resources` Meat decrement.
- The stray `invalidateCampaignCaptives` import was removed cleanly; grepped the file and confirmed it had no other usage there, so nothing else in this file relied on that import.
- Added a one-line comment explaining why this mutation rides the resources tag — good, makes the choice legible for the next person touching this file.

Findings #2-6 from the previous review (service-role client fill, deleted API route, `refreshData()` removal, dead-code claim, settings-save move) were all "verified, no action needed" and are unaffected by this commit.

Only `app/actions/edit-fighter.ts` changed (4 insertions, 2 deletions) — the targeted fix above plus its import line and comment. No new issues.

## Verdict: **Approve**

The one blocking bug is fixed precisely, at the right cache tag, with nothing else disturbed.

---

Djidiouf (approved):

PR #2130, opened 2026-09-14
Added support to be able to exclude fighter types and equipment

joeseos:
## Summary

- Added support to be able to exclude/remove fighter types based on gang subtype. Modifiable in the admin GUI.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- Tested on local with the RPC already applied. If a gang is gang subtype Malstrain Corrupted, they no longer get access to either the Sumpkroc or Zerker

## Database

### Migration status

- [x] Migration already applied

## Risk / rollout

- None


joeseos:
## Code Review

Reviewed against `main` (base `29cb202`, head `7d0770e`). Checked out both branches; `npm run lint` and `npx tsc --noEmit` are clean for every file this PR touches (the handful of pre-existing lint errors elsewhere in the repo are unrelated).

### Findings

**1. `supabase/functions/get_equipment_detailed_data.sql:459-470` — deny logic doesn't cover legacy/affiliation fighter identity (High)**

The grant-side `fighter_type_equipment` join (lines ~410-423, just above) matches a fighter via four paths:
```sql
fte.fighter_type_id = $3
OR fte.vehicle_type_id = $3
OR (gd.legacy_ft_id IS NOT NULL AND (fte.fighter_type_id = gd.legacy_ft_id OR fte.vehicle_type_id = gd.legacy_ft_id) AND $4 = true)
OR (gd.affiliation_ft_id IS NOT NULL AND (fte.fighter_type_id = gd.affiliation_ft_id OR fte.vehicle_type_id = gd.affiliation_ft_id))
```
But the new deny `EXISTS` subquery that's supposed to override it only checks:
```sql
AND (d.fighter_type_id IS NULL OR d.fighter_type_id = $3)
```
It never checks `gd.legacy_ft_id` / `gd.affiliation_ft_id`. Concretely: if an admin sets up a deny scoped to `fighter_type_id = X`, and a fighter reaches equipment access to `X` only through a legacy-fighter mapping or a gang affiliation (both real, actively-used paths elsewhere in this same function), the deny silently never fires — the item stays on that fighter's equipment list even though the admin configured it to be removed. Since this is the core mechanism the PR is adding ("exclude fighter types and equipment"), this gap seems worth mirroring the same OR-branches used by the grant join, or at least a comment explaining why legacy/affiliation identity is intentionally excluded from denies.

**2. `app/api/admin/fighter-types/route.ts` (`ownedSubtypes` check, ~line 576) + `components/admin/admin-edit-fighter-type.tsx` (subtype rule builder) — adding a new subtype and a deny rule for it in the same save fails (Medium)**

`ownedSubtypes` is computed as `submitted ∩ stored` (subtypes already persisted in the DB for this fighter type, intersected with what's being submitted). But the client-side rule builder offers `selectedFighterSubtypes` — the live, not-yet-saved form state — as targets for a blanket subtype deny rule. So: toggle on a brand-new subtype → immediately add a "deny" rule targeting it → Save, and the whole PATCH is rejected with a 400 (`Save the subtype '...' on this fighter type before adding a rule for it`), aborting every other edit in that same save too (by design, per the "validate before any write" comment). The error message is informative enough that an admin can work around it by saving twice, but nothing in the UI hints that this two-step dance is required, and it's an easy sequence to stumble into. Consider either restricting the rule-builder's subtype dropdown to already-saved subtypes, or having the client hold back new blanket-subtype rules whose subtype isn't yet persisted.

### What looks solid
- Both `fighter_type_equipment` and `fighter_type_availability` PATCH paths validate before writing anything (avoids partial six-table saves), and the DB-level `target_chk`/`scope_chk` constraints (from the prior `fighter_type_availability` migration) back-stop the app-level validation regardless.
- `scopeToOwnedAvailability`'s `.or()` filter construction correctly quotes/escapes subtype names for PostgREST syntax (handles names containing parentheses, e.g. "Transport (X)").
- The `grantKey`/`availabilityKey` dedup logic correctly mirrors each table's own unique index (`NULLS NOT DISTINCT`, `excluded` deliberately excluded from the key so a grant and a deny for the same scope collide as intended).
- Good refactor extracting `GangOriginOptions`/`GangSubtypeOptions` into `gang-scope-options.tsx` shared between the equipment and fighter-type editors — no logic changes, straight DRY win.
- Cross-edition cleanup (dropping availability/grant rows whose referenced gang-scope ids belong to a different edition) is handled sensibly in both editors.
- The equipment editor's new "Deny" checkbox correctly disables the all-null case (a deny with literally no fighter and no scope would strip the item from every fighter in the game) while still allowing a fighter-type-only deny with no gang scope.

### Severity
- Finding 1: **High** — silent functional gap in the exact feature this PR ships, for two established fighter-identity paths.
- Finding 2: **Medium** — real workflow friction / confusing rejected save, not a data-integrity issue.

The equipment/fighter-type admin plumbing, dedup logic, and DB-level safety nets are solid, but finding 1 undermines the deny feature for legacy/affiliation-mapped fighters and should be fixed (or explicitly scoped out with a comment) before merge.

---

joeseos:
Re-reviewed the latest commit (c3583dd) against both prior findings.

## Previous findings

**#1 — Deny logic doesn't cover legacy/affiliation fighter identity (High)** — **Resolved.** `supabase/functions/get_equipment_detailed_data.sql`'s deny `EXISTS` subquery now mirrors the grant join's identity branches:
```sql
AND (
    d.fighter_type_id IS NULL
    OR d.fighter_type_id = $3
    OR (gd.legacy_ft_id IS NOT NULL AND d.fighter_type_id = gd.legacy_ft_id AND $4 = true)
    OR (gd.affiliation_ft_id IS NOT NULL AND d.fighter_type_id = gd.affiliation_ft_id)
)
```
This matches the grant-side conditionality exactly, including gating the legacy branch on `$4 = true`. The `d.vehicle_type_id IS NULL` restriction stays in place, now with a comment explaining it's intentional (no deny UI exists for vehicles), so the vehicle-identity OR-branches from the grant join are correctly left out rather than missed. This closes the gap for both fighter-identity paths the finding called out.

**#2 — New subtype + deny rule for it in the same save fails (Medium)** — **Resolved**, via the suggested approach. `admin-edit-fighter-type.tsx` now tracks a separate `savedFighterSubtypes` state (populated only from the loaded DB record) and the rule-builder's subtype `<select>` filters to `selectedFighterSubtypes ∩ savedFighterSubtypes` — i.e. only subtypes that are both currently selected and already persisted. This mirrors the server's own `ownedSubtypes` intersection logic, so a newly-toggled subtype simply can't be picked as a rule target until it's saved, which prevents the failing sequence entirely rather than just improving the error message. The new state is correctly reset alongside `availability` in all four places that clear it (fighter-type switch, cancel, failed detail fetch, dropdown change) — verified by diff, no missed reset path.

Checked whether the equipment editor (`admin-edit-equipment.tsx`) has the analogous "not-yet-saved subtype" hazard, since it also builds grant/deny rules — it doesn't: its subtype options come from a global `fighterSubtypeList` query of already-persisted subtypes, not a live not-yet-saved selection, so this bug pattern doesn't apply there.

Only the two files above changed (11 and 15 lines respectively) — both are the targeted fixes. No new issues.

## Verdict: **Approve**

Both findings are fixed correctly and at the root cause, not just patched around.

---

Djidiouf (approved):

PR #2133, opened 2026-09-16
Fixed so that spyrers in n26 can recover

joeseos:
## Summary

- Fixed so that Spyrers can recover in n26

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested on local

## Risk / rollout

- No risk


joeseos:
@joeseos Review of this PR:

## Summary
Small, focused fix: adds a `spyrerRecovery` capability row to `EDITION_CAPABILITIES` and gates the "Send to Recovery" button on it, so Spyrers can use Recovery in N26 while staying hidden in N23 (where the original blanket `!fighter?.is_spyrer` still applied).

## Findings

**None blocking.** A couple of low-severity observations only:

1. **Low — `types/edition.ts` (new `hasSpyrerRecovery` row) / `components/fighter/fighter-actions.tsx:176`**: Every other edition-gated boolean in this file (`spyrerRecovery` included) has an explanatory doc comment on the `EDITION_CAPABILITIES` row, but the `showGuilderSale` const right above has an inline comment explaining the edition split while the new `showRecovery` const doesn't. Not a functional issue, just a minor readability inconsistency with the sibling line — a one-line comment like `// N23 hides Recovery for Spyrers entirely; N26 allows it via Rig Glitches.` would match the existing style.

2. **Low — `components/fighter/fighter-actions.tsx:344` (Capture button)**: This button is still unconditionally gated by `!fighter?.is_spyrer`, unrelated to the fix here, but worth a sanity check — if N26 Spyrers can also be captured under the rules, the same `hasSpyrerRecovery`-style treatment might eventually be needed there too. Flagging only as a heads-up since it wasn't touched by this PR, not as something blocking this change.

## Correctness / verification
- `EDITION_CAPABILITIES` row follows the established `n23`/`n26` pattern exactly, and `hasSpyrerRecovery` mirrors the other `has*` accessors (e.g. `hasVenatorSkillAccess`).
- `showRecovery = !fighter?.is_spyrer || hasSpyrerRecovery(editionSlug)` correctly preserves N23 behavior (non-Spyrers always see the button; Spyrers never do) while enabling it for Spyrers in N26.
- No server-side logic (`editFighterStatus`, `isStatusIncompatible`, `canBeInRecovery`) gates on `is_spyrer`, so this is a pure UI visibility fix with no backend changes needed — consistent with the stated scope.
- No null/edge-case issues: `hasSpyrerRecovery` safely handles `editionSlug` being `null`/`undefined` via the existing `can`/`answerFor` helpers.

## Verdict
**Approve.**

---

joeseos:
Rules:
_"A Spyrer may never be Captured for any reason, rather they are considered to have suffered Superficial Damage (See page 69)."_

Djidiouf (approved):

PR #2135, opened 2026-09-16
Fixed so that Available to All is an available gang type in custom fi…

joeseos:
…ghter types menu

## Summary

- Fixes so that "Available to All" is available for users when creating a custom fighter type in custom assets.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested on local

## Risk / rollout

- Existing issue is if the user changes the gang type on a custom fighter type in custom assets, it's still cached in the add fighter as it uses useQuery cache keys. Existing issue and acceptable.


joeseos:
@joeseos Review of this PR:

## Summary
Fixes two real backend bugs that were hiding "Available to All" from the custom fighter types menu, plus a small UI cleanup in `custom-fighters.tsx`.

## Findings

1. **Correctness (confirmed fix) — `app/api/fighter-types/route.ts:382-386`**: Good catch. `cf.gang_type_id === gangTypeId` over-matched when `gangTypeId` is `null` (e.g. no `gang_type_id` query param), since every custom-gang-scoped fighter also has `gang_type_id === null` — `null === null` is `true`. The `gangTypeId &&` guard fixes this, and it mirrors the identical guard/comment already used a few lines up in the same file (the `customGangTypeId` skip), so it's consistent with the existing pattern rather than a one-off.

2. **Correctness (confirmed fix) — `app/api/gang-types/route.ts:39`**: Replacing the hardcoded `gang_type_id.eq.b181b2f7-...` with a name match (`gang_type.eq."Available to All"`) is the right fix — I checked `supabase/seed.sql` and confirmed N23 and N26 each have their own "Available to All" row with a different `gang_type_id`, so the old hardcoded UUID only ever matched one edition. Matching by name is consistent with how the rest of the codebase already resolves this row (e.g. `app/api/fighter-types/route.ts:206` already does `.eq('gang_type', 'Available to All')`), and I verified the string casing is used consistently everywhere (seed data, other route, this route), so the exact-match `eq` won't silently miss a differently-cased row.
   - One thing worth double-checking: this endpoint doesn't filter by edition server-side, so with this change a request now gets back *both* editions' "Available to All" rows instead of just one. I traced the only consumer (`components/customise/custom-fighters.tsx:557`) and it already filters the response client-side with `sameEditionForDisplay(gt.edition_slug, editionSlug)`, so no duplicate shows up in the dropdown in practice — but flagging it in case this endpoint gets a new caller later that doesn't do that filtering.

3. **Low / scope creep — `components/customise/custom-fighters.tsx:536-546`**: Removing the `<optgroup label="System">` wrapper is unrelated to the "Available to All" bug (that fighter type was already pulled out into its own top-level block above the groups, both before and after this diff). It's a reasonable simplification on its own, but it now makes "Custom" the only labeled group while system gang types render as an unlabeled flat list — a bit asymmetric UX-wise, and it's bundled into a bug-fix PR without explanation in the PR description. Not blocking, just flagging so it's a deliberate choice rather than a stray edit.

## Other checks
- No null/edge-case regressions spotted; the `gangTypeId &&` guard correctly preserves existing behavior for the normal (non-custom-gang) case where `gangTypeId` is a real id.
- The known caching caveat called out in the PR description (stale `useQuery` cache key after editing a custom fighter type's gang type) is accurate and reasonably deferred as a separate, pre-existing issue.
- Preview build succeeded.

## Verdict
**Approve.** Both backend fixes are correct and well-targeted; the optgroup removal is a minor, low-risk scope addition worth a quick gut-check but not worth blocking on.

---

app/api/gang-types/route.ts
```diff
@@ -35,8 +35,8 @@ export async function GET(request: Request) {
     // Only filter out hidden types if user is not admin
     if (!isAdmin) {
       if (includeAll) {
-        // Include "All" gang type even if hidden, but exclude other hidden types
-        query = query.or('is_hidden.eq.false,gang_type_id.eq.b181b2f7-59f9-452c-84fc-89f183fb8221');
+        // Matched by name, not id: each edition has its own "Available to All" row
```
Djidiouf, line 38: I'm not super fan of this. We saw the issue it caused when Genestealer Infected got renamed for N26.
Maybe it sucks having gang type id hardcoded but is it worst than having something matched by name? I'm not sure

Djidiouf (approved):
Approved but have a look at the comment if you want to reconsider

PR #2136, opened 2026-09-16
Fixed home tab deep links and the sign-in

joeseos:
## Summary

- Home tab deep links now works. Regressed when I previously fixed a lint error, so this fix is basically restoring functionality that were in place before. 

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested navigating back from a campaign page - lands in the campaign tab
- Tested entering URL with campaigns/uuid when logged out. Signing in redirects to the campaign

## Risk / rollout

- No risk


joeseos:
Confirmed finding 1 by hand on the final state of the branch.

I clicked each tab button and checked three things together after each one: the URL, the button highlight, and the rendered section.

- Click Campaigns -> ?tab=campaigns, Campaigns highlighted, Campaigns rendered
- Click Custom Assets -> ?tab=customassets, Custom Assets highlighted, Custom Assets rendered
- Click Gangs -> ?tab=gangs, Gangs highlighted, Gangs rendered
- Settings modal -> Custom Assets -> ?tab=customassets, Custom Assets highlighted, Custom Assets rendered

That last one is the homeTabSwitch path from settings-modal.tsx. The silent regression you flagged, where the URL updates but the content doesn't, is ruled out: all three signals agree in every case.

Worth noting the replaceState call isn't new. updateUrlParam already ran on every tab click before this PR. What changed is which signal drives the re-render, not how the URL is written. I also measured the cost: three tab switches produced zero requests to the Next server (only Supabase CDN image fetches), so there's no RSC round-trip on tab switching.

Cold-load deep links verified too: /?tab=customassets and /?tab=campaigns land correctly, /?tab=bogus and bare / fall back to Gangs, and the campaign breadcrumb lands on Campaigns. No console errors and no hydration mismatch.

Finding 2: removed rather than commented. The redirectPath !== '/' rule now lives once inside signInPath() in utils/auth.ts, so the proxy just calls the helper. The two checks weren't actually equivalent though: the proxy's governed the server redirect URL, while the one in app/sign-in/page.tsx governs whether to render the hidden input at all.

The branch has also grown since this review. Three changes it hasn't seen:

1. proxy.ts: authenticated users hitting /sign-in now redirect to safePostSignInPath(next) instead of always /, which previously discarded the deep link this PR exists to preserve.
2. app/sign-in/page.tsx: dropped the safeNext === nextParam test, which silently discarded any path the URL constructor normalizes (a space becoming %20, for example).
4. utils/auth.ts: signInPath returns a bare /sign-in when the target is /.

The first one makes a redirect target query-derived, so the open-redirect surface is different from what was assessed here. /sign-in?next=//evil.com/x collapses to / with the origin unchanged, but it's worth fresh eyes.

joeseos:
Finding 1 is a false positive. Next patches window.history.replaceState so raw calls are observed by the router — see node_modules/next/dist/client/components/app-router.js:235-278, where Next's own comment reads "Ensure the canonical URL in the Next.js Router is updated when the URL is changed so that `usePathname` and `useSearchParams` hold the pushed values." popstate isn't the only path. This predates the PR: updateUrlParam already used raw replaceState on every tab click.

Verified in a browser on this branch — URL, button highlight, and rendered section all agreed after clicking Gangs, Campaigns, and Custom Assets, and after the homeTabSwitch path from the settings modal. The described failure (address bar updates, content doesn't) does not occur. Three switches also produced zero requests to the Next server.

Finding 2 is correct and intentional: an unrecognized ?tab= renders Gangs and leaves the param alone. The old rewrite never fired on load anyway — that was the bug being fixed.

Djidiouf (approved):

PR #2141, opened 2026-09-18
Fixed gang log errors on campaign gang and user

joeseos:
## Summary

Migration 20260719120000_gate_arb_gang_access_on_accepted.sql added cg.status = 'ACCEPTED' to 25 policies. For gang_logs that was collateral damage: addGangToCampaign inserts campaign_gangs as PENDING when the inviter isn't the gang owner, then logged the join — and the inviter is neither the gang owner nor covered by an ACCEPTED row, so the insert was denied. Policies unchanged; application-layer fix only.

- The join is logged when a gang actually joins, not when it's invited. addGangToCampaign only logs if the row was inserted as ACCEPTED; acceptGangInvite writes the entry when the owner accepts, where RLS passes. It credits the inviter, not the accepter. An unaccepted invite now logs nothing — the owner is already notified by the notify_gang_invite trigger, and ~7% of invites are never accepted, so those gangs were getting a log line for a join that never happened.
- removeGangFromCampaign logged the departure after deleting its own  access, so for a non-owner the row granting the insert was already gone (predates the gate, commit 875c9857). The log now runs before the delete, and only for ACCEPTED rows.
- Removals are verified. All three delete branches use .select('id') and fail on zero rows — the campaign_gangs DELETE policy is narrower than the gang_logs INSERT policy, so a denied removal previously returned success and showed "Removed gang" while the gang stayed in the campaign.
- Comment cleanup. Removed 10 comments claiming the gang owner's user_id is passed "to satisfy RLS policy" — the WITH CHECK only inspects gang_id. The value is still passed; it's correct attribution, not an RLS requirement.

Not included: backfilling the ~1200 accepted gangs missing a campaign_joined since 2026-07-19, and the client-side createGangLog calls in components/battle-session/.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested on local. I will follow up and monitor the error logs in supabase.

## Risk / rollout

- Low. Can be reverted


joeseos:
**Comments on the PR review:**
1 (Medium) — false log on a zero-row delete. Real, but predates this PR: the log previously ran after the delete, and a zero-row delete leaves the row intact and ACCEPTED, so the insert succeeded there too - same false entry, plus a bogus success. Neither suggested fix is possible: once the delete lands the row is gone for everyone, so an arbitrator has no branch of the INSERT policy left to log under; and Only gang owners or admins can delete logs (schema.public.sql:13057) means whoever writes a spurious entry can't remove it. Fixed the reachable cause instead - retry: 2 (will be looked into in a separate PR as this is not the projects convention) on removeGangMutation was re-running a now non-idempotent action and writing up to three campaign_left rows; retry dropped. Double-submit was already blocked by confirmDisabled. The remaining concurrent-removal race is a follow-up, not a blocker.

2 (Low) — .limit(1) vs the fallback delete. Not reachable: addGangToCampaign:85-96 rejects a gang already in the campaign, so there's at most one row per (campaign_id, gang_id). Kept the campaignGangId targeting on the status read; no further change. (The DB doesn't enforce it — that'd be a unique index, separate migration.)
                                 
3 (Low) — campaignGangError only logged. Intentional; the catch immediately below already states the logging is best-effort.

Also: the campaign_gangs DELETE policy isn't from 20260719120000 — that migration creates no policies on campaign_gangs, and the DELETE policy has no ACCEPTED gate (admin / arbitrator / MEMBER owning the gang).

joeseos:
Skärmavbild 2026-09-18 kl  11 02 06

PR #2142, opened 2026-09-19
Add N26 composite battle-session markers with a shared catalog.

Djidiouf:
## Summary

- Add N26 battle-session tokens as outlined shapes with a centred symbol, driven by a shared marker catalog.
- Keep Broken, Gunked, Blaze, and Flesh Wounds as N23-only; Injured is a new boolean; Intoxicated is available on both editions.
- Persist through existing `session_record` condition JSON — no migration.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- [x] N23 battle session: Broken and Gunked still appear; no Status section; Wounds/Flesh Wounds heading unchanged; fighter-row badges stay grey circles
- [x] N26 battle session: Status, Wounds, Injured first then A–Z conditions, Hidden/Revealed on their own row; composite glyphs in the modal and on the fighter row
- [x] Hidden and Revealed still replace each other; Ready/Activated still toggles from the fighter-row action column on both editions
- [x] Confirm conditions, then reload: tokens and numeric wounds survive; fighter-row groups use `∙` for both owner and spectator
- [x] `npm run lint` and `npm run build`

## Risk / rollout

- Low — battle-session UI only; condition storage is unchanged JSON on `battle_session_fighters.session_record`.


Djidiouf:
Medium — shared refactor changes N23's existing fighter-row/modal layout, which the PR claims stays unchanged: inaccuracy regarding the commit message and PR but all good
Low — no glyph fallback for N26-only markers if ever rendered outside composite mode: addressed with inline comment
Low — repeated ternary for composite marker sizing: fixed

joeseos (approved):

PR #2146, opened 2026-09-23
Fixed scroll reset on navigation

joeseos:
## Summary

What
- Fixes pages not opening at the top after in-app navigation. Since the Next 16.3 bump (2fdc1b0c), a new page kept the previous page's scroll position.
- Moves the markup that was copied into all 21 breadcrumb pages into one components/breadcrumb-bar.tsx. Each page now passes only its items, and how each page loads its names is unchanged.
- Removes dead scroll workarounds: data-scroll-ignore, the dev console-warning suppression in app/layout.tsx, experimental.scrollRestoration (it only affects the Pages Router), and the manual window.scrollTo(0, 0) on the user profile page.

Why
- Next 16.3 turned on appNewScrollHandler by default, and the new handler no longer skips fixed-position elements. The breadcrumb slot renders before < main >, so its fixed bar was the first thing Next checked when deciding whether to scroll. The bar always sits 56px from the top, so Next treated the page as already scrolled to the top and did nothing. <main> never got to scroll.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Schema / migration

## How I tested

- Tested on local 

## Risk / rollout

- None


joeseos:
1. Duplicated battle-session label: Leaving it. It's one line that builds a display string in two places. If they drift, the worst case is slightly different label text, and a shared helper doesn't earn its place for that.

2. User profile self-link: Intended. The username linked to the page you were already on, so it's now shown as the current page. I've added it to the PR description.

3. aria-label on links: Intended. The visible text is already the link's accessible name, so "Navigate to X" only repeated it.

4. Scroll fix: Next's new handler doesn't walk the DOM to find the first element. It calls getClientRects() on a React Fragment ref around the slot. React's FragmentInstance.getClientRects only reads the slot's top-level host nodes and never looks inside them (searchWithinHosts = false). The display: contents wrapper has no box, so getClientRects() returns an empty list, not a {0,0,0,0} rect. Next treats an empty result as "not a scroll target" and leaves the scroll signal for <main>; its source comments describe that behaviour explicitly. I tested it too: after scrolling down and navigating, pages stayed at 1610, 1310 and 406px on main, and land at 0 with this change. Back/forward restores scroll position separately from this handler, and prefetching makes no difference to it. We do depend on Next here, as any fix would, but this relies on behaviour Next documents in its own code.

5. Nested landmarks: Intended. There were two nested navigation landmarks with the same label, and now there's one.

Djidiouf (approved):

PR #2148, opened 2026-09-23
feature/cap fighters at 50

joeseos:
## Summary

-   Adds a database trigger that caps each gang at 50 fighters. A gang with more fighters than that slows down the gang page, and in practice it usually signals abuse. 99.9% of gangs have 32 or fewer. The cap has to live in the database because RLS lets users insert fighters directly through the REST API, which bypasses the server actions. add-fighter.ts now shows "Max 50 fighters are allowed" when the cap is hit. The 2 gangs already over 50 keep their fighters but can't add more.                        

- Known limitation: at the cap, beast-granting equipment is bought but its exotic beast is silently not created, whether bought directly or as part of a new fighter's default gear.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [x] Schema / migration

## How I tested

- Not tested yet at trigger and RPC is not running yet. Will be tested once merged and any needed changes will be in a follow up PR.

## Database

### Migration status

## Risk / rollout

- None.


joeseos:
The beast case is a known limitation noted in the PR description. Throwing there would leave an uncharged fighter or stray beasts, and the proper fix is a cap check before any write.

Djidiouf (approved):

